PR #2, opened 2016-08-18
Docs in Markdown

carrodher:
To make it easier to read the docs at GitHub, I have transformed their plain text format to markdown.

The contents of all documents is the same, only the style has changed and corrected some grammatical  and formatting errors.

Modified files:
- Readme (New format)
- Authors (New format)
- Changelog (New format and Unified style dates and changes)
- Copying (New format)
- Install (New format)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,14 +5,14 @@ By Fabien Chereau & al - http://www.stellarium.org
 
 Stellarium comes with ABSOLUTELY NO WARRANTY. See the COPYING file for details.
 
-- [1. INTRODUCTION](#1.-introduction)                
+- [1. INTRODUCTION](#introduction)                
 - [2. INSTALLATION INSTRUCTIONS](#installation-instructions)
-- [3. REQUIREMENT](#requirement)
+- [3. REQUIREMENTS](#requirements)
 - [4. QUICK START](#quick-start)
 - [5. CONFIGURATION](#configuration)
 - [6. THANKS](#thanks)
 - [7. NOTE](#note)
-- [8. FULL REFERENCES & CREDITS](#full-references-&-credits)
+- [8. FULL REFERENCES AND CREDITS](#full-references-and-credits)
 - [9. LICENSE](#license)
 - [10. APPENDIX](#appendix)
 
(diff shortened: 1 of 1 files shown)


carrodher:
Hi, Is there any possibility of continuing with this PR?

cardinot:
Hi @carrodher,
apologies for the delay.

This github repository was just a mirror and that 's probably the reason why your PR hasn't got attention. However, we moved from launchpad 5 days ago, and now we are [officially on github](http://stellarium.org/code/2017/11/24/switch-to-github.html).

Thanks for your work. I do like the idea of moving the contents from plain txt to markdown. But it'd be nice to hear the opinion of other contributors too (@xalioth @alex-w @gzotti ). In any case, would you be interested in reopening this PR?

alex-w:
Just a question: how user should read this README file on Windows (the file is available as part of package)?

gzotti:
The good thing about Markdown is that it is still a plaintext file with sort of ASCII art formatting. 

The simplest working solution is to rename as README.txt for the Windows package please. The same applies for all plaintext files people could potentially want to read: Changelog, GPL, INSTALL, AUTHORS, ... 
With the current renaming as RTF, some "higher level editor" will be called, may ask about which font encoding to use for import (most users will panic at this "scary dialog" stage and never read it!), use proportional font, try to line-wrap, ... this looks always terrible.

Other option: convert to HTML for the Windows package. (Allows hyperlinks, nicer formatting, etc.)

cardinot:
Markdown files are very popular nowadays; I think that Windows users may be used to open it in Notepad or may have a markdown reader. So, I think we should leave the .md files as it is, even for Windows.

In any case, if you think we should increase the compatibility of those files; another option would be to convert them to pdf (or even html, as gzotti suggested), but without deleting/renaming the md files.

gzotti:
I am afraid that only few Windows-only people will know what .md means. Windows does not know the format, and a popup appears "Which program should open the .md?". --> Search Windows Store.

We can absolutely leave the md content as it is, but change the file ending to .txt by the ISS pack script. It would be better than current .RTF. This would be the easiest solution for files which are rarely viewed in the first place...

cardinot:
hmm.. yes, my initial point was to leave the .md files there and just add a copy of them in another windows-compatible format, eg., txt, pdf or html.

However, I agree that a `.txt` copy might be the easiest/best solution now; Also, on a second thought, I'd say that having both `README.md `and `README.txt` sounds unnecessary... So, yes, I agree that, for Windows, we could leave the same content, just changing the file format from `.md` to `.txt`

Issue #11, opened 2017-11-15
Some DSO names are dubious


fmeynadier:
Hello,

I am a bit surprised by some of the names appearing in my favorite planetarium software for some deep sky objects... 

I think this tracks back to commit ee300ad15668aea39aea3b604bef1dabc5f70321

Impacted file :  nebulae/default/names.dat (and possibly translations of it ?)

The indicated source for the names is : the list of 'The 500 Best Deep Sky Objects' by Don Pensack; https://www.cloudynights.com/topic/472872-500-best-dso-list/ 

(later tagged "B500" once comments were allowed, i.e from commit 919221490a660da0559bb529291546176abfa850)

Although the list itself is fine and certainly useful, the "name" column contains mostly nicknames for which I don’t find any primary source. To my opinion they should not appear in Stellarium, which should stick to official identifiers (or more commonly accepted name sources). 

I would suggest to remove the lines with B500 source (I can submit a PR if agreed).

Anyway thanks for the good work on Stellarium !

gzotti:
While the DSO names are still work in progress and I very much insisted that we at least track a reference for some "funnies" and not collect everything from everywhere, I agree that some names require better source reference or more common acceptance. However, common acceptance requires previous publication in widespread journals and software and will likely take 3-20 years, so this is circular argumentation of a tricky kind. E.g. you will likely not find the name "Pillars of Creation" before that Hubble image.
I suggest that you write to Don Pensack and ask whether he can give references in a V2 of his Excel sheet: are the names his own invention, published elsewhere (journal, book, citable sources), learned at a star party from a particular observer group, ...  
I am not sure whether we should add more complications like filtering the list by user-selected sources. If you absolutely dislike some names, you can delete them.

fmeynadier:
Thanks for your answer. Sorry if I missed some discussion on the subject (I did try to browse through the mailing list’s archive but did not find any discussion on that. I’d be happy to read any thread you would point me to).

Before writing to Don Pensack I’d prefer to know if anybody informed him already that his list has been incorporated in Stellarium, which may change a lot the content of my message.

But apart from that : I think a good rule of thumb would be to see if image search engines returns any relevant results. For example, "lawnmower cluster" (=NGC 663 according to B500, which is the one that initially raised my eyebrows) returns zero astronomical results, whereas "pillars of creation" returns nearly 100% NGC 6611/M16 - related results. "Eagle nebula" (which is the label I see currently in that region) is also OK according to this criterion.

[...]

PR #12, opened 2017-11-22
Cleaned up some less known DSO names

fmeynadier:
As discussed in issue #11, I browsed through the DSO names and removed
the ones that were not very commonly used.

Note that the problem has also been spotted here :
https://stargazerslounge.com/topic/303840-where-does-stellarium-get-these-names/

Most of them came from the B500 source but I occasionaly spotted
others.

Methodology : combined google image search on the name, and/or
general search on the identifier to look which nicknames where
proeminent. If unsuccesful : entry removed.

I added a guideline in the file header for future contributors. I also
checked in the source to see how duplicates where handled, and wrote
my findings there too (1st entry = proper name, next = aliases).


alex-w:
Using principle "it's bad because I don't know it" is very bad idea for this case. I fully disagree with this patch because:
1) You are dropping some well-known scientific names for DSO's
2) You are fully ignore well-known non-English names (what was translated to English)
3) You are ignore non-internet sources, e.g. books and magazines (printed)
4) You are ignore objects, who traditionally not used by amateurs for the astrophotography

fmeynadier:
Hi,

I think this is unfair. I did not use the "it's bad because I don't know it" principle. Of course I may have been biased by my experience (I did not check the ones that seemed obviously "good" to me), or mistakenly dropped a line that shouldn’t have been. But at least I tried to use an objective rule.

About your concerns : 
> 1. You are dropping some well-known scientific names for DSO's
This is possible but would then be a mistake. Can you point out an example ? 

> 2. You are fully ignore well-known non-English names (what was translated to English)
As I understand, nebulae/default/names.dat is the reference for translations, so I think it is wrong to put translations from other languages there. Besides if the object has a well-known non-English name I don’t see the point in translating it - just put the original name in default/names.dat, or in a separate, language-specific file (just as for the different sky cultures).

> 3. You are ignore non-internet sources, e.g. books and magazines (printed)
Yes. Because this is easy to check and because I don’t think any well-known name would stay confined in books or magazines and have almost zero occurence on the web in 2017. Again, can anyone think of a counter-example ?
Also it seems preferable to be a little conservative (i.e. miss a name that is indeed used, but scarcely) rather than display, as a primary identifier, a name that is a misinterpretation of a text, or a nickname used by a small group of observers. 

> 4. You are ignore objects, who traditionally not used by amateurs for the astrophotography
On the contrary, I think the "seen on the web" criterion avoids this particular pitfall, to the point of being too much in favor of agencies press releases with catchy names.

Anyway, of course, this is only a proposal and you, maintainers of Stellarium, do as you wish. But I doubt I am the only one bothered by those funny names, I think they make Stellarium look less reliable to advanced users, and will induce beginners in error. Apologies for proposing a solution to this problem, I’ll be happy to see a better one emerge.

alex-w:
Fabien, why you are close this request?

cardinot:
> Fabien, why you are close this request?

@alex-w, I suppose that he closed just to make sure that nobody will try to merge this PR. As the git history was changed after this PR was opened, this PR is now broken and should not be merged through github.

However, the discussion still opened.

If people decide to merge the commit from this PR, then it should be cherry-picked and pushed directly into master. Or @fmeynadier could just fetch the rewritten history and rebase the local changes into the rewritten history; it'd fix the PR, allowing it to be reopened.

fmeynadier:
I did rebase my branch, master on fmeynadier/stellarium is now 1 commit ahead of stellarium/stellarium.

xalioth:
I didn't closed anything manually, I think github noticed that the MR has now 10000 commits and decided it was too much :)

fmeynadier:
By the way : the volume of commits was so big it did not fit into 1 push (maximum package size exceeded). So, after I rebased my local tree, I had to split the push to my github fork in 2, like this : 

```
git push -f origin f580c3dfb5037b9aa4035b5c45eff72595f63259:master
git push -f origin master
```

(I picked a random commit, far enough in the list so the first push stays below the limit).

If you’re still interested I can submit a new pull request.

cardinot:
@xalioth 
Yeah, just saw that it was automatically closed due to the difference between the history from master and the fork...

@fmeynadier 
I'm not sure if it'll be possible to reopen this PR anyway [1]. I guess that you could try to git fetch --all, move to the remote/master branch and then apply your_commit.diff. But I'm not sure if it would fix this PR anyway.

In any case, it seems to be much easier to just open a new PR and continue the discussion from there (copy the link from this PR just for reference). Now we are officially on github and we won't change the git history anymore, so it won't happen again.

Thanks =D

[1] https://github.com/isaacs/github/issues/361

PR #14, opened 2017-11-24
Cleaned up some less known DSO names (bis)

fmeynadier:
This is a re-submission of PR #12, because of the huge number of commits that happened in-between (tree sync from bazaar I suppose).

As discussed in issue #11, I browsed through the DSO names and removed
the ones that were not very commonly used.

Note that the problem has also been spotted here :
https://stargazerslounge.com/topic/303840-where-does-stellarium-get-these-names/

Most of them came from the B500 source but I occasionally spotted
others.

Methodology : combined google image search on the name, and/or
general search on the identifier to look which nicknames where
prominent. If unsuccessful : entry removed.

I added a guideline in the file header for future contributors. I also
checked in the source to see how duplicates were handled, and wrote
my findings there too (1st entry = proper name, next = aliases).

### DIFF
--- a/nebulae/default/names.dat
+++ b/nebulae/default/names.dat
@@ -7,10 +7,6 @@
 # 21 - end string name    proper name of the object (translatable). Format: _("name") [# comment]
 # comment is optional but should be used to indicate source for names. See source list below for keys.
 #
-# In case of duplicates, the first occurence is taken as proper name, next ones as aliases
-#
-# Please include only widely used names here (rule of thumb : should be easy to google and not limited to 1 source).
-#
 # Sources: 
 #   WK:   Well-known name which does not need special source. 
 #         Typically annotated for naked-eye or Messier-like bright objects with names given until the late 20th century. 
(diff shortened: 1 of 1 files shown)


gzotti:
I would not delete the lines, just comment them away for now, and possibly reactivate after some changes. Some people my like to have as many names as possible found somewhere on the net.

It seems we will indeed need a "references preferences" setting that tracks a listbox with "active" and "inactive" allowed references. If the B500 list is so strange, it could then be excluded.  (And I am happy to see that "Caspar the friendly ghost" and other questionable comic characters are not common-place yet but also regarded "silly" by many who find actually time for observing!)

fmeynadier:
I brought them back but commented them, does it seem more suitable to you ? (I also removed the guidelines as it does not seem to be agreed on). I tested it by replacing the names.dat file in my local installation of Stellarium and it seems to work as intended.

Regarding your "references preferences" setting, I would suggest a more versatile mechanism (independently of sky cultures) : one could imagine allowing users to load arbitrary lists of objects from the gui interface, with whatever comment/name they might wish. A simple file with RA/DEC/type(for symbol, or grouped under [...] tags )/name or free-form comment would do. This would allow to load the complete B500 list if one wishes to do so, or plan an observation night, etc… "Common names" could then be provided using this mechanism.

gzotti:
Actually (if I remember correctly) the guidelines were added after the B500 names to prevent further silent addition of such lists. Should be kept.

The DSO names should be independent from skycultures, which usually are naked-eye only. Or actually, theses DSO names are from "western amateur observer astronomy and professional astronomy publicity and press release skyculture". I am not aware of larger DSO lists of non-"western" telescope observers which are genuinely originating from a national/cultural/tribal tradition. (Of course, Pleiades, Orion nebula, Eta Car, Magellanic Clouds and a few open clusters may have some proper names, but in general not more than a handful.)

We have had plans for user lists for years now, and indeed B500 could rather become a user list, whereas the default list should only provide widely accepted names. 

Designing a usable general system for such lists is not easy, and developing time has gone elsewhere into more urgent topics. Requiring coordinates for known NGC or non-NGC DSOs, or HR or SAO or HIP stars, seems nasty.  Sometimes the coordinates are important (historical star lists, or historical DSO lists by Messier, Mechain, Herschel or such) and would require a pointer/symbol at coordinates, optionally linked to the referenced object (if that link is known), in other lists (like "favourite objects") we know which stars/DSOs are meant, so they should just refer those IDs from HIP or HR or whatever star catalog we can understand. There are even cases for both. e.g. "show Tycho's stars" could mean one of:
-(1) Mark HIP stars for which a clear identification in Tycho's catalog exists
-(2) Plot a mark (e.g. circle scaled by magnitude as given by Tycho) and add an anchor line to the HIP star (allows estimate of magnitude and position error).

I would say, your point is taken, we need some system to show or suppress various name catalogs, and I am confirmed in my wish for traceable, referenced names. The problem is that commenting them now will remove the translations, and bringing them back later will annoy the translators because they may have scratched their heads over them already once. 

A hackish way for 0.17 could be a key in config.ini like "ignore_dso_name_sources=B500,WP,???" which would be understood by the DSO name loader to ignore all lines which contain B500 or WP or ??? in the comment part of the DSO file. Some better solution may then be found, maybe before summer 2018 or so. In this sense, I think we should keep the file as-is, but I still thank you for your involvement and thinking about how to make Stellarium better.

Any other opinions?

fmeynadier:
Just to clarify : the guidelines I removed were the ones I added in the previous commit, i.e. the ones requesting to google the names first.

On translations : work already done would not be lost. In my french localized version, names I commented in default/names.dat simply don’t appear any more, if de-commented they would just reappear (as long as the translation file keeps them).

gzotti:
Oops, maybe you wrote the line, sorry. Then we can add this guideline.

Translation files are auto-extracted, and commented-away content is in danger of being removed. So I keep my concern.

I am waiting for comments by other team members. Any more opinions?

leonardcj:
@gzotti depending on you extraction and update/replacement method it is entirely possible to retain your commented-out translation in your PO files.  msgmerge will typically take obsolete strings, move them to the end of the PO file and mark them with #~ from whence they can be resurrected if they are made no longer obsolete.

alex-w:
I've revisiting the names of DSO from various books and works going is not very fast (no google within books :) ), plus I added new option to the GUI to hide an additional names for DSO.

alex-w:
I’m close this PR, but we have a plan to add the manager for common names of DSO

fmeynadier:
Thanks for the info, glad to know there is a way forward that will make everybody happy !

PR #15, opened 2017-11-26
[TelescopeControl] INDI support

asiniscalchi:
It implements a INDI client: you need an external server to connect to.
Tested with INDI telescope simulator and CEM25 driver.

IMPORTANT: it works only when there's only one device connected to INDI.

ADVICE: The plugin needs a deep refactoring.


gzotti:
Thank you, this sounds interesting. Can you please write a user manual entry as well? It should cover the basics for people who have an INDI mount. It would go into guide/plg_atTheTelescope.tex, most likely add a subsubsection of (I assume) less than one page in line 708 just before "RTS2 telescopes". We cannot write it without such a device!

If you feel able to make the deep refactoring of the plugin, you are more than welcome!

asiniscalchi:
Build is failing cause INDI headers are not found.

CMake Error at cmake/modules/FindINDI.cmake:256 (message):
  Could not find INDI include directory
Call Stack (most recent call first):
  plugins/TelescopeControl/src/CMakeLists.txt:1 (FIND_PACKAGE)

gzotti:
I think you could simply update .travis.yml and add installation instructions for INDI headers/libraries.

.travis.yml
```diff
@@ -19,13 +19,15 @@ compiler:
 
 before_install:
   - sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
+  - sudo apt-add-repository -y ppa:mutlaqja/ppa
```
cardinot: Travis uses "Ubuntu 14.04.5 LTS". INDI needs 15.10 or higher [1].

You'll have to install it from source [2,3] and check which dependencies would you need to compile first.

[1] https://launchpad.net/~mutlaqja/+archive/ubuntu/ppa/+packages?batch=75
[2] https://github.com/indilib/indi
[3] https://github.com/indilib/indi/releases/tag/v1.5.0

asiniscalchi:
me vs travis 1-0

alex-w:
I've added some fixes for your branch. How about building the source code on Mac OS X or Windows?

asiniscalchi:
No Windows neither Mac OS X machine here.
Do travis build on those OS ?

gzotti:
I think INDI support should not be mandatory. Like the GPS device support, if library is missing or QT5 module not found, it should just not be built (cmake -DENABLE_INDI=0 or so). And we clearly need instructions (or CMakeFile.txt rules) how to provide INDI support for Windows and Mac, or (as worst case) disabling of INDI on those platforms.

asiniscalchi:
with the support of INDI optional we got all

cardinot:
> me vs travis 1-0

Thanks! can you truncate all theses travis's commits?

alex-w:
@asiniscalchi may be you can fully rewrite Telescope Control pluign (into Device Control plugin for example)?

asiniscalchi:
compiled in win64 and linux.
Can't compile in OSX.

alex-w:
Yes, the version should be updated

alex-w:
Yes, it works on linux and windows! Thank you very much!

But this branch has problem on the exit now - please fix it.

alex-w:
@guillaumechereau could you check it on Mac OS X?

alex-w:
It works on Mac OS X too!

asiniscalchi:
However ... what's the goal of the refactoring of TCP ?

gzotti:
Sounds very good (I have not tried anything so far, though). But please add some usage guidelines into the user guide as requested above!

*You* stated that the plugin needs "deep refactoring". Maybe you can elaborate on what you had in mind? Code seems complicated, maybe. You added INDI support and improved structure, other things? The whole plugin may also need clearer documentation. 

The plugin must work with the dummy (simulator), directly connected telescopes with LX200 and NexStar protocols, RTS2 systems, and to some early telescope servers and the StellariumScope program using some early TCP protocol described at http://stellarium.sourceforge.net/wiki/index.php/Telescope_Control_(client-server). Can you explain (in this user guide entry) to somebody who has never heard of INDI in 2 sentences what INDI provides? I know "something with telescope/instrument control", but there is e.g. no complete support for WIndows (?). Do I understand that your code can connect via TCP/IP to some INDI telescope server which runs on some other computer (mostly Linux, e.g. small SoC like RaspberryPi or Odroid?) and can drive a large variety of telescopes, i.e. is a more general replacement of the other telescope servers. It can then easily become the "recommended solution", and the old telescope servers could possibly be dropped.

On the other hand, the external/indi contains a huge collection of instruments, cameras, rotators, etc. Do we need this library as "client" already just to talk a few sentences of "GOTO"?  Then I would still prefer if INDI remained an optional component. Or is this a start for more instrument control? Other opinions?

guillaumechereau:
Compiles fine on OSX (Qt 5.9.2), and all the tests pass.  Not sure why travis complains about this PR.

PR #21, opened 2017-12-02
using stelMain library

asiniscalchi:
- better compilation time.
- test are testing the lib used to create stellarium bin
- everybody can test their code: just link with stelMain in the test executable


alex-w:
The branch break the GUI :(

PR #24, opened 2017-12-04
Update Chinese Sky Culture

sunshuwei:


sunshuwei:
The six constellations were cancelled in the Ming Dynasty.

Or they are lost.

But they still appear in traditional culture.

They are all located in the South sky. In ancient times, it was difficult to observe the southern sky.
The constellations set up by the missionaries covered them.

So the six constellations can be deleted. Do I need to delete them?

alex-w:
> So the six constellations can be deleted. Do I need to delete them?

No need delete them. You may define line for constellation, who have just one point, for example - 063 1 677 677.

sunshuwei:
Because this constellation has only one star.
If  I do not, the name of the constellation will not be displayed on the window.
The same is true of the official map of China.

Add two screenshots.

alex-w:
> Because this constellation has only one star.
> If I do not, the name of the constellation will not be displayed on the window.
> The same is true of the official map of China.

Example for Beta Ceti:
260  1 3419 3419

sunshuwei:
Yes，Beta Ceti，Its Chinese name is 土司空.
It alone constitutes a constellation, also called 土司空。 
If this does not conform to the specification, please tell me the correct way to handle it.
qq 20171204170156

alex-w:
Star name and constellation name are defined in different files and has a same name - no problem. I said about file constellationship.fab

sunshuwei:
If a Chinese constellation has only one star, there is no second star to define line.
By defining a line with a star itself, the result is that you can see the name of the constellation, but there is no line.

alex-w:
The line with begin and end on one star is a dot :)

gzotti:
If you know that some constellations have been "officially deleted", please remark this in a comment like 
```
# until 1600
```
One of the features I would like to see later is "changing constellations" over time, i.e. when you set year 1500 you may see different constellations than in 1800 or so. In case there is an accurate date (like some official decree with a date), of course you should give this here.

sunshuwei:
1.Yes，it is a dot.
The Chinese constellation is not an area, it contains only the stars with names and add-on star with names .None of the rest belongs to it. 
Usually add-on star cannot be used for define line.
Chinese constellations may be more like asterism, I cannot define line with another star thatdoes not belong to this constellation.
It can ensure that the constellation name and star names are visible.

2.This is to be satisfied and I will add it.
Comment with #
 Although in the civil, the process of the constellation substitution is gradual, it has been carried out earlier than the official announcement. 
The official announcement is that the time for substitution is fixed, ad 1752. The official book Complete Studies of Astronomical Instruments was published in 1752.
The official book confirmed that，officially deleted 6 old constellations and officially introduced 23 New Southern Sky Constellations.

gzotti:
Short translation question @sunshuwei : "Stomach" can mean "belly" or the organ which processes food. This may lead to misleading translations. Which one is it, please?

sunshuwei:
"Stomach"is the organ which processes food.
japanese_moon_stations also has the same constellation, as well as the same translation.
This is a metaphor, its real meaning is to use the stomach analogy granary, are loaded with food.

gzotti:
Thank you! Next request: please change all strings like "Heap of Corpses(In Stomach Mansion)" and add a space before the bracket, like "Heap of Corpses (In Stomach Mansion)".

gzotti:
Another family of words with a few translation possibilities: chariot/battle chariot/five chariots. The usual understanding and translation for "chariot" is already "battle cart".  But "Chariot" out of context can also be a simple merchant's cart, or are these all fast military horse-drawn carts? What would be the difference between "chariot" and "battle chariot", and what kind are the "five chariots"?

sunshuwei:
Zhen（轸）, the original meaning of Chinese characters (thousands of years ago) is the Chariot Cross-Board, but now used to refer to the  complete chariot. It is not clear what kind of use the chariot is. This is an uncommon word and is now used only to denote constellations.

Zhen che（阵车）, the word is clearly a war-fighting chariot. This is a direct translation.

Wu che（五车）, directly means five chariot, but also some experts think it means” five chariots of Five Emperors” , Five Emperors is the five great emperors who lived in the legendary era of 4,000 years ago.

The three translations are from the Hong Kong Space Museum.
In fact these words written in classical Chinese are difficult to understand.
The cultural connotations of the name cannot be explained perfectly, because thousands of years passed, the connotations behind some names are controversial. These are only possible translations.

alex-w:
You may use comments for translators and context for words, of course.

gzotti:
OK, thanks. So I will "de-militarize" the "regular" chariot. 

Just to be sure, is "Zhen" allowed to have 4 wheels, or only 2? The "Zhen che" should have 2 wheels only, right? 

This cross-board seems to be a part of the construction, but I cannot find any clear description. Maybe you could provide an image (link?)?

The Chinese constellations are interesting but very difficult for outsiders, with the whole empire officials structure mirrored, and what seems to be old stories. Some vocabulary is not found via Google, this would benefit from some contextual information as Alexander has suggested. I admit, this makes your work even harder - but also more valuable!

alex-w:
I've merged current updates of files. What about central stars for constellations 063, 092, 150, 166, 170, 264?

gzotti:
@sunshuwei : I think there is a small typo. "Lichpin" should be "Linchpin", right?
What is the "Coiled Thong" really? Google has either stringy underwear or sandals with leather straps.

sunshuwei:
1. constellations 063, 092, 150, 166, 170, 264  were lost in the Ming Dynasty and they do not exist now.
2.Yes, Lichpin is a wrong translation and should be corrected.
3.The image of some chariots can be referenced. In ancient China, there were two wheels in a chariots.
4."Coiled Thong"this is a tool of punishment used to bind and tortured the prisoners.
Sorry, I didn't find this feature before, I thought pull requests was closed and I couldn't use it.
qq 20171210151459
qq 20171205005437
20071202_f1049eab5d476ccca2f99bj1qgezzivy

PR #36, opened 2017-12-12
actived tests that need X

asiniscalchi:
Travis run Xvfb.
Tests of Telescope Control Plugin are active now.


plugins/TelescopeControl/src/INDI/CMakeLists.txt
```diff
@@ -24,4 +24,4 @@ target_link_libraries(TelescopeControl_INDI
 
 target_include_directories(TelescopeControl_INDI INTERFACE ${CMAKE_CURRENT_SOURCE_DIR})
 
-#add_subdirectory(test)
+add_subdirectory(test)
```
cardinot: That's great, thanks.

But I think we should use the ENABLE_TESTING flag here...

Could you please replace this line by:
```
IF(ENABLE_TESTING)
    add_subdirectory(test)
ENDIF(ENABLE_TESTING)
```

PR #78, opened 2018-01-18
Czech sky culture was added.

miroslavbroz:
Dear developers,

I tried to create "Czech" sky culture, with the constellation lines derived from
classical Czech or Slovak books (see the respective description.en.utf8 file, please).
Of course, they are somewhat similar to Western ones, but there are notable differences.
We also use these constellations at the Observatory and Planetarium Hradec Králové,
in our digital projection system (Zeiss Velvet 4+1, 4K resolution, 12m Spitz Nanoseam dome).

I know this is a very small contribution, but I kindly ask you to consider it for the master branch.

Clear skies, Miroslav Broz


gzotti:
Dear Miroslav, 

Thank you for your offer. But from what I can see this is just a Czech/Slovak variant of modern Western constellations, taken from 1-3 concrete modern books. 

This is a case very similar to "Western (H.A.Rey)" (which provides the same 88 constellations just with different lines and names from one popular book), and should be formatted in the same way. 

- It would be better not to name it "Czech" (which implies some genuine traditional sky culture, with constellations at least in significant parts different in name and mythology from the canonized 88 IAU constellations; see "Romanian" for such an example, it has a mix of standard Western plus own tradition, and lots of genuine artwork), but "Western (Hradec Králové)" or so. (Is there a common abbreviation like "Hr.Králové", to make the label shorter?)  Or, if one author is really responsible, call it "Western (Hlad)" or so? (Sorry, there is not enough space in the GUI for 6 names.)

- constellation_names.eng.fab: I have not checked every name, but given that the names seem like just standard constellation names, I think it would be better to just use a copy of the standard file from "Western", this saves translators work: If you want to switch to Czech, the names will likely already have been translated, and most users would prefer to see the names in their respective configured translation. 

- The same is true for the star names: no need for e.g. separate "87937|_("Barnardova hvězda")". This is "Barnard's Star" from the standard list. Please just make a shortened list (your star selection) from the standard list, and only add Czech star names (with English translation in brackets to help translators; see Chinese for example) where they are really "genuinely Czech". If I want to see Czech names, I can just switch to Czech language for skyculture, and the existing translations for stars and constellations will be used.

You could of course add some information on Hradec Králové planetarium, a photo, web link, club info, etc., to show that this skyculture is the one you are using there to successfully explain the constellations.

miroslavbroz:
Dear Georg,

I tried to address the issues: (i) it was renamed to "Western (O.Hlad)" in a similar way as Rey's case;
(ii) I retained English names only because translations indeed work; (iii) I decided to drop star names
entirely as they were mostly translations.

Thank you for your recommendations.

Clear skies, Miroslav

PR #99, opened 2018-02-02
Update stellarium.appdata.xml

GitHaarek:
+ The component ID was not a valid rDNS because of missing {product} and was thereby failing `appstreamcli validate`. Fixed.
Relevant information: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic
+ Norwegian Bokmål (nb) was incomplete, added missing translation 
+ Added category tags
+ Added 4 more screenshots
+ Added faq and translation url
+ Minor reorganizing. Moved the url tags together, moved `<launchable type="desktop-id">` up under the component id tag

### DIFF
--- a/data/stellarium.appdata.xml
+++ b/data/stellarium.appdata.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <component type="desktop-application">
-  <id>org.stellarium</id>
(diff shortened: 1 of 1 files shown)


alex-w:
It's OK for me, but in the future please use transifex for translation - our stellarium.appdata.xml file is translatable as the stellarium-metainfo resource on transifex website.

PR #108, opened 2018-02-26
issue103

EuklidAlexandria:
StelDialog class now limits the maximum size to 0.8x screen size.
LocationDialog is now resizable, the aspect ratio of the LocationMap is preserved.

Maybe the resizing of the pixmap should be implemented in MapLabel class by overriding resizeEvent?
Would it be better to draw the map cursor directly on the pixmap through QPainter instead of a separate label in order to avoid the jumping while resizing?

### DIFF
--- a/src/gui/LocationDialog.cpp
+++ b/src/gui/LocationDialog.cpp
@@ -79,6 +79,11 @@ void LocationDialog::createDialogContent()
 	// We try to directly connect to the observer slots as much as we can
 	ui->setupUi(dialog);
 
+	//enable resizability
+	ui->mapLabel->setMinimumSize(0,0);
+	ui->mapLabel->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Ignored);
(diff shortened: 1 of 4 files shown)


alex-w:
Oops... I was wrong! Sorry.

EuklidAlexandria:
Should it be something like a scroll area or should it automatically zoom in when hovering over it with the mouse?

gzotti:
I think most users would expect some kind of scroll-zoom area which automatically shows slide bars when zoomed in, similar to Google Maps or other such online services. This should keep coordinate conversion reasonably simple. If possible, the zoom should focus (but not center!) on current mouse position, not just on current map center. I mean, when you scroll-zoom in with your mouse pointer close to top right, that area should be better visible, but the mouse position should not be put into center of map by a zoom operation. We will likely have to keep the texture map at 2048x2048 to keep weak computers happy, so the useful zoom range will be limited in any case.

PR #110, opened 2018-03-03
Planets hips

alex-w:


gzotti:
Incredibly great feature! However, a few observations:

After fixing #111, we have the DSS button again. I think this should toggle the "currently active survey", not necessarily switch on DSS if some other survey is active.
Also, some surveys are very bright. There should be a brightness slider in the Surveys GUI. Either one slider, or even one per survey, and you can remember all the settings somehow? These should be stored immediately IMHO. 
Maybe the survey layer should be mixed with the Milky Way?
I have read about local files. Great! I know a photographer who made a survey project. I proposed the Toast system, is healpix better? Please, document (or at least link) where and how one can make a survey dataset.

PR #123, opened 2018-03-25
Macbook touchpad usability fix.

riodoro1:
The master version of stellarium since commit [83072850](https://github.com/Stellarium/stellarium/commit/830728504b2f799deeb436a69ca9808ab22dc267) had an issue on my MBP retina which made the program quite unusable. 

From my investigation it seems that the touchpad on these machines has a very fast refresh rate and due to the way the new display system was implemented every mouse event made the window redraw itself as fast as possible. This left no time to process more events resulting in a very laggy panning of the main view or user interface interaction. 

In order to fix the problem I have made the `StelMainView` respect the maximum fps property it already had and also set the default max fps to the value of 60 which is much more reasonable then 10000. Now, even though my machine is not capable to render stellarium in 60fps the delay between redraws makes the experience acceptable.


gzotti:
Thank you! @guillaumechereau tried to fix that same issue this week, I hope he can compare and commit the best solution.

guillaumechereau:
The commit of Riodoto is actually done on top of my previous fix so I don't know how much better it is cause I haven't tested yet.  It seems reasonable to use the maxfps in the singleShot call, and also to set it to 60 by default, so I am OK with the change and I think we can merge that.  Was there any reason why the max fps was set to 1000 before?

alex-w:
@guillaumechereau are you meant 10000 or 1000? 1000.0/getMaxFps() - it's conversion milliseconds into seconds, 10000 - "unlimited" FPS

guillaumechereau:
No I mean, why was maximum_fps in default_config.ini set to 10000, instead of 60 like what the commit from @riodoro1 changes it to?

gzotti:
60fps is the rate you can usually reach in sync'ed mode, faster fps requires turning off sync in the GPU settings. There are times when you want faster fps if possible (e.g. on large screens flicker can be seen more strongly), but it requires good hardware (high-end GPUs). I have seen about 140fps, 1000 may be a bit optimistic (but hey, give NVidia another year or two?) but the setting is a best-effort setting anyhow.

Keeping default value at 10.000 to signal "no limit" is OK and should not make a difference. Setting default to 60fps will require additional frightening steps for the user (config.ini editing) to set it higher.

I understand from QTimer docs that setting 0 (as Guillaume did) is a signal to "draw immediately at end of the event loop"). The previous solution (which however @guillaumechereau  has fixed) may have operated as @riodoro1 has described (immediate redraw, causing delays). So please one of @riodoro1 or @guillaumechereau should check whether an additional small fps-based delay as introduced by @riodoro1 makes any difference.

gzotti:
What we want to achieve is a delay in the order of 1 frame, right? Wouldn't it be more correct to use "current fps" as base value here?

guillaumechereau:
I don't think, because this line is supposed to work for the transition back to maxfps after a ui event, so we should use maxfps I think.

gzotti:
Ah yes. Was looking on the other line below. Then please merge this if you see the improvement.

riodoro1:
Won't there be a problem with people who have their preferences from the previous version? Their max fps will still be at 10000 and so they will se no improvement on a mac. Also looking at the code now I think the actual fps will never reach the maximum since to the timer delay the usual overhead of drawing and updating is added and this will lower the fps. For example If I set max fps to 18 (something my machine handles fine) the actual fps during movement is around 12.4.

riodoro1:
As I've said in my previous comment the solution with delaying the update caused the max fps to never be reached (by quite a wide margin I'd say). 
This made me realise there is a different solution to simply change the interval of the current timer if the max fps is needed and not act "behind its back".
This additionally eliminates the need to lower current default max fps because the timer event won't be able to redraw faster then the computer can catch up to (because of the `updateQueued` flag), this effectively means that the current default maximum fps setting is unlimited as @gzotti rightly has suggested it should be.
I have updated the branch with the changes.

guillaumechereau:
This looks fine to me.

Issue #102, opened 2018-02-11
screenshots under macOS High Sierra


tgs121346:
I upgraded my Imac's OS from Yosemite to High Sierra, and screenshots no longer work properly,
I just get the upper left quadrant of the screen.
![stellarium-003](https://user-images.githubusercontent.com/36384336/36078605-d793c03a-0f2d-11e8-90db-440075582c55.png)

### Actual Behaviour
attached
[log.txt](https://github.com/Stellarium/stellarium/files/1714849/log.txt)

### System
Hardware Overview:

  Model Name:	iMac
  Model Identifier:	iMac15,1
  Processor Name:	Intel Core i7
  Processor Speed:	4 GHz
  Number of Processors:	1
  Total Number of Cores:	4
  L2 Cache (per Core):	256 KB
  L3 Cache:	8 MB
  Memory:	16 GB
  Boot ROM Version:	IM151.0211.B00
  SMC Version (system):	2.23f11
  Serial Number (system):	D25QC0AXFY14
  Hardware UUID:	6167D67A-8789-53B8-A749-BEB93233C17B

AMD Radeon R9 M295X:

  Chipset Model:	AMD Radeon R9 M295X
  Type:	GPU
  Bus:	PCIe
  PCIe Lane Width:	x16
  VRAM (Total):	4 GB
  Vendor:	AMD (0x1002)
  Device ID:	0x6938
  Revision ID:	0x0000
  ROM Revision:	113-C773AA-731
  VBIOS Version:	113-C773A1-014
  EFI Driver Version:	01.00.731
  Metal:	Supported, feature set macOS GPUFamily1 v3
  Displays:
iMac:
  Display Type:	Built-In Retina LCD
  Resolution:	5120 x 2880 Retina
  Framebuffer Depth:	30-Bit Color (ARGB2101010)
  Main Display:	Yes
  Mirror:	Off
  Online:	Yes
  Automatically Adjust Brightness:	No

### Logfile 
attached

gzotti:
Another 4k or HiDPI problem. Unfortunately no developer has a screen with higher resolution than FullHD. leaving us unable to fix it. (Like shooting into the fog...) Sorry for the inconvenience. Maybe your OS allows some other screenshot mechanism?

tgs121346:
Just to be sure we are on the same page, by screenshot I mean the .png
file generated by Stellarium in response to a cmd-s keystroke.

gzotti:
[...]

PR #141, opened 2018-04-24
Hires screenshot

gzotti:
To continue #102: This is not finished yet, but I am a bit stuck, maybe @guillaumechereau or @xalioth can share some insight. 

I want to make high-resolution screenshots. The various parts of the scene, gui, projection, etc. have to be resized, and certainly I have overlooked something again. I cannot make the button bars appear in the low edge of the screen. Also I don't like the hackish projection parameter patching, is there a more elegant solution?

These hi-res screenshots would be nice to have. I think it is acceptable to have small dialog windows and small text, usually these big images are intended for text/guiless poster prints. Panoramas 16384 pixels wide? No problem on better machines.  Or at least FullHD screenshots on netbooks. Maybe text should be resized? I have not played with the effect of pixelRatio, maybe this helps here?

### DIFF
--- a/guide/app_config_ini.tex
+++ b/guide/app_config_ini.tex
@@ -502,11 +502,14 @@ date\_display\_format & string & date display format: can be \emph{system\_defau
 \begin{tabularx}{\textwidth}{l|l|X}\toprule
 \emph{ID}                 & \emph{Type} & \emph{Description}\\\midrule
 invert\_screenshots\_colors & bool   & If \emph{true}, Stellarium will saving the screenshorts with inverted colors.\\%\midrule
-restore\_defaults           & bool   & If \emph{true}, Stellarium will restore default settings at startup. 
-                                       This is equivalent to calling Stellarium with the \command{--restore-defaults} option.\\%\midrule
 screenshot\_dir             & string & Path for saving screenshots\\%\midrule
+screenshot\_custom\_size    & bool   & \emph{true} if you want to specify particular dimensions next:\\
+screenshot\_custom\_width   & int    & custom width of screenshots. Max.\ available size is hardware dependent!\\
+screenshot\_custom\_height  & int    & custom height of screenshots. Max.\ available size is hardware dependent!\\\midrule
 version                     & string & Version of Stellarium. This parameter may be used to detect necessary changes in config.ini file, do not edit.\\%\midrule
 use\_separate\_output\_file & bool   & Set to \emph{true} if you want to create a new file for script output for each start of Stellarium\\%\midrule
+restore\_defaults           & bool   & If \emph{true}, Stellarium will restore default settings at startup. 
+                                       This is equivalent to calling Stellarium with the \command{--restore-defaults} option.\\%\midrule
 ignore\_opengl\_warning     & bool   & Set to \emph{true} if you don't want to see OpenGL warnings for each start of Stellarium.\\%\midrule
 check\_requirements         & bool   & Set to \emph{false} if you want to disable and permanently ignore checking hardware requirements at startup. 
                                        Expect problems if hardware is below requirements!\\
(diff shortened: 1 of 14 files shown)


guillaumechereau:
As far as I know it's hard to make large size screenshots, and in the current state of the code will involve a few hacks.  I am not sure how to handle the UI rendering properly since this is done by Qt on top of the OpenGL rendering and the code here is already quite complex.

guillaumechereau:
Just tested the branch, I confirm the problem with the bottom bar (on linux).  There is also a small bug in the config dialog: the screenshot size cannot be typed properly, I have to move the cursor after each stroke.  Not sure why.

gzotti:
Config bug: Indeed, I only ever changed the thousands... I have an input mask attached which allows only digits. Probably this is connected. Else, we could make a Spinwheel instead of textboxes, but it's a long spin up to 16000... (Maybe steps of 100 pixels? But what if you want 2048 px for a reason?)
Ah, it works with a QIntValidator instead, will fix that.

Still, the missing bottom toolbar is a mystery to me. The StelGui->forceRefreshGui() should do just that. Or is that a signal to redraw on next render pass, and the screenshot making is just one pass, so it never redraws? I call setGeometry() to avoid the bars drawn inside the image. But it seems forceRefresh() does nothing now.

alex-w:
This feature may cause OpenGL error "Request for more GPU memory than is available" (http://nvidia.custhelp.com/app/answers/detail/a_id/3553), so, maybe we should check number of GPU memory to set the maximum size of screenshot and reduce wanted size to possible values?

gzotti:
Actually I test max framebuffer size, so I thought that's enough. Yes, other tests should be added and image size limited as required.

alex-w:
After some playing with this new feature I found source the "invisibility" of the bottom bar. The main problem is changing geometry of SkyGui without changing geometry of  their parent - StelGui - so in this case all objects outside parent rectangle are cropped.

gzotti:
Good idea, but I see no geometry in StelGui. If I tweak StelGui's parent, StelMainView, I think this is limited by the OS to your screen's resolution (?). Or were you successful and like to show your code?

gzotti:
Yes, this is the problem. I have a clipped button bar with a screenshot of 1220 height (1920x1200 monitor) But is there a solution?

gzotti:
Alright. it's the StelRootItem configured to clip its children... We are getting further!

gzotti:
Yes, that was it! 8192^2 screen shot looks terrific. 16384^2 should work, but only has Milky Way...

gzotti:
I saw that the StarMgr defines star sizes in pixels. Had to add a scale value, but not sure whether this should be applied like now or better applied somewhere else. Result of a large star map looks really nice. It could be used to increase font or button bar size if we want that, or meteor shower radiant symbols, or tweak the distance of the gridlines. It's a matter of taste to change those. The application of the huge images is likely on posters, so adjusting GUI size is not an issue.

@alex-w can you please tell me under which circumstances you have that memory issue?

gzotti:
I see Travis build fails with an unknown GL constant. Actually I don't really do anything else than log the retrieved result, and I don't fully understand the figures, I cannot reproduce problems mentioned by @alex-w . Either disable this test when compiling with clang, or completely?

alex-w:
@gzotti compilation of this branch is broken on OS X

gzotti:
I know, see my previous comment. The break concerns a value request about free GL memory, which apparently does not exist on Macs. The whole test block could be removed or just commented away on OSX. 

Please define "oversized". I had no problems with maximal sizes 16k^2 on a card that qualified itself as capable to do that, and even MESA made 8192^2.  I also had no problems making 8k^2 on my old Ubuntu notebook (Geforce 9800), although it reported no values on the AMD test.

alex-w:
I think it's ready to merge after exclude or resolve problematic code for OSX. I'll try to get memory error again, but I may test in at work only.

PR #142, opened 2018-04-28
Make sure that splash screen shows up immediately

10110111:
We want the splash screen to show its image right after it's show()ed. But on Linux X11 paint event often doesn't come until ~30 ms passes, or sometimes more, so that `app.processEvents();` doesn't show anything. In this case, from user's point of view, the splash screen is created at the same time as main window, which is useless. Thus it's better to make sure that paintEvent handler has been called, and only then to proceed with initialization.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -51,6 +51,7 @@
 #include <QTextStream>
 #include <QTranslator>
 #include <QNetworkDiskCache>
+#include <QThread>
 
 #include <clocale>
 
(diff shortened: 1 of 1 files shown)


gzotti:
Cannot test until next week, but I confirm splash screen sometimes does not show up on Linux (only a "frozen" square of desktop while console output is getting refreshed around it). Thanks for taking care about this.

10110111:
@alex-w
I developed this on Ubuntu 14.04 x86_64 machine with manually compiled Qt 5.5.1. Video chip is an Intel Haswell built-in GPU. I can also see the problem on an LFS x86_64-kernel/x86-userspace machine with GeForce GTX 750 Ti.
For a purer test case, see [this StackOverflow question](https://stackoverflow.com/q/50059520/673852). If you comment out the loop from there, you'll not see the `Loading...` message — only `Finished` after 5 seconds. Same if you replace the loop with `processEvents()` called once, as currently done in upstream Stellarium.

PR #154, opened 2018-05-13
Fix leak of Bayer pattern texture in StelPainter

10110111:


gzotti:
Excellent, thank you! Yes, StelPainter is created frequently.

PR #159, opened 2018-05-18
Improve proportional size textures for large DSO

gzotti:
Also fix rendering of either proportional or outlined DSO.

This adds a few thinner lined DSO textures which look better when scaled up.


gzotti:
I just saw that label placement for large objects is very far away, e.g. North America or Pelican. This should also be fixed. Maybe also depending on outlines vs. symbol.

alex-w:
Sorry, I don't see serious changes for visualization of markers. Maybe changing of the steps will be good here (I meant 35 and 75)? 

Other possible way to resolve problem of proportional hints - draw the hints through OpenGL primitives. I played with it some time ago and I don't like very thin lines for "hint templates".

gzotti:
The changes are rather subtle and should not be too distracting to the eye. Maybe compare screenshots of some rich area in Sco/Sgr. Previously, the hint icons were simply upscaled, and at some point the line thickness becomes ugly as reported in https://bugs.launchpad.net/stellarium/+bug/1485186 / https://bugs.launchpad.net/stellarium/+bug/1507207. Zoom in on a large open cluster now. At some point you see 8 smaller, zooming further in, not 8 but 12 even smaller, sharper dots. (I could change that to have 8..10..12) On large diffuse nebulae (bright&dark) you should also see the difference, though.
Of course, OpenGL lines would be the more elegant solution. Squares would be easy, but ellipses or boxes with fuzzy round edges? I'd leave that to some (as yet unknown) shader hero.

PR #167, opened 2018-05-28
Fixed incorrect line counts in constellationship.fab files.

danjoplin:
A handful of the skycultures constellationship.fab files had an incorrect line count, this updates those files.


gzotti:
Indeed, thanks a lot!

PR #169, opened 2018-05-29
Rise, transit and set times

alex-w:
Added method for calculation of rise, transit and set times for celestial objects (low precision, without interpolation). Information box and AstroCalc tool was updated due the new method.

### DIFF
--- a/src/gui/AstroCalcDialog.cpp
+++ b/src/gui/AstroCalcDialog.cpp
@@ -4104,13 +4104,13 @@ void AstroCalcDialog::selectWutObject()
 
 					Vec3f rts = newSelected[0]->getRTSTime(core);
 					QString sRise = QChar(0x2014);
-					if (rts[0]>=0.f)
+					if (rts[0]>-99.f && rts[0]<100.f)
 						sRise = StelUtils::hoursToHmsStr(rts[0], true);
 					QString sTransit = QChar(0x2014);
 					if (rts[1]>=0.f)
 						sTransit = StelUtils::hoursToHmsStr(rts[1], true);
 					QString sSet = QChar(0x2014);
-					if (rts[2]>=0.f)
+					if (rts[2]>-99.f && rts[2]<100.f)
 						sSet = StelUtils::hoursToHmsStr(rts[2], true);
 
 					ui->labelRiseValue->setText(sRise);


gzotti:
Good idea to include this!
Please align the infostring with Mean&Apparent sidereal time (when using formatting output).

I see a mismatch around the decisive declinations around "object never rises" and actual position above math. horizon. Will look into the code later this week if necessary. Nice to have a difference with/without atmosphere!

English strings: "This object never rises"/"This object never sets" [or just "Circumpolar - never sets"]. 
Maybe write "Too far south - never rises" (resp. "Too far north - never rises" on locations with southern latitudes)
Maybe give upper and lower transits for circumpolar objects?

In a later step, of course planets&moon should show 3-date interpolation (see Meeus, Ast. Algorithms) and even horizon/landscape sampling (with 2 results: math. horizon and landscape)!

As requested, some "go to next rise of selected object" (with/without landscape sampling) scripting/hotkey action would be splendid. The perfect system can say "go to next day when selected object is <1.2> degrees above  <math horizon|photo panorama landscape> <after rising|before setting>."

alex-w:
> Seems the use of getJ2000EquatorialPos() for stars retrieves the star position without proper motion!

Oops... I've fixed it. The proper motion of stars is old and big problem... :'(

>  If possible, 3-date interpolation for planets would be very valuable!

First big question is... where are calculate the data? The second question is... triple updating the screen for Solar system objects. It's not very effective IMHO and due this reason I do not added RTS calculations a long time.

alex-w:
Maybe using term "Upper culmination" for transit will be better?

gzotti:
"Upper culmination" makes sense only for circumpolar objects.

gzotti:
It might make sense to store 3 extra positions for the planets: JDm1 (yesterday, 0UT), JD0 (today 0UT), JDp1 (tomorrow 0UT). Their computation can be triggered when the current date does no longer match JD0, i.e. only once per day. Then computation may be possible using the algorithm from Astr. Algorithms. I have not again looked into details, though.

alex-w:
Georg, in the theory I've prepared RTS data for interpolation, but I have doubts about the correctness of my approach to solving the problemю Could you check the code (esp. getNextRTSTime() & getPreviousRTSTime()!)?

gzotti:
This is just the first fix. Must go into details later.

gzotti:
@alex-w apart from some cleanup this is now a solution which seems to work accurately. However, the times are changing slightly over the day. And I don't see why for locations outside Earth the infostring text is formatted black?

alex-w:
@gzotti I don't see problem with text color

gzotti:
@alex-w Go to Mars and activate "Use Formatting Output". RTS lines are black. Something with the </table> closing too early in the previous infostring group?

alex-w:
Ah! Fixed!

PR #170, opened 2018-05-30
Korean translation for landscape description

ysjbserver:
As same for title, I added translated description files for landscape.

Am I doing right? I'm first time at github...


alex-w:
Thanks a lot!

PR #172, opened 2018-06-02
Added Korean translation

psj8252:
Shadow mapping can not be used on your hardware, check logs for details

Provides more accurate lighting by calculating it per-pixel instead of per-vertex.

Required for bump mapping and shadows!

Determines how the shadows are filtered, in increasing quality from top to bottom.

### DIFF
--- a/po/stellarium/ko.po
+++ b/po/stellarium/ko.po
@@ -2,14 +2,13 @@
 # Copyright (C) YEAR Stellarium's team
 # This file is distributed under the same license as the PACKAGE package.
 # FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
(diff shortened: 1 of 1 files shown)


alex-w:
Seems translators already finished translation those strings on Transifex.com: https://www.transifex.com/stellarium/stellarium/translate/#ko/stellarium

gzotti:
I think in general translations should go over Transifex please, this saves some maintenance work for Alexander. (Unless you have a massive collection of 500+ changes or a completely new language. Of course separate files like any description.ko.utf8 still go via PR.)

psj8252:
Aha, Ok.
I will try that.
Thank you!

PR #173, opened 2018-06-02
Cleaning Korean Star Names

ysjbserver:
I made small fix to Korean skyculture star names. Previous one used unusually long vowel, which don't used on Korean to English writing. I removed overlapping vowels to clear it.


gzotti:
@ysjbserver do these names actually mean something? If possible, to improve understanding of this skyculture, can you please add translations to English? See e.g. dakota sky culture for formatting.

ysjbserver:
@gzotti Great idea! I'll do that as well as adding more star names of Korean.

PR #177, opened 2018-06-05
Ignore warning C4819 in MSVC.

gongminmin:
The locale of my computer is set as Chinese. Building Stellarium with MSVC resulting a lot of warning C4819:

> The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss.

Turn off C4819 to ignore this problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -149,7 +149,7 @@ IF(WIN32)
           # C4305: type truncation
           # C4351: "new" behaviour, member array default initialization. Required since at least C++98, but funny MSVC throws a warning.
           # C4996: deprecated POSIX names (used in zlib)
-          SET(STEL_MSVC_FLAGS "/wd4244 /wd4305 /wd4351 /wd4996")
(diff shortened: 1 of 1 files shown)


gongminmin:
The error when compiling with g++ is "requests 2.18.4 has requirement urllib3<1.23,>=1.21.1, but you'll have urllib3 1.23 which is incompatible." It should be a building infrastructure problem.

gzotti:
The warning says "Save the file in Unicode format to prevent data loss.". This should be the better solution IMHO.

I cannot comment urllib3 1.23. Is this particularly incompatible (else 1.23>=1.21.1 should be fulfilled.) Whose building infrastructure is affected?

alex-w:
@gzotti it's on Treavis-CI

alex-w:
@gongminmin Could you give list of files, who give warnings on your side?

gongminmin:
@alex-w , I have another fix on this, better than just turn the warning off.

@gzotti , sorry I didn't say it clearly. The urllib3 1.23 error is in Travis-Cl.

gongminmin:
Abandon this PR because of #178.

PR #183, opened 2018-06-09
Enables USE_FOLDERS in CMakeLists.

gongminmin:
Groups projects into virtual folders in IDE.


alex-w:
I think names of real and virtual folders should be synced

gongminmin:
@alex-w , sounds good. I'll push a new iteration tomorrow, sync their names.

PR #184, opened 2018-06-09
Some minor changes to increase code readability

gongminmin:
1. No need to call data() before using a QSharedPointer,
2. Use the return value of vector::erase, instead of counting the index,
3. vector::empty() to replace begin() == end().


guillaumechereau:
Looks OK to me.

Just as a note: in general I agree that it is better not to use data(), but there are a few cases where this is needed to optimize the code (for example in Planet::getHeliocentricPos, the data() calls are important).

PR #193, opened 2018-06-13
Fix fullscreen toggle on macos

guillaumechereau:
The problem was that the mouse press event triggers the fullscreen
switch, and the mouse release event is then lost by Qt.  So the user has
to click once on the screen so that the mouse event get processed
properly again.

The fix is was to configure the fullscreen button to trigger on mouse
release event instead of mouse press.  This changes the behavior a
little bit.


gzotti:
I think most GUI toolkits and systems teach in their examples to trigger their actions on button release (so that you can change your mind and move the pressed mouse outside of the trigger area). This never appeared to me, but should we change this globally? Else, at least this change is approved from my side.

guillaumechereau:
@gzotti: yes I also think it would make sense to change the behavior to 'trigger on release' for all the toggle buttons.

PR #196, opened 2018-06-15
C++11 auto to replace complicated hand written types.

gongminmin:
Continue to work on using C++11 features to simplify code.
1. Use auto instead of full type name of iterators,
2. More ranged-based for.


guillaumechereau:
I am not totally against that, so I am fine if this get merged, but to be honest I don't think this is such a needed patch.  I am afraid c++ auto will be used to hide ugly parts of the code instead of trying to clean it up.

Using c++11 range for loop is good, this does make the code easier to read.

PR #197, opened 2018-06-15
Remove gVector, use Vec3d instead.

gongminmin:
Saves a lot of dynamic memory allocations.


guillaumechereau:
Thanks, that a nice cleanup!  I tested briefly on linux, seems to work fine.

PR #202, opened 2018-06-17
Remove the null pointer check after new.

gongminmin:
New no longer returns null even when out of memory.


guillaumechereau:
I am OK with that.

PR #203, opened 2018-06-18
Check struct size by using static_assert.

gongminmin:
Use compile time assert instead of runtime assert.


guillaumechereau:
That's good.

PR #209, opened 2018-06-22
Create Planetary Alignment

hickny:
All major planets will be visible on October 18, 2018 just after sunset from Sydney Australia.

### DIFF
--- a/scripts/Planetary Alignment
+++ b/scripts/Planetary Alignment
@@ -200,4 +200,3 @@ core.setDate("now");
 core.clear("natural");
 LandscapeMgr.setFlagLandscape(true);
 LandscapeMgr.setFlagAtmosphere(true);
-clearLabels(ids);


alex-w:
The last block you may change to `core.goHome();`.

The text is overlapping infostring and this is not very good.

alex-w:
Would you describe by simple words what exactly you want to get by this script?

gzotti:
This event is not a conjunction. It is a nice view, sure, but not a conjunction.
I also don't know why it must be the 18th, you see these objects +/-5days or so around that date.

scripts/Planetary Alignment
```diff
@@ -0,0 +1,203 @@
...
+//
+//
+
+// The following two functions were copied from a script by Schdefon
+
+function blockText(textArray)
+{
+	var fontSize = 30;
+	var lineSep = fontSize * 1.5;
+	y = 250 + (lineSep*textArray.length);
```
hickny: Change the 250 to 420 and the problem should be fixed.

gzotti:
The final clearLabels(ids); must be removed to avoid a crash.

I recommend to place this script example to the website collection of script examples.

hickny:
How do I place script to website collection?

PR #234, opened 2018-07-29
Improvements to the appdata file

cosimoc:
Add OARS rating tags and releases tags, to improve the application presence in application distribution websites and programs such as gnome-software and Flathub.


cosimoc:
See also https://github.com/flathub/flathub/pull/511

PR #274, opened 2018-09-18
Gz fix lunar eclipse

gzotti:
We recently fixed the mis-projection of Earth's shadow during lunar eclipse. However, the colors and brightness are rather ugly. 

0 18 2-lunareclipse klein

This branch brings a better earth shadow texture and brightness balancing. 

gz_fix-lunar-eclipse klein


Atque:
There seems to be a problem with the brightening of the rim when the moon is at perigee. Compare these two screenshots from the July 2018 eclipse (micromoon) and the September 2015 eclispe (supermoon).
lunar_eclipse_bug

gzotti:
There is still room for improvement as I also have seen. I optimized for this July's moon, but should find a sharper criterion, not just elongation, for this. Must compare with shadow size or percentage of occultation. Not this week, but it should happen at some time. Just compare this to the previous state. 
BTW Another thing to fix then is the visual glare estimate, there seems to be a hysteresis when going back and forth about the time when more stars become visible.

PR #498, opened 2018-10-08
Mark functions static and const

rikardfalkeborn:


alex-w:
What exactly you want to get?

gzotti:
There are some definitely non-static functions like Landscape::getTexturePath. Some const or static changes seem OK, though. A merge will likely break things, these have to be checked per case.

rikardfalkeborn:
> There are some definitely non-static functions like Landscape::getTexturePath. Some const or static changes seem OK, though. A merge will likely break things, these have to be checked per case.

I'm not following, are you saying ```Landscape::getTexturePath``` can not be static? I'm probably missing something, but can you elaborate why that's the case?

gzotti:
Oops, case of bad memory. This should indeed work.

Issue #358, opened 2018-09-19
Add solar equator and Laplace invariable planes to Celestial Sphere


alex-w:
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1639336

Converted from blueprint:

Request addition of a line in constellation settings to represent the plane of the solar system (not the ecliptic), which is a line projected on the heavens through the solar equator. In addition, request a similar line drawn to reflect the barycenter (center of mass) of the solar plane, also known as the Laplace plane (see wiki). Both are used by amateur astronomers searching for undiscovered orbiting bodies, as the assumption is often that the majority of orbiting bodies in our solar system lie near these planes. My apologies if I made this request incorrectly.
http://en.wikipedia.org/wiki/Invariable_plane

alex-w:
The solar equator already added in ArchaeoLines plugin

gzotti:
Alex, I am afraid no. The solar equator is a bit off from the ecliptic plane (which is the relevant plane in cultural astronomy and ArchaeoLines) and is not relevant to astronomy before around the 19th century.

alex-w:
The solar equator is called "Current Sun" in the plugin

gzotti:
no, this is the daily track of the sun (just its current declination!) There is also "Current Moon".

alex-w:
Yes, of course. Definition: The solar equator is the latitude on Earth at which the Sun is directly overhead; where the sun is vertically above at midday.

gzotti:
Wel, OP wanted this as plane of the real Solar Equator. Nothing to do with earth. Yours I would call solar zenith passage.

alex-w:
Maybe not :) I've asked Roger for comment of his proposal.

rmorrison0324:
Howdy all.  Rodger here.  Sorry for the delay.  I had trouble getting set in GitHub.

[...]

Issue #357, opened 2018-09-19
Orientation of Jupiter, Uranus and Venus may be incorrect or displaying az/alt wrong


alex-w:
_Original report by Steve:_ https://bugs.launchpad.net/stellarium/+bug/1637417

If you look at this webpage
http://aa.quae.nl/en/reken/zonpositie.html
There is a table under section "15. Comparison with planetary programs" (In middle of the page)which shows comparisons for the azimuth and h altitudes between diff programs.
I went through both tables and noticed Venus, Uranus and Jupiter for stellarium 0.15 varied largly from results of what other programs were showing which generally matched each other by about a degree or 180deg which was predictable.
I verified that stelarium .15 showed the results the table showed.

I also noticed when comparing Jupiter Az/alt from version.14.2 Stellarium and .15 stellarium the results were different for az/alt which seemed odd given that other planets matched between version .14.2 and .15
So it could be just a az/alt info display error or something more serious.

gzotti:
The table is matched pretty close now, sub-degree differences hopefully only come from refined models. 
For Jupiter, rotation rate is a bit difficult. We can easily stick to system II, but this would bring the GRS out of sync with reality, so we adjust the rotation artificially.

Issue #347, opened 2018-09-19
Rotation of Moon inexact


alex-w:
_Original report by Dave Chapman:_ https://bugs.launchpad.net/stellarium/+bug/499971

When simulating Moon phase in millennia far in the past or future (say 2000 years), the face of the Moon is rotated East or West, respectively, relative to its tidally locked orientation. For example, when simulating the Venus-Jupiter conjunction of -1 June 17, the Moon is rotated about 40 degrees East so that Mare Crisium is totally hidden behind the limb. I have found the same bug in other programs, but not all.

```
2009-12-23T17:31:30
/Applications/Stellarium/Stellarium.app/Contents/MacOS/stellarium -psn_0_712878
Mac OS X 10.5
Compiled with GCC 4.0.1
Qt runtime version: 4.5.0
Qt compilation version: 4.5.0
Addressing mode: 32-bit
You look like a Mac user. How would you like to write some system info code here? That would help a lot.
 -------------------------------------------------------
[ This is Stellarium 0.10.2 - http://www.stellarium.org ]
[ Copyright (C) 2000-2009 Fabien Chereau et al ]
 -------------------------------------------------------
Writing log file to: "/Users/test/Library/Preferences/Stellarium/log.txt"
File search paths:
  0 . "/Users/test/Library/Preferences/Stellarium"
  1 . "/Applications/Stellarium/Stellarium.app/Contents/Resources"
Config file is: "/Users/test/Library/Preferences/Stellarium/config.ini"
Cache directory is: "/Users/test/Library/Cachesstellarium"
Sky language is "en_CA"
Application language is "en_CA"
Loading Solar System data ...
Loaded 38 / 38 planet orbits
Loading star data ...
Loading "stars_0_0v0_1.cat": byteswap 0_0v0_1; 5013
Loading "stars_1_0v0_1.cat": byteswap 1_0v0_1; 21999
Loading "stars_2_0v0_1.cat": byteswap 2_0v0_1; 151416
Loading "stars_3_1v0_0.cat": byteswap 3_1v0_0; 434064
Finished loading star catalogue data, max_geodesic_level: 3
navigation/preset_sky_time is a double - treating as jday: 2.45151e+06
Loaded 10051 / 13226 NGC records
Loading NGC name data ...
Loaded 222 / 222 NGC name records successfully
Loaded 88 / 88 constellation records successfully for culture "western"
Loaded 85 / 85 constellation art records successfully for culture "western"
[...]

PR #502, opened 2018-10-12
improved planet axes

gzotti:
AT LAST!

I want to improve the orientation and rotation of planets, especially the Moon. 
Best data for planet rotation come from the IAU WGCCRE09 report, and later updates of that series. However, these data are not compatible to the way Stellarium has computed rotations so far.

It seems this is now compatible to the rest of the world, at least when we have the appropriate data.

The old solution is still used if such data are not present.

There is still some need to cleanup, but functionally it seems good now!

Q: I had implemented Lunar libration from Astronomical Algorithms. This is now unused. Keep as reference or delete?

Related: #347, #357, #358 #877


alex-w:
@gzotti this is may suggest the solution for Moon’s orientation problem: https://github.com/pchev/skychart/blob/98b7ac40b660beb6acd33d3d183e401e9fe76388/skychart/cu_planet.pas#L1226

src/core/modules/NomenclatureItem.cpp
```diff
@@ -227,6 +227,7 @@ QString NomenclatureItem::getInfoString(const StelCore* core, const InfoStringGr
 		QString description = getNomenclatureTypeDescription(nType, planet->getEnglishName());
 		if (getNomenclatureType()!=NomenclatureItem::niUNDEFINED && !description.isEmpty())
 			oss << QString("%1: %2").arg(q_("Landform description"), description) << "<br />";
+		oss << QString("%1: %2").arg(q_("Solar altitude")).arg(QString::number(getSolarAltitude(core), 'f', 1)) << "<br />";
```
alex-w, line 230: what about units?

gzotti: Ah sure. Added.

alex-w (approved):

gzotti:
I had implemented Lunar libration from Astronomical Algorithms before I did the internal (better) solution. This is now unused. Keep as reference or delete these files and the commented-away use in Planet.cpp? I now tend to remove them.

alex-w (approved):
Deleting "old" libration is OK for me

PR #512, opened 2018-10-16
21P/Giacobini-Zinner comet orbit update

vojtechhabarta:
Hello,
Firstly, thanks for this great project, it is very useful.

Recently I wasn't able to find 21P/Giacobini-Zinner comet in the sky using Stellarium and after some investigation I realized that Stellarium doesn't have comet position precise enough. In fact it displays the comet about degree and half away from real position.

I found current orbital data here: https://ssd.jpl.nasa.gov/sbdb.cgi?sstr=21P;orb=1 and when I update the orbital elements in my `C:\Users\me\AppData\Roaming\Stellarium\data\ssystem_minor.ini` file it works well.

So I decided to share this change but I am not sure if other properties should also be changed somehow. I also don't know if these data should be changed like this or if this file is somehow generated automatically from some source.


gzotti:
Indeed, as every observer hopefully knows, orbital elements need frequent updates. 

However, orbital elements of minor bodies are updated in Stellarium programmatically using the Solar System Editor plugin. This is a program feature that has been available for many years now, and is well described in the User Guide and online help.

vojtechhabarta:
Thank you for quick response. Now it was easy to update the comet orbit, no need to manually edit `ssystem_minor.ini` file.

The problem was that I didn't know about this possibility. After some googling I found that there is Solar System Editor but I didn't know it is able to download/import data. It would be nice if Stellarium had some central update system which would indicate that there is something to download - new program version, star catalogs, newer comet orbit data, etc. It would be also good if new Stellarium versions would contain latest data.

Thanks again

alex-w:
From the description of Solar System Editor plugin:

> An interface for adding asteroids and comets to Stellarium. It can download object lists from the Minor Planet Center's website and perform searches in its online database.

This plugin is available for using since version 0.10.6 and he has explanation within the inself GUI. Plus Chapter 12 of  Stellarium User Guide has detailed description for manage MPC data. Who reading the guide?

>  It would be also good if new Stellarium versions would contain latest data.

We cannot guessing what exactly user has in their Solar system data file and we may up-to-date the default list of this data, but this list will be use after resetting the current user data. One of possible solution the problem is adding tool to periodical updating the orbital elements for each asteroid and comete in the current list of minor objects for the Solar System Editor plugin.

gzotti:
Do you mean we (Stellarium team) would have to select, once per week or at least once per month, current orbital elements and distribute to the users like the Satellite plugin does? Good luck doing that, and make everybody happy by selecting just the right comets and asteroids.  Would that overwrite some users' own modified, carefully tweaked element lists? Would you delete outdated elements from comets no longer visible? Would a user be able to recreate a view of a comet or asteroid he has observed 4 years ago? (This needs elements from 4 years ago, when user probably last cared about them... - actually we would need another data entry "epoch" to allow multiple sets of elements per object.) Could we save old elements somewhere? Many details with lots of probable personal preferences would have to be solved here, which is likely why these issues have not been touched so far. A bit of semi-manual interaction should be acceptable. 

Maybe we should add a "Usage" popup on every program start? Or show 100 "tip of the day" FAQs until user disables it or has forcefully been shown and acknowledged to have read all tips at least once?

Issue #523, opened 2018-10-25
Grouped onscreen Information of selected object 


t4saha:
When all information is on then some info displays outside of the screen. Information may be grouped by type (General,  #Position, Orbital, Catalogs, Additional) and shown in separate tabs on screen as a button layer.
![screenshot from 2018-10-25 16-31-09](https://user-images.githubusercontent.com/34002960/47499852-b93c0f00-d87e-11e8-8d73-94cb1a655f52.png)

gzotti:
You seem to have either a too small screen or too large base font size, or both. Can you edit your config.ini:

     [gui]
     base_font_size                         = 13

And if you don't require some information, select what you can omit in the Information tab of the settings dialog (F2). On my FullHD screen, "All" information bits take about half the screen height.

Issue #519, opened 2018-10-25
Menu list slider weird behavior 


pacastro:
### Expected Behaviour
Drag slider up or down on menu list's and the list moves up or down accordingly.

### Actual Behaviour
Drag slider up or down, and menu list moves down or up for a bit and then the movement stop on the slider and the list moves on a weird way. See attached video, it's hard to describe. This problem happens in all the menu list's.

[Screen Recording 2018-10-24 at 10.11.59 PM.zip](https://github.com/Stellarium/stellarium/files/2512895/Screen.Recording.2018-10-24.at.10.11.59.PM.zip)

This is using V0.90.0.16362, if I try with V0.18.2 there is no problem with the slider.

### Steps to reproduce
Using V0.90.0.16362, open any menu that has a list to choose from: location window for example (as the attached video), go with mouse pointer to the slider just right to the list of countries, press mouse button to drag the slider up or down and after a few seconds of dragging the weird behavior starts.

This does not happen with V0.18.2, slider works ok.

### System
* Stellarium version: Stellarium 0.90.0.16362
* Operating system: OSX 10.14
* Graphics Card: Intel HD Graphics 5300, 2.1 INTEL-12.0.34

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/2512918/log.txt)

gzotti:
This is the "kinetic scrolling" which was introduced recently. Don't drag the slider, drag the text inside, like on a smartphone/tablet. 

We may add a config option to the GUI. For now, edit config.ini and set

    [gui]
    flag_enable_kinetic_scrolling          = false

pacastro:
Yep, dragging the text inside works. 

Its nice but slow for moving long list's, grabbing the slider and dragging it fast down/up is faster for searching something you don't remember the name for typing it in...

pacastro:
Edited config.ini and seted kinetoc scrolling flag to false....

Now the slider works OK when grabbed and dragged, AND can drag the text inside the list also.

alex-w:
@gzotti GUI option may be good solution here, but seems introducing this option with state on-the-fly switch will be problematic.

gzotti:
I found a way to do it. :-)

Issue #424, opened 2018-09-26
Option to select font and size


alex-w:
_Original report by Khalid AlAjaji:_ https://bugs.launchpad.net/stellarium/+bug/1505073

font and size can be specified in the config.ini. Can be there an option in the configuration window to select the font and size from the system.

The reason for this request now is that some fonts are behaving better than the default regarding the qt bug: https://bugreports.qt.io/browse/QTBUG-46244 affecting Stellarium panels in Arabic language and reported here: https://bugs.launchpad.net/stellarium/+bug/1498616

Another reason is cosmetic. Some fonts look better.

gzotti:
I have started this. However, some fonts (esp. on the GUI panels) look bad when scaled at runtime, needs some separation.

PR #527, opened 2018-10-26
GUI: fontsize,  kinetic toggle and other improvements

gzotti:
READY TO MERGE! 

A (hopefully) short-lived branch which should provide a few GUI fixes: 
 - [x] Base font size (Scaling for GUI dialogs has also been separated, see discussion below.)
 - [x] toggle for kinetic scrolling
 - [x] Move the color buttons to the left of the respective checkboxes. 
 - [x] Loadable font. 

This relates to #519, #523 and #424

### DIFF
--- a/src/gui/ConfigurationDialog.cpp
+++ b/src/gui/ConfigurationDialog.cpp
@@ -333,6 +333,11 @@ void ConfigurationDialog::createDialogContent()
 	connectBoolProperty(ui->kineticScrollingCheckBox, "StelGui.flagUseKineticScrolling");
 	connect(ui->fontSelectPushButton, SIGNAL(released()), this, SLOT(selectAppFont()));
 
+	ui->fontComboBox->setWritingSystem(QFontDatabase::Any); // TODO: Make this Locale-dependent? Or show another combo to preselect?
+	ui->fontComboBox->setFontFilters(QFontComboBox::ScalableFonts | QFontComboBox::ProportionalFonts);
+	ui->fontComboBox->setCurrentFont(QGuiApplication::font());
+	connect(ui->fontComboBox, SIGNAL(currentFontChanged(QFont)), this, SLOT(setAppFont(QFont)));
+
 	// Dithering
 	populateDitherList();
 	connect(ui->ditheringComboBox, SIGNAL(currentIndexChanged(const QString&)), this, SLOT(setDitherFormat()));
@@ -1710,4 +1715,16 @@ void ConfigurationDialog::selectAppFont()
 	}
 	else
 		return;
+}
+
+
+void ConfigurationDialog::setAppFont(QFont font)
+{
+	QFont oldFont=QGuiApplication::font();
+	int size=oldFont.pixelSize();
+	// MAYBE EXCLUDE SYMBOL FONTS?
+	font.setPixelSize(size); // set to old size!
+	QGuiApplication::setFont(font);
+	StelApp::getInstance().setBaseFontSize(font.pixelSize());
+	// TODO: Store font name in config.ini
 }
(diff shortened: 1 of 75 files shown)


gzotti:
@alex-w , @guillaumechereau , @xalioth  -- Time for discussion please:

All fonts and GUI dialogs have been based on size 13. Some font sizes were hardcoded in the CSS. I changed them to be percents relative to 13. (Except for the time panel, where fixed large size makes sense). 

However, it may even be useful to allow having different screen font and GUI dialog font sizes. Large screen font may help for presentations or weak eyesight (or even the lousy 4k problems?), but may make dialogs uncomfortably large, so some users may like to have a larger screen font size and moderate GUI font size. 

The current solution still has a single base font size, but is prepared to split GUI font size from screen font size. A change in the GUI dialog font size requires reloading the dialog (it will be re-styled in setVisible()), I think this is acceptable. 

The GUI font is controlled by QGuiApplication::font(), the screen font sizes are always set on the fly based on StelApp::getBaseFontSize(). 

The changes to try out the separation are simple: comment away 3 lines in StelApp::setBaseFontSize, and activate line 384 in main.cpp. If we separate, a second Spinbox in the Config panel is simple to add.

I think it would also be possible to allow font selection in the GUI (for #424). Opinions?

src/gui/ConfigurationDialog.cpp
```diff
@@ -1692,3 +1694,19 @@ void ConfigurationDialog::setDitherFormat()
 	conf->setValue("video/dithering_mode", selectedFormat);
 	conf->sync();
 }
+
+void ConfigurationDialog::selectAppFont()
+{
+	QFont font=QGuiApplication::font();
+	int size=font.pixelSize();
```
alex-w: Unused variable

alex-w:
Current implementation of font selector is not very good, because in standard dialog any user will be select font size in points, not in pixels, plus obliquily settings. Maybe we should use pixel-independent scaling for fonts and GUI elements.

src/gui/ConfigurationDialog.cpp
gzotti: Come on - this is work in progress. If you activate line 1706, it will be used. I'd like to get comments on features, not trivialities.

gzotti:
This is just the default font selection panel, for a first quick implementation. If you find ways to disallow size or style changes (also no underline etc.), I would like to suppress those useless options. Also it seems that line 1706 should be activated, as the selected font is loaded in a pretty small size.

gzotti:
I thought about this, but the larger "example" field is nice. We can also write somewhere that only the font, no style, will be switched.

gzotti:
About font selection: The font panel has a "Writing system" selection where the user can limit the font list to those compatible with certain character sets (Cyrillic, Thai, ...). If we decide to not use the larger font panel, we could probably add another dropdown (combo) to do the same.

gzotti:
@xalioth , @guillaumechereau any opinion on this?

alex-w:
I think this line will be good for all cases:
Base font: [Font family selector][size]

xalioth:
Hi Georg,

I just had a look at your branch. First comments are about the process:
 - in a local branch, don't do merge commit, but instead rebase your branch on top of master, we want a linear history
 - don't mix unrelated commits in a such a branch (like airmass fix)

on the features: 
 - the font size change is fine for me, I understand people may want to change that
 - I don't think the solution to bug  #523 is to reduce font size, but rather to reduce the amount displayed object information, which is now ridiculously long. By default, this should be reduced to just a few lines I think.
 - I also don't like that users are allowed to change the font arbitrarily. My feeling is that fonts are a GUI design choice and should not be configurable. Most of the other fonts I tried look completely ugly. For solving the issue #424 (GUI default font looks bad with arabic), a better fix would be to pre-define a good looking font for some languages (and ship it with the program if necessary). User don't want to choose fonts, they just want the GUI to look good.

PR #547, opened 2018-11-10
Update M31 texture and add some DSO textures

sunshuwei:
Update M31 texture and add some DSO textures.
License information has been added to CREDITS.md.


alex-w:
Some textures just removed - why?

sunshuwei:
I have recovered them.
However, ic2948-trev.png, ic2944-virgo.png, they are Running Chicken Nebula, which are not currently used.
n1365.png, fornax cluster already contains it.

alex-w:
It’s very strange.

sunshuwei:
Maybe forgot to delete them before.
The same is true of sadr.png. It is the same object as GammaCygni-vasey.png.

PR #555, opened 2018-11-26
XLSX support

alex-w:
Added support XLSX files instead CSV due ugly support CSV by Microsoft Office on non-English locales


gzotti:
Good idea!
However, would it be possible to also keep CSV for the sake of those who have already developed a workflow? After all, a CSV is a text file which you can AWK or process with other tools. So, the export should allow either CSV or XLSX.

gzotti:
Thanks. It's OK from my side then!

PR #558, opened 2018-11-28
Added 35 DSO textures.

sunshuwei:
Added 35 DSO textures.


gzotti:
Hi!

Thanks for all these images. However, I see many new DSO photos which are rectangular patches with stars and DSO objects. These rectangles are very distracting IMHO. Please, if possible, limit the image area to the DSO object itself, i.e. paint away the field stars after adjusting the image coordinates.

PR #588, opened 2019-01-06
Fix failure on windows for real, see #587

DarthGandalf:


DarthGandalf:
Now this is ready - the failure in travis is unrelated.

PR #610, opened 2019-01-29
 Update and add some DSO textures 

sunshuwei:
Update and add some DSO textures.


alex-w:
Eta Carina texture is oversized - please reduce it on 50-60% (oversized images will not seen on many old systems and it make installation package oversized too).

PR #719, opened 2019-07-20
Make splash screen more informative

10110111:
Now the splash screen provides some clue as to what Stellarium is actually doing during its not so fast loading process. Changing messages let the users feel that the process has not hung, and is still in progress.

I've tried to preserve the version text in the top-left corner, only adding the status messages at the bottom-left side. Here's how it looks now (at one stage of initialization):

new-splash

### DIFF
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -21,6 +21,7 @@
 
 #include "StelCore.hpp"
 #include "StelMainView.hpp"
+#include "StelSplashScreen.hpp"
 #include "StelUtils.hpp"
 #include "StelTextureMgr.hpp"
 #include "StelObjectMgr.hpp"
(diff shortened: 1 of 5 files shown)


alex-w:
Why you are use tr() for splash notifications?

10110111:
Because they are supposed to be user-visible, and the user's system locale may be different from English.

alex-w:
Stellarium do not used tr() for translation and users will be see English text always.

src/core/StelApp.cpp
```diff
@@ -449,120 +455,146 @@ void StelApp::init(QSettings* conf)
...
+	SplashScreen::showMessage(tr("Initializing Hipparcos stars & names..."));
 	StarMgr* hip_stars = new StarMgr();
 	hip_stars->init();
 	getModuleMgr().registerModule(hip_stars);
 
+	SplashScreen::showMessage(tr("Initializing Stellarium Core..."));
 	core->init();
 
 	// Init nebulas
+	SplashScreen::showMessage(tr("Initializing nebulas..."));
```
alex-w: I think “Initializing deep-sky objects...” will be better here.

10110111:
> Stellarium do not used tr() for translation and users will be see English text always.
> 
> Hint: using q_() instead?

Oh, that's unexpected. Will fix shortly.

10110111:
OK, updated.

PR #774, opened 2019-09-25
Small mistake of  words

ultrapre:


alex-w:
I've added fix manually

PR #812, opened 2019-10-22
first commit of this branch: german navigational stars added

### LINKED ISSUE #408: Nautical Navigation Data
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1106745

This feature request from old bug tracker.

Object Data
 - [x] Rise, Meridian Passage and Set times
 - [x] GHA (Greenwich Hour Angle)
 - [x] SHA (Sidereal Hour Angle)
 - [x] LHA (Local Hour Angle)

Twilights (start & end times)
 - [x] civil
 - [ ] nautical
 - [ ] astronomical

_An additional proposal from the old bug tracker:_ indicate beginning and ending times of all twilights for the selected location in the sun data... also moon rise and set data.

larawag:
german navigational stars added

### Description
80 german navigational stars added

Fixes #408

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Linux/Mint 18.3
* Graphics Card:

## Checklist:


gzotti:
ad (b): We must think about correct grammar phrasing when using fixed strings in translations. "... listed in the "Nautisches Jahrbuch"..." sounds terrible if translated into "... tabelliert im Nautisches Jahrbuch". 

ad (e): I have not run the new code yet. If the screen text has no good position yet, like upper-right corner (below a possible Oculars plugin GUI!), we can add it now to the regular infostring. I have added a method "addToExtraInfoString()" method to the StelObject class, so it should be available to any selectable object, and avoids overlap with the "normal" info text display. Just see how the infostring is being created (fragments of HTML, with tables where applicable), format your part an a similar way and add it with that function. This is currently somewhat experimental, and we may think about adding flags about where to position this text, but for now the new info will be added below or near the end of the other data.

Further thoughts: Currently we have a Boolean switch "Use decimal degrees". Maybe this should be changed to allow a selection of (HHMMSSt/DDMMSSt | Decimal Hours/Decimal degrees | Decimal degrees everywhere | DDMMt (decimal minutes)) In addition, negative hour angle [-12..12] vs. [0..24]? I don't know the navigational practice, but DDMMt could also be useful to other users.

However, the text is becoming so much that on small screens it already overlaps the left button bar, so selecting the infostring fields becomes important. Probably navigators will switch off galactic/supergalactic coords and a few other fields, but we should think about combining data, e.g. PA/separation plus deltaRA/deltaDE for binaries into 1 line.

larawag:
Thank you for your quick and detailed response and your hints.

Basically, while coding I tried to keep close to the Observability plugin, where the formatting and the calculation is within the code. Therefore, I did not find the StelUtils formatting code - thanks for the hint. In my branch I already extended it with the DDM format.

About gzotti's further thoughts:

For me, your suggestion to replace the Boolean switch "Use decimal degrees" by a more general choice box for (HHMMSSt, ...) seems to be a major change in StelObject class (also somewhere else?). Could you clarify from your maintainer's point of view internally whether you support this idea before I do some further steps?

Additional question:

i) I tried to run the addToExtraInfoString() method. Unfortunately it does not work. Could you provide me a code snippet example?

ii) I took this code snippet from the Observability plugin:

EquPos = selectedObject->getEquinoxEquatorialPos(core);
		EquPos.normalize();

could you explain me, where to find the definition of EquPos[i]; i=1,2,3? Maybe I could re-implement GHA and SHA in a better way; currently it is calculated with the aid of the current longitude and this seems to me somehow reverse engineered...

(BTW, the code snippet "toUnsigned_HA" was also taken from Observability plugin and is not necessary anymore.)

gzotti:
> About gzotti's further thoughts:

>For me, your suggestion to replace the Boolean switch "Use decimal degrees" by a more general choice box for (HHMMSSt, ...) seems to be a major change in StelObject class (also somewhere else?). Could you clarify from your maintainer's point of view internally whether you support this idea before I do some further steps?

This was more an idea for us (Alex and me) whether we should allow this as third main format. It's OK if you format your GHA etc. strings as you need them. We may change this behaviour later, to harmonize the outputs.  (This would however vice-versa allow display of nav. values in other formats)

>Additional question:

>i) I tried to run the addToExtraInfoString() method. Unfortunately it does not work. Could you provide me a code snippet example?

In Navstars.cpp, lines 250-300, just write 
`    selectedObject->setExtraInfoString(navText + "<br/>");`
for the first, and for the other entries
`    selectedObject->addToExtraInfoString(navText + "<br/>");`

Indeed I will have to make this a bit more fine-grained later to allow extra info per plugin, but this will be the first practical application.  

>ii) I took this code snippet from the Observability plugin:

>EquPos = selectedObject->getEquinoxEquatorialPos(core);
>EquPos.normalize();

>could you explain me, where to find the definition of EquPos[i]; i=1,2,3? Maybe I could re-implement GHA and SHA in a better way; currently it is calculated with the aid of the current longitude and this seems to me somehow reverse engineered...

This is a Cartesian vector into which spherical coordinates can be converted. (BTW i=[0, 1, 2]).
See Object::getCommonInfoString() near line 328 for where the astronomical hour angle is computed. (It's a bit involved depending on whether atmosphere is active and which coordinate format is required.)

larawag:
Thank you, Georg, for your detailed answer concerning my very simple questions. Following your guide, the additional information appeared easily on my screen.

Meanwhile I added a new function 
`QString radToDdmPStr(double angle, const int precision, const bool useD, const char flag);`
to my local copy of StelUntils.cpp, which has an additional flag for a signed, unsigned, "horizontal" and "vertical" return value in the DDM format:
```
		case 'S': prefix = (sign?'+':'-'); break; // signed +/-
		case 'U': break; // unsigned
		case 'H': suffix = (sign?'E':'W'); break; // signed E/W
		case 'V': suffix = (sign?'N':'S'); break; // signed N/S	
```
Do you (Alex & Georg) agree to add such a function directly to StelUtils class?

All the co-ordinates of fixed stars are referred to vernal equinox and it would be necessary to know its LHA and GHA  and maybe also its rise and transit times. From my perspective it would useful to select the equinoxes directly like other objects (stars, planets, satellites) and show the additional data on the screen. How can I make this feature available?

gzotti:
Yes, I think we can accept (and later use) this function in StelUtils. 

On LHA (Local Hour Angle?):  As I understand this should be the very same thing as the astronomical Hour Angle found in the function mentioned above, maybe just expressed in degrees, or negative if in the eastern half. GHA differs from LHA just by geographical longitude.

Or, if the formulation frequently found in coordinate tables, "all coordinates refer to equinox and epoch of J2000.0" or maybe also "... of J2020.0" confuses you: this just says "these coordinates are valid in the J2000.0 [J2020.0] coordinate frame". Else, you could set a Marker (see chapter 3.6 of the User Guide) at the equinox and display data of this point if you really need it for some test. If you need it as an important feature, maybe we can develop a toggle to draw selectable markers.

larawag:
Dear gentlemen,

I implemented the information I regard as important for navigational stuff to the main window as discussed above. I already found (and fixed) the obligatory "principal error" we would receive shortly after first release...

In the last weeks I tried to extend the NavStarGUI. My goal is to create a table like one of those in the AstroCalcDialog. In this new chart, only those NavStars are listed being visible for the current location and time with some additional information. Than I can provide the opportunity to export this list it for example to Excel which enables the user to plan a sailing trip.

On the first glance, that looked easy to me, but now I found out that there are some surprising obstacles (for myself). More precisely, I put the corresponding calculations in the file gui/NavStarsWindow.cpp in order to get a simple access to the Qt-slots. Than I wrote in gui/NavStarsWindow.cpp something like:

```
#include "StelApp.hpp"
#include "StelCore.hpp"
#include "StelLocaleMgr.hpp"
#include "StelModule.hpp"
#include "StelModuleMgr.hpp"
#include "StelGui.hpp"
#include "StelUtils.hpp"
#include "StelObjectMgr.hpp"
```

```
void NavStarsWindow::computeNauticalTwilight(StelCore* core)
{
	Vec3d pos = GETSTELMODULE(StelObjectMgr)->searchByName("Sun")->getEquinoxEquatorialPos(core);
	double ra, dec;
	StelUtils::rectToSphe(&ra, &dec, pos);

	lon = core->getCurrentLocation().longitude;
	lat = core->getCurrentLocation().latitude;

(plus some more lines)
```

But - as shown above - I need access to `StelCore* core` within gui/NavStarsWindow.cpp. Unfortunately I cannot solve this and in any case I get some fatal compiling error.  (I could re-implement the complete NavStar in one single file. But this would create a whole mess in your files. - Maybe not the best idea...) 

Could you give me a hint how you would handle the issue how to get access to the core module? I expected this somewhere in the documentation but - sorry - I failed to find this. I am pretty sure that this is quite simple.

gzotti:
There are dozens of examples in many modules.

Even in NavStarsWindow.cpp you find 

     connect(StelApp::getInstance().getCore(), SIGNAL(configurationDataSaved()), this, SLOT(saveSettings()));


So, StelApp::getInstance().getCore()

PR #867, opened 2019-11-29
Added Pinyin transliteration to Chinese sky culture

sunshuwei:
Added Pinyin transliteration to Chinese sky culture.


sunshuwei:
Pinyin wikipedia
https://en.wikipedia.org/wiki/Pinyin
Google Translate can display pinyin while typing Chinese.

It should be difficult to transliterate into other languages again, and seem to be treated as untranslatable strings.
Pinyin can accurately express the pronunciation of Chinese characters by defining the pronunciation of each Latin letter. The system includes four diacritics denoting tones. It is difficult for other letters to reconstruct a set of spelling rules. It can also be used if denoting tones is cancelled, but it will cause a lot of homophones. (In fact, homophones cannot be completely avoided even with denoting tones.)

PR #880, opened 2019-12-16
Vts plugin

xalioth:
This is a small plugin allowing to run Stellarium as part of the VTS tool from CNES
See https://timeloop.fr/vts/

### Description
No additional dependencies are required
The plugin may not need to be build systematically as it's of no use to most users, but it would be good to keep this code in the master for maintenance.

### Build
Using this plugin in VTS requires stellarium to build it on a CentOS 7 linux. For this, I added a dacker script to generate a final package from any OS. See plugins/Vts/README.txt


gzotti:
Sure. Add-ons which are not in the repo simply vanish or remain unnoticed.

alex-w:
It's fun. Fabien, could you create a script for create package for Windows and demo script for Vts?

xalioth:
Hi Alex,

thanks for merging!

Next time it would be better not to squash the comits in a single one, as
the history is then lost..
Also, your "typo fix" is incorrect, it should really be /app because the
script is meant to be used inside a docker container..

Fabien

alex-w:
Oops... revert dot back

PR #908, opened 2020-01-03
Add the configuration file and shell scripts to translate skyculture descriptions with po4a

yurchor:
It may be useful for other translators.


alex-w:
Please move config file and she’ll scripts into utils directory.

PR #942, opened 2020-01-20
Add Cross-origin (CORS) support to Remote Control plugin

### LINKED ISSUE #941: CORS (Cross-origin resource sharing) for Remote Control plugin
I'm trying to connect to Stellarium using the Remote Control plugin from my website ([https://telescopius.com](https://telescopius.com)), but I'm getting access denied because of CORS. I can circumvent that by installing a Chrome plugin to inject the proper Access-Control-Allow-Origin, but I would like every user to be able to connect to their own Stellarium server without having to install plugins for their browsers.

Is there any reason why CORS is not enabled for the Remote Control plugin, or is it just that there has been no good reason to do it yet?

Enabling this would should be as easy as adding a oneliner with the "Access-Control-Allow-Origin: *" HTTP header to all server responses, so all developers could hook their web apps to local instances of Stellarium.

sebagr:
...so it can be called from web applications in browsers.

### Description
This change will let web applications to call Stellarium using the Remote Control plugin. This was not possible before because of Cross-origin Resource Sharing (CORS) limitations: browsers don't allow websites to make cross-origin requests unless the server returns certain headers. This branch lets the user enable CORS support and enter the web application that will be calling it so no other websites hijack this connection.

No dependencies.

Fixes #941

### Type of change

### How Has This Been Tested?
After opening Stellarium, enable the Remote Control plugin, and configure it. Check the checkbox "Enable CORS for the following origin" and enter the URL of the website that will be accessing Stellarium through the Remote Control plugin - e.g. https://telescopius.com

When the plugin is restarted, such website will be able to send requests to the Remote Control plugin. Conversely, if the checkbox is unchecked, access to Stellarium will be denied by the browser.

**Test Configuration**:
* Operating system: macOS Catalina 10.15.2, Google Chrome 79.0.3945.117 (Official Build) (64-bit)
* Graphics Card: N/A

## Checklist:
- [ ] I have made corresponding changes to the documentation - **it seems there's no documentation about the remote control configuration popup where to add an explanation of this new field. In any case, it should be self-explanatory.**
- [X] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes - **it seems the Remote Control plugin doesn't have any unit tests**
- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/plugins/RemoteControl/src/RequestHandler.hpp
+++ b/plugins/RemoteControl/src/RequestHandler.hpp
@@ -75,7 +75,7 @@ public slots:
 	bool getEnableCors() const { return enableCors; }
 	//! Set the host for which CORS is enabled. Specify "*" to let any website take control.
 	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
-	void setCorsHosts(const QString& hosts);
+	void setCorsOrigin(const QString& origin);
 
 
 
@@ -92,7 +92,7 @@ private:
 	bool usePassword;
 	QString password;
 	bool enableCors;
-	QString corsHosts;
+	QString corsOrigin;
 	QByteArray passwordReply;
 	APIController* apiController;
 	StaticFileController* staticFiles;
--- a/plugins/RemoteControl/src/RequestHandler.hpp
+++ b/plugins/RemoteControl/src/RequestHandler.hpp
@@ -69,6 +69,18 @@ public slots:
 	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
 	void setPassword(const QString& pw);
 
+	//! Sets wether CORS is enabled.
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	void setEnableCors(bool v);
+	//! Returns if CORS is enabled
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	bool getEnableCors() { return enableCors; }
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	void setCorsHosts(const QString& hosts);
+
+
+
+
 private slots:
 	void refreshTemplates();
 
@@ -80,6 +92,8 @@ private:
 
 	bool usePassword;
 	QString password;
+	bool enableCors;
+	QString corsHosts;
 	QByteArray passwordReply;
 	APIController* apiController;
 	StaticFileController* staticFiles;
--- a/plugins/RemoteControl/src/RequestHandler.hpp
+++ b/plugins/RemoteControl/src/RequestHandler.hpp
@@ -73,6 +73,7 @@ public slots:
 	void setEnableCors(bool v);
 	//! Returns if CORS is enabled
 	bool getEnableCors() { return enableCors; }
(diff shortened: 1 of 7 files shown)


gzotti:
Can you please add a label beneath "Enable CORS..." in the config panel, where you give a short description what this does. Something like "Let the specified website access and control your Stellarium. Specify "*" to let any website take control. Do this at your own risk."
Also specify the separator (comma-separated list? or colon? or space?)

A paragraph describing what CORS can do should also be added to the Stellarium User Guide. You find the LaTeX source in guide/plg_interfaces.tex.

plugins/RemoteControl/src/RequestHandler.hpp
```diff
@@ -69,6 +69,18 @@ public slots:
 	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
 	void setPassword(const QString& pw);
 
+	//! Sets wether CORS is enabled.
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	void setEnableCors(bool v);
+	//! Returns if CORS is enabled
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
```
gzotti: warning in L76 probably is not required. Also, this is a const method.

plugins/RemoteControl/src/RequestHandler.hpp
```diff
@@ -69,6 +69,18 @@ public slots:
...
 	void setPassword(const QString& pw);
 
+	//! Sets wether CORS is enabled.
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	void setEnableCors(bool v);
+	//! Returns if CORS is enabled
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	bool getEnableCors() { return enableCors; }
+	//! @warning Make sure to only call this only when the server is offline because they are not synchronized
+	void setCorsHosts(const QString& hosts);
```
gzotti: please document format of the hosts string, including separators if really more than one host is allowed.

sebagr: I just documented this to clarify it's either one host or "*" (any host). More logic is needed if we are to support multiple specific hosts, because we need to adjust the response header to match the requester origin in case it's whitelisted. I think this can wait until it's needed, right?

gzotti: Yes. My short investigation what this all means indicated that no current browser understands more than one address. You may reflect this by using singular in method and argument names. Sorry to be so picky, but it improves understandability.

sebagr: Oh, I see what you mean. I never noticed that because "*" is a possible value which means "all hosts", that's why plural made sense to me. But I'll change it to singular if you prefer, no problem.

sebagr: Agree :)

sebagr: Done. I also took the opportunity to rename "host" to "origin", as it's the proper technical term.

gzotti:
OK, thanks!

PR #946, opened 2020-01-24
Introduce a GimbalOrbit for the "Observer" planets

gzotti:
Another cherry-pick from the Planet Axes branch.

This allows arbitrary viewpoints by keyboard actions.

The current "observer" planets are admittedly rather lame: we must place them high above the ecliptic plane to lose perspective effects. Then light time interferes with e.g. observation of eclipses or shadow phenomena on Jupiter. And we can never observe a planet sideways. 

### Description

This new GimbalOrbit allows setting location manually. On loading "observer" planets they receive this specialized "Orbit" subclass. When observing from such an object, keyboard interaction (Alt-Cursor) allows changing orientation and distance from the parent planet. 

The state of implementation is "not perfect yet", but interesting enough for further development, and could be easily disabled in the master branch if required and merged now.

- Problem 1: Currently the view orientation cannot be properly controlled. We would need an EclipticalJ2000 view frame for meaningful screen orientation, or maybe just a fixed observer "latitude" on our "point planet"? 
- Problem 2: hotkey collision with OS hotkeys in StelMovementMgr. I cannot find an appropriate key combination to change distance from parent. Alt-5/6 are preliminary keys of course.

It can be disabled by disabling a define in Orbit.h

### Screenshot:

gimbalObserver

### Type of change

### How Has This Been Tested?

Just set any "observer" as location planet. Moving to another planet with Ctrl-G works. I have so far not found any critical operational or graphical issue. 

What needs to be done is some way to keep the view aligned with the J2000 ecliptic frame. 

I think even if we don't activate the functionality it can be merged now, feature disabled in Orbit.h, and continued on a later date. 

**Test Configuration**:
* Graphics Card: NVidia Geforce M960 (likely irrelevant)

## Checklist:

- [ ] I have made corresponding changes to the documentation. (This will be required after finishing all related works and activating it for release.)

### DIFF
--- a/plugins/Exoplanets/src/gui/ExoplanetsDialog.cpp
+++ b/plugins/Exoplanets/src/gui/ExoplanetsDialog.cpp
@@ -763,7 +763,7 @@ void ExoplanetsDialog::populateDiagramsList()
 	axisX->clear();
 	axisY->clear();
 
-	const QList<axisPair> axis = {
+	static const QList<axisPair> axis = {
 	{ q_("Orbital Eccentricity"),          0},
 	{ q_("Orbit Semi-Major Axis, AU"),     1},
 	{ q_("Planetary Mass, Mjup"),          2},
--- a/plugins/RemoteControl/src/gui/remoteControlDialog.ui
+++ b/plugins/RemoteControl/src/gui/remoteControlDialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>666</width>
-    <height>525</height>
+    <width>530</width>
+    <height>358</height>
    </rect>
   </property>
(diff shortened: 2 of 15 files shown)


alex-w:
Comment for problem 2: what about Home/End hotkeys?

alex-w:
Comment for problem 1: setting observer "latitude" on our "point planet" after space flight is good step here to avoid finding and selection an planet by manual. But introducing ecliptical frame for good navigation will be best solution here (maybe later, because it's not critical feature at the moment).

gzotti:
> Comment for problem 2: what about Home/End hotkeys?

Yes, that seems to work! I will still prepend Alt to make it consistent. 

Not sure about problem 1 yet, but we can still exclude it from release by that define in Orbit.h.

PR #947, opened 2020-01-26
Enhance extra strings

gzotti:
### Description

The last preliminary cherry-pick from gz_planetAxes.

The extraInfoStrings should allow plugins to add strings into the InfoString, e.g. 
customized coordinates for astro navigation, but all InfoString flag groups can be added to now. 

Strings from scripts are currently added to the end of all InfoStrings, followed by DebugAid strings. The latter can be of great help to display various intermediate values during development. 

TODO: We should probably have smaller virtual methods in StelObject & derived classes instead of StelObject::getCommonInfoString() for the various parts of the InfoString for better granularity. (This can happen also later when required by plugins which need these newly inserted strings.)

### Type of change

### How Has This Been Tested?

A module can insert new info strings e.g. in its draw() method. I am thinking especially of a more complete Astro Navigation plugin which may show new kinds of coordinates (SHA, GHA, ...), the Observability plugin, or some plugin that provides ethnoastronomical details I have not even yet considered. Also the Equation of Time could be modified to add its value only to the Sun's Infostring when selected (or also into the Infostring of whichever object is just shown). 

You can use a script 
```
core.addToSelectedObjectInfoString("some message from a script <br/>");
core.addToSelectedObjectInfoString("another message from script<br/>");
```
to add one or more strings,
`core.addToSelectedObjectInfoString("", true);`
to reset them. 

Currently it is not intended to remove the messages one-by-one, just all of them.
The extra information remains with the selected object and will be shown when that object is selected again. The exception are stars with their non-permanent StarWrapper. 

**Test Configuration**:

## Checklist:

- [x] I have made corresponding changes to the documentation (NOT YET to the SUG. Info should go to the scripting chapter.)
- [ ] I have added tests that prove my fix is effective or that my feature works. (-->tested only locally)

### DIFF
--- a/src/core/modules/StarWrapper.cpp
+++ b/src/core/modules/StarWrapper.cpp
@@ -172,13 +172,13 @@ QString StarWrapper1::getInfoString(const StelCore *core, const InfoStringGroup&
 	{
 		QStringList extraNames=getExtraInfoStrings(Name);
 		if (extraNames.length()>0)
-			oss << q_("Other names: ") << extraNames.join(", ");
+			oss << q_("Additional names: ") << extraNames.join(", ") << "<br/>";
 	}
 	if (flags&CatalogNumber)
 	{
 		QStringList extraCat=getExtraInfoStrings(CatalogNumber);
 		if (extraCat.length()>0)
-			oss << q_("Other catalog numbers: ") << extraCat.join(", ");
+			oss << q_("Additional catalog numbers: ") << extraCat.join(", ") << "<br/>";
 	}
 
 	bool ebsFlag = false;
@@ -219,7 +219,7 @@ QString StarWrapper1::getInfoString(const StelCore *core, const InfoStringGroup&
 		}
 		else
 			oss << QString("%1: <b>%2</b>").arg(q_("Type"), startype) << "<br />";
-		oss << getExtraInfoStrings(flags&ObjectType).join(' ');
+		oss << getExtraInfoStrings(flags&ObjectType).join("");
 	}
 
 	oss << getMagnitudeInfoString(core, flags, alt_app, 2);
@@ -227,7 +227,7 @@ QString StarWrapper1::getInfoString(const StelCore *core, const InfoStringGroup&
 	if ((flags&AbsoluteMagnitude) && s->getPlx ()&& !isNan(s->getPlx()) && !isInf(s->getPlx()))
 		oss << QString("%1: %2").arg(q_("Absolute Magnitude")).arg(getVMagnitude(core)+5.*(1.+std::log10(0.00001*s->getPlx())), 0, 'f', 2) << "<br />";
 	if (flags&AbsoluteMagnitude)
-		oss << getExtraInfoStrings(AbsoluteMagnitude).join(' ');
+		oss << getExtraInfoStrings(AbsoluteMagnitude).join("");
 
 	if (flags&Extra)
 	{
@@ -270,7 +270,7 @@ QString StarWrapper1::getInfoString(const StelCore *core, const InfoStringGroup&
 			else
 				oss << QString("%1: %2 %3").arg(q_("Distance"), QString::number(distance, 'f', 2), ly) << "<br />";
 		}
-		oss << getExtraInfoStrings(Distance).join(' ');
+		oss << getExtraInfoStrings(Distance).join("");
 	}
 
 	if (flags&Extra)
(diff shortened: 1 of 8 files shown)


src/core/modules/StarWrapper.cpp
```diff
@@ -312,9 +331,10 @@ QString StarWrapper1::getInfoString(const StelCore *core, const InfoStringGroup&
 		if (pa<0)
 			pa += 360.f;
 
+		oss << QString("%1: %2%3 %4 %5%6").arg(q_("Proper motion"))
```
alex-w: Missing space char between %2 and %3

gzotti:
Actually the "missing" space was on purpose. If it was available, I'd make a "small space", but if you insist I will add the space.

alex-w:
> Actually the "missing" space was on purpose. If it was available, I'd make a "small space", but if you insist I will add the space.

Yes, all similar data should be in similar format. At the moment format for "proper motion" is inconsistent to all data with unit "mas/yr".

A-j-K:
This PR/Merge has raised a question for me. Currently, while learning the Stellarium codebase, I have a small plugin that I wrote. After rebasing this merge into my branch it failed to compile. No problem, I just had a look at the changes made, read the comments and made the change. This is where it got interesting.

In my plugin I have the following:-

    selectedObject->setExtraInfoString(StelObject::OtherCoord, extraInfo);

However, when selecting an object Stellarium crashed with a read access violation. Again, I did some reading but nothing obvious appeared. So I changed it to thisto see what would happen (as this was the only example of the functions use I found):-

    selectedObject->setExtraInfoString(StelObject::Script, extraInfo);

And this now worked just fine. However, I'm sure I shouldn't be using "Script" here. Also, "DebugAid" did work, other InfoStringGroupFlags didn't. At this point I don't know if the problem is with me or with the code. Some clarification my help me out if I'm missing some knowledge.

reference: 
https://github.com/A-j-K/stellarium/blob/AstroNav/plugins/AstroNav/src/AstroNav.cpp#L108

gzotti:
I have changed quite a lot here. First, for GHA or other "coordinates", you should really use the OtherCoord type. And especially use` addToExtraInfoString(StelObject::OtherCoord, extraInfo)`, and no longer `set....`

Please see `StelObject::getCommonInfoString()` how these strings are then used: Coordinates may be inserted into a table or just shown in lines. If table mode is active, you must prepare those lines in a compatible HTML format. Else the line just has to end in `<br/>`

All other extraInfoStrings with InfoString flags except for Script and DebugAid are auto-deleted after showing them. So they have to be added in some regularly called function like update() or draw() (depending on your plugin).

A-j-K:
When I used ::addToExtraInfoString() the behaviour I observed was that the info is continually added (I use ::draw() ) so I saw my info string repeated over and over which was why I was using ::set....  I have now switched to ::add.

However, I still get the same crash in qmap.h line 89 because it appears to be looking for a tree node that doesn't exist or does exist but points to garbage. I need to grok more on how this works. Just I can't find a "worked example" of doing this in other plugins except the Script plugin (which does work fine when I hijack it's flag).

Callstack:-
InfoPanel::setTextFromObjects() 179
StelObject::removeExtraInfoStrings()  926

gzotti:
Yes, I have seen and removed that string flood. 
Do you have a repo to look into? I tested it with a few strings injected in draw() into the Magnitude and Size extrastrings area.

A-j-K:
@gzotti 
https://github.com/A-j-K/stellarium

Forked from Stellarium, branch "AstroNav"

gzotti:
Ok, a few issues:

1) The extraInfoStrings are a QMultiMap, i.e. you should add the strings one-by-one. 
selectedObject->addToExtraInfoString(StelObject::OtherCoord, extraInfo); If you want to be sure the sequence of strings, concatenating like you do now may still be OK. 
2) The coordinate table as built up in StelObject has 4 fields. Field2 ends with the / sign where needed. You may have to use the css style tags seen in StelObject.cpp 
3) In Table mode, you should not build the table tags yourself. The OtherCoord strings are inserted (again: one-by-one!) as table lines. Other plugins may still add their stuff, adding table tabs will interfere here.
4) My error in StelObject.cpp: please comment away lines 560, 561 and 563. I test the flag, but never set this value... I should fix that soon.
5) Even better: It seems the table is aligned with Ecliptic Obliquity...", i.e. a very long label. I will change the whole thing to create yet another table for the OtherCoords below that. Put this into line 558:

```
	if (withTables)
		 res += "</table>";

	// Specialized plugins (e.g. Astro Navigation or ethno-astronomical specialties) may want to provide additional types of coordinates here.
	//if (flags&OtherCoord)
	//{
		if (withTables)
			res += "<table style='margin:0em 0em 0em -0.125em;border-spacing:0px;border:0px;'>";
		res += getExtraInfoStrings(OtherCoord).join("");
		if (withTables)
			 res += "</table>";
	//}
```


I am on Qt5.13/MSVC2017 and never had a crash.

A-j-K:
@gzotti Thanks for the feedback. I'll go through that tomorrow. Fyi, I'm Qt5.14/MSVC2017_64

A-j-K:
@gzotti 

    "please comment away lines 560, 561 and 563"
This worked and allowed my info lines to appear.

    "res += "<table style='margin:0em 0em 0em -0.125em;border-spacing:0px;border:0px;'>";"
This didn't look good at all so I removed it.

I now understand the QMultiMap use and have split my one long QString into parts.

Regarding the "crash" when using StelObject::OtherCoord flag. This only occurs when using Visual Studio to build a debug build. Release builds appear unaffected. It does, however, mean I can no longer use debug builds. The crash comes from an unhandled exception emitted from Qt5Cored.dll

gzotti:
The table style was taken from the previous table and may be modified at a later date. Your plugin was the final trigger to create these extra strings and let them be weaved into the InfoString.

Regarding the crash, I have no good idea. Maybe create the exception handler where needed.

A-j-K:
Well, I tried adding a try/catch around the while() loop in StelObject::removeExtraInfoStrings() and also configuring Visual Studio to ignore Qt5Cored.dll exceptions but to no effect.

A-j-K:
@gzotti 

The "Additional coordinates (from plugins)" checkbox in the config dialog doesn't function as expected. It's just a small omission as outlined below.


index 0c2b8fd6dc..fc5c91c5e8 100644
@@ -966,6 +966,9 @@
                <property name="text">
                 <string>Additional coordinates (from plugins)</string>
                </property>
+               <attribute name="buttonGroup">
+                 <string notr="true">buttonGroupDisplayedFields</string>
+               </attribute>
               </widget>
              </item>
             </layout>

A-j-K:
If you prefer I just submit a PR for this then let me know. It just seemed such a small omission to an otherwise good commit.

gzotti:
d'oh! Yes, you are right. Thanks!

PR #950, opened 2020-01-27
WIP: Cleanup solar system loader

gzotti:
There were a few remaining changes not really related to planet axes yet, here dissected from the gz_planetAxes branch to clean that out.

- Lots of cleanup related to SolarSystem loading and Orbit simplifications
- also restore/sanitize handling of named moon phases
- Remove unnecessary semimajor orbit axes from Comets and MinorPlanets
- SolarSystem and Planet Docfixes
- Added modern rotational elements (actually including also a preview on coming required changes, but here they are all combined...)

STOP: Something weird currently at Pluto's moons. Must investigate further.

### Type of change

### How Has This Been Tested?

IMHO no big changes over 0.19.3, mostly "documented discoveries" from previous months. Lunar phases now give strings as they may have been originally intended.

**Test Configuration**:
* Graphics Card: n/c

## Checklist:


alex-w:
Code for Moon phase looks weird now - please check it again

src/core/modules/Planet.cpp
```diff
@@ -682,43 +683,53 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
...
+				StelUtils::equToEcl(ra_equ, dec_equ, eclJDE, &lambdaMoon, &betaMoon);
+				StelUtils::rectToSphe(&raSun,&deSun, ssystem->getSun()->getEquinoxEquatorialPos(core1));
+				StelUtils::equToEcl(raSun, deSun, eclJDE, &lambdaSun, &betaSun);
 				core1->setUseTopocentricCoordinates(state);
 				core1->update(0); // enforce update cache for avoid odd selection of Moon details!
+				double deltaLong = (lambdaMoon-lambdaSun)*M_180_PI;
 				if (deltaLong<0.) deltaLong += 360.;
-				int deltaLongI = qRound(deltaLong);
-				if (deltaLongI==45)
+				const int deltaLongI = static_cast<int>(std::round(deltaLong));
```
alex-w: `static_cast<int>(std::round ` is equal to `qRound`

src/core/modules/Planet.cpp
```diff
@@ -682,43 +683,53 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
 			{
 				// For compute the Moon age we use geocentric coordinates
 				QString moonPhase = "";
-				double eclJDE = earth->getRotObliquity(core->getJDE());
-				double ra_equ, dec_equ, lambdaMoon, lambdaSun, beta;
-				StelCore* core1 = StelApp::getInstance().getCore();
-				bool state = core1->getUseTopocentricCoordinates();
+				StelCore* core1 = StelApp::getInstance().getCore(); // FIXME: Why do we need a different reference to the (singular) core here?
```
alex-w: New reference to core need to avoid problem with nomenclature shifts

src/core/modules/Planet.cpp
```diff
@@ -682,43 +683,53 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
...
 				core1->setUseTopocentricCoordinates(state);
 				core1->update(0); // enforce update cache for avoid odd selection of Moon details!
+				double deltaLong = (lambdaMoon-lambdaSun)*M_180_PI;
 				if (deltaLong<0.) deltaLong += 360.;
-				int deltaLongI = qRound(deltaLong);
-				if (deltaLongI==45)
+				const int deltaLongI = static_cast<int>(std::round(deltaLong));
+				if (deltaLongI==0 || deltaLongI==360)
+					moonPhase = qc_("New Moon", "Moon phase");
+				else if (deltaLongI<90)
```
alex-w: Why? Are you using definitions for Moon phases from here: https://www.timeanddate.com/calendar/aboutmoonphases.html

src/core/modules/Planet.cpp
```diff
@@ -682,43 +683,53 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
...
-				if (deltaLongI==270)
+				else if (deltaLongI==270)
 					moonPhase = qc_("Third Quarter", "Moon phase");
-				if (deltaLongI==315)
+				else if (deltaLongI<360)
 					moonPhase = qc_("Waning Crescent", "Moon phase");
-				if (deltaLongI==0 || deltaLongI==360)
-					moonPhase = qc_("New Moon", "Moon phase");
+				else
+					moonPhase = qc_("ERROR IN PHASE STRING PROGRAMMING!", "Moon phase");
```
alex-w: This is should be `qWarning()` - not translatable string for users

src/core/modules/Planet.cpp
```diff
@@ -682,43 +683,53 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
...
 					oss << QString(" (%4)").arg(moonPhase);
 				oss << "<br />";
+
+				// we must repeat the position lookup from above in case we have topocentric corrections.
+				StelUtils::rectToSphe(&ra_equ,&dec_equ, getEquinoxEquatorialPos(core));
+				StelUtils::equToEcl(ra_equ, dec_equ, eclJDE, &lambdaMoon, &betaMoon);
+				StelUtils::rectToSphe(&raSun,&deSun, ssystem->getSun()->getEquinoxEquatorialPos(core));
+				StelUtils::equToEcl(raSun, deSun, eclJDE, &lambdaSun, &betaSun);
+				const double chi=atan2(cos(deSun)*sin(raSun-ra_equ), sin(deSun)*cos(dec_equ)-cos(deSun)*sin(dec_equ)*cos(raSun-ra_equ));
+				oss << QString("%1: %2").arg(q_("PA of bright limb"), StelUtils::radToDecDegStr(StelUtils::fmodpos(chi, M_PI*2.0))) << "<br/>";
```
alex-w: Maybe `Position angle of bright limb` to avoid troubles for newbies?

src/core/modules/Planet.cpp
```diff
@@ -639,8 +639,9 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
 				if (englishName!="Sun")
 					oss << QString("%1: %2").arg(q_("Mean solar day"), StelUtils::hoursToHmsStr(qAbs(getMeanSolarDay()*24))) << "<br />";
 			}
-			else if (re.period==0.f)
+			else if (re.period==0.f) // WILL NEVER BE CALLED! In this case siderealPeriod==0, the enclosing brace will exclude this!
```
alex-w: Why? See: https://en.wikipedia.org/wiki/Chaotic_rotation

Or you removed all `rot_periode=0` from ssystem_major.ini file?

src/core/modules/Planet.cpp
gzotti: Ah yes, thanks. I must have changed that also long ago (before this was qRound).

src/core/modules/Planet.cpp
gzotti: > 
> 
> Code for Moon phase looks weird now - please check it again

The idea is: mark "sharp" phases, but also name the phases in between for much longer time. But in fact, I just see it still can be improved...

src/core/modules/Planet.cpp
gzotti: No, just that e.g. "gibbous" is not sharp at 225 degrees, but anywhere between quarter and full. Likewise for crescent.

src/core/modules/Planet.cpp
gzotti: Right. Rather meant for the developers, who should never see this. I must have been tired...

src/core/modules/Planet.cpp
gzotti: Hmm, I just wanted a shorter label. But OK.

src/core/modules/Planet.cpp
gzotti: Oops, you are right. Probably these lines are from a late night in 2017. 
rot_periode since then defaults to the orbit_period. I must have disabled a few too many lines... I will reactivate the chaotic moons. *-)

alex-w:
>STOP: Something weird currently at Pluto's moons. Must investigate further.

I fear you are using re.period without checks on 0. So, for objects with chaotic rotation we may use special flag in ssystem_*.ini files or do not use re.period in computation when value is equal zero.

src/core/modules/Planet.cpp
alex-w: I think this FIXME maybe deleted or changed to NOTE.

src/core/modules/Planet.cpp
```diff
@@ -479,8 +479,10 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
 		if (EphemWrapper::use_de430(core->getJDE())) algoName="DE430";
 		// TRANSLATORS: Ecliptical rectangular coordinates
 		oss << QString("%1 XYZ (%2): %3/%4/%5").arg(qc_("Ecliptical","coordinates")).arg(algoName).arg(QString::number(eclPos[0], 'f', 7), QString::number(eclPos[1], 'f', 7), QString::number(eclPos[2], 'f', 7)) << "<br>";
+
```
alex-w: No need extra blank line here

src/core/modules/Planet.cpp
```diff
@@ -731,6 +743,7 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
 				oss << QString("%1: %2%").arg(q_("Eclipse obscuration")).arg(QString::number(eclipseObscuration, 'f', 2)) << "<br />";
 				if (onEarth)
 				{
+					// FIXME: This is for Solar eclipses only. Make sure it works for Mercury/Venus transits, or omit it then!
```
alex-w: Why FIXME here? This code made exactly for Solar eclipses on Earth only. Or you want to compute "Eclipse magnitude" for transits?

src/core/modules/Planet.cpp
```diff
@@ -777,6 +789,7 @@ QVariantMap Planet::getInfoMap(const StelCore *core) const
 			map.insert("eclipse-obscuration", eclipseObscuration);
 			if (core->getCurrentPlanet()==ssystem->getEarth())
 			{
+				// FIXME: This is for Solar eclipses only. Make sure it works for Mercury/Venus transits, or omit it then!
```
alex-w: Why FIXME here? This code made exactly for Solar eclipses on Earth only. Or you want to compute "Eclipse magnitude" for transits?

src/core/modules/Planet.cpp
gzotti: I think core1 is not needed. It points to the same thing as core.

src/core/modules/Planet.cpp
gzotti: Look at the weird output during a Venus transit (2004, 2012) or Mercury transit (11.11.2019). Better avoid the line if the Moon is not involved. (just check for Moon distance?)

alex-w: I see no weird data (I see not eclipse data!) in version 0.19.3. So, weird data is result of refactoring?

src/core/modules/Planet.cpp
alex-w: > I think core1 is not needed. It points to the same thing as core.

No problem, but please resolve problem of jumping Moon and lunar nomenclature in this case

src/core/modules/Planet.cpp
alex-w: Oops... I was wrong!

alex-w: OK, I have an idea how to resolve this issue

src/core/modules/Planet.cpp
alex-w: OK, I have an idea how to resolve this issue

src/core/modules/Planet.cpp
alex-w: OK, I've fixed it in master

src/core/modules/Planet.cpp
alex-w: OK, I've fixed it in master

alex-w:
OK, I've fixed getting weird data for eclipse magnitude value during transits of inner planets in master. Georg, could you update code in your branch to resolve conflicts?

alex-w:
Oops, I found the bug! Please run script phobos_phun_3.ssc to reproduce the bug. In version 0.19.3 you may see model or disk of Phobos on the Sun during eclipse. In current master you will do not see Phobos during eclipse and... from the Mars in principle!

alex-w:
OK, I've fixed Phobos problem (in master), but what about other moons near it host planets?

src/core/modules/Planet.cpp
gzotti: Indeed.

gzotti:
> phobos

Oh, that may have been introduced when fixing Pan's view previously. 
What's the smallest distance of any moon from its parent? Hopefully your 0.0001 AU is OK now. It just has to be larger than 0.

The point with the FIXME is of course, as seen from Earth, inside this clause we currently only test the Moon's size.  If Moon is not guilty, the displayed result is meaningless.
The complete solution might involve 
ssystem->getEclipseFactor(StelCore *core, QStringList eclipsingPlanetNames)
and test all returned planetNames.  I might try that later.

alex-w:
> > phobos
> 
> Oh, that may have been introduced when fixing Pan's view previously.
> What's the smallest distance of any moon from its parent? Hopefully your 0.0001 AU is OK now. It just has to be larger than 0.
> 
> The point with the FIXME is of course, as seen from Earth, inside this clause we currently only test the Moon's size. If Moon is not guilty, the displayed result is meaningless.
> The complete solution might involve
> ssystem->getEclipseFactor(StelCore *core, QStringList eclipsingPlanetNames)
> and test all returned planetNames. I might try that later.

I'm already fix both problem - see master ;)

gzotti:
OK, good! Now let's hope there is never a transit together with a solar eclipse :-)

alex-w:
In general this PR is OK for me, except one problem: something strange happens when orbital speed for moons of Pluto computed.

gzotti:
Thanks, yes there is something super weird. Will investigate later.

gzotti:
Of course: KeplerOrbit can return valid speed currently only for solar children. I need to improve the speed function to include mass of the parent.

gzotti:
No, something is weird with orientation of -- as it seems -- Uranus, Neptune, and Pluto, and their moons. I must investigate further. Hopefully this gives some insight what the problem with the axes is. When this is cleared, I will do the merge.

alex-w:
Oops... I didn’t check orientation of planets.

alex-w:
Georg, I fear commit adf7bcfe97bb958c6dbf9fe6be2e55fa76905b46 will broke build of branch

gzotti:
On which platform does it break? Why does it run here, then? The Planet class has the functions which don't need repetition in Comet and MinorPlanet.

alex-w:
Oops... the commit b2bdd10434aa8f3e39db813cccdb1be41b5b2dff already break build branch on linux: /home/travis/build/Stellarium/stellarium/src/core/modules/Orbit.cpp:171:46: error: cannot convert ‘double’ to ‘double*’ for argument ‘3’ to ‘void sincos(double, double*, double*)’
1969  sincos(parentRotAscendingNode, &s_nod, c_nod);
1970

alex-w:
> On which platform does it break? Why does it run here, then? The Planet class has the functions which don't need repetition in Comet and MinorPlanet.

Are you removed getCommon...Name() in SolarSystem class and Solar System Editor plugin?

gzotti:
Ah sorry, my oops. I wrote those blindly.

alex-w:
> Ah sorry, my oops. I wrote those blindly.

I've introduced both methods for comets and minor planets to avoid problem with handling lists of small bodies of Solar system. I didn't test the latest changes, but I fear SSE plugin is broken now and possible partially Solar system class too.

gzotti:
> Are you removed getCommon...Name() in SolarSystem class and Solar System Editor plugin?

No, why?  If you call getCommon...Name() on a MinorPlanet or Comet, it will return the function defined in their classes' parent, Planet::getCommon...Name(). If you want to override the functions, you should consider making them virtual. But I don't see why we should require that here, where the deleted functions just repeated the very same code.

alex-w:
Please try remove any minor planet or comet from Stellarium data through SSE plugin.

gzotti:
OK, just deleted 3 comets and 5 minor bodies. And? I can no longer find Ryugu. Because I deleted it before. Does that break on Linux? (Can try that tomorrow.)

alex-w:
OK, I was wrong - SSE plugin works as expected - thanks!

PR #955, opened 2020-02-01
Navstars improvements

A-j-K:
# Add Nautical Almanac data to extraInfoStrings for navigation stars

### Description
The standard out of the box Nav Stars plugin highlights a set of stars that represent those chosen and found in Nautical Almanacs around the world. However, it only high lights them and doesn't provide any extra data.

This patch adds additional information points that include the intermediate values derived from the Almanac along with the Sight Reduction solution. All these additional data points are displayed using the format found in The Nautical Almanac, DDMM.m

These are:-

* GHAa : The Greenwich Hour Angle for the first point of Aries
* LMST : Local mean sidereal time
* SHA  : Sidereal hour angle of the target object
* DEC : The declination of the target object
* GHA : The Green Hour Angle of the target object
* LHA : The Local Hour Angle of the target object
* LAT : The observer latitude
* LON : The observer longitude
* Hc : The computed altitude as derived via a Sight Reduction (solution)
* Zn : The computed azimuth as derived via a Sight Reduction (solution)  

Additionally, two new configuration checkboxes have been added (default unchecked gives original behavior).

CB: Existing. Only highlight nav stars when the star is visible. Currently, when the plugin is enabled all nav stars are highlighted all the time, even during the day when the star is not visible. Checking this box will only allow the highlight if the star is visible.

CB: New. The modifications allow for Nautical Almanac data for any selectable object. This can be limited to only display data for nav stars, Sun, Moon, and planets.

CB: New. Use "upper limb". Most users of sextants use the lower limb of the Sun or Moon. However, a few use the upper limb. Checking this box ensures the Ho (observed altitude) uses the limb of choice.

Fixes # (issue)

Addresses #408 

### Type of change

### How Has This Been Tested?

The patch includes a Unit Test. The data from which was reduced and tested using pencil and paper using the 2020 Nautical Almanac. Other pencil and paper tests were conducted around various nav stars.

## Checklist:


A-j-K:
Capture1

alex-w:
I got this problem:

> In file included from /home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:34:0:
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:143:7: error: extra qualification ‘NavStars::’ on member ‘extraInfoStrings’ [-fpermissive]
  void NavStars::extraInfoStrings(const QMap<QString, double>& data, QMap<QString, QString>& strings, QString extraText = "");
       ^~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:182:7: error: extra qualification ‘NavStars::’ on member ‘addExtraInfo’ [-fpermissive]
  void NavStars::addExtraInfo(StelCore* core);
       ^~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:186:7: error: extra qualification ‘NavStars::’ on member ‘extraInfo’ [-fpermissive]
  void NavStars::extraInfo(StelCore* core, const StelObjectP& selectedObject, bool withTables);
       ^~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp: In constructor ‘NavStars::NavStars()’:
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:166:7: warning: ‘NavStars::limitInfoToNavStars’ will be initialized after [-Wreorder]
  bool limitInfoToNavStars;
       ^~~~~~~~~~~~~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:164:7: warning:   ‘bool NavStars::upperLimb’ [-Wreorder]
  bool upperLimb;
       ^~~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:68:1: warning:   when initialized here [-Wreorder]
 NavStars::NavStars()
 ^~~~~~~~
In file included from /home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:34:0:
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:164:7: warning: ‘NavStars::upperLimb’ will be initialized after [-Wreorder]
  bool upperLimb;
       ^~~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.hpp:163:7: warning:   ‘bool NavStars::starLabelsState’ [-Wreorder]
  bool starLabelsState;
       ^~~~~~~~~~~~~~~
/home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:68:1: warning:   when initialized here [-Wreorder]
 NavStars::NavStars()
 ^~~~~~~~

Please fix order of variables and remove extra NavStars:: in headers file.

gzotti:
I assume it's just the labels from the tables which navigators are so used to. LMST is also already available without the plugin. (No, I don't want to allow re-labeling by arbitrary plugins...)  But "LON/LAT" and "Zn/Hn" with two coord values each could be an option to save two lines.
However, what we need most with these new data is indeed a description of these fields in the About tab, and a 2-5 pages (?) chapter on astro nav basics in the User Guide. It should at least include the derivation of the new fields. How SHA is connected to RA, etc... Does the plugin already help in navigation (enter a few measurements, derive new location, adjust Stellarium's location), or does it "only" provide the data for the known location?

alex-w:
@gzotti  thanks for notification! I've forgot about Stellarium User Guide.

A-j-K:
Thank you for the feedback. 

* I will add some information to the About tab as requested.
* I will add some information to the User Guide.

As for your questions. I was aware that some information is repeated. However, having discussed how Stellarium could be used by Celestial Nav teachers the first thing to note was "getting all this data into a viewable form is too complicated for most students, please display it in a form that's together and in a format that students will be using, whole degrees and decimal (one place) minutes. This format **must** be consistent for all display values, it's all they will ever use. Please make it so that all other data, except the name, can be switched off and the Nav Data is just shown. The data must include the Ho altitude (apparent Alt) for the current position, Nautical Almanac table inputs (UTC and Ho), table intermediate outputs (GP, which is GHA/DEC) and what that final sight reduction solution should be assuming they followed the Almanac sight reduction procedure correctly (that's Hc/Zn)".

So, that above was my guide to how I presented the information. However, I will adjust so that coord pairs (lat/lon ie AP, gha/dec ie GP, Hc/Zn Almanac computed alt/az) are paired on a single line.

I will also update the user guide to reflect the above. For example, when teaching the celestial object sub-point is important and is called GP (ground point). However, when using the Almanac and following sight reduction, we find GHA and DEC of the target which is the GP. But "GP" doesn't appear in the procedure, only GHA and DEC do. GP is just part of the teaching process to understand what is actually going on.

@alex-w @gzotti Please let me know what you think here with regards to the above "requirement" I was trying to achieve.

gzotti:
OK, thanks for explanations of the "external requirements". I did not know about GP, is that the location where the object is in the zenith? Indeed, some basics on CelNav in the Guide would help clarifying the displayed values. Maybe even a label "GP: GHA/DEC=" would be helpful? 
In the Guide you can also recommend settings in the Config/Info tab to avoid display clutter for navigators (switch off distance, spectral class, or such...)

A-j-K:
> OK, thanks for explanations of the "external requirements". I did not know about GP, is that the location where the object is in the zenith? 

Yes.

> Indeed, some basics on CelNav in the Guide would help clarifying the displayed values. Maybe even a label "GP: GHA/DEC=" would be helpful?

Indeed. I will add information to the user guide that's an introduction to CN and how Stellarium can be used to assist learning.

> In the Guide you can also recommend settings in the Config/Info tab to avoid display clutter for navigators (switch off distance, spectral class, or such...)

I will do that. I can also add a script that initializes the display. In fact, I added some in my PR that allows me to set up Stellarium to run the test cases where I manually reduced using the Almanac to check the numbers. It's also what I based the Unit tests on, a manual reduction.

With regards to the pairing of values such as gha/dec, alt/az, etc. I just spoke with one of my advisors and this was telling... "the problem is most CN students are not astronomers, they are sailors. Whilst astronomers are more than accustomed to coord pairs such as alt/az, CN students work with tables of data. That's why I like the tabulated form I see now. It's how it looks on paper and how they'll actually write it down in their logbooks. It's also how we expect to see it on exam papers. We aren't teaching astronomy but CN".

So, I'm a bit at a loss here on the correct display. I could certainly make it configurable to either see paired coords or tabulated. Your thoughts?

gzotti:
Hm, a config option "compact/separate" would of course be extra comfortable, to add the data without taking too many lines for "regular" ("somewhat interested astronomers") users. 
Your Nav students will then likely configure their program to disable all other astronomical data output, so splitting into separate lines may not be such an issue also on 1368x768 notebook screens.

A-j-K:
> Hm, a config option "compact/separate" would of course be extra comfortable, to add the data without taking too many lines for "regular" ("somewhat interested astronomers") users.
> Your Nav students will then likely configure their program to disable all other astronomical data output, so splitting into separate lines may not be such an issue also on 1368x768 notebook screens.

OK, I will add the ability to allow the user to configure how they want the data to be displayed. That's not a big issue. I will default to "astronomer view", ie paired coords as it is how most Stellarium users will expect. Are we ok with the duplication of data fields? Or should in "paired display view" omit duped fields?

I will start work on the User Guide shortly. LaTex is a learning curve for me. I expect writing the guide will take longer than the code took to write :)  I also need my contribution to the guide to be proofread for grammar and accuracy by others. Luckily I have access to people who can assist me in that. However, this will all take some extra time.

gzotti:
About "astronomer view": Maybe even an extra opt to "omit duplicates" that does not display LON/LAT and Zn/Hn. That may invite a combo box selection, but I don't know what looks best.

LaTeX: Maybe it's difficult to start a new project. However I hope it's not very difficult (conceptually it is approximately as difficult as HTML, just a decade older) to follow examples and add a few \section{...}, and else it's all set up to "make pdf" on the command line. Please see the README in the folder.

A-j-K:
@gzotti @alex-w 

Here are two screen grabs of the display. Which is used is as determined by the user in the config system. If you are happy with this I'll commit this code and move onto extending the About text to provide extra info. Once we are all settled on the UI section I will begin on the User Guide. 

Default View:-
Astroview

Tabulated view (selectable by config, **not** the default out of the box view)
Tableview

gzotti:
I think in the first screen the tags should also be capitalized (like in the Almanac). Else, from my side yes please proceed.

A-j-K:
@gzotti I also noticed GHA doesn't have the leading + sign. I have added that (handling sign is an important part of the sight reduction process and should always be specified, even if the value is always positive). I will capitalize as you suggest.

A-j-K:
> I think in the first screen the tags should also be capitalized (like in the Almanac). Else, from my side yes please proceed.

I decided that SHA was important enough to add to this view. Screengrab added for reference.

Astroview2

A-j-K:
D'oh, I broke it. Fix coming.

axd1967:
@A-j-K I have a core dump with 8d5c2670955f9ba33f4f89b4eda1ce01f4b659ab - I try to check this ASAP.

alex-w:
I've found warning also:
> /home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:448:30: warning: variable ‘infoGroup’ set but not used [-Wunused-but-set-variable]
  StelObject::InfoStringGroup infoGroup = StelObject::OtherCoord;

A-j-K:
> @A-j-K I have a core dump with [8d5c267](https://github.com/Stellarium/stellarium/commit/8d5c2670955f9ba33f4f89b4eda1ce01f4b659ab) - I try to check this ASAP.

Are you building in Release mode or Debug mode? I have found there seems to be a problem in Qt5 libs/headers that causes a core dump/crash when building in Debug mode. I have a workaround but it's not pretty. I really need to isolate the actual cause and figure a proper solution.

axd1967:
My first test was 
```
cmake -DCMAKE_INSTALL_PREFIX=$PWD/ -DENABLE_GPS=0 -DENABLE_LIBGPS=0 -DCMAKE_BUILD_TYPE="RelWithDebInfo" $PROJECT
```
I am now testing with  Debug mode, new userdir, can't start as stellarium crashes immediately, I get this:
```
ASSERT: "parent" in file /home/alex/projects/stellarium/src/git-worktrees/ajk-nav/src/core/modules/SolarSystem.cpp, line 662
Aborted (core dumped)

```
I guess that assert is irrelevant, but I get no other output.

alex-w:
> My first test was
> 
> ```
> cmake -DCMAKE_INSTALL_PREFIX=$PWD/ -DENABLE_GPS=0 -DENABLE_LIBGPS=0 -DCMAKE_BUILD_TYPE="RelWithDebInfo" $PROJECT
> ```
> 
> I am now testing with Debug mode, new userdir, can't start as stellarium crashes immediately, I get this:
> 
> ```
> ASSERT: "parent" in file /home/alex/projects/stellarium/src/git-worktrees/ajk-nav/src/core/modules/SolarSystem.cpp, line 662
> Aborted (core dumped)
> ```
> 
> I guess that assert is irrelevant, but I get no other output.

It's fixed already in master.

A-j-K:
@axd1967  can you try a pure Release build please. That works for me, Debug builds are problematic. At least then I will know I need to continue looking into why Debug build have this issue.

A-j-K:
> I've found warning also:
> 
> > /home/aw/devel/stellarium/plugins/NavStars/src/NavStars.cpp:448:30: warning: variable ‘infoGroup’ set but not used [-Wunused-but-set-variable]
> > StelObject::InfoStringGroup infoGroup = StelObject::OtherCoord;

I will remove this on my next commit. 

@alex-w I am starting work on the User Guide portions. I'd like to know if you're happy with the UI as it stands now (given I added the ability to display CN data in either "Astronomer view" (default) or "Navigator view" (tabular). I'm asking as I prefer to work on documentation when the code is "stable" as it makes writing documentation somewhat less error prone than trying to do both at the same time.

gzotti:
> 
> @axd1967 can you try a pure Release build please. That works for me, Debug builds are problematic. At least then I will know I need to continue looking into why Debug build have this issue.

You should be able to make a 
`git rebase master
`followed by a 
`git push --force
`
so that the latest changes from master (including the bugfix around that too strict assert) will become effective. You should do that anyway before we can merge.
Alternately, just comment line 662 (mentioned in the Assert dump) away.

axd1967:
log files before enabling the plugin, and after restart, activating navstar plugin and selecting **any** sky object 

[log-start.txt](https://github.com/Stellarium/stellarium/files/4147928/log-start.txt)
[log-crash.txt](https://github.com/Stellarium/stellarium/files/4147927/log-crash.txt)

all that the console says is "Segmentation fault (core dumped)".

axd1967:
( @A-j-K  FYI - I also tested the base afe4a846a00044234727bd178c9a33df8ed9257b  of that branch, that seems to work fine for navstars)

alex-w:
The source of crash in method StelObject::removeExtraInfoStrings(); - please rebase to master

A-j-K:
> The source of crash in method StelObject::removeExtraInfoStrings(); - please rebase to master

I was going to put a PR in for this very thing this evening. I will rebase this evening.

axd1967:
a2b8150e40 seems to work.

a few obs:

I compared with Air Almanac (AA) data  2020-01-02 1200UTC and noticed some diffs (I did not check all diffs)

image


Sun:
image


* GHA difference 15"
* DEC diff 4'

Venus:
image

* DEC diff 4'

notice also that the Dec value corresponds to the AA value!

Moon:
image
* I'm not sure why there are such significant diffs with AA
* the moon rise/set values are intriguing, should not be related to this PR, but might be related with the previous diffs

Enif:
image

image
* Dec diff 1.4' 

also
* Pluto is not a navigational source and should not return data. Only Venus/Mars/Jupiter/Saturn are used
* The tabular data in the almanacs is computed for every N minutes, not for every second of the day. students might have to select the correct time and perform interpolation (unless their sighting corresponds to an exact table entry)
* consider inverting DEC/GHA in the output (should be GHA/DEC)
* the output should be completely aligned:
image

* I have a basic knowledge about astronav/cn and could help in the review of the manual (language as well as content).

A-j-K:
@axd1967 Thanks for the feedback.

* Regarding interpolation, students of CN are expected to be able to do this when required so having something to check against is good.
* I will swap GHA and DEC in the tabulated display as you noted.
* The plugin has no control over the alignment, that's handled by StelObject outwith my control. I did mention this to @gzotti that the HTML table view didn't look optimal. I didn't understand the reply.
* Regarding Pluto. I was being lazy to reduce the number of checks. Right now I do isSun or, isMoon or isPlanet. That last test is lazy. I will instead check all names.

Now the difficult part. I am using Nautical Almanacs and I too have seen some inaccuracies. I wanted to run my examples using the 1981 publication as it's the one used in teaching so many many examples online use 1981 (to save the Edu depts having to buy new ones every year).

It's on my list to look at the accuracy of the results. They are mostly "very close". However, they should be within 0.1' (acceptable) and I've seen up to 10' difference (unacceptable). My first thoughts were around some date calculations. It's possible that the UT1 difference from UTC over time does in fact contribute. However, it's stated to be always less than 1 sec from UTC. 

My next goal is going to be a bit grueling in that I run the numbers based on the coords and time that Stellarium supplies the plugin with. Am I using the wrong functions to get data? Or is the data inaccurate? This is what I will have to find out. And that will probably take me some time to crunch the numbers.

gzotti:
I don't know how often I have to repeat why Stellarium still bears version number 0.something. 

1. It does not correct for the aberration of light. This should explain something like around 20" inaccuracies in longitude for the sun, and probably similar values for the planets. Of course the USNO Almanach will therefore provide higher accuracy. 
2. It does not compute the orientation of the planets correctly. Forget about libration for now. I try to finish that first as it seemed moderately easy before I started, but am gritting my teeth on it.

It is all written in Appendix F in the User Guide and in https://github.com/Stellarium/stellarium/wiki/Research.

When these things will at some point have been solved, Stellarium should be accurate enough (I hope: arcseconds. I will not do relativistic effects, starlight bending near the sun or so) for "production cel nav", we will be able to drop the 0, and will have suddenly human resources to improve translations, howtos, skycultures, unexpected (but working) ocular behaviour or bookmark nitpickery and all the other funny ideas that tend to distract our attention from fixing the real accuracy problems.

About the tables: The extra strings of type "OtherCoord" are displayed below the line "Ecliptic obliquity (of date).", which is in itself not inside a table. That means, in table mode I open a table, then add all those strings which must provide HTML tr-td-td-td-td.....-/tr syntax (I don't show the correct tags here, md rendering might mess it up.) and close the table. This makes the cell count independent from the table above, so that these numbers will never align automatically.  The table cell widths are thus controlled by its contents, but as long as you are the only one to provide such extrastrings, you are free when it comes to cell count or other details. You may be able to force some cell width by embedding some css padding instructions. You know more about css than I do.  

I hope nobody else will ever add OtherCoords, or we will have additional mess.

alex-w:
@A-j-K please remove extra degree char for GHA value

alex-w:
I can add some tuning for alignment of additional coordinates after merge

alex-w:
I see all values has format DDMM.m, except GHA and LMST, who have format DDD.ddd - why? For example Air Almanac has format DDMM.m for GHA.

A-j-K:
> I see all values has format DDMM.m, except GHA and LMST, who have format DDD.ddd - why? For example Air Almanac has format DDMM.m for GHA.

I will reformat the GHA of aries to DDMM.m as per the Almanac. I will remove LMST as you won't derive/compute that from the Almanac, it's implicit within the table values. I actually had these like this to assist creating/verifying tests. However, given the previous comments on accuracy, I am redoing the unit test suite to make it easier to create unit tests based on the Almanac. I am doing this to explore the accuracy issues previously noted.

alex-w:
No need remove LMST, just change format to DDMM.m for it

alex-w:
Well, the last fix is needed - text for Stellarium User Guide

A-j-K:
> Well, the last fix is needed - text for Stellarium User Guide

I am starting on the User Guide now. 

But I also need to investigate the accuracy of the results compared to real world Almanac sight reductions. They should match within +/-0.1 minute of arc.

gzotti:
> But I also need to investigate the accuracy of the results compared to real world Almanac sight reductions. They should match within +/-0.1 minute of arc.

They can't. Be happy if they are within 0.33' now, and when we drop the 0 after the aberration fix, we should be within (hopefully) 2''.

gzotti:
What does the "Only highlight when visible" do? 
You may want to add tooltips to the checkboxes for a short help.

A-j-K:
With Stellarium today the highlight circles are drawn around the Nav stars all the time. This includes during the daytime when the star it is highlighting is not visible. Checking this box will only draw the highlight circle if the star is visible in the sky. So, as dawn approaches are stars start to appear, Nav stars are highlighted at that point and not permanently stuck on.

I will add tooltips.

gzotti:
Ah, now I see it. Good feature, thanks! Tooltips will certainly help.

plugins/NavStars/src/NavStars.cpp
```diff
@@ -469,7 +469,7 @@ void NavStars::extraInfo(StelCore* core, const StelObjectP& selectedObject, bool
 		.setJde(jde)
 		.setGmst(get_mean_sidereal_time(jd, jde));
 
-	StelUtils::rectToSphe(&x, &y, selectedObject->getEquinoxEquatorialPosApparent(core));	
+	StelUtils::rectToSphe(&x, &y, selectedObject->getEquinoxEquatorialPos(core));	
 	calc.setRaRad(x)
```
gzotti: Just to be clear on that: "...Apparent()" here means "including atmospheric refraction". To control this via atmosphere switch, you can use "...Auto()". Now refraction is switched off.

A-j-K: Indeed, I didn't want refraction here at all which is why I didn't use Auto().

larawag:
@A-j-K

Within the last days I had the opportunity to take a deeper look into your code. I still like it but now I have some comments and questions:

You supply the computed altitude of the selected celestial object on the main screen, cf. Hc/Zn. If Hc is indeed the computed altitude and azimuth as a solution of Sight Reduction, I cannot see where to define the height of the observer over the sea level. This could be done in the NavStars plugin window. Also I expected the horizontal parallax HP (for Moon or Planets) and the semi-diameter R (for Moon or Sun) in the main window for the selected object.

In addition to that I miss the current time and date of Greenwich on the main screen. From my perspective showing this on the main window is mandatory, especially in cases where your current location is in the vicinity of the 180th degree of longitude.

Maybe you already discussed this issue but when I learned CN we were forced to use the format xx° xx.x' for LAT, Dec, etc. and xxx° xx.x' for LON, etc. and we were forced to write a leading zero (if existing). The idea behind this was to avoid mistakes and so I suggest to use this format also in Stellarium - also for teaching a more robust way to your students and to me ;-) Additionally the data can be vertical aligned on the screen more easily.

To my feeling there is some additional code necessary before a merge into Stellarium is useful. That is the reason why I retracted my pull request #812 last year. At begin of January 2020, you mentioned the first time that you were going to do first steps for implementing CN in Stellarium whereas I already was doing the same. So I tried to ask you for collaboration but I did not get any clear answer. What do you think now about bundling our work? I do have some ideas what I would expect a CN plugin should can do. 

To be consistent to the other lines in `NavStarsCalculator.cpp` you can drop in lines 74 and 75 the `std::` namespace command.

If you need some \LaTeX{} support just let me know.

A-j-K:
@larawag 

> You are going to supply the computed altitude of the selected celestial object, cf. Ho in NavStar plugin window. If Ho is indeed the computed observed altitude and azimuth as a solution of Sight Reduction, I cannot see where to define the height of the observer over the sea level. This could be done in the NavStars plugin window. Also I expected the horizontal parallax HP (for Moon or Planets) and the semi-diameter R (for Moon or Sun) in the main window for the selected object.

First off, I'm displaying Ho, Hc and Zn. As a begin writing the User Guide update I am now toying with the wisdom of displaying Ho at all because it doesn't add anything except confusion (your observation is evidence of just that). So it's likely I will remove it as I make progress with the manual.

As for height of the observer, that's merely a function of what the user chooses in the Location window. Again, there will be information about this in the guide. 

> In addition to that I miss the current time and date of Greenwich on the main screen. From my perspective showing this on the main window is mandatory, especially in cases where your current location is in the vicinity of the 180th degree of longitude.

It is shown? UTC is the first line added in the infoStrings displayed. 

> Maybe you already discussed this issue but when I learned CN we were forced to use the format xx° xx.x' for LAT, Dec, etc. and xxx° xx.x' for LON, etc. and we were forced to write a leading zero (if existing). The idea behind this was to avoid mistakes and so I suggest to use this format also in Stellarium - also for teaching a more robust way to your students and to me ;-) Additionally the data can be vertical aligned on the screen more easily.

Agreed, the leading zero where required is important. I will add that.

> To my feeling there is some additional code necessary before a merge into Stellarium is useful. That is the reason why I retracted my pull request #812 last year. At begin of January 2020, you mentioned the first time that you were going to do first steps for implementing CN in Stellarium whereas I already was doing the same. So I tried to ask you for collaboration but I did not get any clear answer. What do you think now about bundling our work? I do have some ideas what I would expect a CN plugin should can do.

I'd be more than happy to collab. The only reason I continued on was I probably found myself with more time on my hands. But yeah, please do! :) 

The one point I do want to achieve is based on feedback I had from CN instructors. Make sure it's all on the screen. If you have to open another window and start entering data it's not going to be used. What they were looking for was a supporting role. They can "pretend" to be making an observation (that's what Stellarium excels at, it looks superb). But they just wanted the intermediates 
 (such as gha) and the solution (Hc/Zn) right there on the screen. "Instant feedback". Their focus is on the use of the Almanac, not Stellarium. So listing the solution right there without a faff is important. Instructors noted that the use of scripting was a feature they could use to set up a session. I have also used this when testing :)

> To be consistent to the other lines in `NavStarsCalculator.cpp` you can drop in lines 74 and 75 the `std::` namespace command.

Will do.

> If you need some \LaTeX{} support just let me know.

 It doesn't look too bad. \LaTeX{} isn't the issue, writer's block is ;)

--Andy

alex-w:
I see that fixes for guide are in separate branch, so, I'll merge this branch.

A-j-K:
@alex-w Ah yes, I decided to use a separate branch for guide mainly because I have created a number of "temporary" tools to create graphic for the manual. Some of these tolls will turn out to be "not temporary" as I'll want to keep them in case I need to adjust the manual in future (which is possible) . However, I found that these were "cluttering up" my otherwise clean code branch. So I spun that off to work on as a seperate "mini project". I hope to have that done soon. However, for me, code is easier than manuals so I need endless cycles of proofreading. I'd like to make sure my contribution to the User Guide is the best effort.

axd1967:
@A-j-K please shout when you want a review of the manual.

A-j-K:
> @A-j-K please shout when you want a review of the manual.

@axd1967 

I will open a new PR for the guide and I will tag you on it when I do. Thank you for your time and effort, it's appreciated.

A-j-K:
@axd1967 

I should have asked, do you have the ability to compile the manual or would you like me to provide temp links to download the PDF from? I can make them available short term if required unless you can compile it yourself.

axd1967:
@A-j-K I'm new to LaTex, am stumbling across a `make` error 
```! LaTeX Error: File `tikz.sty' not found.``` so I'm now figuring out how to get this tikz package (which apparently drags me into a LaTex update horror story).

don't produce binary (PDF) documents that cannot be annotated (yes, I know it is possible, but nevertheless, that is not manageable). and something that does not contain diffs is impossible to review. so my only choice is to do a diff of these commits (which I now see have already merged ............... **so where can I find all documentation changes related to this issue?** ) and understand this LaTex beast (how elegant it might be). I am trying TexStudio atm.

(so I could ask @gzotti the same question he asled in https://github.com/Stellarium/stellarium/pull/960 ... _what ~editor~ IDE are you using_? but that is an academical question)

https://tex.stackexchange.com/questions/55437/how-do-i-update-my-tex-distribution
https://tex.stackexchange.com/questions/2044/how-to-install-a-current-version-of-tikz

**Comment by gzotti** [comment]
@axd1967 You could read the README.md which resides in the guide directory about the tools that have been used successfully.

**Comment by axd1967** [comment]
I have read that READM...

**Comment by alex-w** [comment]
@axd1967 just try `sudo apt install libpgf6`

**Comment by axd1967** [comment]
thanks - still the same error.
I don't like to install random packages without understanding what goes wrong... StackOverflow is full of well-meant, but wrong suggestions, that end up making a system unstable.
I'll review the diffs for now.

**Comment by alex-w** [comment]
I just use TeXLive from tug.org - https://tug.org/texlive/quickinstall.html

**Comment by axd1967** [comment]
@A-j-K you can find a first review here: https://github.com/axd1967/stellarium/compare/navstars_improvements_guide...axd1967:contrib/review-ajk-doc?expand=1

@alex-w thx but I still get errors. don't worry, I wil try to find out but it will take time, I'll try to update the README too.
image
(and having to install 6 Gb that will take an hour to download/install is **flabbergasting** ... I bet I don't need 99% of this...)

**Comment by gzotti** [comment]
Last time I configured a system it worked out of the box with these packages:

```
sudo apt-get install texlive-base texlive-bibtex-extra texlive-latex-recommended \
texlive-latex-extra texlive-pictures texlive-fonts-recommended biber make
```

axd1967:
yes, that's how it is written in the README. but I first want to untangle whatever LaTex was installed via TexStudio.

A-j-K:
If it's any help I created a Docker image with all the tools required to build the user guide. I then just have Docker Engine installed on my Ubuntu VM and building it becomes simple. I'm actually going to add instructions and a quick way to build the manual using Docker.

As for the manual, I'll be making a new PR shortly to get it rolling.

axd1967:
that is an interesting solution! 

but I see a (for a Linux user) suspect symlink to an .exe in https://github.com/A-j-K/base-dockers/blob/1962c3e9fa7828837528f74420a8d45dae2f018a/stellarium_build/Dockerfile#L63-L64 (wonder if that would work on other systems).

tip: you might also want to separate the packages for stellarium compilation from the packages to build the docs (https://github.com/A-j-K/base-dockers/blob/1962c3e9fa7828837528f74420a8d45dae2f018a/stellarium_build/Dockerfile#L6)

but you just gave me an idea; I am more of a Vagrant user, as that allows me to completely isolate the VM environment from the host env, while Docker relies on host OS - in your case Windows.

A-j-K:
@axd1967 
The reason the symlink is there is because the Makefile for the User Guide expects the executable to be named that way. 

Yes, Vagrant is another option. I just use VMWare to run Ubuntu and install Docker Engine. Then, I build all my tools as images as it makes running them simple without having my Ubuntu image polluted with 1000s of package installs. Each Docker image contains what it needs to get the job done. It also means my building environments are always the same unless I update the image. Just makes being consistent and repeatable a lot easier.

PR #1048, opened 2020-04-14
INDI: Implement telescope syncing

akosiaris:
### Description
Up to now, telescope sync when INDI was used wasn't implemented. Add
support for it. Add a new syncPosition() procedure that will first set
the SYNC ON_COORD_SET switch member and then send the coordinates. Make
sure to unset the switch member at the end.

While at it, make sure that setPosition() procedure is also setting the
TRACK member of the ON_COORD_SET switch to avoid undesired state before
sending the coordinates

Implement the corresponding client function that was a noop up to now

### Type of change

### How Has This Been Tested?

**Test Configuration**: flatpak build based on the official flatpak manifest. 
* Operating system: Debian buster, already running Stellarium using the flatpak build.
* Graphics Card: Intel Iris Graphics 5100. Whatever driver version currently ships with Debian Buster.

## Checklist:

### DIFF
--- a/plugins/TelescopeControl/src/INDI/TelescopeClientINDI.cpp
+++ b/plugins/TelescopeControl/src/INDI/TelescopeClientINDI.cpp
@@ -94,8 +94,26 @@ void TelescopeClientINDI::telescopeGoto(const Vec3d &positionJ2000, StelObjectP 
 void TelescopeClientINDI::telescopeSync(const Vec3d &positionJ2000, StelObjectP selectObject)
 {
 	Q_UNUSED(selectObject)
-	Q_UNUSED(positionJ2000)
-	return;
+	const StelCore* core = StelApp::getInstance().getCore();
+	Vec3d posRectJNow = core->j2000ToEquinoxEqu(positionJ2000, StelCore::RefractionOff);
+	Vec3d posJ2000;
+	StelUtils::rectToSphe(&posJ2000[0], &posJ2000[1], posRectJNow);
+
+	if (posJ2000[0] < 0.0)
+		posJ2000[0] += 2*M_PI;
+
+	double longitudeRad = posJ2000[0] * 12.0 / M_PI;
+	double latitudeRad = posJ2000[1] * 180.0 / M_PI;
+
+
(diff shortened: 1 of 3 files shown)


akosiaris:
Greetings fails with:
```
 Adding message: Thanks for adding your first pull request to Stellarium. If you have questions, please do not hesitate to contact us. to pull request 1048
##[error]Resource not accessible by integration
```

I have no idea why though.

alex-w:
Are you checked it on real hardware?

akosiaris:
> Are you checked it on real hardware?

No, I haven't found the time yet. I 'll do so once the clouds are gone.

plugins/TelescopeControl/src/INDI/TelescopeClientINDI.cpp
```diff
@@ -94,8 +94,26 @@ void TelescopeClientINDI::telescopeGoto(const Vec3d &positionJ2000, StelObjectP
...
+	double latitudeRad = posJ2000[1] * 180.0 / M_PI;
+
+
+	INDIConnection::Coordinates positionJNow = mConnection.position();
+	positionJNow.RA = longitudeRad;
+	positionJNow.DEC = latitudeRad;
+
+	// unpark telescope before slewing!
+	// TODO: Add commands and buttons for park/unpark telescope for all telescopes
+	mConnection.unParkTelescope();
```
d33psky, line 115: I do not think this is the place to call *this* particular command.
It's fine to 'make sure' the mount is unparked before giving it a sync command in another way, but we should not unpark at every sync.

alex-w: The mount will not unpark here if she is unparked already. But possible we should add 2 buttons (Park, Unpark) for all telescopes to allow better management of mounts.

akosiaris:
Managed to find the weather and time to test this. Unfortunately it turns out that the traffic flowing to indiserver from stellarium did not correctly set the SYNC member of ON_COORD_SET switch. I 've submitted #1065 as a tested fix.

PR #1198, opened 2020-08-07
Add Full Sensor Grid to Ocular plugin

mk-dhia:
### Description
Inside Crop overlay add a new checkbox to have the full sensor grid so we can see pixel width on the image and have a fair estimate of detail. I fund my self playing with crop sizes on image to estimate what detail level i would expect from my setup. Now you can see immediately the projected pixel size, check how many pixel is the object
in the process i made a change to the crop to add an upper limit of the sensor size 
Also i have to round the crop size to be multiple of binned pixel

I have changed naming of crop size as it was using Hsize/VSize whereas sensor/binning/fov do use X/Y so i opted for X/Y naming in my variable and i can add a commit to rename all ccdCropOverlayHSize,... to ccdCropOverlayX,... (or use HSize and VSize in my variable name, what ever you prefer)

PS: i'm not used to c++/Qt and i m willing to do the commit correctly (if you find idea worthy) do not hesitate to tell me what to fix/improve

(new option is show pixel grid)
image
with higer resolution sensor zooming in
image

### Type of change

### How Has This Been Tested?
unit test ok
manual test on y laptop

**Test Configuration**:
* Operating system: <Linux Mint 19.2>
* Graphics Card: <Intel, kernel 5.3> 1920x1080

## Checklist:

### DIFF
--- a/plugins/Oculars/src/Oculars.cpp
+++ b/plugins/Oculars/src/Oculars.cpp
@@ -181,6 +181,7 @@ Oculars::Oculars()
 	, equatorialMountEnabledMain(false)
 	, reticleRotation(0.)
 	, flagShowCcdCropOverlay(false)
+	, flagShowCcdCropOverlayPixelGrid(false)
 	, ccdCropOverlayHSize(DEFAULT_CCD_CROP_OVERLAY_SIZE)
 	, ccdCropOverlayVSize(DEFAULT_CCD_CROP_OVERLAY_SIZE)
 	, flagShowContour(false)
@@ -613,6 +614,7 @@ void Oculars::init()
 		relativeStarScaleCCD=settings->value("stars_scale_relative_ccd", 1.0).toDouble();
 		absoluteStarScaleCCD=settings->value("stars_scale_absolute_ccd", 1.0).toDouble();
 		setFlagShowCcdCropOverlay(settings->value("show_ccd_crop_overlay", false).toBool());
+		setFlagShowCcdCropOverlayPixelGrid(settings-> value("ccd_crop_overlay_pixel_grid",false).toBool());
 		setCcdCropOverlayHSize(settings->value("ccd_crop_overlay_hsize", DEFAULT_CCD_CROP_OVERLAY_SIZE).toInt());
 		setCcdCropOverlayVSize(settings->value("ccd_crop_overlay_vsize", DEFAULT_CCD_CROP_OVERLAY_SIZE).toInt());
 		setFlagShowContour(settings->value("show_ocular_contour", false).toBool());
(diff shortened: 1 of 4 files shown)


alex-w:
Thanks for patch!

You marked item “bug fix” - which bug you meant?

alex-w:
You use term “full grid”, but by fact you propose the patch, which show pixel grid - why you did not used term “pixel grid” for methods and tooltip?

P.S. The code require cosmetic changes (a “stylish” changes) and one small addition to support a shortkey to toggle pixel grid.

gzotti:
I tried to set a box size of 32x32 pixels. However, I see the box on screen labeled 32x30, and rendered as 32x10. (i.e., vertically elongated pixels.)
... 
Ah, that was a Binning setting. Maybe this should be somehow added to the box label? ("32px x 30px (Binning 1x3)")

alex-w:
@gzotti I can add binning info into GUI bar of plugin (top right corner of the screen)

gzotti:
I think you have more space in the right bottom corner under the grid box. But either solution is OK.

alex-w:
> I think you have more space in the right bottom corner under the grid box. But either solution is OK.

Yes + I'll preferred to show main properties of CCD near other similar data. In any case showing a binning data is task not for this PR.

gzotti:
I think "full grid" is just in comparison against "just a box". 's/full grid/pixel grid/g'
Ticking off "bug" is IMHO a cosmetic glitch in the PR request.

mk-dhia:
> Thanks for patch!
> 
> You marked item “bug fix” - which bug you meant?

The crop goes beyond the sensor size if you have a sensor of 1000x1000 and you set crop of 2000x2000 the "croped" part will be larger than sensor

mk-dhia:
> I think "full grid" is just in comparison against "just a box". 's/full grid/pixel grid/g'
> Ticking off "bug" is IMHO a cosmetic glitch in the PR request.

Ill fix that

mk-dhia:
> You use term “full grid”, but by fact you propose the patch, which show pixel grid - why you did not used term “pixel grid” for methods and tooltip?
> 
> P.S. The code require cosmetic changes (a “stylish” changes) ...
If you put them in review i fix them or tell me what i should look for
> ..and one small addition to support a keyboard shortcut to toggle pixel grid.
I m not sure that i master  how to add that but i think i could manage it. is there a guidance on what default shortcut shall i put?

plugins/Oculars/src/Oculars.cpp
```diff
@@ -1596,8 +1605,23 @@ void Oculars::paintCCDBounds()
 					a = transform.map(QPoint(static_cast<int>(overlayWidth*0.5f), static_cast<int>(overlayHeight*0.5f)));
 					b = transform.map(QPoint(static_cast<int>(overlayWidth*0.5f), static_cast<int>(-overlayHeight*0.5f)));
 					painter.drawLine2d(a.x(), a.y(), b.x(), b.y());
+					
+					//Tool to show full CCD grid overlay
+					if(flagShowCcdCropOverlayPixelGrid) {
```
alex-w, line 1610: Please move brace on new line

plugins/Oculars/src/Oculars.cpp
```diff
@@ -1596,8 +1605,23 @@ void Oculars::paintCCDBounds()
...
+						//vertical lines
+						for (int l =1 ; l< actualCropOverlayX/ccd->binningX(); l++ ){
```
alex-w, line 1612: Please move brace on new line

plugins/Oculars/src/Oculars.cpp
```diff
@@ -1596,8 +1605,23 @@ void Oculars::paintCCDBounds()
...
+							a = transform.map(QPoint(static_cast<int>(overlayWidth*0.5f- l*pixelProjectedWidth), static_cast<int>(-overlayHeight*0.5f)));
+							b = transform.map(QPoint(static_cast<int>(overlayWidth*0.5f- l*pixelProjectedWidth), static_cast<int>(overlayHeight*0.5f)));
+							painter.drawLine2d(a.x(), a.y(), b.x(), b.y());
+						}
+						//horizontal lines
+						for (int l =1 ; l< actualCropOverlayY/ccd->binningY(); l++ ){
```
alex-w, line 1618: Please move brace on new line

alex-w:
Hmm... should be crop box have a size 2x if binning 2x2 is used? Looks like bug in our old code. @gzotti ?

alex-w:
> I m not sure that i master how to add that but i think i could manage it. is there a guidance on what default shortcut shall i put?

I’ll add this part after merge.

gzotti:
> 
> 
> Hmm... should be crop box have a size 2x if binning 2x2 is used? Looks like bug in our old code. @gzotti ?

According to the GUI label "Overlay size (px)", native pixels are OK. Or relabel to "Overlay size (binned px)" and define the box in current binning settings. For the grid, it is indeed good to show the pixel grid with current binning setting (and clarify also in the screen label). Not sure what other software titles have here, or whether there is any standard recommendation to follow?

alex-w:
I have a strange question - why pixel grid do not use a full CCD frame?

mk-dhia:
> I have a strange question - why pixel grid do not use a full CCD frame?

That was my first implem. however you get red rectangle when you are zoomed out (unless you use very low resolution sensor ). i found it more useful when just applied to the crop so i still see the global image

alex-w:
OK, let’s merge it and ask users to test this feature  with feedback :)

gzotti:
Yes, I agree using the grid only for the crop area is better. Its application is likely planetary hi-res work.

PR #1206, opened 2020-08-09
Ocular plugin improve for sensor grid in overlay crop

mk-dhia:
### Description
Following #1198
Modify the cross in the center , rotating it by 45° so we can distinguish we do not confuse line of the sensor grid with line from cross (i do not know if we consider this a breaking change, i see what was needed was the location of center and not the axes of cross https://bugs.launchpad.net/stellarium/+bug/1404695 )
Add a text when the crop size is different from the one inputted this could be the case for 2 reason: the crop size is larger than sensor size (but there is a ratio limit that prevent showing any message anyway. the second case when crop size is not multiple of binning (example 100px with binning 3, the crop size will be 99px). moved the resolution from bottom right to left and show the message "Adjusted for binning" in the bottom right when the crop size is adjusted.

Fixes # (issue)

Crop 131x131 binning 4 both axes crop size adjusted to 128x128
image
When binning is 1 in both axes (no adjustment done)
image
When zooming in you do net get confused by the line of the cross
image
PS: Sorry for the low quality screenshot. when cropping image manually lines get compressed in wired way
### Type of change

- [x] (Meh!) Breaking change (fix or feature that would cause existing functionality to change) 

### How Has This Been Tested?
Unit tests passes
manual testing

**Test Configuration**:
* Operating system: Linux Mint 19;2
* Graphics Card: Intel, Kernel 5.4 res 1920x1080

## Checklist:

### DIFF
--- a/plugins/Oculars/src/Oculars.cpp
+++ b/plugins/Oculars/src/Oculars.cpp
@@ -1724,8 +1724,15 @@ void Oculars::paintCCDBounds()
 								.arg(QString::number(actualCropOverlayY, 'd', 0))
 								.arg(qc_("px", "pixel"))
 								.arg(QChar(0x00D7));
-						a = transform.map(QPoint(qRound(overlayWidth*0.5f - painter.getFontMetrics().width(resolutionOverlayText)*params.devicePixelsPerPixel), qRound(-overlayHeight*0.5f - fontSize*scaleFactor)));
+						a = transform.map(QPoint(qRound(-overlayWidth*0.5f), qRound(-overlayHeight*0.5f - fontSize*scaleFactor)));
 						painter.drawText(a.x(), a.y(), resolutionOverlayText, static_cast<float>(-(ccd->chipRotAngle() + polarAngle)));
+						if(actualCropOverlayX!=ccdCropOverlayHSize || actualCropOverlayY!=ccdCropOverlayVSize)
+						{
+							//we only show the message if crop is smaller than sensor anyway (ratioLimitCrop). so only binning cause adjustment message to be chown
+							QString adjustedCropText = qc_("Adjusted for binning", "abbreviated in the plugin");
+							b = transform.map(QPoint(qRound(overlayWidth*0.5f - painter.getFontMetrics().width(adjustedCropText)*params.devicePixelsPerPixel), qRound(-overlayHeight*0.5f - fontSize*scaleFactor)));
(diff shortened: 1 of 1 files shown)


alex-w:
Rotation of cross is OK, but using string "Adjusted for binning" on bottom right side is problematic for 2 reasons: the string and resolution info is overwritten (so, unusable) + average word on some lanuages is more length than English average length of words. Of course using abbreaviation will be unusable too, because many users just will not understand this abbreviation. I think the string "Adjusted for binning" may be move to upper side of crop overlay and limit for visualization of info adjusted too. Or just add the special marker a additional to resolution info and add some notes for this marker into Help window and Stellarium User Guide.

@gzotti please your opinion here...

gzotti:
Agreed, it should only be 1 string. Maybe "128px x 64px @ 1x2", with the "@ ..." only added if binning is not 1x1.

alex-w:
@gzotti this is not binning notification, but adjust of crop overlay notification ;) For example binning 2x2 for crop 100x100 will not cause adjusting, but not binnings 2x3, 3x2 or 3x3.

gzotti:
Yes, OK. Just formulate something meaningful and short that explains a mismatch between configured and displayed pixel raster. Whatever it is, it should go into one corner, to avoid an overlap of left/right corners when the box is small. Try to avoid too much text on screen.

mk-dhia:
* i could go with "*" on size and some documentation (need pointer to doc to update)
* another option would be in the configuration dialog show a message if the chosen value is not aligned with binning, so in case of mismatch when you are back in the config tab you see a message say real size will be xx xx accounting for binning. i probably need some help to do that
* last option i see, is to keep crop size with pixel at edge cut.

alex-w:
I prefer first solution. Please see method OcularDialog::initAboutText() and /guide/plg_atTheTelescope.tex file to update documentation.

@gzotti ?

alex-w:
OK, I've added your fix for cross with small additional improvement and I've added a special marker for text in adjusted for binning mode (with additionals for About dialog). So, this PR was superseded by commit fb3f30a970c8151fb64aee7883d086b59efb135b and I'll close it.

Thanks for patches and ideas!

mk-dhia:
than

> OK, I've added your fix for cross with small additional improvement and I've added a special marker for text in adjusted for binning mode (with additionals for About dialog). So, this PR was superseded by commit [fb3f30a](https://github.com/Stellarium/stellarium/commit/fb3f30a970c8151fb64aee7883d086b59efb135b) and I'll close it.
> 
> Thanks for patches and ideas!

Thanks @alex-w . i had a tough work week i was planning doing modification today. sorry for the delay

PR #1220, opened 2020-08-17
Remove QSignalMapper

gzotti:
Thanks to Guillaume to find the right syntax. This branch should now compile on all CI platforms. 

### Description

In preparation for the upcoming Qt6 we must get rid of several obsolete classes. One is QSignalMapper, used for signal/slot connections with several senders and trivial arguments. These can be avoided by using Lambda functions. I could cleanup the Oculars plugin in https://github.com/Stellarium/stellarium/commit/eb7b297d4821cfa98356a81850883d07b39c1cc7, but further changes are more difficult.

Currently there are only 5 files affected:
- StelActionMgr.{h|cpp}
- StelModule.{h|cpp}
- StelScriptMgr.cpp

As soon as I try to use the new variant of addAction() in StelScriptMgr.cpp, I trigger link errors. Any solution to this?
The branch should be used to first solve this, and then remove the last QSignalMappers in the TelescopeControl plugin.

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10, Qt 5.13, MSVC2017. 

This must compile on all CI platforms, obviously. 

## Checklist:

### DIFF
--- a/plugins/TelescopeControl/src/TelescopeControl.cpp
+++ b/plugins/TelescopeControl/src/TelescopeControl.cpp
@@ -189,7 +189,6 @@ void TelescopeControl::init()
 			shortcut = QString("Ctrl+Shift+Alt+%1").arg(i);
 			text = q_("Abort last slew command of telescope #%1").arg(i);
 			addAction(name, section, text, this, [=](){abortTelescopeSlew(i);}, shortcut);
-
 		}
 		connect(&StelApp::getInstance(), SIGNAL(languageChanged()), this, SLOT(translateActionDescriptions()));
 
--- a/plugins/TelescopeControl/src/TelescopeControl.cpp
+++ b/plugins/TelescopeControl/src/TelescopeControl.cpp
@@ -57,7 +57,6 @@
 #include <QString>
 #include <QStringList>
 #include <QDir>
-#include <QSignalMapper>
 
 #include <QDebug>
 
@@ -160,11 +159,6 @@ void TelescopeControl::init()
 		reticleTexture = StelApp::getInstance().getTextureManager().createTexture(":/telescopeControl/telescope_reticle.png");
 		selectionTexture = StelApp::getInstance().getTextureManager().createTexture(StelFileMgr::getInstallationDir()+"/textures/pointeur2.png");
 
-		QSignalMapper* slewObj = new QSignalMapper(this);
-		QSignalMapper* syncObj = new QSignalMapper(this);
-		QSignalMapper* slewDir = new QSignalMapper(this);
-		QSignalMapper* abortSlew = new QSignalMapper(this);
-
 		//Create telescope key bindings
 		/* StelAction-s with these key bindings existed in Stellarium prior to
 			revision 6311. Any future backports should account for that. */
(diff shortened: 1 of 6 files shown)


guillaumechereau:
I will look at this in more details this week.
Since the implementation is using std::function, wouldn't it be easier to directly use it in the header instead of a template?

guillaumechereau:
On a side note: is QSignalMapper really going to be deprecated in Qt?  I can't find information about it online?

gzotti:
Thank you! Any solution that has QSignalMapper removed and replaced by Lambda connections is fine. The Template may be required to use that "Functor" name,  but if you know how to specify a Lambda function as argument in StelAction::connectToObject(QObject* obj, Functor functor), please do what is required. :-)

Do you mean just to use std::function as argument? Great if that would really be so simple! I have never seen std::function...

> On a side note: is QSignalMapper really going to be deprecated in Qt? I can't find information about it online?

The Qt help files describe it as obsolete ("kept in to let old code work") without giving clear instructions for replacement. This is a sure candidate for removal. And we were using it only in 3 (now: 2) modules.  But you find tons of discussion since ~2017 (?) how to replace it by Lambdas with C++11 on various fora (stackexchange &friends). Lambdas were new to me, but are pretty neat. However, template syntax was never my best friend. I tried with inline in header, or no inline with separate implementation, but none was successful so far. Compiler error message called for an explicit instantiation (or similar) missing, but how to specify that properly in this case was beyond my expertise. I gained some ideas with Qt's own qmenu sources and action connections, but just not enough.

guillaumechereau:
One possible issue with lambda is that the connection might not be removed when the connected object is destroyed, whereas with Qt signal this is taken care of automatically.

As for template vs std:function I guess it is a matter of style.  I think I prefer to use std:function in the headers as the type for the lambda instead of a template.  It is a bit less generic but make the code easier to read in my opinion.  Not sure if that would pose other issue though.

Otherwise, the only usage of QSignalMapper I see is in StelScriptMgr.  Wouldn't it be possible to remove it without changing current StelAction api?

gzotti:
You can specify a context object. AFAICR (from some forum) the Lambda will then not be called if this is deleted.  Please see the help entry for  QMetaObject::Connection QObject::connect(const QObject *sender, PointerToMemberFunction signal, const QObject *context, Functor functor, Qt::ConnectionType type = Qt::AutoConnection)
 
How do you specify the function argument? as "std::function<void()> functor" ?

As I wrote above, we use the last QSignalMappers in StelScriptMgr and in the TelescopeControl  plugin.

guillaumechereau:
OK.  I got a simple implementation that seems to work:

```
StelAction* StelActionMgr::addAction(const QString& id, const QString& groupId, const QString& text,
				      std::function<void()> lambda,
				      const QString& shortcut, const QString& altShortcut,
				      bool global)
{
	StelAction* action = new StelAction(id, groupId, text, shortcut, altShortcut, global);
	connect(action, &StelAction::triggered, lambda);
	return action;
}
```

Then the code in StelScriptMgr.cpp becomes:

```
void StelScriptMgr::initActions()
{
	StelActionMgr* actionMgr = StelApp::getInstance().getStelActionManager();
	for (const auto& script : getScriptList())
	{
		QString shortcut = getShortcut(script);
		QString actionId = "actionScript/" + script;
		auto callback = [this, script] { runScript(script); };
		actionMgr->addAction(
				actionId, N_("Scripts"), q_(getName(script).trimmed()), callback, shortcut);
	}
}
```

I haven't checked in the telescope plugin but I think it should work as well.

gzotti:
@guillaumechereau Brillant! That did it :-)

Issue #84, opened 2018-01-22
lunar positions mismatch


alex-w:
### Bug Description
There is a slight error in our ELP2000 lunar positions. I am pretty sure it has been here for years, only now we want to go into higher accuracy.

Compare photo and simulation from the same location. The star (at 2 o'clock in the photo) reappeared seconds before 04:52:12 where Stellarium has it reappearing more than a minute earlier. Camera time was corrected to NTP-adjusted PC, so cannot be more than 1-2sec off. DeltaT setting default near 69s.

Obviously related:
* Eclipse 2017 Bug https://bugs.launchpad.net/stellarium/+bug/1275092

Likely also related:
* Forum discussion https://sourceforge.net/p/stellarium/discussion/278769/thread/374439f7/#9c2b/5620/0694/19fe/9a9b/02a7/e123/e498/fe8b/fb2a

### Original report
https://bugs.launchpad.net/stellarium/+bug/1509659

gzotti:
To be re-confirmed after implementation of aberration...

gzotti:
![TLE20150928_reference](https://user-images.githubusercontent.com/18099873/124269257-f3cfcc80-db3a-11eb-80ae-ad46f7e7c3ed.jpg)
![stellarium-000](https://user-images.githubusercontent.com/18099873/124269334-0a762380-db3b-11eb-9335-53c644870c41.jpg)
Finally coming close!!!

PR #1259, opened 2020-09-11
missing horizon_line_color IS drawn - in black

axd1967:
### Description
Two landscape files say that the horizon line will not be drawn:
https://github.com/Stellarium/stellarium/blob/4a6bd4f0660c33f909c73ed9c7a8a91a3e1c8096/landscapes/geneva/landscape.ini#L15-L16

however, this is the result when one comments out the parameter:
(polygon line thickness exaggerated to show the result)

image

Fixes # (issue)

### Type of change

### How Has This Been Tested?

tested on Stellarium 0.20.2.18226 [master]

**Test Configuration**:
* Operating system: KUbuntu 20.04

## Checklist:

### DIFF
--- a/landscapes/geneva/landscape.ini
+++ b/landscapes/geneva/landscape.ini
@@ -12,7 +12,7 @@ polygonal_angle_rotatez=0.00001
 
 ; Fill color for this landscape. It gets darkened at night
 ground_color = .15,.45,.45
-; Color for the line. It will not be drawn if this is not defined. This color will not be dimmed at night.
+; Color for the line. It will be drawn in black if this is not defined. This color will not be dimmed at night.
 horizon_line_color =  .25,.15,.15
 ; specify a minimum brightness value 0..1 to have the ground always visible.
 minimal_brightness = 0.15
--- a/landscapes/zero/landscape.ini
+++ b/landscapes/zero/landscape.ini
@@ -17,7 +17,7 @@ polygonal_angle_rotatez=0.00001
 ; ground_color = .35,.95,.25
 ; dark green
 ground_color = .15,.45,.05
-; Color for the line. It will not be drawn if this is not defined. This color will not be dimmed at night.
+; Color for the line. It will be drawn in black if this is not defined. This color will not be dimmed at night.
 ; horizon_line_color =  .85,.45,.15
 ; specify a minimum brightness value 0..1 to have the ground always visible if you also select the minimal landscape brightness setting in the GUI.
 minimal_brightness = 0.15


gzotti:
Nah, this was the effect of changing, elsewhere, the behaviour of colors defined in config files. A negative value used to mark "invalid", and is now clamped to 0 (black). Thanks for recalling this to my attention. However, I must reject it, and solve the "don't draw when invalid color" in some other way.

gzotti:
I have actually solved this on July 5th. How old is this beta? (Honestly I don't know how to date build numbers...)

axd1967:
> I have actually solved this on July 5th. How old is this beta? (Honestly I don't know how to date build numbers...)

I'm using recent betas (and I've noticed that it has happened a few times that I report issues that seem to have been solved long time ago). When I write `master` I'm fully aware that this can cause confusion. my intention is to question this scheme.

But I notice with interest that you are questioning the numbering scheme. IMHO a good numbering scheme is important, and some improvements are possible. I have a few issues with the current way the software version strings are chosen. 

For example, I sometimes encounter RC. This means "release candidate"; but I don't find the associated git tag, so to me this is smelly. An RC is close to an actual release, and is preceded by alpha and beta releases.  (See also https://github.com/Stellarium/stellarium/wiki/Release-Schedule which explains the 0.YY.N scheme, and which is OK).

Better is not to ask devs to check some RC, but to check/test a SHA1, and have the SHA1 reported in issues.

The multi-digit suffix is a bit smelly: 0.20.2.**18252** - where is this coming from? Probably some kind of build identifier, maybe a manually managed integer (that is not versioned...); I would rather expect - and recommend a SHA1.

I don't like `0.20.2.18226 [master]` either. the SHA of the commit that has been tested is far better, but I'm too lazy to go into the source to fetch the correct SHA; I would expect to find it in de Help/About dialog.

42414bdc37223d2458c1abd6970bc705092c718b mentions `v0.20.2.18252/v0.20.3RC2`, so this is a commit that represents a version too (in fact two related version, rather confusing); if really needed, this should be made explicit by creating a git tag: `0.20.3-rc3+18252` (using the syntax as specified by SemVer; and `rc` is easier to read than `RC`, call it an opinion).

[Semver](https://semver.org/) would be a very good solution. There, a commit SHA could be appended: `0.20.2+sha.42414bdc`

Semver offers enough syntax to specify alpha, beta, and release candidates. To blend this with SemVer, this would become (dropping the `X` from SemVer `X.Y.Z`)
- 0.YY.N-alpha
- 0.YY.N-beta
- 0.YY.N-rc.N
- 0.YY.N.Y.Z
- 0.YY.N.Y.Z-alpha
- 0.YY.N.Y.Z-beta
- 0.YY.N.Y.Z-rc.N
etc

This assumes that `0.YY.N` is a major (seasonal) release.

Although not essential, I have taken the habit of storing all git tags in a namespace: `v/0.20.3`; in the (quite unlikely) instance that source code repositories might need to be merged, then the operation will be relatively simple as tags can then very easily be renamed. It also makes some git GUI tools easier to use (such as SourceTree that uses the folder approach).

I would drop the leading `0.`, as Stellarium is unlikely to ever reach `1.0`; as a result, the `0` has become noise. I am aware of the issue, but `20.2` is just as good as `0.20.2`. The really interesting bit is what is added after the `YY.N`: `YY.N.Y.Z`. (see the example `0.19.1.1` [on the wiki](https://github.com/Stellarium/stellarium/wiki/Release-Schedule)).

gzotti:
It might surprise you, but the leading 0 will stay until planet axes are fixed (#502) and lunar occultations are more accurate (#84). These are IMHO the last showstoppers. 
I would then declare and celebrate one release 1.0==21.x (hopefully!), and then continue with 21.x. 

The question is: Do you see this issue with a build from current master? You can always add a human-readable calendar date.

gzotti:
BTW, from your semver: 

> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Another argument to keep it at 0 for the time being...

axd1967:
Yes, I understand the reason behind the '0`; IMHO it is not really relevant. Once a product is released to the public, we start with 1.

But @alex-w this is very nice to see: `Stellarium 0.20.3-2290e7e5de [master]` ! great!
image

A calendar date is also a way to version the code; often it presents less fuss than having to decide if it is a major or minor release. But Semver is interesting to indicate some extra info for those who are interested (many - if not most - users don't really care not understand).

gzotti:
We tried behind the scenes. The build number must also stay as 4th part due to how Windows Version numbers are created. But we added (short) sha1 now.

To find out the REAL subject of this bug report, I repeat: 
> The question is: **Do you see this issue with a build from current master?**

axd1967:
> To find out the REAL subject of this PR, I repeat:
> 
> > The question is: **Do you see this issue with a build from current master?**

So sorry - I didn't register your question correctly.
2290e7e5de no longer shows this issue, thanks!

gzotti:
(Superseded at time of report.)

axd1967:
> (Superseded at time of report.)

@gzotti with all due respect, I just recompiled 0.20.2 and that version still shows this behaviour (a black line is drawn when the landscape.ini property is commented). even if this is no longer important. I'll reinvestigate later what the exact cause was; let's assume for now that my build system has an issue.

gzotti:
You reported on Sept. 11, but I had fixed it on July 5. Report on releases or on current master.

PR #1266, opened 2020-09-14
Fix Chinese sky cultures

sunshuwei:
Fix two stars.
Recover lost files (description.zh_CN.utf8). It is missing for unknown reasons.
Fix a determinative star


alex-w:
Please check 0.20.3RC2

PR #1285, opened 2020-09-26
Oculars: Better handling of manual and auto limit (Fix #400)

### LINKED ISSUE #400: persistence of star magnitude limit
_Original report by Tig la Pomme:_ https://bugs.launchpad.net/stellarium/+bug/1722384

Good evening,

When I set the star magnitude limit, it is applied but not persistent, the limit magnitude becomes 9,63 every time I change ocular or restart the program.

gzotti:
### Description

This provides consistent handling of manual and automatic limits for stellar magnitudes. 

- When enabling ocular view, the plugin-own magnitude limit is used. 
- Both limits and limitation flags are duplicated for independent Main and Oculars toggling.
- If in addition user selects auto-limit, this will be used instead. 
- Switching on auto-limit activates the limitation flag and computes limit mag, but does not replace the plugin-own manual limit value. 
- Switching off auto-limit reverts to manual ocular limit but keeps keeps manual limit active. 
- Switching off (manual) limit while oculars view is active deactivates auto-limit
- The value for the manual limit is not permanently overwritten when using autolimit.

Fixes #400

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Qt 5.13, MSVC2017, 64bit

## Checklist:

- [X] I have made corresponding changes to the documentation
- [X] New and existing unit tests pass locally with my changes


alex-w:
A quick check:
a) Disable limit magnitide in View dialog and save settings
b) Disable automatic limit magnitude in Oculars plugin
c) Enable eyepiece view and check View dialog (Sky tab) - the limit is enabled and have value 12 - why???

gzotti:
Did you do (a) while looking through the eyepiece?
12 is the (arbitrary) default manual ocular limit value, line 605.

alex-w:
No, I just check this point before eyepiece mode was enabled

alex-w:
> 12 is the (arbitrary) default manual ocular limit value, line 605.

But any value shouldn't be apply when all limits are disabled...

gzotti:
The limit can be displayed, but just not activated. There are limitation flags for the main application and for viewing with ocular. Additionally, the auto-flag overrides the limit, but does not store the manual value permanently (you can toggle auto/manual).

checking out again for another test...

gzotti:
No, all is as intended and described above. If I switch off limitation in ocular mode, it stays switched off at next start.  The displayed value depends on whether you are in ocular mode or not. 

When you enable auto-limit and then disable it, the manual limit stays on. This is intended. You either want to limit, in principle, or not. If you want to go from autolimit to no limit, you have to disable autolimit and main GUI limit (while in ocular mode), or only main GUI limit (while in ocular view). Disabling the latter while in ocular view also disables autolimit.

alex-w:
See screenshots:
stellarium-000
stellarium-001

a) Both limits are disabled
b) Eyepiece mode is enabled and I see that magnitude limit is enabled and re-defined - why?

gzotti:
Because the ocular-mode limit flag was stored in modules/ocular.ini.

This has three relevant entries:

autolimit_stellar_magnitude=false
limit_stellar_magnitude_oculars_val=9.70
limit_stellar_magnitude_oculars=false

Switch of (manual) limitation while in ocular mode. Toggle ocular view, check that it remains switched off, even if you switch it on while not in oculars mode.

alex-w:
Some user has defined 12" Dobsonian telescope in Oculars plugin (with no defines any limits by whole planetarium for magnitudes) and enable eyepiece view - now he see stars up to 12 magnitude, what is very wrong.

gzotti:
Then he switches off the manual limitation of stars. The default is false (Oculars.cpp line 604.) I don't know why it is true on your computer. (Maybe old settings from the time before the auto-limit?) Changing this setting discrepancy costs exactly one mouse click. (Not even another "save settings".)

alex-w:
> Then he switches off the manual limitation of stars _while in oculars view mode_. The default is false (Oculars.cpp line 604.) I don't know why it is true on your computer. (Maybe old settings from the time before the auto-limit?) Changing this setting discrepancy costs exactly one mouse click. (Not even another "save settings".)

Screenshots from Stellarium with default settings + disable auto-limits for oculars:
stellarium-002
stellarium-003

Another problem (default settings):
stellarium-004
stellarium-005

gzotti:
Yes, **this is intentional**! Please read the description.

alex-w:
> Yes, **this is intentional**! Please read the description.

I guess limit_stellar_magnitude_oculars_val should have default value at least 21, because limit up to 12 magnitude when users are not expected any limits is very strange behavior of tool.

alex-w (approved):
Thanks!

gzotti:
I deleted my config and oculars.ini. The default starts with an auto-limit (why? we could switch that off instead!). 

If user switches off auto-limit, there remains a manual limit, by intention. If 12 is not what user wants, he/she can change it. Or switch it off entirely. I don't see a problem. If you think that an installation default of 20 or 25 is more useful, so be it.

alex-w:
OK, please merge the solution

axd1967:
@gzotti  I compiled 29b061ea29 which contains https://github.com/Stellarium/stellarium/commit/a90a207625c953f427bcd50f7e0e12bdbbfcc0f5 and have a question

There is something I don't understand: why is the sky limit linked to the ocular mode (which is logical), but not the SSO mag limit?
- when ocular mode is off, sky, and SSO limits are as selected by the user
- when ocular mode is on, sky limit is changed (related to ocular parameters in use), but the SSO mag limit is not. As a result, a faint planet will not appear when an ocular is active.

test:
1. ocular mode off
1. set sky limit to 6.0 (in fact this is independent from the SSO limit, but it illustrates the different behaviour)
1. choose a SSO limit that barely hides a planet such as Uranus
1. activate ocular mode
1. observe that the sky limit has temporarily been increased (which is OK)

BUT: the SSO limit has not changed - as a result the faint planet is still not visible.

gzotti:
When we first dealt with changing mag limits, we only had 1 value to play with. Now we have 3. While there are potentially millions of stars clipped off prematurely, only a few tens of SSO may be involved. 

If you look closely, we can have manually set limits for naked-eye and ocular mode, and an optional auto-limiter for ocular mode. This only adapts the stellar magnitudes cutoff. Indeed, SSO and DSO may need adaptations also, but take care, as surface magnitude may be important here. If I ever found the time to get up from my screen and view through an eyepiece again, it might annoy me enough to think of further steps. 

If you want, you can extend the system now in effect for the stars and add and complete handling of a second and third pair of limit magnitudes along the lines of magLimitStarsMain/magLimitStarsOculars and the two flags. Probably the manual limit is enough for SSO and DSO, unless we have a formula that correctly deals with surface mags.

PR #1295, opened 2020-09-28
Use system versions of dependencies

DarthGandalf:
### Description

I've noticed that src/external contains many third-party libraries. And, as usual with bundled dependencies, they are outdated. This PR is a first approximation of how to make stellarium use external dependencies instead.
For zlib, there is a cmake switch to use system one or bundled one. I doubt whether there's a value in keeping bundled ones at all, especially as it creates an incentive to add more libraries to the dir.

Qxlsx requires an upstream patch to be able to install it (https://github.com/QtExcel/QXlsx/pull/114)

I left qcustomplot as is for now, because qcustomplot 2 changed its API a bit, so that Stellarium doesn't build with it. (Axis tick and tick label generation was completely refactored there, so the function called doesn't exist anymore, but there is now a class for a similar functionality)
Also I left glues/libtess because I couldn't find an alive upstream with same API: https://github.com/memononen/libtess2 looks somewhat different.

### Type of change

### How Has This Been Tested?
By installing the dependencies (newest versions of them!) separately, patching stellarium to use them (this PR), and verifying it still runs. See https://github.com/gentoo/gentoo/pull/17639

**Test Configuration**:
* Operating system: Gentoo Linux, 5.4.60-x86_64
* Graphics Card: NVidia RTX 2070S, driver 450.66

## Checklist:


axd1967:
interesting. not sure whether this applies here: an error that projects sometime do is to include external contributions into `master`. unfortunately, this way future external changes become difficult - if not impossible - to merge. the solution is the "vendor branch" concept.
strictly speaking this should apply to all plugins; an indication that this is the correct approach is the fact that plugins have their own version strings.
(to implement this, every plugin requires its own master branch (eg `plugin/satellites/master`), that is then merged in master, usually via a feature branch. updates to the plugin are done in the plugin's branch, that is merged to master)

this allows to apply local changes to a plugin without disturning the plugin's master branch, as well as keep all plugins in one single repo.

DarthGandalf:
@axd1967 that's an interesting idea, but it's independent from what I'm doing here.

> an error that projects sometime do is to include external contributions into master

This is what I'm trying to fix. Indilib is not a part of Stellarium, but for some reason its copy is here in stellarium's master branch. Same for other dependencies in `src/external/`

> the solution is the "vendor branch" concept.

From packaging perspective, it's not different from having the dependency right in master. You're still relying (vendoring) on a particular (likely outdated) version of the library, while there is another version available (installed) on the system.

> strictly speaking this should apply to all plugins

As plugins are parts of the same project, I don't mind them being here. It becomes different if plugins can be *installed* separately and independently from each other, e.g. as separate .so/.dll files, and that opens a possibility to have plugins coming from different sources, but it also opens lots of new questions: ABI compatibility between such plugins, for one. I'm not diving into that.

gzotti:
Plugins could be unbundled if anybody found out how to build them dynamically on Windows. This seemed to have been possible with Qt4, but something was lost years ago. Not that we would be flooded by contributions...

axd1967:
> @axd1967 that's an interesting idea, but it's independent from what I'm doing here.
> 
> > an error that projects sometime do is to include external contributions into master
> 
> This is what I'm trying to fix. Indilib is not a part of Stellarium, but for some reason its copy is here in stellarium's master branch. Same for other dependencies in `src/external/`

ah! I was too lazy to look what exactly was copy-pasted; but indeed, this is very smelly. copy-paste is evil :-)

each of these modules under `src/external` should live on its own `vendor/<lib>/master` branch. meanwhile I see that there have been substantial local changes made, so it could be less trivial to import recent versions. there is a need for vendor branching here: 
- reimport the most recent versions - but using vendor branching; apply vendor tags:
  - `vendor/libindi/1.8.6`,  (Stellarium forked off 1.5.0 ...)
  - `vendor/qtcompress/1.0`, 
  - `vendor/QCustomPlot/2.0.1`, 
  - ...
- patch local changes (if necessary) in the master version of the merged branches (if needed; ideally, issues should be reported to and solved by the original library developers)

this will allow to painlessly import updates to the external code, _as well as apply changes that might rejected by the library owners_ (at least, this will work until a definitive fork is no longer avoidable).

> From packaging perspective, it's not different from having the dependency right in master. You're still relying (vendoring) on a particular (likely outdated) version of the library, while there is another version available (installed) on the system.

Remembering that it is sometimes important to stick to an older/outdated version; a move to a more recent version should be tested in a branch. but a package manager is usually parameterised with a version number of an external system to use/retrieve, so the effect is the same.


(http://svnbook.red-bean.com/en/1.7/svn.advanced.vendorbr.html - applies to Git as well)

axd1967:
> Plugins could be unbundled if anybody found out how to build them dynamically on Windows. This seemed to have been possible with Qt4, but something was lost years ago. Not that we would be flooded by contributions...

of course, if Stellarium is the only place where these plugins live, then there is no real need to use a vendor approach.

alex-w:
Alexey, will you support and build all those libraries for Windows and macOS now?

DarthGandalf:
> Alexey, will you support and build all those libraries for Windows and macOS now?

Good question. The answer is no, therefore I shouldn't just remove them from here.
Okay, will add cmake switches instead. Also will try to replace these dirs with git submodules, so at least it'd be easier to update them.

axd1967:
> Okay, will add cmake switches instead. Also will try to replace these dirs with git submodules, so at least it'd be easier to update them.

Hmm... git submodules (and subtrees) are not easy to use, but could be a solution *if* the other libraries are also versioned in Git.

the **vendor branch approach** is agnostic on how the library is versioned elsewhere.

1. (official) https://git-scm.com/book/en/v2/Git-Tools-Submodules#Issues-with-Submodules
2. (old) https://medium.com/@uttamkini/sharing-code-and-why-git-submodules-is-a-bad-idea-1efd24eb645d
3. (old) https://codingkilledthecat.wordpress.com/2012/04/28/why-your-company-shouldnt-use-git-submodules/
4. (2015) https://medium.com/@porteneuve/mastering-git-submodules-34c65e940407
5. https://devcenter.heroku.com/articles/git-submodules (suggests "vendoring" but not detailed enough - a copy-paste is suggested, which is not how the vendor branch approach works)
1. (2016) https://www.philosophicalhacker.com/post/using-git-submodules-effectively/

personally I would recommend against using git submodules because they are tricky and require some git experience.

DarthGandalf:
> one advantage of vendor branching is that it is very simple and intuitive to maintain.

Only if you're familiar with this uncommon pattern. This doesn't prevent you from accidentally modifying files there outside of the vendor branch, and such modifications make it harder to update the library in future. Such modifications already happened here, though mostly it's formatting changes which don't change the meaning of code.

> personally I would recommend against using git submodules because they are tricky and require some git experience.

I can add a few lines to README about how to update dependencies. It's not like it ever happens now, with files copied here.

> some links

Most of the issues there are with submodules used for subprojects of the same project, like in your plugins example. That doesn't apply with truly external dependencies like here.

I've also just discovered https://cmake.org/cmake/help/latest/module/FetchContent.html so perhaps it will work for this case instead of submodules. I'll look more into it.

axd1967:
> This doesn't prevent you from accidentally modifying files there outside of the vendor branch, and such modifications make it harder to update the library in future. Such modifications already happened here, though mostly it's formatting changes which don't change the meaning of code.

Such (accidental) modifications are precisely the reason why vendor branches exist. Such modifications _have_ to be done in the `master` branch, _never_ in the vendor branch.

>  It's not like it ever happens now, with files copied here.

famous last words?

DarthGandalf:
>> It's not like it ever happens now, with files copied here.
>
> famous last words?

That's exactly what I'm fixing here, you know. To make it possible/easy to update.

DarthGandalf:
I'm shelving this change for now, until either the deps become available in a simpler way, or until required cmake version can be at least 3.14

axd1967:
please consider meanwhile to deal with future external code additions by following the vendor branch pattern; pending any better mechanism, this will allow for easier updates of external code while allowing for internal changes (if needed).

it is not difficult. for each vendor a branch `external/<vendorname>` will be created that will receive the external code, **unmodifed**. this branch will be tagged with vendor tags `vendor/<vendorname>/<vendorversion>` and will be merged to `master` (probably via a issue or feature branch); only on `master` (or feature branch) will local changes occur. Ideally, local changes are also coordinated with the vendor (via issues submitted to the vendor), because future updates of vendor code are likely to produce conflicts.

only when the external code can no longer be merged in, will a decision be taken to fork the vendor by deleting the vendor branch. from that point on we are in the situation that currently exists. this should be avoided as much as possible.

the existing modules under `external` could be vendored, but IMHO because of the effort required this should be attempted only when re-importing a recent version of the external sourcecode: that's the point to start over, in the correct (vendor branch) way.

this is a way to keep the external code future proof, and allow Stellarium to benefit from bug fixes, new features, etc.

DarthGandalf:
The correct way would be to convince vendors to package their software to vcpkg or conan, and then install the dependencies of stellarium from there. Then it will be trivial to point it to either the version in the system, or installed by those packages managers. And it allows easy updates to newer versions.

Vendor branches don't have that benefit.

axd1967:
> The correct way would be to convince vendors to package their software to vcpkg or conan, and then install the dependencies of stellarium from there. Then it will be trivial to point it to either the version in the system, or installed by those packages managers. And it allows easy updates to newer versions.

Package managers are great for code that doesn't change, but I'm not sure it offers tools in case an external library is changed locally.

But I doubt that you will be able to convince the vendor of e.g. qtcompress to do this. With a vendor branch, there is no issue at all. Even better: even if the vendor uses sloppy (or zero) versioning, you can still cleanly track it.

Another important point is to include some basic documentation that is kept in the `master` branch near (or even in, yes!) the source of external code: a developer needs to know from where the external code came from, and where new versions can be found. In the case of qtcompress this is not immediately necessary, as it is (and always should be) nicely mentioned:
https://github.com/Stellarium/stellarium/blob/c086b368a4ae9b9bd4c9dcb13f84a001da98438e/src/external/qtcompress/qzip.cpp#L3-L4

except that it is not clear which version was taken (and some searching still has to be done). 

(probably https://code.qt.io/cgit/qt/qt.git/plain/src/gui/text/qzip.cpp - and there we see a 2015 **QT4** version, while Stellarium is toting a 2013 version... and we wait for https://bugreports.qt.io/browse/QTBUG-3897)

that's where the vendor tags should have helped.

in the case of qxlsx I couldn't immediately find a source, which makes it difficult to update it.


> Vendor branches don't have that benefit.

Sure vendor branches do allow for easy updating: see https://www.roe.ch/Git_Reference
This requires some basic Git understanding; while a package manager can be far more complex to work with.

But again, code that need not be changed locally is probably best served via a package manager or some mechanism that pulls in external code.

Another snag: in case external code is pulled in, a subtle issue is whether to pull in a **master or latest version**, or a tagged version. Sometimes, a latest version can increase the risk for new bugs. A **tagged version** needs to be manually updated from time to time (after testing).

And then there is the case of becoming dependent on external code and not having a local copy. What if the external code disappears or changes dramatically?

Personally I'm wary of such external dependencies and believe in local copies. External libraries work smoothly, until they disappear or introduce major changes (and they sometimes? often? depend on a functional internet connection).

DarthGandalf:
> But I doubt that you will be able to convince the vendor of e.g. qtcompress to do this

I don't know, I haven't tried. Why do you think so? Package managers are getting popularity.

> I'm not sure it offers tools in case an external library is changed locally.

It doesn't need to be changed locally. Here it's changed locally just to make it easier to be bundled. Which is moot if it's not bundled.

And if it does need to be changed locally, it's a fork, and should be treated appropriately.

> Another important point is to include some basic documentation …  a developer needs to know from where the external code came from, and where new versions can be found

Oh, definitely!

> it is not clear which version was taken

Yes, and package manager can make it clear which version is used. With vendor branch it's possible, but require more strict git discipline (just like with submodules)

> This requires some basic Git understanding; while a package manager can be far more complex to work with.

The problem is that *basic* git understanding is not enough. Someone will definitely put something to a wrong branch, which will be hard to fix later. With a package manager, the place where to put it is clear. There is documentation available for them, just like for git.

> Another snag: in case external code is pulled in, a subtle issue is whether to pull in a master or latest version, or a tagged version. Sometimes, a latest version can increase the risk for new bugs. A tagged version needs to be manually updated from time to time (after testing).

Sure, no doubt. This is the case either way.

> And then there is the case of becoming dependent on external code and not having a local copy. What if the external code disappears

Valid point. If upstream disappears, it's an issue anyway. Sometimes there are mirrors.

> or changes dramatically?

That's what versions are for

DarthGandalf:
> > Vendor branches don't have that benefit.
>
> Sure vendor branches do allow for easy updating

I specifically meant an easy way use the system-installed library instead of the bundled one. Updating may be easy, sure, that was not my point.

axd1967:
> It doesn't need to be changed locally. Here it's changed locally just to make it easier to be bundled. Which is moot if it's not bundled.

I didn't look in detail, but following is an example of a far more than a trivial change

https://github.com/Stellarium/stellarium/commit/4adb4567ec66afabbc8efddb6c4398c43809c740

Even those local changes that you mention to make packaging easier will make updates to a package more difficult (except, of course, when you ditch the local code and import a specific version).

> 
> And if it does need to be changed locally, it's a fork, and should be treated appropriately.

I disagree. Vendor branching allows very well to use a library that is maintained/updated externally AND apply local changes. This is the core of the concept of vendor branching: you can stay up-to-date and at the same time introduce - admittedly small - changes.

> Yes, and package manager can make it clear which version is used. 

Absolutely - this is the great strength of a PM.

> With vendor branch it's possible, but require more strict git discipline (just like with submodules)

Much easier than git submodules!  But agreed, it requires discipline.

> 
> > This requires some basic Git understanding; while a package manager can be far more complex to work with.
> 
> The problem is that _basic_ git understanding is not enough. Someone will definitely put something to a wrong branch, which will be hard to fix later. 

Wrong: such changes have to be approved. It is relatively trivial to check that the import of a specific version (on the associated branch) was done correctly (just repeat the procedure, and nothing new should appear).

> With a package manager, the place where to put it is clear. There is documentation available for them, just like for git.

Totally agree. It's "baked in".

DarthGandalf:
> I didn't look in detail, but following is an example of a far more than a trivial change

libtess doesn't have upstream, so this copy is a fork. But such fixes should go upstream in normal cases.

> except, of course, when you ditch the local code and import a specific version

That's the plan

> Wrong: such changes have to be approved.

By whom? :) By the same people whose discipline is required?

axd1967:
> > Wrong: such changes have to be approved.
> 
> By whom? :) By the same people whose discipline is required?

By the official maintainers: @gzotti and @alex-w . They carry the responsability for Stellarium. Because if things go wrong, they get he flak.

DarthGandalf:
> By the official maintainers

What I mean is that humans make mistakes. With vendor branching like you describe, there's nothing to help maintainers notice that there is something wrong. And they need to remember about this vendor branching pattern when they are doing various changes themselves too.

axd1967:
> > By the official maintainers
> 
> What I mean is that humans make mistakes. With vendor branching like you describe, there's nothing to help maintainers notice that there is something wrong. And they need to remember about this vendor branching pattern when they are doing various changes themselves too.

Strictly speaking they don't need to remember much, as local change are done on `master`. it's only when external changes come into play that they have to checkout the vendor branch, explode the new version, and merge to (probably) an issue branch.

I suspect that [RemoteControl](https://github.com/Stellarium/stellarium/tree/master/plugins/RemoteControl/webroot) suffers from the same vendor issues: it looks likely that some code there came from the outside world. (eg https://github.com/Stellarium/stellarium/blob/master/plugins/RemoteControl/webroot/js/jquery.ui.touch-punch.js VS https://github.com/rotundasoftware/jquery-ui-touch-punch/blob/b357c2cbe5d294996d0229adc6bd8c21efa3f3cd/jquery.ui.touch-punch.js ; see also  https://github.com/furf/jquery-ui-touch-punch/network )
(Note also incidentally how the version number has been ignored - this is another small yet nontrivial issue of not fully mastering the vendoring of code)

axd1967:
See also https://github.com/Stellarium/stellarium/discussions/1856

axd1967:
Re-reading the conversation, I have hidden my vendor branch comments to this issue as here the intention is to use specific tools for this use case. [Vendor branching](https://github.com/Stellarium/stellarium/discussions/1856) is a good and simple to use tool for generic cases or where advanced vendor mechanisms are not available. It deals with the copy-pasta problem in a simple way.

But we do need a technically sound way to avoid the copy-pasta issue which leads to code rot.

axd1967:
It looks like the best way is simply to test if compilation still works on the different platform?

DarthGandalf:
> It looks like the best way is simply to test if compilation still works on the different platform?

I knew that without updating cmake to at least 3.14 it does not work. I saw that CI uses ubuntu 18.04, and was afraid of updating it because there might be some other reason for it which I'm not aware of.

But I think you're right, and I can just update the PR so that CI uses the newer cmake, since @gzotti and @alex-w don't seem to be against the increased requirement for cmake version.

gzotti:
Don't count my silence as agreement, I am just too busy with my own aims and actual improvements to also deal with changing one functioning way of working with another here, even if there may be dev benefits in the long run.
 
cmake min versions depends on supported systems. I still am on 3.13 but could upgrade easily. On some otherwise supported OSes it may not be that easy. (Ubuntu, Raspbian?)   Qt6 will require cmake 3.16, or even 3.21 (https://doc-snapshots.qt.io/qt6-dev/cmake-get-started.html)

DarthGandalf:
> On some otherwise supported OSes it may not be that easy. (Ubuntu, Raspbian?)

On Ubuntu the update is easy: https://cmake.org/download/ provides the binaries for major platforms, there are also PPAs, etc.

On Raspbian Buster release (Debian oldstable) the debian's buster-backports repo [can](http://marksrpicluster.blogspot.com/2019/12/add-buster-backports-to-raspberry-pi.html) be added, which has cmake 3.16. Newer raspbians (aka Raspberry Pi OS) have new enough cmake already.

Is there anywhere a full list of supported systems I can check?

axd1967:
maybe we can also check what the OS is of the currently active developers?

an alternative is also to craft a VM from which a build can be triggered; I'm thinking of VirtualBox as the simplest solution (rather than Docker).

DarthGandalf:
> an alternative is also to craft a VM from which a build can be triggered; I'm thinking of VirtualBox as the simplest solution (rather than Docker).

I'm surprised to see this suggestion from someone who wants simplicity :)

Maintaining such VM is a lot of overhead. In fact, such VM already [exists](https://github.com/Stellarium/stellarium/blob/master/.github/workflows/ci.yml), but it's not any useful for users who want to build stellarium on their system

DarthGandalf:
I see you updated cmake requirements to 3.16 recently. So I can now resurrect this PR. Will try to do that soon

alex-w:
Note: QCustomPlot is dropped from dependencies now

alex-w:
Note to the `CPM` module: no need adding special cmake rule for adding it - just put the code of module into `cmake/modules` directory

alex-w:
What about rebase the PR?

DarthGandalf:
I find it easier to fix conflicts with merge instead of rebase. It's merged with yesterday's master now. If you wish, it can be squashed to a single rebased commit.

For now I'm debugging the error happening on windows build, so not ready yet anyway

DarthGandalf:
This is done. Do you want me to turn this into a single rebased commit?

alex-w:
Could be `CPM.cmake` moved into `cmake/modules` directory (no need extra cmake rule for adding module)?

Please check LGTM log - probably `external/CMakeLists.txt` has errors

P.S. Did you tried use latest stable INDI client locally?

DarthGandalf:
LGTM just couldn't merge the files, even though there are no conflicts.

> Could be CPM.cmake moved into cmake/modules directory (no need extra cmake rule for adding module)?

CMAKE_MODULE_PATH only affects which directories are searched by `include()` and `find_package()`. It doesn't automatically include everything from it. Still moved there though.

> P.S. Did you tried use latest stable INDI client locally?

On linux (gentoo) it works, when used from the system package. 1.9.6 is the current version available. I didn't check whether it works on windows.

BUILDING.md
```diff
@@ -56,6 +56,15 @@ your distribution. Here's a list.
 - [Graphviz](http://www.graphviz.org/) - required to build the API documentation and include fancy class diagrams
 - [libgps](https://gpsd.gitlab.io/gpsd/index.html) - if you want to build Stellarium with GPS support (Linux/macOS only)
 
+### Optionally bundled dependencies
+
+If these are not found on the system, they are downloaded automatically.
+
+- [QCustomPlot](https://www.qcustomplot.com)
```
alex-w, line 63: Please remove this line, because QCustomPlot is not used anymore

alex-w (approved):
This pull request is OK for me, but please fix BUILDING.md file

@gzotti your opinion?

BUILDING.md
DarthGandalf: Oh, right.

Removed QCompress too, since switched to the internal QGui's one.

gzotti:
I can test only later next week. No real opinion so far. If it works and is well described how to operate/update, it should be OK.
Note that descriptions about what changed here (and how they influence development in the future) need to be added to BUILDING.md. Write this clear enough for a novice developer just starting with git.

DarthGandalf:
I don't think "how to update" belongs to BUILDING.md. If you just want to *build* from source, you'll notice no difference, except that if the package is found on system, it'll be used, and if it's not then it'll be automatically downloaded by cmake.

How to update or how to add a new library: prefer `find_package()` if it's some common library or `CPMFindPackage()` if you want to add the fallback of downloading it. But then it depends on the library. If it provides a good CMakeLists.txt, by default `CPMFindPackage()` will just use it. But if there are some issues with it, there is an option to only download/unpack, and then you can deal with files in it however you want. https://github.com/cpm-cmake/CPM.cmake has more docs.

gzotti:
I just want to make sure that after merge of this PR which stems from your experience, this part of experience is also transferred to where new developers will be able to find it and build on top of this. I.e., it needs to go to the development docs, whether that is BUILDING.md or even a new MAINTAINER_BUSINESS.md I don't care. Until you wrote it here I was unaware of CPM.cmake, and when this PR is closed this relevant bit of info may be hard to find.

axd1967:
(related to [vendor_branching](https://github.com/Stellarium/stellarium/issues?q=vendor_branching))

PR #1332, opened 2020-10-17
Ephemeris plotting improvements

### LINKED ISSUE #1315: A few requests for ephemeris visualization (AstroCalc)
Not urgent!

I wish for a few improvements of the ephemeris visualization in AstroCalc.

1. Being able to change the symbol of the marking. E.g. a cross, a star, a ring, general comet symbol etc.
2. Being able to show markings only when labled. This means that I can choose to show e.g. Hale-Bopp's position from 1997-01-01 to 1997-12-31, with 1 day intervals, but only show the markings at e.g. 10 day intervals. This means the line keeps smooth, but without markings in every breaking point. See this image for a suggestion:
https://upload.wikimedia.org/wikipedia/commons/f/fc/Comet_Hale-Bopp_starmap_1997.png
3. Being able to use more customizable labels. For example, I would like to show only day and month, or year and month, or just day and hour.
4. Perhaps some antialiasing of the lines and symbols?

gzotti:
### Description

I would like to have better-readable code, together with some nice enhancements suggested in #1315.

So far,

- Added comet icon for ephemeris points plotting
- cleaned up the code a bit
- indicated locations of crucially missing documentation. @alex-w I hope you can add some here.
- added plotting the comet icon rotated in screen space so that the tail points away from the solar location. This works fine for projections without discontinuity. For those, e.g. Cylindrical or Mercator, the direction is wrong when e.g. the sun is near the left edge and the comet near the right. I am unsure how to solve this properly. Do some manual screen-size modulo detection (beware of scaling issues!), or are projections smart enough to detect this somehow?

Maybe, even the AstroCalc, which has grown into a spectacular but huge and hard to understand module, should be refactored, computing separated from the GUI, and the Ephemeris parts taken out from SolarSystem. Maybe not in this branch, though.

Issue #1315 

### Type of change

### How Has This Been Tested?

Plotted a few comet ephemerides and checked visually where the tail points to.

**Test Configuration**:
* Graphics Card: NVidia Geforce 960M. 

## Checklist:

- [x] I have made corresponding changes to the documentation (No SUG changes yet)


alex-w:
But comet icon isn't used yet...

> However, I did not reverse engineer the code deeply enough to identify whether solar positions are available when we want to plot rotated comet icons.

Just expand the struct Ephemeris by adding position angle data for comet tail ;)

gzotti:
I started with that comet icon before having a look into the code how the ephemeris plot actually works. Icon usage can be trivially switched when we know the object to display is a comet. Probably a dedicated colorIndex would be enough to detect this? But it seems the list should either always include solar data for the same dates, or alternatively, the Ephemeris struct should be enlarged to include a solarAngle (for all objects, even if at first only used for comets). 
I just wanted to also point out that these things should please be documented reasonably complete to understand the intent of the developer.  Maybe you have something in mind to develop ~ by February, and if I restructure things, it would break your ideas.

gzotti:
OK, I will add solarAngle to the Ephemeris struct and add a new colorIndex value to mark comets.

alex-w:
> I started with that comet icon before having a look into the code how the ephemeris plot actually works. Icon usage can be trivially switched when we know the object to display is a comet. Probably a dedicated colorIndex would be enough to detect this? 

No. But adding info about type of object into Ephemris struct will help you.

> But it seems the list should either always include solar data for the same dates, or alternatively, the Ephemeris struct should be enlarged to include a solarAngle (for all objects, even if at first only used for comets).

Yes, adding solarAngle or positionAngle into struct will help you

> I just wanted to also point out that these things should please be documented reasonably complete to understand the intent of the developer. Maybe you have something in mind to develop ~ by February, and if I restructure things, it would break your ideas.

No specific ideas

alex-w:
> and add a new colorIndex value to mark comets.

Why? Do you want to see any comets with separate color of markers and lines always? But in this case you should add two 2 colorIndex

gzotti:
I either have to add another new element to select "comet or other icon", or use a dedicated value in the existing elements, like a dedicated colorIndex.

A first result is here.
stellarium-039

However, depending on azimuthal or equatorial mode, and where on screen the object is located, I have to add parallactic angle (probably easy) and some other projection-dependent angle.

alex-w:
> I either have to add another new element to select "comet or other icon", or use a dedicated value in the existing elements, like a dedicated colorIndex.

Georg, the colorIndex is used for connection between color buttons in the GUI and color of markers/lines - please do not use it for define the type of markers.

gzotti:
OK, then I need another element in Ephemeris, although now the color index, when out of bounds, just falls back to default marker color. Getting that extra angle is definitely more difficult.

alex-w:
> OK, then I need another element in Ephemeris, although now the color index, when out of bounds, just falls back to default marker color. Getting that extra angle is definitely more difficult.

In any way why you do not define another flag for ephemeris (like is was used for "smart dates") and use getPlanetType() to select texture or define required name (index?) of the texture in AstroCalcDialog::generateEphemeris() (like colorIndex). 

Adding 2 extra fields into struct to get correct data is best way IMHO

gzotti:
I wanted to avoid extra fields when even the extra color is a working solution, but really may even require adding solar position to Ephemeris, to make the angle computation in screen space. But maybe only tomorrow...

src/gui/AstroCalcDialog.cpp
```diff
@@ -253,7 +253,7 @@ void AstroCalcDialog::createDialogContent()
 	// TODO: Switch a QDateTimeEdit to StelDateTimeEdit widget to apply wide range of dates
 	// NOTE: https://github.com/Stellarium/stellarium/issues/711
 	//QDate min = QDate(1600, 1, 1);
-	QDate minDate = QDate(1582, 10, 15); // QtDateTime's minimum date is 1.1.100AD, but appears to be always Gregorian.
+	const QDate minDate = QDate(1582, 10, 15); // QtDateTime's minimum date is 1.1.100AD, but appears to be always Gregorian.
 	QString validDates = QString("%1 1600/1/1 - 9999/12/31").arg(q_("Gregorian dates. Valid range:"));
```
alex-w, line 257: Please update tooltip too...

alex-w (approved):
OK

gzotti:
Do you want to add the requested documentation before we merge? Also, do you have a good idea how the false rotation can be avoided in Cylindrical/Mercator/... projections?

axd1967:
Looks nice!

You probably saw it coming, and it is likely easier said than done: the comet info includes an approximate tail length information; isn't it more interesting to draw "just" a simple line with that length, rather than use a symbol?

gzotti:
The comet icon represents "comet with possible tail into that direction", not "we guarantee (or at least prognosticate) the comet tail had this or that length in the sky". 

As written before in #1315, the tails rendered by the Comet class are modelled in 3D space. While an estimated tail length in km or AU can be queried, the length of the tail in degrees as seen by the observer can not. You can suggest a PR that does this, though.

alex-w:
> Do you want to add the requested documentation before we merge? 

OK, but... maybe really move all ephemeris stuff from SolarSystem class into VisualizeEphemeris class?

gzotti:
@alex-w It may be good to add the missing bits of doc now (really only a few lines, but I just don't know all the details, so you should do that please), merge the working feature (if you are happy with the mentioned borderline bug around cylindrical projections; or we still find a solution?), and then for the next step forward create a new branch to disentangle the AstroCalc and SolarSystem bits into a new VisualizeEphemeris class. The SolarSystem and VisualizeEphemeris can/must of course be friends, but I feel these monstrous classes need some restructuring and please, just a few bits of non-trivial description. (I mean: "setStopValue(); // sets stop value" is not enough. What stops here? [this method does not exist, but I have seen similar cases])  Maybe some of the AstroCalc class can also move to VisualizeEphemeris. This needs some quiet week to get a good overview how these classes interact, what's GUI and non-GUI, etc. If we want better participation from outside, clear structuring and documentation is important.

Issue #1313, opened 2020-10-06
Calendar options


goedible:
Hi.  I can imagine calendar options that would allow people to try different types of calendars.  For example with lunar calendars like the Chinese, Vietnamese, Hindu, and Thai calendars.
I see this a powerful aid in teaching the importance of calendars and how they can help us relate to cosmic influences.
By the way... stellar work so far!

gzotti:
Yes, I can also imagine that. Even without cosmic influences.
This is an obvious candidate for a plugin. Volunteers, go ahead.

goedible:
Thanks for lettimg me know George.  I would like to volunteer however I
have very limited programing skills.  I dropped out after my first java
course in college.  Do you have any suggestions where I should be surfing
the Web to look for programers?

gzotti:
I don't know if "surfing the web" will yield a result. I think most people here are obsessed developing things they urgently need themselves, but don't have the time doing it all. (We do have daytime occupations...) A calendar plugin has been on my long unwritten list of want-to-have-myselves for years. My first two algorithm books are about 2 armlengths away. Just that other things I need more urgently turned out to be trickier than expected.

Other programmers take money to implement what somebody else wants from them. Doing it takes some time to carefully study how the program works and how to best read and write dates. It should not be too difficult. But would you pay 2-3 months (estimated) for a professional "business" developer at full rate, with tax&insurance?

In principle you could also even just implement some date control as multi-calendar web interface, an extension of the RemoteControl plugin. My own Javascript is too limited for that path, though.

goedible:
It would perhaps motivate me to make more money.  I would pay if I thought
that I could afford it.  I am unfamiliar with salaries and wages for
programers, however at this time I can not aford to pay someone minimum
wage for more than a week and a half so I think I'll need some time to save
up.

PR #1339, opened 2020-10-21
Add plugin ExtraCalendar

rayw000:
### Description

Plugin `ExtraCalendar` displays multiple calendars (extensible, currently only support Chinese calendar) on the bottom-right corner. It helps user know how different calendars work with the influence of solar/lunar/etc.

Partially fixes #1313

Thanks to https://github.com/isee15/Lunar-Solar-Calendar-Converter/ 

### Screenshots (if appropriate):
image
image
image
image

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

### How Has This Been Tested?

- Test environment
  - macOS 10.15.7 Catalina
  - cmake 3.18.3
  - Qt 5.15.1
  - g++ Apple clang version 12.0.0 (clang-1200.0.32.2)
- Test steps
  - Make directory `build` under project root and enter into it
  - Run command `CMAKE_PREFIX_PATH=/usr/local/opt/qt cmake ..`
  - Run command `make -j4 && ./src/stellarium`
  - Open Configuration window, see if `ExtraCalendar` appears in the plugin tab
  - See if the description of`ExtraCalendar` plugin works as expected
  - See if `ExtraCalendar` plugin works as expected
  - Change default language of Stellarium, see if `ExtraCalendar` changes its name label as expected
  - Increase timespeed, see if `ExtraCalendar` updates correctly on local time 00:00AM
  - Press `+` and `-` to forward/backward dates, see if `ExtraCalendar` updates correctly
  - Press `F5`, change to date out of 1901~2099, see if `ExtraCalendar` disappears as expected
  - Go back to current time, see if `ExtraCalendar` shows again as expected
  - Change date to which is a Chinese leap month, see if `ExtraCalendar` works as expected
- Affected files
  - `/CMakeLists.txt`
  - `/po/stellarium/zh_CN.po`
  - `/src/core/StelApp.cpp`

**Test Configuration**:
* Operating system: macOS 10.15.7 Catalina
* Graphics Card: Intel Iris Plus Graphics 1536 MB

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
[...]

PR #1343, opened 2020-11-01
Calendars plugin

gzotti:
### Description
The calendars plugin provides a secondary data interface with a multitude of computable calendars. 

- Julian
- Gregorian
- ISO week
- Maya
- Aztec

Future additions
- Chinese
- French Revolution
- Indian
etc.

Fixes: #1313 
Related to #1339

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 20H2, Qt5.13
* Graphics Card: NVidia Geforce 960M (irrelevant)

### Issues: 
+ Our most important source for calendars provides date conversion (for noon) only, and day switches not at midnight will have to be estimated by the user. 
+ I try to follow the book as closely as possible to avoid problems. Of course, porting Lisp to C++ has some funny considerations, but on the way I see the elegance of QVector....

## Checklist:


alex-w:
Hooray! I can compile it now and he is fun!

gzotti:
I think this first version is ready for a first merge and availability in 0.20.4 while keeping the branch active for further development.

alex-w:
Please update po/stellarium/POTFILES.in file

axd1967:
- looks nice
- I'm not sure the choice of colour (white) will always work. I think there are many places in Stellarium where a semi-transparent (or even totally opaque) background would be better
image
- "what does the date range display?" =>  the notation " 8 - 12" is confusing, the format is odd (DDD, D - M (mmm) - YYYY); at least during the alpha phase, I'd stick to ISO-like formatting (Y->M->D) and wait how the Anglosaxon world will react. (tip: https://en.wikipedia.org/wiki/ISO_8601#Week_dates and https://en.wikipedia.org/wiki/ISO_week_date => e.g. 2019-W52-1. this is a good occasion to introduce the users to ISO date formats )
- the right/left alignment could maybe be improved to a fixed tabular alignment - the text jumps while the user scrolls through time (which is cool).

and finally - I'm looking forward to a Chinese calendar :-)

gzotti:
@axd1967 this is a very early preview. Call it fast beta for 0.21.0. 
- I am aware of the "odd formatting". But how would I transport that September is the ninth month in ISO format? I need both, number and name.
- The background and color should be identical to the usual object infostring panel. But I was sure somebody would be complaining about that. 
- Also on placement: It is an HTML table (Label/data), now placed into the corner. I intend to later keep a forced left-shift to avoid jumping too much. But these are obvious cosmetics. First things first.
- Time will bring more calendars. Just by itself...

alex-w (approved):
Thanks for good work!

P.S. Please check CodeFactor suggestions

axd1967:
> I am aware of the "odd formatting". But how would I transport that September is the ninth month in ISO format? I need both, number and name.

In fact the plugin now uses a format referring to "ISO format" but then uses a non-ISO format ("ddd, week W, YYYY"), which is smelly: this should be improved so that at least this format follows ISO format YYYY-`W`W-D. 

Also, ISO format does not use day / month names. So it's not clear why - at least in this case - there is a need for names.

Remember also that almost everywhere in Stellarium, there is a logical date/time ordering (Y>M>D>H>M>S), and I guess nobody has ever complained about that.

image

gzotti:
I don't refer to ISO format, but ISO week number, in the ISO Week calendar which gives GregorianYear/Weeknumber/WeekdayName. The traditional historical European calendar elements definitely need their names. 
The format is in no way finalized and yes, looks suboptimal. Just propose something that works and does not omit any data items.

axd1967:
If I were to choose, I would follow the ISO logic but not mention "ISO" because that imposes a specific format that the plugin cannot follow.

So I'd propose YYYY - MM (mmm) - DD (ddd).

axd1967:
An additional issue is that it might later be handy to copy-paste data.

The ISO format should be respected as much as possible so that the user does not need to fiddle with the copied string, so at least a space should separate the official format from the additional (dat/month) data.

e.g.
- **YYYY-MM-DD (ddd, mmm)**
- **YYYY-Www-D (ddd)**

alex-w:
IcelandicCalendar.(c|h)pp files are missing

gzotti:
Ouch, I did it again :-P

Atque:
@gzotti I built from your latest commit. If many plugins are loaded, the calendar information collides with the bottom bar and makes the text unreadable. See this screenshot:

info_collide

Is there any setting to lift it up, so it is completely above the bottom bar?

gzotti:
Not in this first version. Even on FullHD screens there is also already the problem that the main object info may collide with the left menu bar if all info bits are requested. To mitigate these issues, historians or other users who need the calendars are able to e.g. disable Exoplanets, quasars, satellites plugins, or deactivate night view, galactic grid or even the quit button.

ultrapre:
GJ, @gzotti 

In MSVC, std::runtime_error need #include <stdexcept> by https://github.com/Stellarium/stellarium/blob/d910d398924c8361109c6d9685c8896d28ab7e85/plugins/Calendars/src/Calendars.cpp#L168

gzotti:
Really? I am using MSVC, this part of the code was taken from other plugins, and our CI systems don't complain. Does it fail for you, and with which MSVC?

ultrapre:
Well, I build for MSVC 2019 Qt5.15.2

gzotti:
Can you compile the ArchaeoLines plugin without warnings? It does also not include stdexcept.
I am still on Qt5.13 and MSVC2017. Qt5.15.0 behaved badly, and I need MSVC2017 for other tasks. Maybe we need those includes now with MSVC2019?

ultrapre:
It is difficult to find source includes. Some Class include stdexcept, and plugins include it directly or not, and it also effects.

ultrapre:
Qt 5.15 is final LTS version for Qt5, and it also worth to be included in docker. Maybe it is time to add a github ci for 5.12 and 5.15, the both LTS Qt5.

gzotti:
We will certainly switch to Qt5.15.2 at some point. 5.15.0 and .1 had issues with insane memory allocation during translation. I understand there will be no open-source version later than Qt5.15.2 now that Qt6 has appeared with all its challenges we will have to face.

Atque:
@gzotti I really like this plugin. A few notes though (I'm sure you've seen them):
-Icelandic calendar says "error" on the day display when the year is set to 0 or less.
-The Olympiad calendar has errors in displayal or cardinal numbers. 21th, 2th and so on.

gzotti:
- According to the CC book this calendar was introduced only in 1700. I expect things like this can happen with this early stage of the plugin. Later I may just write "invalid date". 
- Unfortunately there is no function I am aware of to provide correct cardinality terms in multiple languages.  We may have to live with such flaws for some time, but I hope it can be understood.

axd1967:
@gzotti some calendars  (eg coptic, egytpian, maya) use native names ; some of these should not be translated, correct?

for example (nl)
- Quauhtli (Eagle) should be translated to Quauhtli (Adelaar)
- but Choiak (just a random pick) should remain untranslated?

Also, I had a small translation issue: how to translate "%4th" to (in my case) nl ? "%4e" is an approximation, but is rejected by Transifex

gzotti:
* Native names usually remain untranslated. Some languages use different character sets, so then there may be some form of transliteration. The phonetic value of the native name should not be changed, though. Of course, "eagle" etc. shall be. 
* There may be canonical translations of Aztec month names from your language, these should be preferred over own translation attempts.
* About cardinal numbers: there is no solution ATM (create one?!) for multilinguistically appropriate cardinals, but things like "2th" should be understandable from the context.

Atque:
@gzotti About the cardinal numbers: Is that even necessary? I'm not sure about English, but in Swedish, you can omit the equivalent of "th". Some languages may be more strict with this though.

gzotti:
@Atque it is up to the translators to find a solution that reasonably works with the restriction that only a simple ordinal number is placed into the string. And yes, some languages want ordinals in this context. In some, it is as easy as placing a dot after the number. Others like English need value-dependent characters, which is as it seems an unsolved problem. I just read that Swedish has no differentiation. So be it. Currently "information" counts more than "polished style", though.

axd1967:
Is it OK to keep the native Roman month names in Dutch? The names are readable by Dutch users (I guess as well in other languages).

gzotti:
I prefer to keep this very peculiar Latin way of expressing calendar dates in the Julian Calendar of Roman form unchanged in Latin, apart from transliteration into a different character set where required. If I have "January" in the Julian calendar and "Ianuarii" (genitive form of "Ianuarius"), I hope most users are capable of drawing conclusions about the linguistic relationship and origin of month names in the European languages. Placing a Dutch month name into the Latin string may look awkward.

PR #1361, opened 2020-11-12
Improve planets' halos size in relation to stars'

### LINKED ISSUE #367: halo size mismatch between stars and planets
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1595463

Anonymous from
https://sourceforge.net/p/stellarium/discussion/278769/thread/9bac03cd/?limit=25#082e/b52c

Indeed the halo sizes for bright stars and much brighter planets appear about equally large. Planets brighter than the brightest stars may deserve a slightly larger halo.

Atque:
Yet another try at the bothering problem with planet halos being significantly smaller than their stellar counterpart. It also helps visualizing how bright Venus is in relation to Jupiter. My last try had a problem where the Moon's halo stayed on screen at too small FOV's, but this try fixes that.

(kinda) Fixes #367

### Known issues
Stars are still brighter than planets when the relative scale is set to a high value (>1.5 or so). To fix this, I think we need a whole new way to draw planet halos.

improve_halos

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Irrelevant
* Graphics Card: Irrelevant

## Checklist:

### DIFF
--- a/src/core/StelSkyDrawer.cpp
+++ b/src/core/StelSkyDrawer.cpp
@@ -521,11 +521,11 @@ void StelSkyDrawer::postDrawSky3dModel(StelPainter* painter, const Vec3f& v, flo
 	// so that the radius of the halo is small enough to be not visible (so that we see the disk)
 
 	// TODO: Change drawing halo to more realistic view of stars and planets
-	float tStart = 3.f; // Was 2.f: planet's halo is too dim
+	float tStart = 3.f; // Was 2.f: planet's halo is too dim. Atque 2020-11-12: No need to change these anymore. It appears that this has to do with halo size vs FOV (?).
 	float tStop = 6.f;
 	bool truncated=false;
 
-	float maxHaloRadius = qMax(tStart*3.f, pixRadius*3.f);
+	float maxHaloRadius = qMax(tStart*6.f, pixRadius*3.f); //Atque 2020-11-12: Careful, if tStart*6.f is too big (tStart*10.f or something), the Moon gets a ridiculously big halo.
 	if (rcm.radius>maxHaloRadius)
 	{
 		truncated = true;
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -2711,6 +2711,8 @@ void Planet::draw3dModel(StelCore* core, StelProjector::ModelViewTranformP trans
 			haloColorToDraw.set(haloColor[0], static_cast<float>(pow(0.75, extinctedMag)) * haloColor[1], static_cast<float>(pow(0.42, 0.9*extinctedMag)) * haloColor[2]);
 		else
 			haloColorToDraw.set(haloColor[0], static_cast<float>(magFactorGreen) * haloColor[1], static_cast<float>(magFactorBlue) * haloColor[2]);
+		if (this==ssm->getMoon())
+			haloColorToDraw*=0.6f; //GZ: Attempt to make lunar halo less glaring...
 
 		core->getSkyDrawer()->postDrawSky3dModel(&sPainter, tmp.toVec3f(), surfArcMin2, getVMagnitudeWithExtinction(core), haloColorToDraw);
 


Atque:
Another "issue" (rarely seen, may not be important enough to fix): The planet halos seem fine up to about -6-7. This means that from Jupiter, Io (-9.9) will look smaller than the SN1006 (-7.5). This is easily fixable by setting _tStart*6.f_ to 10 or some even higher value. The problem is that the moon looks really, really bright from earth with relative size set to 1.

Here a full moon at 60 FOV:
moontoobig

Opinions @alex-w, @gzotti? I personally think this problem is too rare to even care about. But once you set the relative scale to something more pleasant, say 0.65, it looks fine even with crazy numbers.

alex-w:
I've little played with your branch tonight and the view of the halo for planets looks better than previously. I see some overbright halo for the Moon, but this is a very old problem.

gzotti:
Sorry, no time these days for details. We could of course limit the max. brightness for the moon halo. Then the minuscule sphere remains visible, surrounded by a large but not white disk.

Atque:
@alex-w I'm glad you like it. Do you think 6 or 10 is the better? 10 fixes so that Jupiter's moons' halos are bigger than the brightest supernova (SN1006) at relative scale 1.
@gzotti I like the idea of the moon being a special case. But I can't help with that, unfortunately.

alex-w:
6 is good for me.

Atque:
Okay, then I will keep the changes.

gzotti:
Ad-hoc attempt in Planet.cpp, starting at line 2709. Add two lines as in:
```

Vec3f haloColorToDraw;
if (this==ssm->getSun())
	haloColorToDraw.set(haloColor[0], static_cast<float>(pow(0.75, extinctedMag)) * haloColor[1], static_cast<float>(pow(0.42, 0.9*extinctedMag)) * haloColor[2]);
else
	haloColorToDraw.set(haloColor[0], static_cast<float>(magFactorGreen) * haloColor[1], static_cast<float>(magFactorBlue) * haloColor[2]);
if (this==ssm->getMoon())
	haloColorToDraw*=0.6f; // Attempt to make lunar halo less glaring...

```

Probably we should use a dedicated "Moon Halo" texture for a slightly larger aureole.

Atque:
@gzotti Thanks a lot for that solution. I tried it and it looks okay to me. Now you can see the lunar phase even at 60+ FOV; something many (inexperienced) users complain about in Stellarium ("Why does the moon not show phases in Stellarium?").

Should I add this to the PR?

gzotti:
I think it's OK for me.  @alex-w ?

alex-w:
> Should I add this to the PR?

Yes, please

alex-w (approved):
OK, thanks!

10110111:
What are these halos supposed to represent? An atmospheric effect ([aureole](http://www.instesre.org/Solar/AureolePhotography/Aureole.htm))? Or simply diffraction in the lens ([bloom](https://en.wikipedia.org/wiki/Bloom_(shader_effect)))?

Atque:
@10110111 How else to represent difference in brightness, than to increase the size of the texture used? When I look at the sky,  Venus appears bigger than e.g. Sirius (hence the word magnitude, from Latin _magnus_ 'big').

10110111:
OK, so this is not an atmospheric effect, but something observer-dependent—like bloom. That's all I wanted to know, thanks.

gzotti:
The "halos" around the planets are used to represent them as star-like objects when not magnifying enough to show them as textured illuminated spheres. A textured illuminated sphere of invisibly small visible diameter would simply vanish from screen. When you zoom out enough, even the moon's sphere appears too small to be discernible. Now that we found out where the lunar halo is drawn, we could adapt it even more to take its magnitude into account or whatever. Some phase angle dependent brightness value or even halo texture of variable diameter may be nice.

axd1967:
in the night of 2021-08-12 I could clearky see Jovian sats in a 7x50; there was less "smear/halo" than is shown here:

image


the two moons (Ganymedes, Io) were well visible, I can't recall I could see Europa (r): that was the equivalent of this:

image

In other words, the halo of Jupiter was a bit too large. (I'm aware this is a tricky issue.)

I tried the other magnitude algorithms, they were less satisfactory ; the 2013 and 2018 gave the closest approximation.

(note that I do not have perfect vision.)

Atque:
@axd1967 The halo size is good compared to that of the stars (previously, Sirius and a few other stars looked brighter than Jupiter and even Venus). To be frank, the stars look quite ugly when the relative size is set to 1. A better halo texture or a shader might fix that. I have a repository that you can try (and improve of you find better balance).

axd1967:
I did not experiment much with absolute/relative star sizes - but I now notice that the relative scale also applies to planets, and following is a much better approximation of what I saw:

image

Atque:
@axd1967 Maybe it would be possible (and not too difficult) to separate the relative sizes of planets and stars? So that you can choose if you want to view planets as large, and stars as small.

This is beyond my programming capabilities.

gzotti:
The halo sizes of stars and planet used to be different, to avoid planet sizes becoming too large. This is the result of a "harmonisation" I soon was not really happy with...

Note that we use distinct settings for naked-eye and ocular view, so you should be able to configure your favourite ocular view.

Atque:
> The halo sizes of stars and planet used to be different, to avoid planet sizes becoming too large. This is the result of a "harmonisation" I soon was not really happy with...

The change is most noticeable when the relative size is large, i.e. close to 1 or even higher. When giving a realistic view, such high values are unrealistic because bright stars in reality do not look like bright blobs. 1 is good when finding constellations, and large planets is no problem then.

gzotti:
A setting of 1/1 is considered "normal" and should be an optimal compromise. Sure, the brightest stars are large, but yes, this is for identification of constellations. We don't aim at reproducing ordinary photographs (with oversaturated pixels). Maybe think of Akira Fuji style. 
And I really dislike having to tweak this setting now frequently, e.g. as said, for the Galileans or other planet close-ups.

Atque:
> A setting of 1/1 is considered "normal" and should be an optimal compromise. Sure, the brightest stars are large, but yes, this is for identification of constellations. We don't aim at reproducing ordinary photographs (with oversaturated pixels). Maybe think of Akira Fuji style.
> And I really dislike having to tweak this setting now frequently, e.g. as said, for the Galileans or other planet close-ups.

I have tried, and not found a better balance or texture to help getting stars to look better than they already do. So it seems we need a whole new way to solve the realism part.

About the tweaking, maybe this can be solved by splitting the planet and stellar drawing into two separate options. Draw planets as 0.5 while stars at 0.75 or whatever preference the user wishes. I leave this task to someone else though.

axd1967:
I was tempted to make an issue of it, now I'll leave it here as a related comment:

### Expected Behaviour
No significant change in light intensity; I would expect the satellite to keep its magnitude.

### Actual Behaviour

Abrupt change in intensity

image

One second later, Io lights up:
 
image

Same issue as #1804, by the way.


ps - and just to play with the settings, one then gets something totally unrealistic ...

image

axd1967:
I was tempted to make an issue of it, now I'll leave it here as a related comment:

### Expected Behaviour
No significant change in light intensity; I would expect the satellite to keep its magnitude.

image
image



### Actual Behaviour

Abrupt change in intensity

image

One second later, Io lights up:
 
image

Same issue as #1804, by the way.


ps - and just to play with the settings, one then gets something totally unrealistic ...

image

PR #1371, opened 2020-11-19
Use python3

alex-w:
Let’s try switch travis to use python3 by default on linuxes


alex-w:
No need more... outdated

PR #1383, opened 2020-11-25
Added JPL DE430 for unit tests

alex-w:
I've added new one step for CI/Coveralls action (downloading JPL DE430 ephemeris) to allow use unit tests for JPL DE430 ephemeris in additional to VSOP87 ephemeris. This step add extra 10 seconds to all time of running action (for example updating installed packages add extra 5 minutes).


alex-w:
@gzotti is it OK for you?

PR #1384, opened 2020-11-27
Drop the XPM version of the app icon

pinotree:
The application icon is already provided as PNG in different sizes,
and installed in the global hicolor icon theme. Hence, the XPM version
of it can be dropped, as it is redundant.


alex-w (approved):
Thanks!

pinotree:
Whoa, that was fast... thanks!

PR #1387, opened 2020-11-29
text fix for meteor showers plugin

ultrapre:
disambiguation. diff with main search dialog.

And, there may be similar situations in other place.


alex-w (changes requested):
Please fix line 148 too

alex-w (approved):

PR #1388, opened 2020-11-29
Add a config option to use multisampling for rendering "Planet"s

10110111:
## Description

This adds a new option, `video/multisampling`, that lets one enable multisampling antialiasing for rendering planet-like objects (`class Planet`) and specify number of samples to use.

Currently there's no GUI for this. Moreover, if one were to add a GUI control, this control should tell the user to restart Stellarium for changes to take effect, because the OpenGL format is applied at the creation of `StelGLWidget`. Implementation of it without requiring a restart would likely need quite a bit of work.

## Why would one want MSAA in Stellarium?

Just look at the following animation [no-msaa.mp4.gz](https://github.com/Stellarium/stellarium/files/5612648/no-msaa.mp4.gz). Note the borders of the Moon (the same would be with the Sun etc.): not only are they jagged, but the motion of the object makes these jaggies even more prominent and ugly. And here's what it'll look like after the patch: [with-msaa.mp4.gz](https://github.com/Stellarium/stellarium/files/5612650/with-msaa.mp4.gz).

## Some warnings for users
Be careful when choosing the value for the option. On my system with GeForce GTX 750 Ti Linux `nvidia` driver 390.116, if I set the setting to something ≤8, everything is OK. But if I try a higher value, NVidia driver appears to enable FXAA instead of (or in addition to) MSAA, which ruins sharpness of the GUI widgets and text. And that's even if the value is less than `GL_MAX_SAMPLES` (which is 32 on my system).

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -254,6 +254,7 @@ Planet::Planet(const QString& englishName,
 	  hidden(hidden),
 	  atmosphere(hasAtmosphere),
 	  halo(hasHalo),
+      multisamplingEnabled_(StelApp::getInstance().getSettings()->value("video/multisampling", 0).toUInt() != 0),
 	  gl(Q_NULLPTR),
 	  iauMoonNumber(""),
 	  positionsCache(ORBIT_SEGMENTS * 2)
@@ -2620,6 +2621,11 @@ void Planet::draw3dModel(StelCore* core, StelProjector::ModelViewTranformP trans
 		transfo2->combine(Mat4d::zrotation(M_PI_180*static_cast<double>(axisRotation + 90.f)));
 		StelPainter sPainter(core->getProjection(transfo2));
 		gl = sPainter.glFuncs();
+
+        if(multisamplingEnabled_)
+            gl->glEnable(GL_MULTISAMPLE);
+        else
+            gl->glDisable(GL_MULTISAMPLE);
 		
 		// Set the main source of light to be the sun
 		Vec3d sunPos(0.);
@@ -2688,6 +2694,8 @@ void Planet::draw3dModel(StelCore* core, StelProjector::ModelViewTranformP trans
 
 
 		core->setClippingPlanes(n,f);  // Restore old clipping planes
+        if(multisamplingEnabled_)
+            gl->glDisable(GL_MULTISAMPLE);
 	}
 
 	bool allowDrawHalo = true;
--- a/guide/app_config_ini.tex
+++ b/guide/app_config_ini.tex
@@ -793,7 +793,8 @@ screen\_x        & int    & Horizontal position of the top-left corner in window
 screen\_y        & int    & Vertical   position of the top-left corner in windowed mode. Value in pixels, e.g. \emph{0}\\%\midrule
 viewport\_effect & string & This is used when the spheric mirror display mode is activated. Values include \emph{none} and \emph{sphericMirrorDistorter}.\\%\midrule
 minimum\_fps     & int    & Sets the minimum number of frames per second to display at (hardware performance permitting)\\%\midrule
-maximum\_fps     & int    & Sets the maximum number of frames per second to display at. This is useful to reduce power consumption in laptops.\\\bottomrule
+maximum\_fps     & int    & Sets the maximum number of frames per second to display at. This is useful to reduce power consumption in laptops.\\%\midrule
+multisampling    & int    & Sets the number of samples to use for multisampling antialiasing. 0 disables antialiasing, higher values increase smoothness and degrade performance. Too high a value may result in excessive blurriness of the GUI.\\\bottomrule
 \end{tabularx}
 
 \subsection{\big[viewing\big]}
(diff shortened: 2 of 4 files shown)


gzotti:
Looks good, thank you!
For now, this can remain a (user guide documented!) entry. 
- What are really useful values? any [0...8] ? Or [0, 2, 4, 8]? These could be the allowed values in a drop-down combo or so. 
- What happens with 1 vs. default 0?

10110111:
> What are really useful values? any [0...8] ? Or [0, 2, 4, 8]? These could be the allowed values in a drop-down combo or so.

I think this is system-dependent. At first I hoped it would be [0..GL_MAX_SAMPLES], but it appeared that good values are not all of these (due to the silent addition of FXAA by `nvidia`). So for GTX 750 Ti it's [0..8]. On more powerful GPUs larger values might also be usable.

> What happens with 1 vs. default 0?

The value of 0 disables multisampling completely (i.e. via `glDisable`). The value of 1 enables no-op multisampling (uses single sample per fragment).

gzotti:
Yes, I have seen that. So if 1 looks like 0, 1 is useless and users will be rather confused than happy. Maybe we should use a dropdown: off/2/4/8, next to the dithering setting?

10110111:
> Maybe we should use a dropdown: off/2/4/8, next to the dithering setting?

I don't think it's useful to exclude 3, 5, 6, 7. Maybe better a spinbox. Plus, users' machines may have higher values usable, and it would be counter-productive to restrict their usage.

10110111:
BTW, a spinbox can have a special value to mean "off", see [`QSpinBox::setSpecialValueText`](https://doc.qt.io/qt-5/qabstractspinbox.html#specialValueText-prop).

alex-w:
Could we get value of GL_MAX_SAMPLES for graphic card to define maximum value of spinbox?

10110111:
> Could we get value of GL_MAX_SAMPLES for graphic card to define maximum value of spinbox?

I think this would be a reasonable maximum, yes. Technically, it's just a matter of a call to `glGetInteger` (needs a GL context).

gzotti:
But if >8 causes problems, I would prefer simply not to have it. Else I know at least one who will complain.

10110111:
It seems we can't easily find out the actual HW limits using OpenGL. But we

1. can ask Vulkan API, see `vulkaninfo | sed -n '/framebufferColorSampleCounts/,/framebufferDepthSampleCounts/p'`
1. know that it's only nvidia problem
1. if Vulkan query fails, just use `GL_MAX_SAMPLES`.

This is of course another whole problem to get to this info: I don't know whether the problem is present on Windows (thus whether there's any need to query Vulkan info there), and on Linux I suppose it'd be best to avoid parsing the output of `vulkaninfo` (because its format changes over time), so we'd have to load `libvulkan.so.1` and try to resolve the necessary API function there.

10110111:
As for

> But if >8 causes problems, I would prefer simply not to have it

On my Intel UHD Graphics 620 I have `GL_MAX_SAMPLES=16`, and there are no problems when I use this value. So restricting to 8 would be actively harmful to users of similar GPUs.

src/core/modules/Planet.cpp
```diff
@@ -2620,6 +2620,11 @@ void Planet::draw3dModel(StelCore* core, StelProjector::ModelViewTranformP trans
 		transfo2->combine(Mat4d::zrotation(M_PI_180*static_cast<double>(axisRotation + 90.f)));
 		StelPainter sPainter(core->getProjection(transfo2));
 		gl = sPainter.glFuncs();
+
+        if(StelApp::getInstance().getSettings()->value("video/multisampling", 0).toUInt() != 0)
```
alex-w, line 2624: I think this part should be moved into init() and here just use some variable to speed-up comparison and avoid extra reading config.ini data

src/core/modules/Planet.cpp
```diff
@@ -2688,6 +2693,7 @@ void Planet::draw3dModel(StelCore* core, StelProjector::ModelViewTranformP trans
 
 
 		core->setClippingPlanes(n,f);  // Restore old clipping planes
+        gl->glDisable(GL_MULTISAMPLE);
```
alex-w, line 2696: No need run glDisable if glEnable was not used for GL_MULTISAMPLE

alex-w (changes requested):
Please improve commented parts of code

src/core/modules/Planet.cpp
10110111: Moved to the constructor

alex-w (approved):

alex-w:
I think we can skip GUI option at the moment and ask users to do some experiments

alex-w (changes requested):
@10110111 please add info for config option into Stellarium User Guide

10110111:
@alex-w Please check that it renders as desired. I couldn't check locally because my TeXLive installation lacks some packages.

alex-w (approved):
Thanks!

PR #1411, opened 2020-12-15
guide: remove duplicate line

jackschmidt:
The user guide, chapter 18.2.4, Astronomical Phenomena, Stars,
Spectral Type & Luminosity Class, contained the following line twice:

```
in figure~\ref{fig:colourmag} (which is technically a colour/magnitude
```

This commit removes the duplicate copy.


alex-w (approved):
Thanks for fix!

PR #1419, opened 2020-12-20
a few suggestions for calendar formats

axd1967:
Suggestions for improvements

image

- follow ISO format more closely
- try to use less confusing formats, pending better proposals and maye use of locale (see https://en.wikipedia.org/wiki/Calendar_date#Advantages_for_ordering_in_sequence)

(Julian: I suspect that AD/BC needs to be separated fro the year number; also, avoid dots in the abreviation).

### DIFF
--- a/guide/plg_calendars.tex
+++ b/guide/plg_calendars.tex
@@ -14,24 +14,24 @@
 \subsection{Introduction}
 \label{sec:plugin:Calendars:Introduction}
 
-The calendar dates in the main program behave like most other astronomical software titles:
+The calendar dates in the main program behave like most other astronomical software utilities:
 
 \begin{itemize}
 \item Dates are given in the Gregorian calendar for all dates
   beginning with October 15, 1582.
 \item All earlier dates are given in the Julian Calendar in its
   finalized form by Augustus. Historically, only dates beginning with
-  March 1st, 4 A.D. coincide with historically recorded dates: the
+  March 1st, 4 AD coincide with historically recorded dates: the
   Roman priesthood messed up the 4-year count introduced by Julius
   Caesar and counted leap years every third year. Augustus decreed to
-  omit leap days from 12 B.C. to 4~A.D to move the seasons to where
+  omit leap days from 12 BC to 4 AD to move the seasons to where
   Julius Caesar had placed them.
 \item Given the errors in the Julian calendar, simulation in early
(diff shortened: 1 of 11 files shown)


axd1967:
with ISO day of week

image

alex-w:
> with ISO day of week
> 
> image

Alex, do you see no difference between ISO week and ISO date?

axd1967:
> > with ISO day of week
> > image
> 
> Alex, do you see no difference between ISO week and ISO date?

There is a difference between ISO week format ("YYYY-Www-D") and ISO date format ("YYYY-MM-DD").

Maybe the plugin primarily wants to convey the "ww" part of the official ISO week format?

Which reminds me that I didn't read the tex file. I'll take a look there, maybe the answer is there.

alex-w:
ISO Week != ISO Week with weekday

See [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) for details

gzotti:
"ISO week"  in this context is nothing about standardized formatting of a date string but about the standardized counting of weeks.

axd1967:
@gzotti  please consider the suggested changes (especially the user guide still contains errors that I corrected). 

especially:

- use of A.D. / B.C or A.U.C is not recommended. avoid dots in abbreviations (see Wikipedia). in this particular case, it will save on pixel space too
- BCE can as well be written BC as far as I know
- there still are several typos
- according to Wikipedia , the correct spelling is " Haab` ", not "Haab"; "Uayeb" is also known as " Wayeb' ".

gzotti:
* Our CC source book and other historical literature is consistently using with-dot notation. Also, A(.)D(.) is as it seems more often written in front of the number. I currently write it after. Such details may change as the plugin is evolving. 
* B.C.E.=Before Common Era. Some people prefer to not be reminded of a Christian era, using a more neutral term. 
* I may have a further look into typos and improvements for the SUG later this week, but now I have other tasks and priorities. Now I will let the translators do their work before version 0.20.4 will materialize.
* Spelling and transliteration systems differ by target language and, as it seems, development of literature. I am using a citable reference. As good as WP is, in this case our source book wins for now. However, for the Zoroastrian months (omitted from CC) I had to make the compromise to use CC's transliteration (diacritics) scheme and not the early-20th-century diacritics used by Ginzel.

PR #1442, opened 2020-12-28
Reduce the size of some texture files

sunshuwei:
Most of them have a small angular size on the celestial sphere, but have a 1024×1024 resolution.
Reduce the size
There is almost no effect on the display effect after reducing the texture size.


alex-w (approved):
Thanks!

gzotti:
Just a remark: If reducing file size is a topic here, we'd better use JPG textures with not too high compression.

PR #1449, opened 2020-12-31
CN_SC fix

ultrapre:
Those unused 0 values should be removed. They also caused bugs cross platforms (crash).


alex-w:
> Those unused 0 values should be removed. 

If you want delete them, then why you propose partial solution?

> They also caused bugs cross platforms (crash).

windows, Mac, linux, bsd... haven't problems for those "unused" constellations. Why android have it?

ultrapre:
Maybe I need more time to debug and find why. Removing those lines actually also follows the logic.

ultrapre:
Ummm, maybe it is a bad PR.

PR #1395, opened 2020-12-09
[style fix] QColorDialog

ultrapre:
fix style of QColorDialog.
Although style fixed, QColorDialog cannot be embed into MainView like other. Any way?

image

PR #1451, opened 2020-12-31
Fix styles of QMessageBox and QColorDialog

ultrapre:
Similar PR as #1395

image

image

image

image

image


alex-w (changes requested):
The text in QMessageBox is very difficult to read with this style - please add QSS rule to resolve this problem.

ultrapre:
> The text in QMessageBox is very difficult to read with this style - please add QSS rule to resolve this problem.

add new rules into .css or remove fix for QMessageBox?

alex-w:
Probably add new rule into .css will be better solution

ultrapre:
@alex-w  How about this:

image

image

alex-w:
Good! Maybe same background style apply to color chooser also?

ultrapre:
Okay, wait for moment

alex-w (approved):

ultrapre:
https://github.com/Stellarium/stellarium/commit/5a7beec0ceea7af4504eb129deb1ca67d7eadff7#commitcomment-45571558

PR #1468, opened 2021-01-10
Make StelToneReproducer's xyYToRGB.glsl core function reusable

10110111:
I'd like to avoid duplication of the tone reproducer's code when I add the second atmosphere model, so these commits make it reusable (and more readable).


alex-w (approved):
Thanks!

10110111:
Mmm.. I tried hard to keep these two commits separate to make `git blame` easier to use if needed...
OK, too late to undo the squashing.

alex-w:
ouch... sorry

PR #1469, opened 2021-01-13
Fix display gamma application in Tone Reproducer

10110111:
The main commit here fixes incorrect application of display gamma in the Tone Reproducer's _xyY_ to _RGB_ conversion. Namely, instead of applying gamma correction to luminance _Y_, it's now applied to _R_,_G_,_B_ components separately. This does change the colors of the sky: now they will be not as oversaturated as they used to be. Basically, now the photopic-mode chromaticities will be rendered exactly as the sky model computes them.

In addition to this, the never-used C++ version of the same conversion is **deleted**. This is for the sake of maintainability: this code is not used anywhere, it's unlikely to ever be, and this means that if a change in it introduces a bug, we'll never know—due to inability to test.

And a bit of refactoring: a small bit of redundant tone mapping code that got into atmosphere vertex shader is also removed.

Please don't squash these commits when merging: they are quite separate in their purpose, and squashing them will make this part of git history harder to use. If you need them to be rebased, just tell me, and I'll do it.


gzotti:
Thanks for this. I have not tested the new code yet, hopefully on the weekend. The StelTonemapper is based on a few papers from the literature. Can you please cross-check and note deviations from those models in the hpp file? I really aim at making the source code more traceable. 

As I can see we can do a Rebase&merge in the web interface, this should keep the commits separate.

alex-w:
Hmm... the sky is almost gray - did you expected it?

10110111:
> Hmm... the sky is almost gray - did you expected it?

It's less saturated, for sure. But when the Sun is high, I get a blue color that looks much more realistic to me than what current Tone Reproducer yields. When it's low, then indeed it's a bit grayish. But it's just wrong in another way: to me evening skies in Stellarium have never looked realistic.

Anyway, this is what the Preetham's model, as currently implemented, predicts. Maybe one could play with its parameters (I've not tried) to make it better at small solar elevations, but I wouldn't bet on good results (I'm quite skeptical of this model).

gzotti:
Indeed Preetham's model has its problems (you may know my short review many years ago). I have made a branch years ago (gz_AtmosphereTweaks) where I wanted to fine-tune and allow playing with the nonintuitive constants. It would also allow making the weirdest of atmospheres (working in CIE XYZ (or Yxy) allows interesting non-intuitive effects like red solar areas in a green sky). However the branch is pretty outdated now and I had hoped was soon to be obsolete when I saw your work :-)

alex-w (approved):
Well, the code works and works well, but for me the view of the evening/morning sky has very strange saturation. On the other side I never saw the sky, which was created by Preetham's model and standard saturation.

data/shaders/atmosphere.vert
```diff
@@ -47,22 +47,14 @@ void main()
 	highp vec3 unprojectedVertex = skyColor.xyz;
 	highp float Y = skyColor.w;
 	highp float x,y;
-	if (Y>0.01)
```
gzotti, line 50: Deleting this distinction disables the separation between day and night vision. I have to go into more details here and dig out the original papers, but to my understanding our current code is close to the published models, and the colors look like most other implementations on various game engines.

src/core/StelToneReproducer.cpp
```diff
@@ -89,56 +89,5 @@ void StelToneReproducer::setWorldAdaptationLuminance(float _Lwa)
 	alphaWaOverAlphaDa = alphaWa/alphaDa;
 	term2 = (stelpow10f((betaWa-betaDa)/alphaDa) / (M_PIf*0.0001f));
 	lnTerm2 = std::log(term2);
-	term2TimesOneOverMaxdLpOneOverGamma = std::pow(term2*oneOverMaxdL, oneOverGamma);
-}
-
```
gzotti, line 94: Just a few days ago we had a request to add sky brightness etc. data to the mouse position coordinate plugin, which I basically agree to add when the numerical data make sense. This should deliver Y [cd/m^2] (coming from Schaefer's sky model?) and could of course deliver also Yxy or CIE RGB reference data if this code is still available. (There may be some use in RemoteControl connected apps to be yet developed.) I think it is simpler to call such a function for it than to get samples from the GPU (?) So please, instead of deleting, double check to see if this is the standardized CIE formulation, and update documentation.

data/shaders/atmosphere.vert
10110111: No, this just removes the redundancy with respect to `xyYToRGB.glsl`, where this same check is already done, and where x,y values are overridden.

src/core/StelToneReproducer.cpp
10110111: I'm not sure what you mean. This hunk just replaces one cached term (that includes `...^(1/γ)`) with another (that doesn't include it).

gzotti: My concern is that by removing the whole conversion function (xyYToRGB()) may remove functionality that we were asked to make more accessible just a few days ago. I recommend double-check to ensure these are current CIE RGB values, document more clearly, and indeed implement that lookup/feedback in the mouse coordinate plugin (like a virtual sky brightness sensor).

alex-w: @10110111 see issue #1404 for details

10110111: @gzotti 
As I see in #1404, it asks for _luminance_. Luminance is the _input_ for `xyYToRGB()`. Moreover, even if someone asked us for RGB values (be it sRGB, CIE RGB or whatever else kind of RGB), Tone Reproducer is a bad object to ask for them.

Why? Because it's not meant to provide anything objective at all. It's supposed to take objective data (CIE xyY color coordinates) and try to simulate _perception_ of these colors, adapting them to the user's display device and the assumed user's color adaptation. To this end, it uses XYZ-to-sRGB conversion matrix in one case, does blue-shifting in other, or simply substitutes a constant blue in the rest of cases. Plus gamma correction.

This means that, if we wanted to use `xyYToRGB()` to satisfy #1404, we'd better stop and just return the Y parameter as the luminance requested. If we had another feature request that asks for chromaticity, we should take the x,y,Y parameters and do the conversion to whatever color space desired. But in no case should we even _touch_ Tone Reproducer for these purposes. It could just as well not exist at all, without any effect on these features.

gzotti: Yes, this one only asks for luminance. But we could provide other data as well when we already have them. SB else (not on github) asked me for a "sky color query".  This may simply sample Stellarium's sky, but I am not yet sure what best to return (I have currently no time to go into this). CIE Yxy, CIE XYZ, CIE RGB, sRGB, ...?

10110111: If you want true absolute colors, the best is either xyY or XYZ. Any RGB variant, I suppose, is not too useful: sRGB may in general become negative for very saturated colors, and also has its gamma, CIE RGB doesn't seem to be used anywhere. XYZ is always non-negative and widely used. xyY is when you want to separate chromaticity from luminance in the XYZ coordinates.

Anyway, removal of Tone Reproducer's `xyYToRGB` doesn't affect this in the slightest.

alex-w:
@gzotti this PR is OK for you?

gzotti:
Please give me another week, I need to study the relevant papers, but have no time these days (deadlines...).

gzotti:
I am really sorry I have to postpone to the next release. Something looks too weird to approve, but I cannot investigate this now.

10110111:
This seems to be going nowhere. Maybe we could do it in a bit kludgy way, but with the benefit of not affecting the old atmosphere model: have a separate tone reproducer for the new model, with custom shaders etc. Not nice architecturally, but might at least let us move forward. Eventually we might just ditch the old model with the old tone reproducer altogether.

Or are you planning to review this soon?

gzotti:
Yes, sorry for the endless delay!

I have indeed started to read into the sky-related code again last weekend, but was soon blocked by other review tasks (and indeed real world tasks) now. I hope to have this solved by end of November. The situation now is unclear. The current blue looks good and familiar, but your new greyed-out sky colors look bad and nothing like any other Preetham sky implementation I have ever seen. But I cannot currently say why.

The old (Preetham) sky model must stay for the benefit of less capable systems (OpenGL ES2, Geforce 8xxx etc.) There are still enough such systems.

gzotti:
I have dug out the relevant papers and started re-reading, but won't make it by Dec.22. I still have not found what exactly makes the tonemapped sky look so terrible in this change.

gzotti:
I have now updated my old branch and modified it to reproduce both the old tonemapping (with extra gamma, more colorful) and your new tonemapping. It also can switch between pure Preetham and Preetham+Schaefer, and users can play with many parameters. I would prefer to merge that one, and see how then Bruneton will behave with both ways of Tonemapping.

10110111:
https://github.com/Stellarium/stellarium/pull/1469/commits/e368c924d4c5af0342b4b80bee2ed268dee11a78 that isn't present in your branch still remains valid. It does reduce mix-up of functionality, while not affecting visual output.

gzotti:
Indeed --- fixed.

10110111:
OK, please let me know (maybe at #624) when I can start rebasing my [showmysky](https://github.com/10110111/stellarium/tree/showmysky) branch against the new `master`.

Issue #1350, opened 2020-11-05
Lunar magnitude formula incorrect?


Atque:
Comparing Stellarium to other planetaria (mainly Cartes du Ciel, but also some others such as CyberSky 5) shows that Stellarium is the odd man out when it comes to displaying the lunar apparent magnitude. A full supermoon has about -12.44 in Stellarium, while in the other programs the value is more like -12.7. The difference is more prominent when you compare full vs half moon. In Cartes du Ciel, the half moon is about -10, while in Stellarium its in -11. That means the difference between a full and half moon in Stellarium is about 1.4 magnitude (a factor 3.8 difference), while in the other programs it's 2.7 (a factor 12.2 in difference).

Stellarium also seems to lack the light boost effect of a full moon, which is present in the other programs I mentioned. The difference between a 95% and 100% illuminated moon in Stellarium is about 0.12 (factor of 1.11), while in Cartes du Ciel, it's a difference of 0.8 (factor 2.12).

This graph shows lunar magnitude as a function of its phase:
![alb01](https://user-images.githubusercontent.com/37028368/98210072-4c18e500-1f40-11eb-9972-2ad55d5f9868.jpg)

I guess there are better formulae online, but sadly I can't provide one.

gzotti:
Yes, there must be better formulae, but sadly we have not found them yet. The current computation for the Moon is still the generalized model from Stellarium's early ages. Just develop an algorithm from your graph...

Atque:
@gzotti It "should" be fairly easy considering several planetaria use the same formula. I'd love to help but I think my abilities are inadequate.

I have no idea how to validate this since I lack mathematical and programming experience apart from a few university courses, but a simple Google search gave this formula (Allen, 1976):
https://astronomy.stackexchange.com/questions/10246/is-there-a-simple-analytical-formula-for-the-lunar-phase-brightness-curve

It has the brightness boost.

gzotti:
Thanks. These are at least some hints. Of course, the PASP paper may also contain a solution.

gzotti:
Next reference for you or SB else to locate: Whitaker, E. A. 1969, in Analysis of Apollo 8 Photography and Visual Observations, NASA SP-201, 38.

10110111:
> Edit: It appears I can't read properly. It does NOT have the brightness boost. I better keep on searching.

[...]

Issue #1115, opened 2020-06-06
Error with lunar eclipse texture and magnitudes


Atque:
Using the latest beta (0.20.1.17986), we learn that a neat addition to the infostring has been added: umbral and penumbral magnitudes. Very nice.

But there is a problem. They don't get along with the eclipse texture nor the (visual) magnitudes given in the infostring.

To see what I mean, set the time to 2019-01-21, 04:41:16 (UTC) (location on earth is irrelevant), Look at the values in the infostring; the moon is supposed to be completely within the umbra, yet the texture and the visual magnitude values suggest that it is not. The time fits the "official" values on almost exactly, which means the umbral magnitude is correct.

See the screenshot:
![stellarium-012](https://user-images.githubusercontent.com/37028368/83940060-12ea1680-a7e2-11ea-8154-9762b0d75fc9.png)

The article on Wikipedia gives the total eclipse start at 04:41:17 UTC, so Stellarium is just 1 second off! 
https://en.wikipedia.org/wiki/January_2019_lunar_eclipse
But the texture and the visual magnitude gives the total eclipse start at 04:43:25 UTC, more than 2 minutes off.

gzotti:
Unlike the razor-sharp edge of the Moon during a total Solar eclipse, Earth's umbra during Lunar eclipse has no clear edge. My last round of adjustments (late 2018) pretty closely matches my own photographs. 
![TLE20180727_GZ_crop (Individuell)](https://user-images.githubusercontent.com/18099873/83941115-03220080-a7e9-11ea-8b99-da8037eaf8bc.jpg)
![TLE20180727_Stellarium_crop (Individuell)](https://user-images.githubusercontent.com/18099873/83941118-087f4b00-a7e9-11ea-9cc4-35b0b7e5e263.png)

It may be easy to define the borders geometrically, simulating the visual appearance is much more difficult. From other things like stellar or planet occultations I know that Stellarium is rather 1 minute away. When this can be solved, some further fine-tuning of the eclipse texture and other minor issues can be performed.

The magnitude during eclipse is a rough guess. Can you supply us with a comprehensible, reproducible model that fits better?

Atque:
@gzotti 
> The magnitude during eclipse is a rough guess. Can you supply us with a comprehensible, reproducible model that fits better?

[...]

Issue #430, opened 2018-09-26
Display of earth's shadow


alex-w:
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1628544

From Anonymous: https://sourceforge.net/p/stellarium/discussion/278769/thread/9e2158ee/?limit=25#8d79

An optional display of earth's shadow (umbra and penumbra) projected into space that could be activated/deactivated to plan and view lunar eclipses would be desirable.

PR #1479, opened 2021-01-24
New planet magnitudes

### LINKED ISSUE #574: New planet magnitude calculation has appeared
There has appeared a new way of calculating planetary magnitudes. Mercury and Venus seem to be the ones most affected by this new formula.

The paper:
https://www.sciencedirect.com/science/article/pii/S221313371830009X?via%3Dihub

Stellarium could have this as an additional option for the 5 ways of displaying mangitudes given at the moment.

gzotti:
### Description

This branch should add a few planet magnitude improvements:
- A new paper from Mallama&Hilton (2018). 
- A proper magnitude evaluation for the Moon.
- Umbra and penumbra circles.
- Eclipse magnitude in Infostring now in percent
- Earth shadow enlargement using either traditional (2%) or Danjon method.

Fixes #574, #1350, #430.

It does not solve #1115. There is an unknown positional mismatch between numerical and graphical operations, which is a separate problem.

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 20H2
* Graphics Card: irrelevant

We can 
- observe new planet magnitudes. Esp. Mars showing albedo-dependent magnitude changes
- note the first implementation of a more realistic Lunar magnitude
- see Lunar eclipse details: a small increase of the umbra edge seems to give a better match.
- The umbra circle sizes have been derived as geometrical circle drawn at Lunar distance and using the Besselian size formulations from the Explanatory supplement. There is currently a slight visual difference between the numerical InfoString, shadow visualisation (undocumented) and my geometrical circles results which is hard to trace. Maybe we need to solve aberration and possibly other light time effects before this mismatch can be closed.

## Checklist:


gzotti:
@alex-w I think you made the shadowFactor in Planet.cpp lines ~2084: Can you please maybe also work out a factor for penumbra for a Lunar eclipse? Currently the opposition surge adds more light than penumbra takes away, which seems odd.

Atque:
@gzotti I know this is draft and WIP an all, but I just wanted to report a bug:
I get some REALLY weird magnitudes for Jovian moons when viewed from Jupiter. They reach magnitude -70 (!).

gzotti:
OK, thanks. Yes, I dared to apply the Mallama algorithm for observers not on earth, but the Jovian satellites were taken from some other source, and this may have to be adapted. Thanks for mentioning, this needs to be fixed of course.
 
Do you agree with the Lunar behaviour now?

Atque:
@gzotti Yes, the lunar formula seems to agree very well with other sources now (even better, actually, since CdC and other planetaria don't take the distance factor into account). The opposition surge makes the sky incredibly bright, which agrees with my own impression of a full moon vs 90%-ish moon.

Atque:
@gzotti You are probably aware of this, but the new magnitude algorithm doesn't take eclipses into account (Jovian, Saturnian and Martian moons seen from respective parent body).

gzotti:
Indeed there was a case still open for non-terrestrial locations that did not care for shadows. TNX.

alex-w:
@gzotti sorry for late answer. Yes, I've added the code for compute the shadow factor, but this code was ported from old version of Stellarium (0.12). Probably it was added by @xalioth or @guillaumechereau

Atque:
@gzotti I get some weird readings for Jupiter when viewed from Saturn. A 0.2% illuminated Jupiter shines at magnitude -3.7, which is clearly incorrect. Is there some kind of phase factor not taken into account?

gzotti:
> @gzotti sorry for late answer. Yes, I've added the code for compute the shadow factor, but this code was ported from old version of Stellarium (0.12). Probably it was added by @xalioth or @guillaumechereau

No problem, I made my own now with pencil&paper and some geometry.

gzotti:
> 
> 
> @gzotti I get some weird readings for Jupiter when viewed from Saturn. A 0.2% illuminated Jupiter shines at magnitude -3.7, which is clearly incorrect. Is there some kind of phase factor not taken into account?

The paper by Mallama&Hilton provide phase functions which end near 150°. Given its principally higher accuracy and phase functions extending widely beyond what's visible from earth, I allow its use also from other planets. The only alternative I see for now is going back not applying this magnitude model for any location but earth/moon, and revert, for all places beyond earth/moon, to the albedo/diameter/phaseangle model the origin of which I don't know. It is OK to be picky, but please consider this as "science currently has no answer for such case". 

You may also again find a few seconds of mismatch between lunar eclipse factor and the visual contact with the shadow indicator circles. The latter use a slightly different way to work out the shadow geometry. On third reading I can try to apply the usual way to derive shadow diameters in arcseconds and convert back to kilometers for drawing. 

I am a bit unsure about the magnitude curve during Lunar eclipses. May deserve another look, but I must really do other things this week.

Atque:
> The paper by Mallama&Hilton provide phase functions which end near 150°. Given its principally higher accuracy and phase functions extending widely beyond what's visible from earth, I allow its use also from other planets. The only alternative I see for now is going back not applying this magnitude model for any location but earth/moon, and revert, for all places beyond earth/moon, to the albedo/diameter/phaseangle model the origin of which I don't know. It is OK to be picky, but please consider this as "science currently has no answer for such case".

I thought it might be a limitation. This algorithm is much better overall.

> You may also again find a few seconds of mismatch between lunar eclipse factor and the visual contact with the shadow indicator circles. The latter use a slightly different way to work out the shadow geometry. On third reading I can try to apply the usual way to derive shadow diameters in arcseconds and convert back to kilometers for drawing.

Isn't this simply an issue caused by the absence of aberration computation?

gzotti:
>Isn't this simply an issue caused by the absence of aberration computation?

I am not 100% clear about the reason for the apparent asymmetry. What really surprised me is that the numerical computation (which uses apparent positions) and the shadow simulation (which should use the same numbers) appear to disagree by a few seconds. Also the circle should match in a very symmetric way, but does not, by some arcseconds. But indeed, we should implement all things around aberration, and then reconsider these deviations. I think at least for illustrative purposes the current set of improvements is quite good.

alex-w (approved):
Thank you very much for the code!

PR #1547, opened 2021-03-14
core/StelTexture: read image backwards, dont flip

ajeddeloh:
When loading images, instead of flipping the images upside down then converting then, just read them backwards.

### Description
I was profiling Stellarium and found that the image load flips the image upside down then loads it. Instead of flipping it, just read it backwards. There's no perceptible change in loading speed on my system since I suspect it's IO bound anyway, but the new code is simpler.

No dependency changes.

### Type of change

### How Has This Been Tested?
I first removed the code to flip the image and ran stellarium to ensure it resulted in the images being upside down. I then switched the code to load backward and confirmed they were back to being right side up.

Ran all the automated tested as well (e.g. `make test`).

I don't think this change affects other areas of the code.

**Test Configuration**:
* Operating system: Arch Linux (up to date as of today) with wayland
* Graphics Card: Intel, Model (HD Graphics 530), i915

## Checklist:
- [x] (N/A) I have made corresponding changes to the documentation
     **I'm not sure what this should look like or where it should live. I'm happy to add a test, just let me know where it should live/what it should test.**
- [x] (N/A) Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/core/StelTexture.cpp
+++ b/src/core/StelTexture.cpp
@@ -300,19 +300,10 @@ QByteArray StelTexture::convertToGLFormat(const QImage& image, GLint *format, GL
 	ret.reserve(width * height * bpp);
 	QImage tmp = image.convertToFormat(QImage::Format_ARGB32);
 
-	// flips bits over y
-	int ipl = tmp.bytesPerLine() / 4;
-	for (int y = 0; y < height / 2; ++y)
-	{
-		int *a = reinterpret_cast<int *>(tmp.scanLine(y));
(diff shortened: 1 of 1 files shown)


ajeddeloh:
> - [ ]  Are possibly unused includes removed?

- [x] Yes

Looks like qSwap is part of QtAlgorithms which isn't included anyway.

alex-w (approved):
It’s ok for me. @gzotti?

gzotti (approved):
Fine for me, thanks!

PR #1550, opened 2021-03-16
Fix compatibility of ToneReproducer shader with GLES

10110111:
GLSL ES requires that each shader contains main() function. This makes
it impossible to have multiple shaders of the same type (e.g. vertex) in
a GLSL ES program.

Since Stellarium needs compatibility with ANGLE, this patch works around
this limitation of GLSL ES by concatenating the two shader sources
before compiling.

NOTE: I haven't actually tested this on Windows/ANGLE. I've only made sure that it runs without problems on my Linux-based system.

### DIFF
--- a/src/core/modules/Atmosphere.cpp
+++ b/src/core/modules/Atmosphere.cpp
@@ -27,6 +27,7 @@
 #include "StelFileMgr.hpp"
 #include "Dithering.hpp"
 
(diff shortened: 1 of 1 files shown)


alex-w (approved):
Tested on Windows 7 in all modes (OpenGL, ANGLE, MESA) - all works good

PR #1573, opened 2021-03-28
ArchaeoLines additions

### LINKED ISSUE #1555: Arctic and antarctic lines
In older starmaps and star globes, there is often markings for arctic and antarctic circles in the sky. They correspond to the zenith passage of said circles on Earth.

It might be useful to add them to the archaeolines plugin.

gzotti:
### Description

A few niceties in the ArchaeoLines plugin. 

* Add polar circles
* Add two custom altitude lines
* Use AngleSpinBoxes for the angles of the custom lines
* Allow selection of custom locations from Stellarium's location list
* Allow picking azimuth/altitude/declination from the currently selected object
* Set plugin version number equal to the main program version (static plugin, so I see no use for a different number)

Fixes #1555
Supersedes #1447

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 64bit 20H2
* Graphics Card: Nvidia Geforce 960M (irrelevant)

## Checklist:


axd1967:
the plugin version number is an interesting issue.

it could indeed be set to the main program version number, but in that case it has become obsolete (remember DRY) and should be removed, it has become unneeded, noise, it confers a false sense of separate versioning.

IF a user can install a plugin separately , THEN a separate plugin versioning scheme IS needed.

anyway, it is not trivial for an average user to install a plugin separately. 

a true versioning would also call for a distinct master branch per plugin, so that true parallel development is possible. in reality, the plugin code is implicitly subject to the versioning scheme of the main application.

(so IMHO the plugin versioning should be removed _globally (for all plugins)_ and therefore as a _separate_ issue.)

axd1967:
as far as I could see, the plugin depends on a equatorial coordinate system; while I fiddled with #1573 , I thought of adding a _line of constant altitude_, as a simple way to define a quick and crude horizon (typically 10-15 degrees), but this might require a bit more plumbing.

alex-w:
> the plugin version number is an interesting issue.

This is NOT an issue.

> it could indeed be set to the main program version number, but in that case it has become obsolete (remember DRY) and should be removed, it has become unneeded, noise, it confers a false sense of separate versioning.

This is may be very strange for you, but you MAY build plugins separately as dynamic (for linux at least).

> IF a user can install a plugin separately , THEN a separate plugin versioning scheme IS needed.

User CAN, but you want to DELETE versioning, because by default we do not publish all plugins as separately libraries.

> anyway, it is not trivial for an average user to install a plugin separately.

Copy & paste file - is it very difficult?

> a true versioning would also call for a distinct master branch per plugin, so that true parallel development is possible. in reality, the plugin code is implicitly subject to the versioning scheme of the main application.

Why? Why we should introduce over difficult solution, which will require in 5 times more resources for development, distribution and support?

> (so IMHO the plugin versioning should be removed _globally (for all plugins)_ and therefore as a _separate_ issue.)

Seriously?

gzotti:
1-2 lines of constant altitude are - as you prefer to express - a separate issue which could however still be addressed in this branch. Any horizon should however be modeled as "landscape". 

The plugin versioning may have been more important when dynamical plugins were developed. These ceased to work on Windows with Qt5, or nobody has ever cared to solve whatever problem there is. Of course, some plugins with their own config files may use that to add and read versioned config files, and the version number remains on the previous release number during releases. I agree, else I also see no need for them, but this is not the place to discuss.

gzotti:
Now I wonder: ARE there separate repos out there with separate dynamic plugins for Linux, does somebody release open- (or even closed-) source plugins for Linux on a regular base, or is this a purely hypothetical discussion?

alex-w:
> Now I wonder: ARE there separate repos out there with separate dynamic plugins for Linux, does somebody release open- (or even closed-) source plugins for Linux on a regular base, or is this a purely hypothetical discussion?

Right now I cannot share examples for series 0.20.x, but I saw dynamic plugins for Stellarium (linux + windows) in github. Just for protocol: we can restore support dynamic plugins on Windows if we switch to building Stellarium through MinGW GCC. Dynamic plugins for MacOS always was very painful and probably 3rdparty dynamic plugins for MacOS was not exist.

Switching to static plugins was by 2 reasons: building plugins for all platforms and distribution plugins to end users. The nightmare for support those plugins followed from second item.

alex-w (approved):
It's OK for me, thanks!

axd1967:
@gzotti thanks for the two custom altitude lines! it can also help in selecting good navigation stars (@A-j-K )

gzotti:
Sure. I never needed them in practice, but they don't hurt either. But I hope nobody needs to activate all lines (plus those from the main program) at once...

Issue #1272, opened 2020-09-17
Aberration of light


alex-w:
A small positional inaccuracy is caused by not properly dealing with aberration of light. This definitely ruins accuracy with the beginnings of the seasons and may also cause issues with lunar occultations. Correction will require warping transformation matrices similar to what has been done for refraction. For demonstrations of the effect it should be possible to exaggerate it.

gzotti:
Note that a recent typofix in the nutation matrix (submitted by an attentive user!) fixed the season times. However, I don't know ATM what causes the tens of seconds error for lunar occultations. There is a small figure mismatch for the Moon which we also don't model. In any case, Aberration correction is IMO one of the 2 stopping requirements for 1.0 (the other being planet axes).

PR #1626, opened 2021-04-29
Annual aberration

### LINKED ISSUE #84: lunar positions mismatch

gzotti:
### Description

One of the last long-standing astronomical issues is the non-computation of annual aberration of stellar and planetary positions. 
This is a periodic shift of positions by about 20 arcseconds (if on earth), not much, but annoying enough to cause errors
for stellar occultations by the moon of about 1 minute. 

Preconditions to this improvement have been made much earlier, esp. making the velocity vector accessible. For the Moon (ELP82B) the velocity vector is now created by calling lunar position twice with 10s time difference and taking the difference. 

This branch introduces the annual aberration only. Diurnal effects would have to process the topographic motion vector of the observer location for ultimate, subarcsecond corrections. These are currently not included. 

* aberration is switchable (added property in StelCore, checkbox in configuration dialog)
* it is possible to exaggerate it (0-5x) for demonstrations. Higher exaggeration unfortunately causes graphics/clipping problems. (You are cordially invited to solve them!)

Of course, all classes of celestial objects must be processed now. The change is not large for simple objects, though. For DSO photos, it was much trickier. Planets needed changes in several parts of the code.

### Objects solved
- [X] Stars
- [X] DSO icons
- [x] DSO icon orientation
- [x] DSO images
- [x] DSO outlines
- [X] Milky Way
- [x] Planets
- [x] Comet tails
- [x] Moon
- [x] Sun
- [x] Markers
- [x] SkyImages
- [x] Pulsars
- [x] Quasars
- [x] Exoplanets
- [x] Novae
- [x] Supernovae
- [x] Constellation artwork
- [x] Constellation Labels
- [x] Constellation boundaries
- [x] Asterism Labels
- [x] TOAST
- [x] HiPS
- [x] Object selection (inverse-aberr click position into the sky...)
- [ ] (please complete this list with missing objects!)

### Not required:
- Satellites
- Meteor showers
- Zodiacal Light (glued to the sun)
- NavStars (auto-change)

Fixes #84, #1272, #1115

See image linked at the original report of #84. With this fix, the photo is matched nicely. 
Critical test of course involve solar and lunar eclipses and occultations. The original "preliminary" fix for Solar eclipses may have to be adapted now, also earth shadow for lunar eclipses (#1115) is related and fixed. 

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 20H2/21H1 (irrelevant)
* Graphics Card: NVidia Geforce 960M (irrelevant)

Comparison against JPL HORIZONS is showing final deviations of typically 0.01s/0.1" only, in the present era. This is where I personally say "really good enough for my purposes". Historical values differ slightly also if compared in DT, probably from some different models for earth orientation.

Visually tested several Lunar eclipses against predictions in almanachs. 

## Checklist:


alex-w:
> shall aberration be switchable (add property in StelCore, checkbox in configuration dialog)

Yes, for didactic purposes.

> shall it be possible to exaggerate it (e.g. 0-1000x?) for demonstrations?

This is good question! I guess the answer is “yes” and probably similar behavior should be added for parallax for didactic purposes.

gzotti:
Hmm, it is quite easy to exaggerate aberration by multiplying the velocity vector, will do that. However, parallax results from pure geometrical projection which comes for free in OpenGL. We would have to mess around with - e.g. - scaling the astronomical unit. I cannot say in how many places some scaling operation implicitly assumes AU and will have to be adapted. I leave this to you :-)

alex-w:
Shall the coefficient for exaggeration of aberration have one value (1000x for example) and toggler for enabling/disabling it?

gzotti:
No, I am just working on it (almost done): Toggle for principal enabling, and double spin for setting 0...1000. Yes, I know, somebody else will want to exaggerate a million times... :-)

Atque:
This is awesome. Will planets be included? I can't see them on the list of objects affected, but surely, they are affected in some way (at least Stellarium's planetary positions are off a few seconds of arc compared to JPL Horizons).

gzotti:
I think I must add aberration for planet as well. However, the change in stars alone fixed the occultation issue, so what about the lunar position? Does the Lunar ephemeris include aberration already? There will also likely be changes around the preliminary fixes which I did for Solar eclipses before the American eclipse (2017). 

@Atque You could help us building a reference dataset: Can you please prepare a list of geocentric planet positions from JPL, including aberration and nutation, maybe 4 positions per planet, several months apart. Also positions of the Sun and the Moon. 

Regarding earth's shadow: I wonder if the solar position is assumed to be at 0/0/0 for the shadow algorithms. If the shadow is also "aberred" this would explain the asymmetry you reported.

Atque:
@gzotti You mean like this?

[JPL_ephemeris.txt](https://github.com/Stellarium/stellarium/files/6401512/JPL_ephemeris.txt)

axd1967:
Allow me to suggest to group the settings of `simulate light speed` (a View option) and `include aberration` (a Tools option) closer together, as they are related to intricate light effects.

gzotti:
They are indeed connected, but light time correction is an issue supported for solar system objects, while aberration and nutation are fine coordinate corrections applicable to all objects. I am likely not going to deal with relativistic effects though. As the program evolves, settings may occasionally move. I can indeed only think of rare cases to switch off light time correction. 

Question: Are shadows of moving objects curved?

Atque:
@gzotti What about "permanent" markings in the sky, such as the galactic center/anticenter, permanent grids etc? Should they be affected by aberration?

gzotti:
Hmm... The grids and poles are not affected. I think galactic coordinates are also only defined as rigid rotation against the J2000 grid, so the galactic center marker (defined in terms of galactic coordinates) would also not undergo aberration, although the visible objects (milky way texture) of course would. (And does already.)

gzotti:
Bah, now planet moons are doubly-aberred. Seems I need another member variable...

gzotti:
@Atque can you please repeat the ephemeris lookup for the 4 Galilean satellites of Jupiter, for the same dates? I am pretty lousy at JS scripting, but will attempt to make a script that reproduces the list.

Atque:
@gzotti Sure, here you go.
[ephemerides_jovian_moons.txt](https://github.com/Stellarium/stellarium/files/6759653/ephemerides_jovian_moons.txt)

gzotti:
Thanks. We are coming really close now!

Sun
2021-01-01T00:00:00 : RA: 18h46m53.605s  DE: -22°59'57.223''
2021-06-02T00:00:00 : RA: 4h40m40.920s  DE: +22°11'9.109''
2021-11-01T00:00:00 : RA: 14h25m34.795s  DE: -14°25'30.993''
2022-04-02T00:00:00 : RA: 0h45m0.083s  DE: +4°50'3.816''
Mercury
2021-01-01T00:00:00 : RA: 19h17m59.443s  DE: -24°20'58.113''
2021-06-02T00:00:00 : RA: 5h35m35.590s  DE: +22°40'50.631''
2021-11-01T00:00:00 : RA: 13h25m55.568s  DE: -6°47'18.898''
2022-04-02T00:00:00 : RA: 0h43m5.375s  DE: +3°21'44.904''
Venus
2021-01-01T00:00:00 : RA: 17h18m28.447s  DE: -22°26'1.562''
2021-06-02T00:00:00 : RA: 5h57m1.469s  DE: +24°21'39.220''
2021-11-01T00:00:00 : RA: 17h40m41.747s  DE: -27°5'5.308''
2022-04-02T00:00:00 : RA: 21h52m25.372s  DE: -12°4'41.434''
Moon
2021-01-01T00:00:00 : RA: 8h23m36.792s  DE: +23°1'20.880''
2021-06-02T00:00:00 : RA: 22h46m54.139s  DE: -13°27'21.167''
2021-11-01T00:00:00 : RA: 11h18m43.708s  DE: +9°57'41.960''
2022-04-02T00:00:00 : RA: 1h21m51.333s  DE: +5°40'1.584''
Mars
2021-01-01T00:00:00 : RA: 1h40m18.146s  DE: +11°20'41.181''
2021-06-02T00:00:00 : RA: 7h45m3.736s  DE: +22°37'21.006''
2021-11-01T00:00:00 : RA: 13h55m55.309s  DE: -11°22'15.197''
2022-04-02T00:00:00 : RA: 21h31m33.540s  DE: -15°56'58.436''
Jupiter
2021-01-01T00:00:00 : RA: 20h20m44.626s  DE: -20°0'44.334''
2021-06-02T00:00:00 : RA: 22h15m51.654s  DE: -11°42'45.279''
2021-11-01T00:00:00 : RA: 21h41m29.273s  DE: -15°1'8.906''
2022-04-02T00:00:00 : RA: 23h30m29.240s  DE: -4°17'8.182''
Saturn
2021-01-01T00:00:00 : RA: 20h15m50.412s  DE: -20°10'28.693''
2021-06-02T00:00:00 : RA: 21h4m25.239s  DE: -17°24'46.638''
2021-11-01T00:00:00 : RA: 20h39m27.716s  DE: -19°14'59.678''
2022-04-02T00:00:00 : RA: 21h39m14.502s  DE: -15°1'42.175''
Uranus
2021-01-01T00:00:00 : RA: 2h18m27.826s  DE: +13°21'27.510''
2021-06-02T00:00:00 : RA: 2h40m37.837s  DE: +15°11'44.705''
2021-11-01T00:00:00 : RA: 2h42m31.601s  DE: +15°19'13.498''
2022-04-02T00:00:00 : RA: 2h42m25.315s  DE: +15°22'19.689''
Neptune
2021-01-01T00:00:00 : RA: 23h19m18.387s  DE: -5°33'12.401''
2021-06-02T00:00:00 : RA: 23h36m13.265s  DE: -3°46'42.921''
2021-11-01T00:00:00 : RA: 23h27m30.354s  DE: -4°46'19.713''
2022-04-02T00:00:00 : RA: 23h38m17.593s  DE: -3°34'18.997''
Io
2021-01-01T00:00:00 : RA: 20h20m41.433s  DE: -20°0'57.712''
2021-06-02T00:00:00 : RA: 22h15m44.013s  DE: -11°43'33.233''
2021-11-01T00:00:00 : RA: 21h41m21.653s  DE: -15°1'52.258''
2022-04-02T00:00:00 : RA: 23h30m27.933s  DE: -4°17'20.348''
Europa
2021-01-01T00:00:00 : RA: 20h20m34.166s  DE: -20°1'24.890''
2021-06-02T00:00:00 : RA: 22h15m55.574s  DE: -11°42'23.493''
2021-11-01T00:00:00 : RA: 21h41m41.686s  DE: -14°59'58.180''
2022-04-02T00:00:00 : RA: 23h30m29.113s  DE: -4°17'3.228''
Ganymede
2021-01-01T00:00:00 : RA: 20h20m33.157s  DE: -20°1'30.853''
2021-06-02T00:00:00 : RA: 22h15m57.194s  DE: -11°42'7.480''
2021-11-01T00:00:00 : RA: 21h41m47.820s  DE: -14°59'25.005''
2022-04-02T00:00:00 : RA: 23h30m31.231s  DE: -4°17'0.950''
Callisto
2021-01-01T00:00:00 : RA: 20h20m42.731s  DE: -20°0'56.314''
2021-06-02T00:00:00 : RA: 22h15m55.413s  DE: -11°42'15.082''
2021-11-01T00:00:00 : RA: 21h41m58.247s  DE: -14°58'23.930''
2022-04-02T00:00:00 : RA: 23h30m53.158s  DE: -4°14'14.969''

Not bad, eh? :-) 
However, click-selecting planets currently does not work :-( Searching...

@Atque, for historical application, can you please repeat such lists for e.g. years 1600, 500, -500 and -2500? (Or if this is some configurable URL, maybe you can provide this? :-)

Atque:
@gzotti This is very, very nice. Great work!

Here's the list. Sadly, JPL Horizons can't provide ephemerides for the outer planets for the given dates. For Jupiter and its moons + Uranus, I set it to 1601-01-01 to 1602-08-31, and for Saturn I had to put 1750-01-01 to 1751-08-31. For Neptune, there is no ephemeris available earlier than 1801-01-01.

Also, do note that JPL Horizons uses the BC/AD system. This means I have input 2500 BC and not -2500, and so on.

[ephemerides_historical.txt](https://github.com/Stellarium/stellarium/files/6759693/ephemerides_historical.txt)

Atque:
I noted that neither Earth nor Venus seem to show aberration occasionally when seen from Mercury or other bodies very close to the Sun (such as C/Kirch 1680 or C/2011 W3 Lovejoy).

gzotti:
@Atque many thanks! I was not aware HORIZONS was so "limited" (oh well, they don't have to control spacecraft in those times... :-)

I would like to provide this reference list in Appendix F, to answer a few related questions which appear several times a year. For this I'd need to know the settings you called for in HORIZONS. These are geocentric apparent positions for the equinox of date, including precession, nutation, aberration, right? Anything else that could be configured there? Does HORIZON compute light deflection by Solar gravity or some relativistic stuff that we never cared for so far? I don't know yet where the 0.1 or 0.2" mismatch comes from, but TBH if this is a random error which does not get larger with time, I don't care too much. If some bias rotation is missing, maybe we can still find it. But there are also other places in the project which have cried for attention for quite a while now...

gzotti:
Uh, Earth indeed shows no aberration when seen from Mercury. Weird. :-(

Atque:
@gzotti I was also surprised, especially since the DE431 extends from-13200 to +17020.

I use this interface, if you want to play around with the settings yourself:
https://ssd.jpl.nasa.gov/horizons.cgi

It assumes a geocentric view, with equinox of date, with nutation, aberration and other corrections applied.

Atque:
> Uh, Earth indeed shows no aberration when seen from Mercury. Weird. :-(

Only occasionally. Try looking at Earth from Mercury with 100x aberration, on 2021-10-10, and scroll through the dates.

gzotti:
Uh, yes, I remember there was some caching available for positions. Maybe, when the same time is called from a new location, some old data are retrieved. Brrr...

Atque:
@gzotti If it is to any help in investigating, it seems to happen when Venus and Earth are close to the ASP as seen from Mercury.

gzotti:
Sounds weird. thanks. Needed a break. I am now just writing my table comparison. Will continue hopefully next week.

gzotti:
@Atque the ASP was in charge only indirectly, as being the opposition point. When time difference was small enough, the third round of computation was skipped, which only then actually would have brought in aberration correction. Now I force aberration already in rounds 2 and 3. Maybe I will allow hiding round 3 behind an extra flag, though, as it does indeed seem to cost some computation time...

Atque:
@gzotti Great to see that you found a solution. I noticed that aberration would be needed for asteroids and comets too (these are not mentioned in the PR).

gzotti:
Asteroids and comets are subclasses of planets, so they are included with the planets.

gzotti:
@Atque I have added the comparison table to the User Guide. To exclude systematic errors unrelated to DeltaT, it may however be helpful if you can repeat your table creation with TT dates and compare those with Stellarium's result for "no DeltaT correction". I have read enough of the docs there to see we are using different models for precession and nutation, of course this will also result in slight differences, but hopefully not in diverging drifts. It's a pity the IAU2000B nutation is only "secure" for mas accuracy in 1995-2050 compared to the full IAU2000A, with no information given for its behaviour outside. What happens if I extend our using it from +-500 years to e.g. +-5000 years?

Atque:
@gzotti I’ll have it ready hopefully by tomorrow.

Atque:
@gzotti It appears the interface does not allow changing or disabling Delta-T. Would it be sufficient if I create tables with Delta-T values given?

gzotti:
The interface reads: 

> Times may be specified as calendar dates and optionally times (e.g. "YYYY{BC|AD}-MMM-DD {hh:mm} {UT|TT})", 

so presumably just adding TT should do the trick? If that really does not work, yes, adding a particular DeltaT column is also OK. (Maybe if you manage to get TT, still output DeltaT please!) 

TNX, G.

Atque:
> so presumably just adding TT should do the trick? If that really does not work, yes, adding a particular DeltaT column is also OK. (Maybe if you manage to get TT, still output DeltaT please!)

That appears to work. I'll have it ready in an hour or so.

BTW: I noticed your latest commit in planetary aberration causes the Moon to be aberrated just as much as stars, meaning occultation accuracy is lost.

gzotti:
Hm? That eclipse occultation should still work, or not? I am just struggling with getting rid of the Solar eclipse hack, which should now be solved, but isn't. I had some other things to do now, though.

Atque:
> Hm? That eclipse occultation should still work, or not? I am just struggling with getting rid of the Solar eclipse hack, which should now be solved, but isn't. I had some other things to do now, though.

I haven't tested with eclipses, but the Aldebaran occultation of 6th November 2017 is about 40 seconds off (just like in the master-branch). It was only one or two seconds from CdC's time before the latest commit on planetary aberration.

Here's the ephemeris file using TT, and with Delta-T's given:
[ephemerides_historical_TT.txt](https://github.com/Stellarium/stellarium/files/6832763/ephemerides_historical_TT.txt)

gzotti:
@Atque Still trying to understand why the Moon is better not affected by refraction computations/corrections. Any ideas?

Atque:
@gzotti I can "see" visually what the problem is: The Moon has double aberration it seems. Its own, correct aberration (very small because of the relatively slow orbital velocity), AND the aberration from the Earth orbiting the Sun (i.e. "stellar" aberration). This causes the Moon to be placed incorrectly in relation to stars, giving bad occultation times. It is very noticeable when toggling aberration when the Moon is in frame. Zooming in on stars very close to the lunar limb, you can see that the Moon indeed has a very small aberration in relation to the stars too, which I would assume is the correct value.

You see similar phaenomena on other planets. Jupiter vs Io, Saturn vs Mimas etc.

I can't follow the code to see what the problem is though, but in case you hadn't noticed this, it might help in investigating.

gzotti:
Uhm, I have made some changes, but not yet on Github. I wanted to push when this seemingly trivial issue would have been fixed. There should now be no way to have double aberration.

Atque:
@gzotti Well, the Moon is clearly affected both by aberration from the Earth orbiting the Sun, and from its own relative veolocity around the Earth. It should only be the latter, if I'm not mistaken.

Before you fixed the rounding errors (which caused aberration to be ignored when viewing from Mercury at some points), this problem did not appear, and occultation times were more or less perfect.

gzotti:
Presumably fixing that error enabled the aberrationPush for the Moon, just like for the other planets where the error was only visible occasionally. 
The following idea is not visible in the relevant literature: For "own" moons (i.e. Moon for Earth, Jovian Moons if observing from Jupiter etc) we may only take the relative speed between observer planet and that moon? Likewise the same for observing planets from moons. But I fail to see the physical explanation. The speed of the observed object should not matter!

Atque:
> Presumably fixing that error enabled the aberrationPush for the Moon, just like for the other planets where the error was only visible occasionally.
> The following idea is not visible in the relevant literature: For "own" moons (i.e. Moon for Earth, Jovian Moons if observing from Jupiter etc) we may only take the relative speed between observer planet and that moon? Likewise the same for observing planets from moons. But I fail to see the physical explanation. The speed of the observed object should not matter!

I'm not a mathematician, nor am I an astronomer or physicist, but wouldn't a terrestrial observation point ignore the stellar (annular) aberration of all objects orbiting the Earth? The Moon and the Earth orbit the Sun at similar speed, and the Moon orbits the Earth at about 1 km/s, which is very slow compared to the 30 km/s of the orbit of the Earth around the Sun. The same goes for other planets and their moons.

gzotti:
Yes, this idea is inviting, but really contradicts the literature. Let's see how that works out.

Atque:
@gzotti Perhaps this paper can help:
https://ntrs.nasa.gov/api/citations/19700008007/downloads/19700008007.pdf

Scroll down to "Planetary aberration".

gzotti:
I had also found this myself. Yes, you see my problem? One of many similar descriptions.
I assume there is a tiny principal error in my attempts from working with heliocentric, not barycentric velocities. But this should be mostly irrelevant when we are dealing with ~0.25 (?) arcsec accuracies.

Atque:
It seems only the Moon is being affected. I get good equinox/solstice times with your latest commit, and I see there is aberration for the Sun included. Planets are fine, Sun is fine, asteroids and comets are fine (although tails are not aberrated), but the Moon is weird. I compare to Cartes du Ciel, because it gets the occultation times perfectly (and it supports DE440/441).

The difference is about 20 seconds of arc (when comparing the 2017-11-06 occultation of Aldebaran), which funnily enough is the same as if aberration were not to be accounted for.

gzotti:
I had derived my own hacked solution for the sun in 2017, but want to get rid of it now, in favour of documented models. But the new issue with the moon shot me off the track...
Somebody else told me he had the same problem with the moon, but he never went into other planet moons seen from those planets. I may as well just exclude the Moon, assuming the aberration is already built into the ephemerides.
The current state again solves #84, and plays another occultation from an older almanach (2019) exactly within its given accuracy (0.1min). 

Another occultation computed with Occult in September is played 3s late (1arcsecond). Hmm, the end is 4s early. Not sure what lunar diameter they used? Patch to allow our lunar scaling with 4 decimals. Scale moon by 1.0015 yields 1s accuracy. 

I guess I will just exclude the Moon, and apply known models for the rest.

Atque:
> 
> 
> I had derived my own hacked solution for the sun in 2017, but want to get rid of it now, in favour of documented models. But the new issue with the moon shot me off the track...
> Somebody else told me he had the same problem with the moon, but he never went into other planet moons seen from those planets. I may as well just exclude the Moon, assuming the aberration is already built into the ephemerides.
> The current state again solves #84, and plays another occultation from an older almanach (2019) exactly within its given accuracy (0.1min).
> 
> Another occultation computed with Occult in September is played 3s late (1arcsecond). Hmm, the end is 4s early. Not sure what lunar diameter they used? Patch to allow our lunar scaling with 4 decimals. Scale moon by 1.0015 yields 1s accuracy.
> 
> I guess I will just exclude the Moon, and apply known models for the rest.

It would be surprised if it's included in the ephemeris, because it does not differ that much from VSOP87. This is some other problem, maybe similar to the one with the Sun that was solved in 2020 (sign error, I believe).

Could it be possible to only disable stellar aberration for the Moon when seen from Earth? Otherwise, the Moon will show no aberration when viewed from other solar system bodies.

gzotti:
There was a sign error in the source paper I had used for nutation which somebody identified last year. 
Yes, "disabling nutation for the moon seen from earth" is possible and may be the way to go for now. And future astronauts will have to report accurate occultation timings for the rest.

Atque:
@gzotti Okay, upon investigating, I compared SkySafari pro 6 to Stellarium. (SkySafari is the only software that I'm aware of that allows disabling only aberration. CdC only allows disabling both aberration and nutation). It seems that occultation times are still rather accurate without aberration, which suggests that aberration might not be the solution to inaccurate lunar positions. I think you should keep stellar aberration for the Moon, because the problem does not seem to lie with aberration at all.

gzotti:
If "don't apply extra correction" solves the Lunar positional error, I will apply this. There will just be no "Lunar position without aberration". Most people are interested in the apparent positions (which they can observe). If somebody then comes out with the real explanation, we will all be still happier :-)

Atque:
The real solution will probably also solve #1115.

gzotti:
#1115 is even trickier. I currently cannot find where the shadow axis is defined! (Light and shadow maps are very different beasts here.)

Atque:
I tried your latest commit, and interestingly enough, the coverage percentage fits more or less perfectly when aberration is off.

gzotti:
I have not touched the numbers yet, but this should be easy to fix (famous words :-). There are many bits and pieces of code to adapt, and the visible earth shadow code is what will cost some more time to comprehend.

gzotti:
Forget the shadows for now. Something else happened, and the numbers are off again from the list of July 4th :-(. Aaargh!

Atque:
> Forget the shadows for now. Something else happened, and the numbers are off again from the list of July 4th :-(. Aaargh!

I get pretty accurate results (within 0.1-0.2 seconds of arc for recent dates). Or are you trying a newer commit?

gzotti:
Not sure what or why I am seeing the nonsense, but I will investigate today. Just that the result from my scripted comparison to your JPL lists is currently much worse than in July.

gzotti:
Alright, false alarm. My scripting capabilities... Joyfully just a stupid error between setting time as UT vs. DT...

gzotti:
Alright, the shadows and shadow circles look good now (#1115). I have not checked whether anything is to be done in the numerical shadow sizes yet.

Atque:
@gzotti The shadow seems fine, but the brightness boost is off a bit.

gzotti:
Yes, that, and the magnitude drop is still unbalanced.

Atque:
@gzotti I just noticed that disabling/enabling lighttime correction does not seem to change planetary positions, only their rotation. Is this intentional?

gzotti:
@Atque Switching on aberration implies light time correction. At least this is what I had in mind. Apparently rotation can still be computed without, thanks for noticing. Now, is that useful or a bug?

Atque:
They are two separate phaenomena, so I think keeping them separate, if possible, would be better.

gzotti:
They are separate, but if you see the literature, there is always a chain of corrections. "First, get light time right for the major iprovement, and then fix aberration as fine-tuning step". Do you think it is necessary to apply aberration to a non-light-time-corrected position? Or disable aberration when light-time correction is off?

I agree, under normal circumstances, light-time correction should never be switched off. Switching off may however be useful to e.g. show a Lunar eclipse from the SolarSystemObserver when this is very far (light hours) away. Else the time is distractingly wrong. However, I have not yet checked how this appears now. :-P

gzotti:
@Atque The problem with separating: aberration depends on the planet's distance, which we find iteratively. 
We could still switch off light time correction, and at the same time disable aberration for the planets. @alex-w ?

Atque:
> We could still switch off light time correction, and at the same time disable aberration for the planets.

That might be the best solution.

alex-w (changes requested):
Please use \newFeature{v0.21.2} in User Guide for consistency

gzotti:
The best way to enforce a format in LaTeX is by formatting the macro, not the input... Now all is consistent.

alex-w (approved):
Thank you very much!

Shall we change the current version of the planetarium?

gzotti:
Well... about 6 years ago I said "only two things are missing: planet axes and aberration". But now we see the Qt6 pressure, so I think we should wait with "1.*" for Qt6. I.e. next big task: Move to Qt5.15.2, get rid of QtScript and other things missing in Qt6, upgrade to Qt6. Then we drop the zero.

I will do the merge in about 2 hours. Shall I squash to one big commit or add all commits piecewise?

alex-w:
> Well... about 6 years ago I said "only two things are missing: planet axes and aberration". But now we see the Qt6 pressure, so I think we should wait with "1.*" for Qt6. I.e. next big task: Move to Qt5.15.2, get rid of QtScript and other things missing in Qt6, upgrade to Qt6. Then we drop the zero.

OK, good plan, but please update links to milestones in PR/reports also

> I will do the merge in about 2 hours. Shall I squash to one big commit or add all commits piecewise?

Well, I prefer save history of changes to simplificate investigation of possible problems in future.

gzotti:
I think the milestone can stay at 1.0, as this was a precondition. But we can achieve earlier... (?)

PR #1521, opened 2021-02-21
Oculars plugin: Introduce a 5-value pluginMode property

gzotti:
### Description

The mode switching and zooming code in the Oculars plugin is hard to follow, with a series of intermingled side effects.
This work here should consolidate the state switching. However, it seems the menu buttons in the top-right GUI panel still need boolean properties. I have moved those as "slave properties" with private getters/setters.

- better understandable mode switching with one centralized method.
- no more entangled side effects
- cleanup not finished yet. For now I have commented away the previous functions for easier comparison. 
- Any missing switching, storage and re-setting of main program settings should be done in the setPluginMode(). 

The other, even larger restructuring is the introduction of a dedicated Finder mode which replaces the hybrid isBinocular() from Ocular mode.  This adds a button to the menu bar. However, Binoculars and Finders are collected in the same class "Finder" given that their properties are specified in the same way. 

Fixes #1509, #356

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10
* Graphics Card: irrelevant

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1633, opened 2021-04-30
Oculars: begin code cleanup

treaves:
### Description
This is the start of code cleanup.  But as there is so much to do, it's going to take too long to do it at once.  This needs to be merged into master before more work can cinque; I just lost a few days of work due to a bad merge from upstream.


alex-w:
Please check CI/Linux and CI/AppVeyor reports.

gzotti:
Interesting new syntax, I have never seen the "auto XY() --> type" . Are there real benefits? Cannot say. Makes reading harder for me. 

Do I understand I shall definitely no longer bother with #1521. Fine for me, as long as our intended functionality does not suffer: 

* separate view settings (store+reconfigure a few visibility settings on activation of Ocular or Sensor) 
* optional auto-estimate of telescopic visibility conditions
* store current state of those and restore the cached main view settings on exit to main view.

treaves:
> Please check CI/Linux and CI/AppVeyor reports.

Yes, those build setups are broken. Not the code.

treaves:
> Interesting new syntax, I have never seen the "auto XY() --> type" . Are there real benefits? Cannot say. Makes reading harder for me. 
> 
> 
> 
> Do I understand I shall definitely no longer bother with #1521. Fine for me, as long as our intended functionality does not suffer: 
> 
> 
> 
> * separate view settings (store+reconfigure a few visibility settings on activation of Ocular or Sensor) 
> 
> * optional auto-estimate of telescopic visibility conditions
> 
> * store current state of those and restore the cached main view settings on exit to main view.

There are benefits to the trailing returns. 

Yes, I intend to do that functionality and more.

axd1967:
> Interesting new syntax, I have never seen the "auto XY() --> type" . Are there real benefits? Cannot say. Makes reading harder for me.

https://stackoverflow.com/a/11215414/605463

alex-w:
> > Please check CI/Linux and CI/AppVeyor reports.
> 
> Yes, those build setups are broken. Not the code.

Syntax errors in the code are not code problem. Nice!

alex-w (changes requested):
Please split Q_DISABLE_COPY_MOVE(Class) on 2 macro - Q_DISABLE_COPY(Class) and Q_DISABLE_MOVE(Class)

treaves:
> Do you really need to use macro Q_DISABLE_COPY_MOVE(Class)?

I'll looks at the two separate macros tomorrow. But, yes, they need to be used an any subclass of QObject. And to your other comment, the two build failures are not do to issues in my code; you don't use a semicolon at the end of a macro. Is that why you suggest using the two macros?

alex-w:
> I'll looks at the two separate macros tomorrow. But, yes, they need to be used an any subclass of QObject. 

For what? Which reasons for use this macros? 

> And to your other comment, the two build failures are not do to issues in my code; you don't use a semicolon at the end of a macro. Is that why you suggest using the two macros?

Just forget about 2 macros. You use macros, which was introduced in Qt 5.13 and now 2/3 of CI instances and all instances for building binary packages are broken.

treaves:
> > I'll looks at the two separate macros tomorrow. But, yes, they need to be used an any subclass of QObject.
> 
> For what? Which reasons for use this macros?
> 
> > And to your other comment, the two build failures are not do to issues in my code; you don't use a semicolon at the end of a macro. Is that why you suggest using the two macros?
> 
> Just forget about 2 macros. You use macros, which was introduced in Qt 5.13 and now 2/3 of CI instances and all instances for building binary packages are broken.

Ah.  I did not see they were introduced so recently.  They are needed because of [the rule of three/five/zero](https://en.cppreference.com/w/cpp/language/rule_of_three).  I'll replace them for the time being.

gzotti:
You can test for Qt versions and include/define what's possible. At some point we will have to step up to 5.15 and replace QtScript usage, before we can prepare further moving on to Qt6.

treaves:
> You can test for Qt versions and include/define what's possible. At some point we will have to step up to 5.15 and replace QtScript usage, before we can prepare further moving on to Qt6.

Ya, that's what I pushed earlier.

treaves:
> Do you really need to use macro Q_DISABLE_COPY_MOVE(Class)?

Did you see I put this into a guard?

alex-w (approved):
Well, it works.

alex-w:
All changes extract to oculars-ng branch

PR #1677, opened 2021-05-18
treaves 2

### LINKED ISSUE #1673: Crash when any device added into Oculars plugin
### Expected Behaviour
Adding the devices (eyepieces, lenses, telescopes, CCDs) into list of devices

### Actual Behaviour
When user press button "Add" in the Oculars plugin the planetarium is crashes.

### Steps to reproduce
1. Open Oculars plugin config dialog
2. Open any tab for devices (eyepice, lens, CCD, telescope)
3. Press "Add" button

### System
* Stellarium version: 0.21.0-b0c26ce
* Operating system: Ubuntu Linux 20.04
* Graphics Card: NVIDIA GeForce GTX 1050 Ti/PCIe/SSE2, driver 460
* Screen type (if applicable): 1920x1080

treaves:
Continued cleanup

### Description
This focuses on the Oculars module itself.Work on the properties and the member instance variables for those properties.

Marked StelDialog::connectXXXProperty as deprecated.  These never should have been added, as they prevent compile-time error checking.  Standard connect syntax is what should hav been used.  Eventually I'll clean all of this up along with StelAction.

fixes #1673


alex-w:
Do you want to remove StelPropery and StelAction support for what? Just for understanding: right now you broke a support features of Oculars plugin by 3rd-party software, like N.I.N.A. and AstroPhotography Tool. Probably Oculars is not available through Remote Control plugin also.

treaves:
> Do you want to remove StelPropery and StelAction support for what? Just for understanding: right now you broke a support features of Oculars plugin by 3rd-party software, like N.I.N.A. and AstroPhotography Tool. Probably Oculars is not available through Remote Control plugin also.

No, not remove StelProperty & StelAction; just the way connect() is called.  Qt changed to the new connect() methodology for a reason; to get compile-time error checking.  No application should have methods commented with:

> //! @warning If the action with \c propName is invalid/unregistered, or cannot be converted
	//! to the required datatype, the application will crash

alex-w:
I didn’t compile the code and, of course, I didn’t check how it works now ;)

But 3rd-party software will not be work, because names of properties is changed. Maybe this is not critical, if we notify the developers of NINA and APT about changes.

Could you check issue #1673?

gzotti:
The connect*Property() helpers were a massive improvement in code clarity, saving 2-3 lines of connecting back and forth. Yes, you have to spell the property correctly. 
It seems you found a better way for the connections. This may not have been available in ~2015. But please don't just say "don't use", clarify what to use instead.

treaves:
@gzotti Correct, the functor-based connect was not present in 2015.  I've updated the deprivation message.

alex-w (approved):

PR #1696, opened 2021-05-29
sat plugin: qsplitter in sat list view

axd1967:
### Description
- A splitter has been added in the satellite tab
- tabbing order improved
- a reminder added in the Description box

I often wanted to resize the relative layout of the satellite tab because I wanted to see more of the groups. 
Also, I wasn't aware I could attach notes to satellites, so I highlighted this in the dialog.

image

### Type of change

### How Has This Been Tested?

- compare behaviour before/after
- diff review
-

**Test Configuration**:
KDE Plasma Version: 5.18.5
Qt Version: 5.12.8
Kernel Version: 5.4.0-73-generic
OS Type: 64-bit
Memory: 15.3 GiB of RAM

## Checklist:


alex-w (changes requested):
> Also, I wasn't aware I could attach notes to satellites, so I highlighted this in the dialog.

This feature (editable description for satellites) was added many years ago.

The text for placeholder is not very good IMHO and it should be rephrased and please avoid using round brackets for it.

@gzotti ?

gzotti:
I must admit I rarely use this plugin. I hope I can review the new functionality later today, but have other duties right now.

First sight: I don't object the bracketed "(add notes)", it implies there could be something interesting to add, while there is (likely) nothing actually stored yet.

For future PRs: please use screenshots in English.

axd1967:
> For future PRs: please use screenshots in English.

of course, sorry!

plugins/Satellites/src/gui/satellitesDialog.ui
```diff
@@ -1019,7 +1078,7 @@
               <bool>false</bool>
              </property>
              <property name="placeholderText">
-              <string notr="true"/>
+                <string>(add personal notes here)</string>
```
gzotti, line 1081: For me bracketed text is OK. It is used also in other programs. Behaviour looks good. Whether it's this or only "(add notes)" is a matter of taste.

gzotti (approved):
For me it's OK. I cannot comment the "improved tabbing" as I never used it. As soon as I tab into the text input, I input tabs there...

plugins/Satellites/src/gui/satellitesDialog.ui
alex-w: Ok

alex-w (approved):

axd1967:
The tab action can be changed with a property: TAB then just changes the focus.

I'm an avid keyboard user, so I was hesitating to enable this, because adding tabs in text could be interesting too, but it indeed stops further tabbing via keyboard.

There are more issues with tabbing, apparently the focus is not well visible, and there is probably an existing Qt bug that prevents toggling checkboxes with spacebar. Maybe Windows users can confirm that this is not a bug in Windows; but in KUbuntu it is a bit disappointing.

gzotti:
Maybe the GUI has been designed and implemented to hide tabbing artifacts (highlighting current item etc.) all those many years ago? Most of the GUI look&feel is legacy, and tab-related issues are indeed not my focus of interest.

Issue #1586, opened 2021-04-04
What quantity is actually meant by "light pollution luminance"?


10110111:
Working on integrating light pollution model into Stellarium, I'm struggling to understand the API of `LandscapeMgr`. Namely, there's a function, [`LandscapeMgr::setAtmosphereBortleLightPollution(int)`](https://github.com/Stellarium/stellarium/blob/8a4dc037de3515354a9f8a9ab322c953454cb05c/src/core/modules/LandscapeMgr.cpp#L1147), which calls [`LandscapeMgr::setAtmosphereLightPollutionLuminance(float)`](https://github.com/Stellarium/stellarium/blob/8a4dc037de3515354a9f8a9ab322c953454cb05c/src/core/modules/LandscapeMgr.cpp#L1135) with the argument that apparently converts Bortle index into some kind of luminance:
```cpp
qMax(0.f,0.0004f*powf(bIndex-1, 2.1f))
```
In math form it'd be

4×10<sup>-4</sup>(*n*-1)<sup>2.1</sup>,

where *n* is the Bortle index.

But when I take the "Approx. SQM" values from the [Wikipedia article on Bortle scale](https://en.wikipedia.org/wiki/Bortle_scale) and convert them to cd/m<sup>2</sup> using the formula [at this page](http://unihedron.com/projects/darksky/magconv.php), I get the following results (assuming the formula above yields value in cd/m<sup>2</sup>, which is expected due to [this comment](https://github.com/Stellarium/stellarium/blob/8a4dc037de3515354a9f8a9ab322c953454cb05c/src/core/modules/Atmosphere.hpp#L79)):

![plot](https://user-images.githubusercontent.com/6376882/113517297-bbdeda80-9587-11eb-8fcc-e5b634751906.png)

Here the gray area is the expected range of luminances from the table in Wikipedia, and the blue curve is the output of the above expression from `LandscapeMgr`.

As you can see, the line doesn't really match the area. Even if I try multiplying the `LandscapeMgr`'s expression by a constant (assuming it's just some proportional units that are wrong), it obviously won't fit in the expected area, here's an example of the factor 0.2:

![plot](https://user-images.githubusercontent.com/6376882/113517254-70c4c780-9587-11eb-9ed6-8da418096251.png)

OK, SQM includes some background luminance, so we might want to subtract the minimum value of luminance corresponding to magnitude 22 to get light pollution itself. Here's the result:

[...]

PR #1723, opened 2021-06-06
WIP: Switch from Bortle class to luminance as the quantity of light pollution

10110111:
This PR addresses #1586.

### WIP

This PR is not complete, but I'd like to get some feedback on it before I proceed. What's planned but not yet done is:

* GUI as proposed in https://github.com/Stellarium/stellarium/issues/1586#issuecomment-813623246 (currently it's just a 7-digit `QDoubleSpinbox`)
* User Guide should reflect the change in the interface, particularly for the Landscape config files.

### Description

Bortle scale has several disadvantages:

1. It was designed as a guide for humans, not input for machines. From this follows the next disadvantage.
2. It's discrete: each point on the scale corresponds to a range of NELM values, which makes it imprecise a program's input.
3. It's subjective: based on naked-eye limiting magnitude that depends on observer's visual acuity.
4. It isn't as widely used (and thus validated) as photometric units.

OTOH, luminance is the quantity that defines all the relevant effects of light pollution.

This commit makes luminance in cd/m² the input parameter, leaving for backwards compatibility the option of supplying Bortle class in Landscapes. Some changes to calculations had to be introduced, as listed below.

1. `StelSkyDrawer::getNELMFromBortleScale` was reimplemented in `StelCore::bortleScaleIndexToNELM` using a bit different representative values of NELM - to simplify the comment documenting the choice. The values are almost the same: each old value except one will be reproduced by adding 0.05 to the new ones.
2. Calculation of `B_mpsas` in `Nebula.cpp` was trying to convert from Bortle class to magnitude scale using some formula from Schaefer paper, the same paper as I used in `StelCore::luminanceToNELM`, but without explaining the steps to get the formula. The old expression form `Nebula.cpp` (which is widespread over the Internet) doesn't work near NELM==8 (see the discussion at https://astronomy.stackexchange.com/q/43313), while the one I used does. Anyway, `Nebula`'s logic is tied to luminance, so if the user supplies luminance instead of Bortle class, it'll be as intended.

For some reason, stars/init_bortle_scale setting defaulted to 3 while StelLocation::DEFAULT_BORTLE_SCALE_INDEX is 2. I've retained these preferences.

Technical note: I had to use `QVariant` instead of an `optional` type because Stellarium isn't using C++≥17 (C++17 has `std::optional`), nor anything from Boost (which has `boost::optional`). The other alternative was to make my own `optional`, but that seemed like overkill.

### Type of change

## Checklist:

### DIFF
--- a/src/core/StelCore.cpp
+++ b/src/core/StelCore.cpp
@@ -1156,7 +1156,7 @@ void StelCore::moveObserverToSelected()
 				loc.state = "";
 				loc.longitude=ni->getLongitude();
 				loc.latitude=ni->getLatitude();
-				loc.bortleScaleIndex=1;
+				loc.lightPollutionLuminance = 0;
 
 				moveObserverTo(loc);
 				objmgr->unSelect(); // no use to keep it: Marker will flicker around the screen.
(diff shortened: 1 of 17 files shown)


alex-w (changes requested):
I see in the log:
```
QObject::connect: No such slot StelSkyDrawer::setBortleScaleIndex(int)
QObject::connect:  (receiver name: 'StelSkyDrawer')
```

Please change the default step for luminance input box

10110111:
> I see in the log:

Hmm, where could it come from?.. I don't have any such in the log. I don't even find anything like `connect.*setBortleScaleIndex` int the sources...

> Please change the default step for luminance input box

What could a good step value be?

gzotti:
That's a tricky GUI issue. The Bortle scale may offer itself as human-readable and easier to remember/estimate external representation of luminance values which may not be as intuitive. I cannot say whether more users are used to discussing finding a "class 3 sky" than "XY cd/m²". Others in my vicinity are rather discussing their NELM directly. So, it may be good to use cd/m² inside the program, but keep the Bortle scale as GUI manipulator, and add a table of values to the User Guide. Maybe we can use graduations in tenths of "Bortle Units" for finer adjustments? Or even alternatives, i.e. offering one both interfaces, configurable. Or a slider which moves cd/m² but in addition shows Bortle values.

10110111:
> So, it may be good to use cd/m² inside the program, but keep the Bortle scale as GUI manipulator

Yeah, this is the next step. As I said above all, this is not yet done. The main point of the current RFC is about the infrastructure changes. The final more complex GUI will just use the infrastructure developed here.

alex-w:
> > Please change the default step for luminance input box
> 
> What could a good step value be?

I think it should be reduced to the first significant digit

gzotti:
Not sure (sorry, did not read into this topic for some years now). Should this not be logarithmic?

10110111:
> Not sure (sorry, did not read into this topic for some years now). Should this not be logarithmic?

It should, but spinboxes are limited in their behavior. Once again, the current version is minimal, it's meant to just show that the infrastructure changes don't break _functionality_. The UI part is irrelevant here.

Well, if no one here is willing to review the "invisible" side, I guess I should close this PR and come back when the final GUI is ready. I just didn't want to rework everything if the infrastructure appears to be unsatisfactory.

gzotti:
Sorry, not really unwilling, but too occupied with "real" work these weeks. Really sorry, I still must review your other PR which looks visually suspiciously bad to me. 
You can continue development in this PR and set it from grey draft to green "merge candidate" (I don't see the correct word now...). I doubt I will be able to touch this part of the code in June.

alex-w:
> What could a good step value be?

Is this question actual still?

10110111:
> Is this question actual still?

No, and it wasn't from the beginning. I was trying to remove distractions and make actual review possible. But now I'm not sure this PR makes any sense.

10110111:
Subsequent development of this issue is in [10110111:light-pollution-gui](https://github.com/10110111/stellarium/commits/light-pollution-gui) branch. I'll open a PR when I consider it ready.

PR #1816, opened 2021-07-26
Observing lists

Jocelyn1109:
### Description
new version of bookmarks: ObservingList

### Type of change
See specifications here: https://github.com/Stellarium/stellarium/issues/976

### How Has This Been Tested?
Manual unit test

**Test Configuration**:
* Operating system: Linux Ubuntu 20.04.2 LTS
* Graphics Card: GeForce RTX2060 6Go

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

Issue #976, opened 2020-02-17
Bookmarks becomes Observing List


Jocelyn1109:
Hi,
I create a ticket to formalize the development of this new feature. 
This ticket will be attached to the pull request.
The bookmarks will be completely redone to become observation lists.

Here the requirment :

 The Bookmarks tool will be deep refactored and visually splitted on 2 window - the first windows for using lists of bookmarked objects and the second list for adding/editing lists of objects. S
[Doc1.pdf](https://github.com/Stellarium/stellarium/files/4213417/Doc1.pdf)
[Doc1.pdf](https://github.com/Stellarium/stellarium/files/4213423/Doc1.pdf)
[Doc2.pdf](https://github.com/Stellarium/stellarium/files/4213424/Doc2.pdf)

ee sketches in the attachment.

Some comments:

- 1) We may store all lists in one file - for example observingLists.json - and this file just storing list of lists;  

- 2) Any list should storing 2 additional hidden properties for manage import/export of the data - unique ID and data of latest changes. Unique ID already implemented for current bookmarks.

- 3) Export and import tool should export/import list of bookmarks as currently implemented (subset of observingLists.json file)

- 4) All list shouldn't be stored translated data - english only names of designations. Translation should be applied on stage of visualization (creating table of view) data. Formal we may store only English name (or designation) and coordinates of object (currently in bookmarks we use coordinates to create a custom marker for object, who is not finded in Stellarium's objects) - and adding the coordinates (for finded objects - current coordinates), magnitude and abbreviation of constellation into table on stage of visualization of list.

- 5) "Empty" observingLists.json should store one list without objects - Bookmarks.

- 6) All files in UTF-8 and we may use special extension for separate lists (an exchange format of data) - .sol (stellarium observing list) or .skylist, or something else.

- 7) In future we may add tool for import observing lists from stellarium.org website (online import data) - of course if this tool will be very appropriated by users and we'll get feedback.

- 8) At the moment we store the location and date of observing data for each bookmark item and this is not very frequently use by users. But if we'll feedback we may add date and location data for all items of list as property for list: adding 2 fields into window 2 for define location and date/time and 2 checkboxes in first window for enable/disable applies those data.

- 9) Selecting object should be implemented by double click on list of object (+maybe adding separate button for go to object as currently implemented in Bookmarks tool) 

- 10) Maybe adding column "Type" (Type of object) will be useful. In any way we should use model to manage the data (like in Bookmarks at the moment).

[...]

PR #1826, opened 2021-07-31
some suggestions for the new observing list feature

axd1967:
@Jocelyn1109  here are some suggestions for improvements for #1816 . feel free to use this PR as you wish.

### DIFF
--- a/src/gui/ObsListCreateEditDialog.cpp
+++ b/src/gui/ObsListCreateEditDialog.cpp
@@ -1,5 +1,5 @@
 /*
- * * Stellarium
+ * Stellarium
  * Copyright (C) 2020 Jocelyn GIROD
  *
  * This program is free software; you can redistribute it and/or modify
@@ -119,7 +119,7 @@ void ObsListCreateEditDialog::createDialogContent()
     } else {
         // case of edit mode
         isCreationMode = false;
-        ui->stelWindowTitle->setText ( "Observing list modification mode" );
+        ui->stelWindowTitle->setText ( "Observing list editor" );
         loadObservingList();
     }
 }
@@ -148,12 +148,13 @@ void ObsListCreateEditDialog::styleChanged()
 */
 void ObsListCreateEditDialog::setObservingListHeaderNames()
 {
+    // XXX there is a lot of copy-pasta from ObsListDialog - this will lead to maintenance headaches!
     const QStringList headerStrings = {
-        "UUID", // Hided column
+        "UUID", // Hidden column
         q_ ( "Object name" ),
-        q_ ( "Object Name I18N" ), // Hided column
+        q_ ( "Object name I18N" ), // Hided column
         q_ ( "Type" ),
-        q_ ( "Right ascencion" ),
+        q_ ( "Right ascension" ),
         q_ ( "Declination" ),
         q_ ( "Magnitude" ),
         q_ ( "Constellation" ),
@@ -163,7 +164,6 @@ void ObsListCreateEditDialog::setObservingListHeaderNames()
 
     obsListListModel->setHorizontalHeaderLabels ( headerStrings );
 }
-
 
 /*
  * Add row in the obsListListModel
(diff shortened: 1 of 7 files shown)


src/gui/ObsListCreateEditDialog.cpp
```diff
@@ -119,7 +119,7 @@ void ObsListCreateEditDialog::createDialogContent()
     } else {
         // case of edit mode
         isCreationMode = false;
-        ui->stelWindowTitle->setText ( "Observing list modification mode" );
+        ui->stelWindowTitle->setText ( "Observing list editor" );
```
Jocelyn1109, line 122: @axd1967 I changed "modification mode into "editor mode". Is it ok for you ?

axd1967: _(If you allow me to be honest, it feels slightly wrong? Rest assured, it makes ME feel bad to have to write this, because it's so much easier to criticize (even in a positive way) than to develop a working solution; so please bear with me. The core issue, having different bookmark lists, remains a significant step forward because users had to juggle with a lot of bookmark files)._

Either there is _one_ window that has several modes (in which case the user is shown the mode), or there are different windows that each do a specific (and probably complex) task.

In this case, the user is interacting with the **observing list editor**. Another indication that there is something smelly is that the user switches to a different mode _and to a new dialog_.


I think that the real problem is due to the near-identical UI of two slightly different windows (which has a heavy smell of "copy-pasta"). IMHO a single window is enough that can show as well as edit the current observing list; the list selector selects the currently active list; the dialog allows all operations.

image

(`(!)` = guarded, `(?)` = do we really need a close button ? maybe, if there is a save/undo feature available, otherwise ESC or the top-right close buttons are fine )

I suspect that Qt allows to type text in a dropdown list, which might be a nice way to allow the user to create a new list in an intuitive way. You see this in action in the Astrocalc dialog:
image

(But I am prepared to admit that there might be a smell of short-sightedness to these suggestions, so I'm also looking forward to inputs of others.)

axd1967:
Notice another advantage of a minimal interface: I can lodge it in a corner and jump to objects:

image

axd1967:
In fact , the UI could be further simplified:


image


- list management buttons are together
- (no more close button, as the dialog has a close button anyway)

axd1967:
- another important use case is to be able to **copy** such a list, as this allows to reshuffle bookmarked objects, try things, etc. se we need a "copy" button too.
- a less obvious but **extremely powerful** change is to allow the user to **rename** a list by **simply editing the name in the list dropdown/combobox**. See https://doc.qt.io/qt-5/qcombobox.html#editable-prop for details.

axd1967:
The dialog can be further simplified:

image


CRUD:
- to create a new list, select the empty slot in the dropdown (so we keep an empty slot) and type a name. this will create a copy of the previously selected list (and create a new empty slot)
- to rename the list, edit the list name in the dropdown
- to delete the list, click Delete (! =  action that is guarded with a confirmation dialog)

axd1967:
I will close this PR because it is not meant to be merged at all, but is a review with comments and suggestions.

However, the comments remain actual.

Jocelyn1109:
Hi, currently we need to focus on what was defined in the first specification ([#976](https://github.com/Stellarium/stellarium/issues/976)), fix the bugs to be able to release a first version and wait for user feedback. then we will be able to make other deliveries with modifications in the bookmarks. We have to proceed step by step otherwise it will be impossible for me to integrate everything.

Jocelyn

axd1967:
I understand, and I agree with the principle; but the current state is not mature for release, and I have the impression that the requirements are not ok.

In fact, requirement "viii" states _"At the moment we store the location and date of observing data for each bookmark item and this is not very frequently use by users"_ : this is WRONG and a major regression.

I'm sorry but I do firmly believe this should not be released.

Alternatively, this could be released as a SEPARATE feature, living TOGETHER with the current bookmark functionality; in that case, I can live with the present state, as I can continue to use the existing Bookmark feature.

Sorry to cause this hassle, this is not personal, but purely based on the present state of this issue.

Issue #678, opened 2019-05-07
Daylight-color-mode


ratmalwer:
Today I was out in the field with my notebook to look what objects could be seen at night. Unfortunately the sun was shining bright. So I could not see anything on my screen (Stars, compass point or horizon). Even I turned the display on maximum brightnes.

Could it be possible to make a daylight-mode? I suggest inverse colors or better just black on white. This should affect the fullscreen including stars and texts. Preferably even the menues in an oldfashioned way without opacy.

Kind regards 
Markus

axd1967:
Totaly agree on this issue, and seeing this tagged "opinion" and "wont fix" almost stopped my heart and has started wondering what's happening here. I almost missed the "PR accepted" note...

personally I'm struggling a lot with this theme, and I'm afraid I'm not alone. my eyes have lived over 50 years and are a bit worn...

the dark grey theme is a major UI issue . Take the astro calc window as an example: the graphs are almost unreadable (especially the thin red line on grey background, I would put my hand in fire that a UI expert would stumble over this, and I wonder how the color-blind percieve this). 

![image](https://user-images.githubusercontent.com/3529789/63294627-1c45d900-c2cb-11e9-91c2-00a1c57148e5.png)

 and one has to look very hard to see those see extra, faint lines: I've inverted the colors in an attempt to bring thise out:


![image](https://user-images.githubusercontent.com/3529789/63295191-5663aa80-c2cc-11e9-9d19-16b9a3f830ca.png)

![image](https://user-images.githubusercontent.com/3529789/63295273-8743df80-c2cc-11e9-956e-28bf9092798f.png)

Probably they were chosen to operate optimally in darkened conditions, which is understandable.

But those graphs are inherently meant to be used for planning purposes - i.e. during bright daylight.

I guess this a Qt theme that should become an option/setting; someone wanting to consult such graphs in dark condition can always enable night vision mode instead. 

Which, oddly, currently (0.19) provides a better viewing experience...

![image](https://user-images.githubusercontent.com/3529789/63294623-14863480-c2cb-11e9-9d01-1cca60099d2f.png)

 This should not be tagged "won"t fix" because quite rude, but, for example, "too expensive for the moment". and/or it could be put in a backlog pipeline, for example.

(and unfortunately unrelated, but the tooltips try to be helpful but fade away ar too quickly!)

alex-w:
[...]

PR #1834, opened 2021-08-14
keyboard shortcut editor: focus/tabs

axd1967:
### Description
A recurring issue is that I need to click in the search field of the dialog to search for an action, and the tab order is odd (in fact, not specified, but grown over time).

This PR puts the initial focus in the search box, and improves the tab order.

An easy way to edit shortcuts is to search for a part of their name.
A SHIFT TAB allows the user to select a matching action, and the user can then TAB forward into the editor fields.

(Small note: on Linux, the BackspaceButtons do not seem to show a focus marker, but the focus is present and the buttons can be pushed via ENTER rather than SPACEBAR.)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating System: Kubuntu 20.04
KDE Plasma Version: 5.18.5
Qt Version: 5.12.8
Kernel Version: 5.4.0-80-generic
OS Type: 64-bit
Memory: 15.3 GiB of RAM

## Checklist:

### DIFF
--- a/src/gui/ShortcutsDialog.cpp
+++ b/src/gui/ShortcutsDialog.cpp
@@ -335,6 +335,9 @@ void ShortcutsDialog::createDialogContent()
 	QString style = "QLabel { color: rgb(238, 238, 238); }";
 	ui->primaryLabel->setStyleSheet(style);
 	ui->altLabel->setStyleSheet(style);
+
+	// set initial focus to action search
+	ui->lineEditSearch->setFocus();
 }
 
 void ShortcutsDialog::polish()
--- a/src/gui/ShortcutsDialog.hpp
+++ b/src/gui/ShortcutsDialog.hpp
@@ -95,7 +95,7 @@ private:
 	static bool itemIsEditable(QStandardItem *item);
 
 	//! Apply style changes.
-	//! See http://qt-project.org/faq/answer/how_can_my_stylesheet_account_for_custom_properties
+	//! See https://wiki.qt.io/Technical_FAQ#How_can_my_stylesheet_account_for_custom_properties.3F
 	void polish();
 
 	QStandardItem* updateGroup(const QString& group);
@@ -103,7 +103,7 @@ private:
 	//! search for first appearance of item with requested data.
 	QStandardItem* findItemByData(QVariant value, int role, int column = 0) const;
 
-	//! pointer to mgr, for not getting it from StelApp every time.
+	//! pointer to mgr, to avoid getting it from StelApp every time.
 	class StelActionMgr* actionMgr;
 
 	//! list for storing collisions items, so we can easy restore their colors.
--- a/src/gui/shortcutsDialog.ui
+++ b/src/gui/shortcutsDialog.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>654</width>
+    <width>737</width>
     <height>436</height>
    </rect>
   </property>
@@ -297,6 +297,18 @@
    </slots>
   </customwidget>
  </customwidgets>
+ <tabstops>
+  <tabstop>shortcutsTreeView</tabstop>
+  <tabstop>lineEditSearch</tabstop>
+  <tabstop>primaryShortcutEdit</tabstop>
+  <tabstop>primaryBackspaceButton</tabstop>
+  <tabstop>applyButton</tabstop>
+  <tabstop>altShortcutEdit</tabstop>
+  <tabstop>altBackspaceButton</tabstop>
+  <tabstop>restoreDefaultsButton</tabstop>
+  <tabstop>restoreAllDefaultsButton</tabstop>
+  <tabstop>closeStelWindow</tabstop>
+ </tabstops>
  <resources>
   <include location="../../data/gui/guiRes.qrc"/>
  </resources>


formtapez:
This has to be merged!
I wrecked my program settings so many times by hitting CTRL+F and start typing without the focus in the search field. So annoying...

axd1967:
> This has to be merged!
> I wrecked my program settings so many times by hitting CTRL+F and start typing without the focus in the search field. So annoying...

Note that this is a different issue - I created a separate issue for this annoyance.

gzotti:
That's an interesting thing. I was not even aware of the "search" line before. 
However, while the cursor blinks in the entry lines, on Windows there is no hint on which button would now be active when spacebar would be pressed. This is likely by design, and our original GUI designer did not have keyboard button navigation by tabbing in mind when designing our GUIs. The sequence therefore appears currently rather more dangerous than useful, sorry. I will add the focus line though, or even two of the three changed files, thanks for the URL update docfix.

axd1967:
When the cursor blinks, the focus is where the cursor is, so spacebar will not work other than to add spaces in the input field. In such a case, the user has to press TAB in order to move to the next UI item.

About tabbing: the additions made over time turned the tabbing order into a mess because the successive developers simply overlooked or were not aware of that issue, so this has to be rectified anyway. This is the moment to think over the tabbing order.

BTW, I noticed that in Ubuntu, it is very hard to see where the focus is (#678 ) if it is not in an entry field.

We can't rely on what previous developers intended to do or didn't think of; that way, nothing can be changed any more... They might very well never have thought of things that creep up today.

gzotti:
I am principally aware of how tabbing works to switch from field to field. It is just that outside from input fields, there is no visual feedback to where the focus now is. At least not on Windows, and you just wrote also not on Ubuntu. In other programs button labels may get highlighted, the label text bounding box dotted, or something else. If I activate your sequence, it works going from input field to input field (with blinking cursor), but after that there is no clue as to what would happen if I now press space or enter. Do I press some delete key, restore to defaults, or reset all? Therefore it is better to not have any half-working sequence and just not use the tabs.

axd1967:
You are right - the absence of focus markers is not good. But I believe that is a matter of finding the right setting somewhere in Qt, because I can't accept that this was overlooked by the Qt team. But at first sight, it is [not looking good](https://stackoverflow.com/questions/58067679/qtquick-controls-2-12-does-not-show-focus-on-buttons/69329055#69329055).

Nowadays, when the user presses TAB, there is already no way either to know where the focus is, so keyboard users are discouraged by the messy tab logic

image

at least, with this patch the experienced user can expect (and will discover) a more logical ordering. with the main idea of this patch all being about being able to jump between tree item, search line, and editor line.

image

There are no real dangers - the "Restore all Default" is guarded, and the Restore can be undone by memory.

axd1967:
How about we meet in the middle: we allow tabbing to "safe" widgets only. 

image


Personally I think this is not OK, but it is better than leaving the tab order in its current state.

Then, a keyboard user has bare minimal access to essential actions, and should still use the mouse for the other actions, so everybody is happy.

(By the way - the branch is wrongly named "search/focus" but should be "shortcuts/search", sorry for that.)

alex-w:
@gzotti adding the order of shortcuts for keyboard navigation is good for vim/emacs fans and old-school users - this is my opinion for generic case or situation. Let's see current versions of implementation: all orders looks weird for me, because all of them ignores the location of focus. IMHO obviously start point for order (number 1) should be in input field for search a shortcuts. The next point in input field for primary shortcut and etc.

axd1967:
be aware that _order_ does not require that the most important field has to be _number one_ in that order; that is programmed by setting the initial focus, set wherever it is the most important for the UI. e.g. in the screenshot above, initial focus should be on number 2 in the tab order.

the order should be logical top-down, left-right.

and there is indeed an issue where buttons do not get a focus indication; but as long as the button is not dangerous (eg important and without confirmation dialog) , there is no stringent reason to skip buttons because this goes against the very idea of tab ordering: to provide a keyboard interface (that is much faster than a mouse interface!). keyboard users soon discover how often they have to tab in order to activate a button, irrespective of focus. and users that do not use the keyboard are not bothered about the tab order anyway.

axd1967:
Styling is probably (I have no experience in the stylesheet aspect of Qt) the ideal solution, but the current PR solves the problem for now, in a practical way.

And in successive iterations, we can review the tab policy in other UI; this can also be done every time someone touches a UI.

gzotti:
@axd1967 can you please rebase this from current master and PR to gui-show-focus? Tab order makes sense if checkboxes and buttons indicate focus and react to space or enter. I added a SpaceCheckBox which can be placed in tab order and pressed with Space. I think buttons also need a subclass to be pressed with Space.  This can be developed in that branch.

axd1967:
I will check this.
But tab order also makes sense without visual indication, although ergonomically bad.

axd1967:
LGTM
there are two observations:

1. I would use an extra pixel width for the highlighting frame, the focus is a bit difficult to see
2. I would be careful with ENTER, as it can sometimes be useful as an alternative to TAB ([Qt](https://doc.qt.io/qt-5/focus.html#tab-or-shift-tab))
2. ENTER is often also a default key for a _dialog_, irrespective of which element currently has focus: eg "get orbital elements" in the "import data" dialog, or "calculate ephemeris" in the Astrocalc dialog could benefit from this, even though this is currently not implemented in those dialogs. (but indeed, in the Stellarium confirmation dialog, ENTER does indeed also trigger the button having the focus)
2.  SPACEBAR (rather than ENTER) typically acts as a replacement for a click action for elements having focus: the back/delete buttons are such example.

Some unrelated observations: 

1. pressing the `primary/altBackspacebutton` behaves a bit odd - the underlying logic tries to be too clever by shifting the focus back to the `xShortcutEdit` (and select the entire field); the focus should remain on the back button. in other words, here a keyboard user is handicapped compared to a mouse user.
2. probably a fresh bug: filtering does not seem to work as previously, I try to bisect that problem

PR #1835, opened 2021-08-14
MPC import: improve sort of URLs

axd1967:
also
- remove duplicate DOU MPEC
- readability

image

after:

image

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating System: Kubuntu 20.04
KDE Plasma Version: 5.18.5
Qt Version: 5.12.8
Kernel Version: 5.4.0-80-generic
OS Type: 64-bit
Memory: 15.3 GiB of RAM

## Checklist:

### DIFF
--- a/plugins/SolarSystemEditor/src/gui/MpcImportWindow.cpp
+++ b/plugins/SolarSystemEditor/src/gui/MpcImportWindow.cpp
@@ -374,11 +374,13 @@ void MpcImportWindow::populateCandidateObjects(QList<SsoElements> objects)
 	//QHash<QString,QString> defaultSsoIdentifiers = ssoManager->getDefaultSsoIdentifiers();
 	QHash<QString,QString> loadedSsoIdentifiers = ssoManager->listAllLoadedSsoIdentifiers();
 
-	//Separate the objects into visual groups in the list
+	//Separate the objects into visual (internally unsorted, anyone?) groups in the list
+
 	//int newDefaultSsoIndex = 0;
-	int newLoadedSsoIndex = 0;
-	int newNovelSsoIndex = 0;
-	int insertionIndex = 0;
+	int newLoadedSsoIndex = 0; // existing objects
+	int newNovelSsoIndex = 0; // new objects
+
+	int insertionIndex = 0; // index of object to be inserted next
 
 	QStandardItemModel * model = candidateObjectsModel;
 	model->clear();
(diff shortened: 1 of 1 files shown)


plugins/SolarSystemEditor/src/gui/MpcImportWindow.cpp
```diff
@@ -1038,39 +1040,49 @@ void MpcImportWindow::loadBookmarks()
...
-	bookmarks[MpcMinorPlanets].insert("MPCORB: orbits from the latest DOU MPEC", "https://www.minorplanetcenter.net/iau/MPCORB/DAILY.DAT");
-	bookmarks[MpcMinorPlanets].insert("MPCORB: elements of NEAs for current epochs (today)", "https://www.minorplanetcenter.net/iau/MPCORB/NEAm00.txt");
-	bookmarks[MpcMinorPlanets].insert("MPCORB: orbits from the latest DOU MPEC", "https://www.minorplanetcenter.net/iau/MPCORB/DAILY.DAT");
-	bookmarks[MpcMinorPlanets].insert("MPCAT: Unusual minor planets (including NEOs)", "https://www.minorplanetcenter.net/iau/ECS/MPCAT/unusual.txt");
-	bookmarks[MpcMinorPlanets].insert("MPCAT: Distant minor planets (Centaurs and transneptunians)", "https://www.minorplanetcenter.net/iau/ECS/MPCAT/distant.txt");
-	int start = 0;
-	int finish = 52;
+	// NOTE: this list is reordered anyway when loaded
+
+	bookmarks[MpcMinorPlanets].insert("MPC's list of bright minor planets at opposition in 2018", 		"https://www.minorplanetcenter.net/iau/Ephemerides/Bright/2018/Soft00Bright.txt");
```
gzotti, line 1045: I haven't used this in a long time. ;-) 2018 could be bumped to 2021, or can we even add all years 2018..2022 if MPC offers it? (Keeping a few older years makes sense to re-investigate old images or so.)

alex-w: The link is still actual

gzotti: Oops, are there no more updates of these annual lists from MPC?

axd1967: Yes, but it is a duplicate of line 1049 unless I'm mistaken?

gzotti: Indeed. But you have removed the duplicate already.
I hope nobody ever loads a full set of even just 10000 objects :-P

alex-w: > Oops, are there no more updates of these annual lists from MPC?

No

gzotti:
It seems this may re-sort the list on first call only. With my existing bookmarks.json file it keeps showing this:
stellarium-006

Still, it seems not to break other things, so we could as well merge it.

gzotti (approved):
For me it looks OK.

axd1967:
> It seems this may re-sort the list on first call only. With my existing bookmarks.json file it keeps showing this:
> stellarium-006
> 
> Still, it seems not to break other things, so we could as well merge it.

indeed, and that's a workaround for existing configurations.

which leads to another question: I did not check whether updates in the data will be absorbed by users that already have a `bookmarks.json` file; in other words, do such users benefit from future changes in the original list, or is the bookmark file copied once, only when it does not yet exist?

axd1967:
that was an academic question, but is irrelevant for this PR :-)

PR #1852, opened 2021-08-28
Planet RTS

### LINKED ISSUE #554: Objects' raise/set time changes during night
### Expected Behaviour
Raise/culminate/set times are consistent during night

### Actual Behaviour
Raise/culminate/set times change during night

stellarium-006
stellarium-007

### Steps to reproduce
- pick star on the east
- notice its set time
- speed up time
- notice set time when star is on the west
- there is a ~4 minutes difference

### System
* Stellarium version: stellarium-0.90.0.16394-win64.exe
* Operating system: Windows 7
* Graphics Card: Intel(R) HD Graphics Family

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/2609115/log.txt)

gzotti:
### Description

This should solve the inaccurate rise/set times given for planets, esp. the Moon. It is based on the method given by Meeus  but with some optimisations. 

Fixes #554

### Type of change

### How Has This Been Tested?

Observe Moon (or Sun, or planet) on screen, see behaviour and times.
I have still to make tests around reaching circumpolarity and other funny things.

There is still the problem that sometimes rise time for the next day or setting time for the previous day is found. 

Still not fully tested: behaviour on other planets. 

The fine corrections are also applied to StelObject.

**Test Configuration**:
* Operating system: Win10 21H1 
* Graphics Card: Nvidia 960M (irrelevant)

## Checklist:


alex-w:
By the fact `getRTSTime` and `computeRTSTime` methods have same behavior and I think we can move code from `computeRTSTime` into `getRTSTime` method and removed first one. The second problem: to compute RTS time of the Sun for twilight we need change the height of celestial body (+ users ask this feature for AstroCalc functions) - so, maybe add new one parameter for `getRTSTime` method with default value?

alex-w:
> Maybe exclude DebugAid flags from Release builds?

DebugAid should be available in Debug mode only

alex-w:
A quick test: RTS time for planets give very strange results - like a double application of the time zone shift

gzotti:
re1: One is public, the other is not. But that's easy to decide...
re2: Yes, target altitude should be possible. An advanced solution (possible around 0.14, maybe 2022...) would be taking the landscape horizon into account. This may be possible with some binary search in hour angle. But not in this branch.
re3: DebugAid: this comment is old, just the bot found it... I could exclude it from real STELLARIUM_RELEASE_BUILDs, but not Qt's "release" mode. However, these in-screen data displays are targeted for developing particular features, so they have to be commented away (or deleted) at end of branch development.  
re4: (strange) I had hoped the interpolation wraparound zone fix would remedy that. Need another look. 

I tried the "goto next setting" etc. functions on hotkey which are really nice in principle. While this version is usually much better (esp. Moon), the case where the solution provides times for the previous or next day needs to be discovered and solved. There are however 2 special cases: Sidereal day is 23h56m, so a star may rise twice on a single calendar day. And with moonrise/-set ~25 hours apart, the Moon may not rise/set on some particular date. If anybody has a worked-out solution or a better algorithm than Meeus (how dare you!), this could really help. We actually need to "find nearest event" (maybe with a way to flag it "will occur on next day only")  and "goto next moonrise" (not "find moonrise date from algorithm, then set current date with that time").

alex-w:
> Still untested: behaviour on other planets.

Well, the calendars and duration of the day on extraterrestrial locations is separate big problem. :(

alex-w:
> I tried the "goto next setting" etc. functions on hotkey which are really nice in principle. While this version is usually much better (esp. Moon), the case where the solution provides times for the previous or next day needs to be discovered and solved. There are however 2 special cases: Sidereal day is 23h56m, so a star may rise twice on a single calendar day. And with moonrise/-set ~25 hours apart, the Moon may not rise/set on some particular date. If anybody has a worked-out solution or a better algorithm than Meeus (how dare you!), this could really help. We actually need to "find nearest event" (maybe with a way to flag it "will occur on next day only") and "goto next moonrise" (not "find moonrise date from algorithm, then set current date with that time").

Well, the special cases for the Sun on high latitudes also need to remember. :)

gzotti:
Yes, indeed.  This PR is currently only a partial fix. I think I should invent a different algorithm that starts with the current hour angle, find nearest transit, then estimates for rise/set from semidiurnal arc, with final solutions from binary search or those correction terms given by Meeus. And if sun rises above horizon in Polar summer, it will have a rise and transit time, but no more setting this day.  The "next setting" call would switch to 24 o'clock of that day.

Not sure, do we accept this (maybe some more fixes?) for 0.21.2 and see to devise sth more complete ~ end of year?

alex-w:
I think we may postpone implementing the special cases to other release.

alex-w:
Well, I found another issue - the latest 2 commits solve time zone shift, but now user can see time over 24 hours. Plus duration of day is negative now.

gzotti:
Didn't see that, must check with your timezone ;-)

alex-w:
It's OK now, thanks!

gzotti:
If you agree, I will squash&merge.  Or should I add the abbreviated version to StelObject, then the rise/set times will no longer move during the day.

alex-w (approved):

alex-w:
> Or should I add the abbreviated version to StelObject, then the rise/set times will no longer move during the day.

Please do it

gzotti:
Uh, big problem (crashes) when moving to Mars :-(
I could do a "if location on earth: this new solution, else old solution". Or give up this branch for now and do something entirely different as described above.

alex-w:
Probably you shouldn’t compute RTS when planet is SpaceShip.

P.S. I didn’t test this PR on extraterrestrial locations :(

gzotti:
A yes, spaceship observer may be the real problem.

gzotti:
This needs an entirely different approach. No time in the next 2 weeks. :-(

alex-w:
This patch fixed crash for spaceship location: https://github.com/Stellarium/stellarium/commit/607ce46ac24039ccf370ec82a14753035ecd5edb

gzotti:
Thanks! I have thought about a different implementation now, which should avoid the next/following day problem. Hopefully later in September, not sure if already for 0.21.2.

alex-w:
Important note: my patch is replaced your latest commit

gzotti:
Good! :-) My astro club wanted such a function for years...

alex-w (approved):

gzotti:
@alex-w How shall we report circumpolarity to the InfoMap? Return transit time (or lower culmination), or use a verbal annotation?

alex-w:
Probably verbal annotation will be enough here

alex-w:
> Uh, I'm not sure if I'm missing something here, but selecting the Moon will give different rise/transit/set values over different times of the same day. Is that intentional?

No, this is wrong

gzotti:
There is a slight change in time by a few minutes if you are using topographic correction. For geocentric computation is should be stable. Given atmospheric vagaries the difference should be negligible. We accept better-working solutions in code :-)

PR #1855, opened 2021-08-31
Switch to v1.4.3 of QXlsx (v1.4.3-stel.0)

alex-w:
QXlsx was updated to version 1.4.3 with patch to use our copy of qtcompress library

### DIFF
--- a/src/external/CMakeLists.txt
+++ b/src/external/CMakeLists.txt
@@ -93,6 +93,10 @@ add_library(qcustomplot_stel STATIC EXCLUDE_FROM_ALL ${qcustomplot_SRCS})
 target_link_libraries(qcustomplot_stel Qt5::PrintSupport Qt5::Widgets)
 SET_TARGET_PROPERTIES(qcustomplot_stel PROPERTIES FOLDER "src/external")
 
+# QXlsx
+# Source code: https://github.com/QtExcel/QXlsx
+# v1.4.3-stel.0:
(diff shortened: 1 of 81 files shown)


gzotti:
Anything in particular I could test?

alex-w:
Please generate xlsx files (through AstroCalc for example) and open these files in various electronic tables. I've tested it witih LibreOffice Calc, MS Excel and Apple Numbers - everything OK.

axd1967:
(this is an example where the **vendor branch** mechanism can be used.)

gzotti:
The buttons in AstroCalc with labels "Save positions", "Save ephemeris", etc. could be clearer when labeled "Export positions...", "Export ephemeris..." etc. The actual export seems to work well.

alex-w:
> The buttons in AstroCalc with labels "Save positions", "Save ephemeris", etc. could be clearer when labeled "Export positions...", "Export ephemeris..." etc. The actual export seems to work well.

OK, I'll rename the text of labels tonight

alex-w:
@gzotti OK, renaming is done

gzotti:
Thanks, this just looks better. 

I am not fully sure: Is the code else just imported with new versions (with copyright notes removed by the original author!), or did you change something? In the first case, I approve. Else the required copyright notes should remain in the code. And probably some 1-line README with URL of origin or so? Or is the origin of this directory documented elsewhere?

PR #1880, opened 2021-09-11
Exclude outdated solar system objects

### LINKED ISSUE #1564: Exclude outdated comet data in ephemeris computations
### Expected Behaviour
Don't include comets beyond their valid time range in ephemeris computations

### Actual Behaviour
Found during trial around #1563: When searching for phenomena around Jupiter's moons in 2021, I can find conjunctions with comets with orbital data from the 1960s or so. 

### Steps to reproduce

Load 1000 comets data file. 
Use AstroCalc/Phenomena to search for phenomena around Jupiter's moons, e.g. June to September 2021.

### System
* Stellarium version: 0.21.0-RC1
* Operating system: Windows (irrelevant)
* Graphics Card: (irrelevant)

alex-w:
### Description
Excluding the Solar system objects with outdated data from ephemeris, positions and phenomena computations

Fixes #1564

v0.21.1:
stellarium-006

v0.21.1-122fe98:
stellarium-007

### Type of change

### How Has This Been Tested?
Load 1000 comets data file.
Use AstroCalc/Phenomena (AstroCalc/Positions or AstroCalc/Ephemeris also) to search for phenomena around Jupiter's moons, e.g. June to September 2021.

## Checklist:

### DIFF
--- a/src/core/modules/Planet.hpp
+++ b/src/core/modules/Planet.hpp
@@ -222,6 +222,8 @@ public:
 	virtual double getAngularSize(const StelCore* core) const Q_DECL_OVERRIDE;
 	virtual bool hasAtmosphere(void) {return atmosphere;}
 	virtual bool hasHalo(void) {return halo;}
+	//! Returns whether planet positions are valid and useful for the current simulation time. E.g. outdated orbital elements for Kepler orbits (beyond their orbit_good .ini file entries) may lead to invalid positions which should better not be used.
+	//! @note for major planets and moons this method will be always return true
 	bool hasValidPositionalData();
 	float getAxisRotation(void) { return axisRotation;} //! return axisRotation last computed in computeTransMatrix().
 
--- a/src/core/modules/Planet.hpp
+++ b/src/core/modules/Planet.hpp
@@ -225,12 +225,12 @@ public:
 	//! Returns whether planet positions are valid and useful for the current simulation time.
 	//! E.g. outdated orbital elements for Kepler orbits (beyond their orbit_good .ini file entries)
 	//! may lead to invalid positions which should better not be used.
-	//! @note for major planets and moons this method will be always return true
+	//! @note for major planets and moons this method will always return true
 	bool hasValidPositionalData(const double JDE);
 	//! Returns JDE dates of presumably valid data for positional calculation.
-	//! For the major planets, this is always (std::minreal, std::maxreal)
+	//! For the major planets and moons, this is always (std::numeric_limits<double>::min(), std::numeric_limits<double>::max())
 	//! For planets with Keplerian orbits, this is (epoch-orbit_good, epoch+orbit_good)
-	Vec2d getValidPositionalData();
+	Vec2d getValidPositionalDataRange();
 	float getAxisRotation(void) { return axisRotation;} //! return axisRotation last computed in computeTransMatrix().
 
 	///////////////////////////////////////////////////////////////////////////
--- a/src/core/modules/Planet.hpp
+++ b/src/core/modules/Planet.hpp
@@ -222,9 +222,15 @@ public:
 	virtual double getAngularSize(const StelCore* core) const Q_DECL_OVERRIDE;
 	virtual bool hasAtmosphere(void) {return atmosphere;}
 	virtual bool hasHalo(void) {return halo;}
-	//! Returns whether planet positions are valid and useful for the current simulation time. E.g. outdated orbital elements for Kepler orbits (beyond their orbit_good .ini file entries) may lead to invalid positions which should better not be used.
+	//! Returns whether planet positions are valid and useful for the current simulation time.
+	//! E.g. outdated orbital elements for Kepler orbits (beyond their orbit_good .ini file entries)
+	//! may lead to invalid positions which should better not be used.
 	//! @note for major planets and moons this method will be always return true
 	bool hasValidPositionalData(const double JDE);
+	//! Returns JDE dates of presumably valid data for positional calculation.
+	//! For the major planets, this is always (std::minreal, std::maxreal)
+	//! For planets with Keplerian orbits, this is (epoch-orbit_good, epoch+orbit_good)
+	Vec2d getValidPositionalData();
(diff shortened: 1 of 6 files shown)


axd1967:
The user would benefit from being informed about the problem, as it is not obvious that ephemeris data is outdated (as we can judge from several issues that are reported), especially for comets/asteroids, as these require the user to open the Solar System Editor, which is a tad hairy to manipulate. All too often the users **assume** that the data is always up-to-date, which is obviously NOT the case. So until an automatic update is possible, the user should be aware of the problem.

at least emit a `QWarning()` in the log file for each object that is outdated. 
It looks like this could be a good location for such a warning:
https://github.com/Stellarium/stellarium/blob/bc5ee585fc7b0b5f496a5afa340ac5c2cbbc25e3/src/core/modules/SolarSystem.cpp#L852

nicer but much more expensive is to (also) highlight outdated objected in the tabular GUIs (eg grey/italic?)

src/core/modules/Planet.hpp
```diff
@@ -222,6 +222,7 @@ class Planet : public StelObject
 	virtual double getAngularSize(const StelCore* core) const Q_DECL_OVERRIDE;
 	virtual bool hasAtmosphere(void) {return atmosphere;}
 	virtual bool hasHalo(void) {return halo;}
+	bool hasValidData();
```
gzotti, line 225: Please document what "valid" means in this context! Maybe rename to hasValidOrbitData() or hasValidPositionalData()

alex-w: Good suggestion! 

Done.

src/core/modules/Planet.hpp
```diff
@@ -222,6 +222,7 @@ class Planet : public StelObject
 	virtual double getAngularSize(const StelCore* core) const Q_DECL_OVERRIDE;
 	virtual bool hasAtmosphere(void) {return atmosphere;}
 	virtual bool hasHalo(void) {return halo;}
+	bool hasValidPositionalData();
```
gzotti, line 225: Honestly, we need better documentation! The name is now better, but this should explain to other developers (or ourselves, in 4 months from now) why we need it, or what this test can be used for. E.g.

//! returns whether planet positions are valid and useful for the current simulation time. E.g. outdated orbital elements for Kepler orbits (beyond their orbit_good .ini file entries) may lead to invalid positions which should better not be used.

alex-w: OK, you are right. I'm added notes also...

src/gui/AstroCalcDialog.cpp
```diff
@@ -4473,6 +4480,19 @@ QMap<double, double> AstroCalcDialog::findClosestApproach(PlanetP& object1, Stel
 	QMap<double, double> separations;
 	QPair<double, double> extremum;
 
+	if (object2->getType()=="Planet")
```
gzotti, line 4483: Hmm, do you also see the inefficiency?
For a simple date validity test you set the whole program state twice. 

Please do the following: extend
bool Planet::hasValidPositionalData(const double jde)
{
	if (pType<isObserver)
		return true;
	else if (orbitPtr && pType>=isArtificial)
		return static_cast<KeplerOrbit*>(orbitPtr)->objectDateValid(jde);
	else
		return false;
}

And then you can simply test here:

if  ( (!planet->hasValidPositionalData(stopJD)) || (!planet->hasValidPositionalData(startJD)) )
     return separations;  // empty result.

And probably this is not even what we need here? Imagine some object with orbit_good=100 (days)

You make an ephemeris for 250 days which enclose the orbit's validity range. The object would be excluded with the current solution! Maybe add a test for (startJD+stopJD)/2? Or sample every 50 days between startJD and stopJD? Or is there a limit or typical value for (stopJD-startJD)? If this is always something like 10 days, you could exclude the object if both dates are invalid.

alex-w: I've added JDE as parameter to reduce number of calls, but for second proposal I think it should be postponed to the future releases, because this subsystem should be refactored after pencil and paper works.

gzotti: OK. The inefficiency is solved. 
The next question is just what this function should do. Is this logic correct? Whenever at least one limit of the search interval is outside the validity range for the planet, the planet is discarded. I see a danger of another error. If you need to merge this now, leave a FIXME with this question for later, and maybe somebody reports an error. Else please don't merge, and let's solve this with better tests for the overlapping cases. (We may need to add some really simple date validity queries to the Planet.)

axd1967: please add the `QWarning()` logging, or some other warning mechanism. the `QWarning()` is not much asked, and will help if a user opens an issue; the log will explain what happened.

thank you!

alex-w: > The next question is just what this function should do. Is this logic correct? 

Yes, but I'll change it as first iteration of excluding the data.

gzotti:
I looks much faster, but I may have found another bug.
stellarium-007
Eclipse between Jupiter and Neptune? With more than 3° separation this cannot be an occultation. Maybe Jupiter in transit as seen from Neptune? (Goto Neptune, but seen no transit). 
This is a separate issue though. 
Also, I think shadow transit checks can safely be reduced to the respective planet's moons, not things like comet and Jupiter. Yet another separate issue.

alex-w:
I found today this issue too and this is puzzled me...

gzotti (approved):
I think the purpose of the branch is fulfilled. Newly found old weirdnesses which we now have found should be fixed in other issues/PRs. (0.21.1 finds the same things, so that's not a problem of this PR.)

alex-w:
I'm work on fix of founded issue - just testing the fix right now in this branch

alex-w:
OK, please check the fix

gzotti:
OK. No more eclipses Neptune/Jupiter. 
However, running 0.21.1 and this branch side-by-side, the shadow transit events at Jupiter differ: One finds an event for Io, the other finds Europe, or vice versa. Both are correct in what they are finding, but just seem not to find all events. The search intervals in the phenomena tab overlap but are not identical, if that matters. (It should not!)

alex-w:

PR #1885, opened 2021-09-14
Greatest Digression

gzotti:
### Description

This little branch adds azimuths and hour angles of greatest digression relevant for circumpolar objects. These are the maximally reachable azimuths.

I place these two lines of information into the RTS infostring display group, but this can be discussed, either create a new group or move elsewhere?

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 21H1
* Graphics Card: Nvidia Geforce 960M (irrelevant)
 
## Checklist:

### DIFF
--- a/src/core/StelObject.cpp
+++ b/src/core/StelObject.cpp
@@ -688,6 +688,10 @@ QString StelObject::getCommonInfoString(const StelCore *core, const InfoStringGr
 			double az=asin(cos(dec_equ)/cos(latitude)); // azimuth (eastern)
 			// TRANSLATORS: Greatest Eastern Digression is the maximum azimuth for circumpolar stars
 			QString event(q_("Max. E. Digression"));
+			// TRANSLATORS: azimuth (abbrev.)
+			QString azStr=(withDesignations? qc_("A", "celestial coordinate system") : qc_("Az.", "celestial coordinate system"));
+			// Translators: hour angle  (abbrev.)
+			QString haStr=(withDesignations? qc_("h", "celestial coordinate system") : qc_("HA", "celestial coordinate system"));
 			if (latitude<0.)
 				az=M_PI-az;
 			if (StelApp::getInstance().getFlagSouthAzimuthUsage())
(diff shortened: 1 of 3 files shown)


axd1967:
Nice.

I tested this and _at first_ there seemed to be something wrong (location +51.033148  +3.711431, see purple circle for circumpolar area ):

Edasich works fine
image


Same for Eltanin
image


Alkaid does not:

image

Then I understood this is for objects inside the _zenith passage circle_ (I added that after finding out; the Archeo plugin is an awesome plugin), not necessarily for circumpolar objects, so that latter term might better be avoided - **also in the source code**.

Or, more likely, there is a bug in the implementation and it should (logically) apply to _all_ circumpolar objects?

I don't know about the literature, but is it needed to restrict this to circumpolar objects?
If I'm correct, every object that travels above the horizon can have such extremes - the range from east to west?

image

gzotti:
Indeed. Yes, these things occur for stars with upper culmination between pole and zenith. Thanks for playing, I will fix the documentation. 
The thing is in fact limited to objects in the polar area. Only these stars have a moment of vertical climb or fall where they reach those extreme azimuths.

src/core/StelObject.cpp
```diff
@@ -677,6 +676,67 @@ QString StelObject::getCommonInfoString(const StelCore *core, const InfoStringGr
 			res += q_("Polar dawn") + "<br />";
 		else if (rts[0]<99.f && rts[2]>99.f)
 			res += q_("Polar dusk") + "<br />";
+
+		// Greatest Digression: limiting azimuth and hour angles for circumpolar stars
+		// This may fill 2 lines where RTS just says "circumpolar"
+		double dec_equ, ra_equ;
+		StelUtils::rectToSphe(&ra_equ,&dec_equ,getEquinoxEquatorialPos(core));
+		const double latitude=static_cast<double>(core->getCurrentLocation().latitude)*M_PI_180;
+		if (((latitude>0.) && (dec_equ>=latitude)) || ((latitude<0.) && (dec_equ<=latitude)))
```
alex-w, line 685: Why? On my latitude (53 degrees) I'll see objects as circumpolar with declination since 37 degrees and up to 90 degrees. In your code I'll see objects as circumpolar with declination since 53 degrees, but this is wrong.

alex-w (changes requested):
Please fix circumpolar checks

src/core/StelObject.cpp
gzotti: The code is OK, but the documentation is faulty. This only is for objects culminating between pole and zenith.

alex-w: OK, please fix the docs and merge the code :)

Issue #1546, opened 2021-03-13
Extend or remove 100 000 years date limit


gotgutgit:
Please let the user choose dates **beyond 100 000 years** in the past or in the future. I'd like to model constellation changes based on currently known proper motion data for major stars. Maybe add a warning about accuracy or a checkbox somewhere in the configuration to remove this limit.

gzotti:
Unfortunately the proper motion data are not good enough to warrant this. Even 100.000 years are too much IMHO, rather stay within 40.000. (See Appendix F in the User Guide.) There are some online programs which already show stars from Gaia which should be much better suited for that purpose.

gotgutgit:
> There are some online programs which already show stars from Gaia which should be much better suited for that purpose.

@gzotti Could you please point me at some of these programs? Actually, the best one I could find so far is Stellarium. :)
And will Stellarium include data from Gaia & Hipparcos which were used in making of this 450 000 years video? https://www.space.com/37148-orion-constellation-changes-over-450000-years.html

gzotti:
Currently I cannot find more than Google, it's not my field of immediate interest. 
As much as I'd like to have a better star catalog myself, somebody would have to actually develop it. Stellarium's development is based on voluntary work by currently 2 people who are more than busy with their own focal tasks.

PR #1888, opened 2021-09-14
star trails

gzotti:
### Description

This little branch adds an interesting "wow" effect by not blanking frames. You can configure a keyboard shortcut to toggle this, and then render star trails.

This is a bit "hacky"! One problem are extended objects: Zodiacal light, Milky Way, atmosphere, DSO images need to be switched off first. This is handled automatically and easy to solve.

The worst problem is that the scene becomes too bright too quickly when any extended objects would be plotted.  A solution would be to not just "forget" blanking, but add a slight dimming of the scene. This should then result in an afterglow effect where star trails don't get infinitely long. Maybe the dimming can be made configurable. 

Due to the nature of this method, screenshots have to be made with operating system tools, not Stellarium's own screenshot renderer.

OpenGL/Qt experts please continue on this. Change StelCore::preDraw() and implement the methods in the StelViewportFaderEffect class, or maybe directly in StelCore::preDraw().

Screenshot (221) (Klein)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 21H1; Windows 11 25H2
* Graphics Card: Nvidia Geforce 960M; Geforce RTX 3070Ti
 
## Checklist:

### DIFF
--- a/guide/ch_advanced_use.tex
+++ b/guide/ch_advanced_use.tex
@@ -819,12 +819,12 @@ In \file{config.ini}, it is stored in
 sky_background_color  = 0.000000,0.000000,0.000000
 \end{configfile}
 
-\subsection{Star Streaks}
-\label{sec:CommandLineOptions:Special:StarStreaks}
+\subsection{Star Trails}
+\label{sec:CommandLineOptions:Special:StarTrails}
 
 One of the oldest\newFeature{26.2}  and previously simplest techniques from the time of film-based night photography is the recording 
 of Earth's rotation by long-time exposure with an untracked camera. 
-This will, for example, identify the Polar Star as just not quite on the north celestial pole, but else show thousands of circular star streaks. 
+This will, for example, identify the Polar Star as just not quite on the north celestial pole, but else show thousands of circular star trails. 
 While an exposure time of many hours is possible under dark skies, when the sky is too bright by remaining twilight or light pollution, 
 the exposure time meanwhile is often limited to minutes, else the picture is just flooded with unwanted background. 
 
@@ -838,9 +838,10 @@ The trick implemented in this special mode draws one frame in ``normal'' mode, a
 which cause only smeared images like screen labels. A moderately long ``exposure'' can therefore still successfully show constellations or star labels.  
 
 This is an option probably only used by instructors or presenters and therefore not preconfigured for beginners. 
-To make use of it, you must assign a keyboard shortcut to the action ``Toggle background clearing''
+To make use of it, you must assign a keyboard shortcut to the action ``Toggle star trails''
 found in the ``Display Options'' key shortcut actions (see \ref{sec:gui:help:hotkeys}).
 
+Note that screenshots have to be made with system means while this mode is running. 
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
--- a/guide/ch_advanced_use.tex
+++ b/guide/ch_advanced_use.tex
@@ -802,6 +802,48 @@ Note that Spout use disables any multisampling setting (see Appendix~\ref{sec:co
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
+\section{Show effects} 
+\label{sec:CommandLineOptions:Special:Show}
+
+\subsection{Sky Background}
+\label{sec:CommandLineOptions:Special:ColoredBackground}
+
+Stellarium is sometimes used for entertainment or artistic music performances. 
+From this community a request was placed to allow a non-black background. We can do that, but you must revert the color yourself.
+Setting a background color is not available in the usual GUI to avoid inadvertent changes that later confuse users. 
+You can play with the sky background in the Text User Interface (see section \ref{sec:plugins:TUI:commands}, feature 5.26). 
+In \file{config.ini}, it is stored in
+
+\begin{configfile}
+[color]
+sky_background_color  = 0.000000,0.000000,0.000000
+\end{configfile}
+
+\subsection{Star Streaks}
+\label{sec:CommandLineOptions:Special:StarStreaks}
+
+One of the oldest\newFeature{26.2}  and previously simplest techniques from the time of film-based night photography is the recording 
+of Earth's rotation by long-time exposure with an untracked camera. 
+This will, for example, identify the Polar Star as just not quite on the north celestial pole, but else show thousands of circular star streaks. 
+While an exposure time of many hours is possible under dark skies, when the sky is too bright by remaining twilight or light pollution, 
+the exposure time meanwhile is often limited to minutes, else the picture is just flooded with unwanted background. 
+
+Stellarium can be made to simulate such images by a somewhat ``dirty trick'': Usually the image area is cleared between frames, 
+then the Milky Way, stars, Zodiacal light, grids, planets, atmosphere, landscape etc., and finally GUI panels are rendered into the scene. 
+By not clearing the image area, we can just add up everything that is rendered on-screen, like in a long-time exposure. 
+However, just as cars, airplanes, satellites or people with flashlights can ruin your awesome star streak photo, 
+some elements usually visible in Stellarium's sky would also create unwanted views. 
+Image a celestial grid or the Milky Way rotating with the sky, and twilight stacking all up and flooding the long-time exposure. 
+The trick implemented in this special mode draws one frame in ``normal'' mode, and then suppresses those elements and other details 
+which cause only smeared images like screen labels. A moderately long ``exposure'' can therefore still successfully show constellations or star labels.  
+
+This is an option probably only used by instructors or presenters and therefore not preconfigured for beginners. 
+To make use of it, you must assign a keyboard shortcut to the action ``Toggle background clearing''
+found in the ``Display Options'' key shortcut actions (see \ref{sec:gui:help:hotkeys}).
+
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
 \iffalse
 
 %% NOTE: We cannot provide accurate documentation on this topic at this time. 
@@ -849,7 +891,7 @@ Piper runs a web server which allows integration to \program{speechd}.
 
 With Stellarium it’s very easy to project the sky on a dome and create your own planetarium.
 
-\subsubsection{360° Projection}
+\subsection{360° Projection}
 You can use either a very expensive fisheye lens projector -- or a standard projector with a spherical mirror.
 Not only is spherical mirror projection less expensive, 
 but it produces a higher resolution image on the dome\footnote{\url{https://domeclub.com/#PxComp}},
@@ -872,7 +914,7 @@ it's quite easy to build your
 own\footnote{\url{https://docs.worldwidetelescope.org/diy-planetarium-dome/1/dome/#setting-up-the-projector-and-mirrors}} 
 -- and you can use a cheap convex safety mirror for testing before investing in a first-surface mirror.
 
-\subsubsection{Projector Specifications}
+\subsection{Projector Specifications}
 A major challenge with digital planetariums (360° cinemas) is the cross-reflection of light on the dome. 
 This is why projector \emph{contrast ratio} and \emph{colour saturation} are more important than brightness (lumens). 
 Light `bouncing' around the dome causes the planetarium image to `wash out' (lose colour) -- 
@@ -882,7 +924,7 @@ Similarly, the higher the contrast ratio, the sharper and crisper the planetariu
 4K projectors are now very affordable, allowing you to deliver the same high-quality visuals in your small
 dome as large planetariums!
 
-\subsubsection{Computer Configuration}
+\subsection{Computer Configuration}
 The best way to operate a planetarium is with a dual display arrangement, i.e., 
 computer screen as the primary (main) display, and projector as the secondary (extended) display.
 
@@ -893,7 +935,7 @@ the dome (even with no application open), use a high-resolution image of stars a
 To function as a planetarium, Stellarium should be configured to open in full screen mode on the extended
 display (see below).
 
-\subsubsection{System Configuration}
+\subsection{System Configuration}
 To minimise image distortion at the back of the dome, it’s worth carrying out the slightly complex
 \program{Meshmapper} configuration\footnote{\url{https://paulbourke.net/dome/meshmapper}}. 
 This creates a custom distortion file for your specific system (i.e., your dome
@@ -901,7 +943,7 @@ diameter, stand height, spherical mirror dimensions);
 and the image improvement will be remarkable\footnote{\url{https://domeclub.com/images/Meshmapper-1152w.jpg}}. 
 This custom distortion file (‘warp mesh data file’) can then be used with dome applications like Stellarium.
 
-\subsubsection{Stellarium Configuration}
+\subsection{Stellarium Configuration}
 To enable planetarium spherical mirror projection, the \file{config.ini} file needs to include the following:
 
 \begin{configfile}
@@ -962,7 +1004,7 @@ star_twinkle_amount = 0.50
 \end{configfile}
 
 
-\subsubsection{Further notes}
+\subsection{Further notes}
 This feature is rarely used by the current developers and cannot be tested on standard hardware. 
 See ongoing discussion on Github\footnote{\url{https://github.com/Stellarium/stellarium/issues/1914} and \\
 \url{https://github.com/Stellarium/stellarium/issues/3535}} and get involved.
--- a/guide/ch_advanced_use.tex
+++ b/guide/ch_advanced_use.tex
@@ -831,7 +831,7 @@ the exposure time meanwhile is often limited to minutes, else the picture is jus
 Stellarium can be made to simulate such images by a somewhat ``dirty trick'': Usually the image area is cleared between frames, 
 then the Milky Way, stars, Zodiacal light, grids, planets, atmosphere, landscape etc., and finally GUI panels are rendered into the scene. 
 By not clearing the image area, we can just add up everything that is rendered on-screen, like in a long-time exposure. 
-However, just as cars, airplanes, satellites or people with flashlights can ruin your awesome star streak photo, 
+However, just as cars, airplanes, satellites or people with flashlights can ruin your awesome star trail photo, 
 some elements usually visible in Stellarium's sky would also create unwanted views. 
 Image a celestial grid or the Milky Way rotating with the sky, and twilight stacking all up and flooding the long-time exposure. 
 The trick implemented in this special mode draws one frame in ``normal'' mode, and then suppresses those elements and other details 
@@ -839,7 +839,12 @@ which cause only smeared images like screen labels. A moderately long ``exposure
 
 This is an option probably only used by instructors or presenters and therefore not preconfigured for beginners. 
 To make use of it, you must assign a keyboard shortcut to the action ``Toggle star trails''
-found in the ``Display Options'' key shortcut actions (see \ref{sec:gui:help:hotkeys}).
+found in the ``Display Options'' key shortcut actions (see \ref{sec:gui:help:hotkeys}). 
+While the mode is on, tapping \key{Ctrl} resets the trails. 
+It can make a nice ending scene in a presentation: show a moderately slow moving sky, and activate it.  
+The sky will fill with curves, hopefully impressing your audience. 
+You can also utilize it when talking about the problem of satellite constellations now destroying the sky. (Use realistic view, not icons, for satellites.) 
+Here your time speed should not be too fast, else satellites will just show up as dot chain. 
 
 Note that screenshots have to be made with system means while this mode is running. 
 
--- a/plugins/AngleMeasure/src/AngleMeasure.cpp
+++ b/plugins/AngleMeasure/src/AngleMeasure.cpp
@@ -289,6 +289,8 @@ void AngleMeasure::drawOne(StelCore *core, const StelCore::FrameType frameType, 
 //! Draw any parts on the screen which are for our module
 void AngleMeasure::draw(StelCore* core)
 {
+	if(!core->getFlagClearSky())
+		return;
 	if (startPoint.normSquared()==0.0) // avoid crash on switch-on, lp:#1455839
 		return;
 	if (lineVisible.getInterstate() < 0.000001f && messageFader.getInterstate() < 0.000001f)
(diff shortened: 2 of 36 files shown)


guillaumechereau:
I tried to complete the viewport effect code, using a blending function with alpha = 0.08:
    	 GL_CONSTANT_ALPHA, GL_ONE_MINUS_CONSTANT_ALPHA

I can get this result:

image

While the trails are not very clear, this has the nice effect of not affecting the non moving parts of the screen, like the landscape.

Should I commit in this branch?

gzotti:
Yes, this looks very good for a start, thank you. I may be then able to play around with different alphas and maybe still different blending functions, and resetting the screen when view direction was changed etc. The non-affection of the landscape is very welcome!

guillaumechereau:
OK I'll commit what I have, feel free to modify it.  Unfortunately the only way to keep the landscape stable is to use a blending function of the form:    color = new * t + prev * (1 - t), and instead of the accumulating effect, we get something that looks more like a blur.  Not sure how to improve that.

The other problem here is that we apply the effect to everything, including the labels and UI, so it makes the interface clumsy.

gzotti:
Thank you, it is certainly a progress. But the blur is strange, indeed.  

I think we can auto-hide the GUI while the effect is active, or do similar things. It's mostly a show-off effect.

I wonder how alterBuffer would ever be applied in a ViewportEffect?

luzpaz:
Bump

gzotti:
Some modules are auto-inhibited now: gridlines, labels, planet and DSO hints, planet halos, Milky Way, Zodiacal light. 

Now it depends on what users expect. 
The best use is to switch off landscape and press Ctrl-Alt-C. This works for a few spectacular effects in a demo, but can run into really ugly white-outs if not applied with caution. 

I would like to see landscape being replotted and NOT run into white-out, however I am unable to see an easy solution. 

Also, changing view direction or fov changes should re-initialize (clear) the view. This would need a few new signals. Is the overhead worth it?

I cannot get @guillaumechereau 's solution to show again. :-(

10110111:
Why not just draw all the cosmic items into a separate FBO for accumulation, and blit on each frame into the main render target after rendering into this FBO?

gzotti:
Sounds like a plan :-) 
I think by now you have the better overview about which buffers exist and how they correlate with each other. (This architecture should be described somewhere!!!) It may also be related to your longer-term ideas around using physical radiance units and post-process effect filtering (blur up bright stars/planets) which should also be applied to the sky only. (And not on labels, hint icons, ... or landscapes.) I will be glad if you bring in your skills here.

The net view effect I envision in this mode should be "interactive foreground" (i.e. landscape should redraw well), stars should not run over the landscape (I tried several blending options, all but current looked bad), and on change of view direction, this "sky buffer" should reset to black (or background color). Dithering should be applied only to the final view (Currently it of course accumulates and looks bad as well). An extension could be that the buffer is dimmed instead of blanked, so that the star trails fade away after e.g. 10...60 seconds. 

If possible, gridlines or DSO markers, or even the MilkyWay/ZL could be made visible again without streaking. (I.e., not put into that "celestial" FBO. (But another FBO for MW/ZL/DSO textures/DSS/HiPS is yet another indirection, may be overkill, so IMHO it's OK to just switch them off). 


This all is meant to be for stronger systems. The "low graphics mode" running on memory limited systems should not be further encumbered with yet another FBO.

10110111:
> This architecture should be described somewhere!!!

Well, it's still gonna change when I split the rendering between layers to make labels etc. separate from physical objects and tone mapping.

> Dithering should be applied only to the final view

This must always be true, since it's a conversion from full color depth to the monitor space.

> If possible, gridlines or DSO markers, or even the MilkyWay/ZL could be made visible again without streaking. (I.e., not put into that "celestial" FBO. (But another FBO for MW/ZL/DSO textures/DSS/HiPS is yet another indirection, may be overkill, so IMHO it's OK to just switch them off).

What would a star streak represent if the Milky Way remains at one position? Sounds unphysical. If you want to simulate a long-exposure photo, all celestial objects should move and be smeared. Otherwise you could just as well draw the stars as lines (similar to orbits).

> The "low graphics mode" running on memory limited systems should not be further encumbered with yet another FBO.

The low graphics mode shouldn't receive any new graphical features IMO. Only fix the existing bugs (except the ones requiring a serious rework).

gzotti:
> > This architecture should be described somewhere!!!
> 
> Well, it's still gonna change when I split the rendering between layers to make labels etc. separate from physical objects and tone mapping.

Yes, but just as code evolves, docs can evolve and later versions can supersede current docs. If somebody wants to join right now, it's super hard without some information on what goes where. (It has always been like that, though...) 
 
> > Dithering should be applied only to the final view
> This must always be true, since it's a conversion from full color depth to the monitor space.

Sure. Just in the current state, it accumulates, so I must switch it off.

 
> > If possible, gridlines or DSO markers, or even the MilkyWay/ZL could be made visible again without streaking. (I.e., not put into that "celestial" FBO. (But another FBO for MW/ZL/DSO textures/DSS/HiPS is yet another indirection, may be overkill, so IMHO it's OK to just switch them off).
> 
> What would a star streak represent if the Milky Way remains at one position? Sounds unphysical. If you want to simulate a long-exposure photo, all celestial objects should move and be smeared. Otherwise you could just as well draw the stars as lines (similar to orbits).

The MW and other extended features could either continue rotating with the stars as they are newly plotted, or stop at the place when this mode was activated, or be smeared. In the last case, it must dimmed down however, or the sky will be white within seconds. Therefore I currently switch it completely off.

> 
> > The "low graphics mode" running on memory limited systems should not be further encumbered with yet another FBO.
> 
> The low graphics mode shouldn't receive any new graphical features IMO. Only fix the existing bugs (except the ones requiring a serious rework).

OK for me.

dromeuf:
Hi, I'm looking to make an illustration with a long exposure effect of circumpolar star trails. Does this feature work in Stellarium 24.2 ? In the documentation I can only find the trail planets. Kind regards,

dromeuf:
Georges, I did read the thread you created in 2021. I didn't know whether you had completed this experiment. Thank you for all your great work on the Stellarium project.

dromeuf:
Further to our discussion, here is the solution that I found last night to illustrate the circumpolar rotation at the northern horizon of the site studied and the movement in the sky of the star Vega in 4000 BC which rubs the horizon of the place of observation :
- stop animation
- set the date
- capture screen with capture tools
- change the date (min)
- capture screen.... n times...
- Use StarStaX https://markus-enzweiler.de/software/starstax/ for stacking capture screen

IllustrationStarsTrailsStellarium

I read in the latest Stellarium 24.3 user guide that the representation of the proper motion is good to within 100,000 years but that the ‘Carte du Ciel’ software will be more accurate than Stellarium. In 2021 Georg you indicated in #1546 + or - 40,000 years. By comparing with ‘Carte du Ciel’ on the star Vega, I have the impression that the proper motions of Stellarium are those of the Hipparcos 2007 catalogue. In 6,000 years, Vega's proper motion is 35'. When I calculate the angular distance between the coordinates for the date between Stellarium and ‘Carte du ciel’, I find a difference of about 4'. I don't have the explanation because I haven't had time to examine the details of the calculations made by the 2 software packages: precession model, mean coordinates, etc.

gzotti:
I did not intend to say Stellarium is accurate within &pm;40.000 years. The settable date range is &pm;100.000 just to have round numbers. The available models concentrate on present times. A few like precession declare date range (exceeding the 100.000), no information is given for nutation, so we invented the limit given in the Guide. VSOP87 is recommended for -4000...+8000, the DE[34]31 solutions -13.000...+17.000. It should be clear that linear proper motion is not enough for accurate positions over tens of thousands of years, still the _idea_ of motion is often more important than the actual figure change. Just don't trust it to the last decimal place! (Don't ask us to reduce the number of displayed digits in prehistory!)  And in a previous (rejected) PR a suboptimal way to fix the current problem can still be seen. 

Currently a new star catalog is in preparation (#3982). therefore I will not compare anything that involves the current, known-to-be-somewhat-flawed solution.

dromeuf:
In my case and for the purpose mentioned, the precision is sufficient for visual illustration. Thanks again and congratulations on all the work done by the Stellarium team. It's also good news if your synergy focuses on updating star catalogs. I've read #3982

gzotti:
We're getting somewhere! Run, switch off atmosphere, set some timelapse and press Ctrl-Alt-C. 
Not sure whether I should remove the attempted ViewportEffect.

gzotti:
The basic appearance is to freeze the last frame (including lines, Milky Way, even atmosphere, and add star streaks as if there was no atmosphere.  I think it has its applications in demos or teaching.  

Technically, this is now continuing my first attempt of "forgetting" the framebuffer reset. However, to allow a nicely filled background, one "regular" frame is filled, and only starting with the next draw cycle the "large" modules (MilkyWay, Zodiacal Light, gridlines, etc.) are suppressed. Any movement, zoom or time jump resets the scene. 

Closing Dialogs should also reset the scene, but only Search dialog seems to behave. Not sure about the others.
EDIT: Solved.

gzotti:
Any comments here? I'd probably take away Guillaume's Viewport Effect as I have not continued that path. Else, it works well for me. IMHO this is a feature best used by instructors either demonstrating daily motion, or just as closing show effect. By default you cannot accidentally trigger it but have to configure your hotkey.

alex-w:
> Any comments here? I'd probably take away Guillaume's Viewport Effect as I have not continued that path. Else, it works well for me. 

OK

> IMHO this is a feature best used by instructors either demonstrating daily motion, or just as closing show effect. By default you cannot accidentally trigger it but have to configure your hotkey.

Hmm... what about some "presets", like disabling atmosphere and SSO, when star trails effect is enabled?

gzotti:
Try it. As described above, Atmosphere, Milky Way, DSO images are rendered just once, after that only stars and SSO are rendered and accumulate the long-time exposure. This avoids a white-out that else happens within seconds.

alex-w:
> Try it. As described above, Atmosphere, Milky Way, DSO images are rendered just once, after that only stars and SSO are rendered and accumulate the long-time exposure. This avoids a white-out that else happens within seconds.

At the moment it draw the trails for artificial satellites too - IMHO this shouldn't be visible to users.

gzotti:
You can actually nicely illustrate the nightmare for photographers.
Screenshot (12)

For Satellite icons and orbit/shadows: indeed these should be suppressed.

Meh - disabling icons with the switchover is not possible, a first image would show them. Switching off icons is left to the operator.

alex-w (approved):

10110111:
In star trails mode the bottom of the Guereins landscape gets overexposed.

gzotti:
Ah, thanks for catching this. I never use this but sure, old_style needs to work.

10110111:
Generally, the implementation seems to be too invasive. Can't we just draw the few items that we want to accumulate after the first frame containing everything has been drawn? I.e. decide whether we want to draw landscape, grids, etc. at the level of module choice, rather than put dozens of `if(flagClearSky)` everywhere?

gzotti:
Well, I just reverted this "global" selection in StelApp in the "forall(modules):draw()" loops to not have too many exception paths and have one common pattern: a flag check at begin of each module's draw().  

It turned out it has to be more invasive than wanted in 2021, but that's the difference between "I'd like to have" and "what has to be done for it" (to avoid ugliness).

I can re-enable the draw exclusion in StelApp based on module name string comparison to silence most affected modules if you think this is the better way, but I fear that's slower than checking a Boolean flag.  For stars and planets with all their labeling it remains more complicated in any case.

10110111:
No, the hybrid way (as was before your last commits) is of course even
worse than what we have now.
My idea was not to filter, but rather to only call the modules we are
interested in. But I guess the current intermixture of the scene and labels
doesn't let us do it elegantly...

gzotti:
So, do you have a better solution for now? In 2023 you proposed changing the overall architecture. This has not happened, so I continued like this.

10110111:
> In 2023 you proposed changing the overall architecture.

Oh yes I remember. I still want to do it, but have lots of other things to do too.

> So, do you have a better solution for now?

Let me look at your implementation for a bit. If I don't come up with anything better, I guess it will do for now.

BTW, do you intend to squash all these commits? There are many fail-and-fix cases here that would better be squashed together.

10110111:
Somehow the bottom part of the old-style landscape is transparent for the Moon and the Sun:

Screenshot_2026-05-16_22-23-51

Internally-created screenshots also don't behave as one might expect:

stellarium-020

gzotti:
Actually I think this can be squashed into one new minor feature. 
Landscapes: Oops... will have a look now.
Screenshots: SUG 6.5.2, last sentence. If you find an easy way to do that, fine.

gzotti:
Somehow I don't find the decisive difference in the shader code between decor and bottom parts. What works for the sides does not work at the bottom panel! I can only make it work in lowGL mode which also looks OK for me.

gzotti:
@10110111 Are you OK with switching to lowGL, or can you identify the problem?
(With the lowGL solution, I cold revert the shading var tweaks.)

10110111:
So I have fixed the ground issue now.

There are some glitches when panning around during daytime with star trails enabled: the atmosphere flickers, getting sometimes overexposed, other times completely absent. On disabling star trails the atmosphere doesn't immediately reappear for some reason. (I'm using ShowMySky with `atmosphere_resolution_reduction` enabled if this matters).

Generally, the current implementation is hard to follow. Not least because of the confusing name of the flag, but also because it's hard to even tell which objects are supposed to be smeared over the screen. (Is it only stars, planets, and satellites?)

Maybe a better way would be to do as follows:

* Prepare a framebuffer for accumulation of the items we want to see smeared.
* Make (and save) a list of pointers to the modules that should appear in this framebuffer (extracted from the list with the right call order).
* Make (and save) two lists of the remaining modules (first for the ones before the smeared modules, second for those after).
* When rendering:
  - draw the first batch of modules (those before the smeared ones)
  - switch FBO to our accumulation buffer
  - draw the modules we want to smear
  - switch back to the target FBO
  - blit the accumulation buffer
  - draw the rest of modules
* On panning clear the accumulation buffer.
* On screen resize also resize and clear the accumulation buffer.

Of course, the accumulation buffer doesn't have to be created and managed until the user enabled star trails.

This way only the handful of smeared modules will have to deal with the star-trail mode (disabling their labels), the rest will work just as usual (except maybe for the atmosphere that we may want to suppress in this mode).

As a bonus, we'll get back screenshots (except those with a custom size), and avoid the GUI artifacts (currently the bottom panel with the time rate button leaves trails too).

gzotti:
> So I have fixed the ground issue now.

Many thanks. I don't know where blindness struck me. 

> There are some glitches when panning around during daytime with star trails enabled: the atmosphere flickers, getting sometimes overexposed, other times completely absent. On disabling star trails the atmosphere doesn't immediately reappear for some reason. (I'm using ShowMySky with `atmosphere_resolution_reduction` enabled if this matters).

Yes, admittedly it's not perfect, just good enough for some effects. I do not understand why the atmosphere sometimes accumulates 2-3 frames (running into white-out) before being silenced, and why it then sometimes remains off. 
 
> Generally, the current implementation is hard to follow. Not least because of the confusing name of the flag, but also because it's hard to even tell which objects are supposed to be smeared over the screen. (Is it only stars, planets, and satellites?)

The name just describes what happens technically, not why. I hope the documentation string is complete enough for understanding. 

Everything that causes large-area brightening must be disabled or the screen turns white within seconds. In 2021-23 my idea was to dim down the accumulated image to limit brightness, which would have also meant limited trail length. Guillaume's filter effect went into this direction but also blurred the view.  In the end only stars (and other star-like extras: Novae, Supernovae), planets, satellites remained. But the other modules should be visible at the onset of the drawing, and the landscape must still cover the ground (if enabled). 


> Maybe a better way would be to do as follows:
> [...]

I can roughly follow this in theory, but it vastly exceeds my OpenGL proficiency, sorry. I wanted to have this feature finally available showing something not the least for satellite annoyance awareness. 

> This way only the handful of smeared modules will have to deal with the star-trail mode (disabling their labels), the rest will work just as usual (except maybe for the atmosphere that we may want to suppress in this mode).

As said, the other modules including atmosphere (awesome twilight composites!) should draw one initial snapshot. E.g., constellations can show stick figures from which star lines grow. Of course, with your mode it would mix "current" view and recorded trails, that's another mode that promises to look interesting. Can we think of this as future version 2 of this feature?
 
> As a bonus, we'll get back screenshots (except those with a custom size), and avoid the GUI artifacts (currently the bottom panel with the time rate button leaves trails too).

I keep the menus visible or switch them off entirely (Ctrl-T). Pressing Ctrl resets the scene.

10110111:
I don't quite get the logic of sky clearing. Why is there this magic number 3? Why not any other? Besides, `flagClearSkyOnce` is not actually a flag, since it ranges from 0 to 3.

And, I think `flagClearSky` should be renamed to something like `isAccumulatingStarTrails` (with inverse condition), because it's not about sky clearing generally, but rather a hack exclusively for star trails (given that only a few modules actually draw something when this flag is reset).

gzotti:
The naming grew historically. I thought a second flag would do for resetting, but it showed I need a 3-2-1-0 countdown to handle the initial 1-shot drawing completely. So, the counter could be renamed, ok. How does `frameClearingResetCounter // for Star Trail Effect ` sound? 
The flagClearSky, again, was the intention in 2021 to signal that at frame start the frame should be reset, as normal case. When disabled, frame is not reset, unless (flag)ClearSkyOnce is not 0, in which case there is some localized logic to handle the reset sequence.   If you think this should be renamed in some inverted sense and handled with inverted logic to be better readable, OK for me.

10110111:
But still, can you explain the idea behind this 3-2-1-0 sequence? Why can't this be done in only two stages (everything once, then accumulation of the few items)?

gzotti:
Reducing (Starting) the counter back to 2 causes overly frequent white-out with the daylight sky when pressing Ctrl to reset the trails.

alex-w:
@gzotti @10110111 any news?

gzotti:
Sorry, conference week. It is as described in the User Guide a hackish approach, but it works usually. It will probably only be found by users that stumble over it while reading the advanced options in the SUG. Separating that into several buffers may be the technically cleaner way, but is beyond my knowledge. I would like this showpiece effect "in", obviously accepting improvements if e.g. @10110111 finds time. I could also add yet another manually modified config option to hide the feature for beginners.

10110111:
If I don't find time to add any improvements here by the end of the next week, I guess this can then go in as is. Otherwise it'll have too small time to be tested before release.

gzotti:
The occasional atmo white-out or absence may be caused by the overall brightness estimator getting confused. Must have another look. Weekend, hopefully...

10110111:
This crashes unconditionally in debug mode with assertion failure "singleton" in `StelApp.hpp:132`:

```
#8  0x00007ffff4ec1366 in QMessageLogger::fatal
    (this=<optimised out>, msg=0x7ffff522aad8 "ASSERT: \"%s\" in file %s, line %d")
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/corelib/global/qlogging.cpp:883
#9  0x00007ffff4ebf450 in qt_assert
    (assertion=<optimised out>, file=<optimised out>, line=<optimised out>)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/corelib/global/qassert.cpp:105
#10 0x0000555555fca49f in StelApp::getInstance ()
    at /home/ruslan/src/stellarium/src/core/StelApp.hpp:132
#11 0x000055555644053f in StelRootItem::setSize (this=0x555558afbdd0, size=...)
    at /home/ruslan/src/stellarium/src/StelMainView.cpp:381
#12 0x000055555642fa99 in StelMainView::resizeEvent (this=0x7fffffffd740, event=0x7fffffffcee0)
    at /home/ruslan/src/stellarium/src/StelMainView.cpp:725
#13 0x00007ffff6df750e in QWidget::event (this=0x7fffffffd740, event=0x7fffffffcee0)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:9153
#14 0x00007ffff6e5d866 in QFrame::event (this=0x7fffffffd740, e=0x7fffffffcee0)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/widgets/qframe.cpp:521
#15 0x00007ffff4f878da in QCoreApplicationPrivate::sendThroughObjectEventFilters
    (receiver=receiver@entry=0x555558a518a0, event=event@entry=0x7fffffffcee0)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/corelib/kernel/qcoreapplication.cpp:1243
#16 0x00007ffff6d9d48f in QApplicationPrivate::notify_helper
    (this=<optimised out>, receiver=0x555558a518a0, e=0x7fffffffcee0)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qapplication.cpp:3294
#17 0x00007ffff4f8add8 in QCoreApplication::notifyInternal2
    (receiver=0x555558a518a0, event=0x7fffffffcee0)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/corelib/kernel/qcoreapplication.cpp:1106
#18 0x00007ffff4f8ae1d in QCoreApplication::sendEvent
    (receiver=<optimised out>, event=<optimised out>)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/corelib/kernel/qcoreapplication.cpp:1546
#19 0x00007ffff6df0ebe in QWidgetPrivate::sendPendingMoveAndResizeEvents
    (this=this@entry=0x5555589a6360, recursive=recursive@entry=false, disableUpdates=<optimised out>, disableUpdates@entry=false)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:7982
#20 0x00007ffff6df4ab7 in QWidgetPrivate::show_helper (this=this@entry=0x5555589a6360)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:8033
#21 0x00007ffff6df55c3 in QWidgetPrivate::setVisible (this=0x5555589a6360, visible=<optimised out>)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:8394
#22 0x00007ffff6df4f9e in QWidgetPrivate::showChildren
    (this=<optimised out>, spontaneous=<optimised out>)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:8482
#23 0x00007ffff6df4ad3 in QWidgetPrivate::show_helper (this=this@entry=0x555558d17910)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:8039
#24 0x00007ffff6df55c3 in QWidgetPrivate::setVisible (this=0x555558d17910, visible=<optimised out>)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:8394
#25 0x00007ffff6de337b in QWidget::showFullScreen (this=0x7fffffffd740)
    at /usr/src/qt6-base-6.9.2+dfsg-1ubuntu1/src/widgets/kernel/qwidget.cpp:3056
#26 0x0000555556438136 in StelMainView::setFullScreen (this=0x7fffffffd740, b=true)
    at /home/ruslan/src/stellarium/src/StelMainView.cpp:1500
#27 0x0000555555fc65f0 in main (argc=1, argv=0x7fffffffdaf8)
    at /home/ruslan/src/stellarium/src/main.cpp:477
```

10110111:
> This crashes unconditionally in debug mode

Fixed by 14c74c47221a9cc43d7d25671854247871bb9fe9.

gzotti:
Thanks!

Issue #1804, opened 2021-07-18
satellite plugin: ISS dot "disappears" behind Moon


axd1967:
### Actual Behaviour

![image](https://user-images.githubusercontent.com/3529789/126066841-20c361ed-8919-4b5f-925f-1278d3842e02.png)

![image](https://user-images.githubusercontent.com/3529789/126066852-73054174-ceb2-46f6-8f26-a919a0f86969.png)

### System
* Stellarium version: 0.21.1
* Operating system: 
* Operating System: Kubuntu 20.04
KDE Plasma Version: 5.18.5
KDE Frameworks Version: 5.68.0
Qt Version: 5.12.8
Kernel Version: 5.4.0-77-generic
OS Type: 64-bit
Processors: 8 × Intel® Core™ i7-10510U CPU @ 1.80GHz
Memory: 15.3 GiB of RAM

### Logfile 
If possible, attach the logfile `log.txt` from your user data directory. Look into the Guide for its location.

alex-w:
@axd1967 please see screenshot 2

axd1967:
@gzotti @Atque @treaves I please help me. I find the behaviour of @alex-w extremely abusive.

for_wishlist

alex-w:
> @gzotti @Atque @treaves I please help me. I find the behaviour of @alex-w extremely abusive.

Really? Alex, please see your screenshot and read your report:
> ISS dot "disappears" behind Moon

P.S.  I've marked ISS especially for you on your screenshot
![stellarium-iss](https://user-images.githubusercontent.com/88731/126121652-58acff15-a666-432a-ae39-491e8de36494.jpg)

Now please explain where I was "extremely abusive"

gzotti:
Absolutely what we intended to have when somebody asked to show a transiting black object in front of the Lunar disk.

axd1967:
(The abuse is that I have eyes and can see the label, yet you point it out to me as if I'm blind.)

OK, found the code here:
https://github.com/Stellarium/stellarium/blob/9af5e2eeacc35cd2959533110eb97e016122546e/plugins/Satellites/src/Satellite.cpp#L933-L935

The related issue is https://github.com/Stellarium/stellarium/issues/1186 and was implemented in https://github.com/Stellarium/stellarium/compare/f5af8d9c2a7d21f25690a6406ba0decc860ea41b...6eca0ffd88759489510d097a3a31fea9cbf2ad13

This is the **intention** of the current code:
[...]

PR #1906, opened 2021-09-22
include vendor branching strategy for developers

axd1967:
### Description
See https://github.com/Stellarium/stellarium/discussions/1856

### DIFF
--- a/VENDOR_BRANCHING.md
+++ b/VENDOR_BRANCHING.md
@@ -1,0 +1,67 @@
+# Vendor branching : importing external artifacts
+
+Stellarium will often be enriched with text/data/source files that are **copy-pasted** from places outside of Stellarium.
+
+Examples:
+
+- geonames data (https://www.geonames.org/recent-changes.html), https://github.com/Stellarium/stellarium-data/tags
+- quasar data (https://github.com/Stellarium/stellarium/blob/master/plugins/Quasars/util/quasars.tsv)
+- Almagest data (minor fixes, of course - this is essentially frozen data)
+- various libraries (https://github.com/Stellarium/stellarium/tree/master/src/external)
+
+The problem with copy-pasting external data is **code (or data) rot** (and also: "*copy-paste is evil!*").
+
+The external version of such artifacts will often continue to evolve, but these modifications obviously will not be reflected in our repository, thus leading to artifacts slowly getting totally outdated (hence the term "code rot"). 
+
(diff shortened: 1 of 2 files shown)


alex-w:
In mid-summer I said that I'll disregard all future Alex's activity (reports, proposals, comments and suggestions) due his abusive behavior, but probably I should make an one exception right now and leave my comment here.

This proposal is strongly inappropriate for me as a maintainer and as developer, because this strategy introduces many problems and zero solutions, and this PR with "strategy docs" should be declined. A complete disregard of the opinion of the core developers and real maintainers of the Stellarium package isn't good behavior and experience.

axd1967:
for anyone interested in "abusive behaviour" please see https://github.com/Stellarium/stellarium/issues/1804#issuecomment-885693896.

this proposal solves recurring issues that are explained in detail in https://github.com/Stellarium/stellarium/discussions/1856.

axd1967:
"don't fix what ain't broken"... but there is no rule that prohibits improvements. in fact, that's part of Agile behaviour: continually reviewing and improving ways of work. and in this case, there are already signs that Stellarium will benefit from such a procedure. also, there is no obligation to use it; but there is a clear need to be aware of it.

VENDOR_BRANCHING.md
```diff
@@ -7,6 +7,7 @@ Examples:
 - geonames data (https://www.geonames.org/recent-changes.html), https://github.com/Stellarium/stellarium-data/tags
 - quasar data (https://github.com/Stellarium/stellarium/blob/master/plugins/Quasars/util/quasars.tsv)
 - Almagest data (minor fixes, of course - this is essentially frozen data)
+- HTC algorithms (Helene, Telesto, and Calypso (Lagrangian satellites of Dione) - taken from ftp://ftp.imcce.fr/pub/ephem/satel/htc20/htc20.f ? )
```
gzotti, line 10: Only in the unlikely case that we directly add their unmodified Fortran code to the project...

axd1967: that's called Murphy's Law :-)

(and also, when one never tries, one will never become proficient in the pattern, so better try now than never)

axd1967:
I'm clearly not really respected here, despite years of professional experience.

(@treaves @xalioth @Atque @Jocelyn1109 @guillaumechereau @A-j-K @abalkin )

axd1967:
(related to [vendor_branching](https://github.com/Stellarium/stellarium/issues?q=vendor_branching))

PR #1919, opened 2021-09-24
Fixed Chinese constellation boundaries.

sunshuwei:
Fixed Chinese constellation boundaries.

Issue: when the "Select single constellation" option is selected, the constellation boundaries will be misaligned and and the adjacent constellation will be displayed.


alex-w (approved):
Thanks!

PR #1959, opened 2021-10-09
Fix stellar appearance of lunar occultation

### LINKED ISSUE #1924: Don't draw stellar halos if the star is obscured
Now that stellar aberration is implemented and we have very accurate lunar positions in relation to stars, it might be advisable to hide an occulted star's halo completely when the Moon or whatever obscures it. The screenshot below shows an occultation of 19 Piscium by the Moon, and the star's halo is still visible, despite the star actually being hidden behind the lunar disc.

stellarium-112

I think this has been up for discussion earlier, but there is no formal report of it.

gzotti:
### Description

Currently stars are drawn as "halo textures". When occulted by the Moon (or other planet, but it's most obvious here!), a star should instantly vanish (or light up again) when the moon covers its centre.  

The current attempt here is a way not to do it, though! I tried to set up a moon-sized SphericalCap and exclude stars from rendering. This comes close, but as the Moon is rendered as tessellated sphere (polygonal mesh), it may appear too small by a few arcseconds. 

The real way to go would probably involve rendering the Moon first with a stencil buffer, then drawing all stars except those hidden by the stencil. It may even be useful to draw the complete SolarSystem with stencil buffer before rendering the stars, to catch the effect in all cases. 

Thinking of it, we use a Z buffer already to mix planets and moons. So, just draw solar system before stars, and test Z buffer for star rendering. 

Any OpenGL experts here to take over? 

Fixes #1924

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Nvidia Geforce 960M 

## Checklist:

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -3767,7 +3767,7 @@ float Planet::getVMagnitude(const StelCore* core, double eclipseFactor) const
 double Planet::getAngularRadius(const StelCore* core) const
 {
 	const double rad = (rings ? rings->getSize() : equatorialRadius);
-	return std::atan2(rad*sphereScale,getJ2000EquatorialPos(core).norm()) * M_180_PI;
+	return std::asin(rad*sphereScale/getJ2000EquatorialPos(core).norm()) * M_180_PI;
 }
 
 
--- a/src/core/modules/StarMgr.cpp
+++ b/src/core/modules/StarMgr.cpp
@@ -46,6 +46,7 @@
 #include "StelUtils.hpp"
 #include "StelHealpix.hpp"
 #include "StelObject.hpp"
+#include "SolarSystem.hpp"
 
 #include <QGlobalStatic>
 #include <QTextStream>
@@ -1380,6 +1381,20 @@ void StarMgr::draw(StelCore* core)
 	// Prepare a table for storing precomputed RCMag for all ZoneArrays
 	RCMag rcmag_table[RCMAG_TABLE_SIZE];
 	const float starStreakScale = core->getFlagClearSky()? 1.0f:0.6f;
+
+	// Try to filter away stars hidden by the Moon!
+	bool filterMoon=false;
+	PlanetP moon = GETSTELMODULE(SolarSystem)->getMoon();
+	Vec3d moonPos=moon->getJ2000EquatorialPos(core);
+	const double moonRadius = moon->getEquatorialRadius() * moon->getSphereScale();
+	double angularSize = atan2(moonRadius, moonPos.norm());
+	moonPos.normalize();
+	SphericalCap moonCap(moonPos, cos(angularSize));
+	for (auto  cap : viewportCaps)
+	{
+	    if (cap.intersects(moonCap))
+		filterMoon=true;
+	}
 	
 	// Draw all the stars of all the selected zones
 	for (const auto* z : std::as_const(gridLevels))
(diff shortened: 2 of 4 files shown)


gzotti:
@10110111 maybe you can look into this one? (See also #1924). 

The solution attempted here is not good, as the spherical cap misses the lunar edge by potentially a few pixels. 

Should we invert plotting solarsystem and stars and plot stars when their locations are not covered in the Z buffer? (Caveat: semitransparent rings of Saturn and Uranus) The "overbleeding" effect mentioned by @guillaumechereau in #1924 can indeed even look better than covering part of the star "halo" by the Moon. 

Attn, this may also need to change drawing the other object classes from various plugins (QSR, PSR, SN, ...)

Other ideas?

10110111:
Ideally, I'd go a very unusual (and longish) route to solve this: render the whole scene into a floating-point framebuffer in physical units of brightness (i.e. with correct relative brightness, sizes, and without halos), and then apply a glare effect, like e.g. a Gaussian-blurred version on top of the initial render. 

All the other ways like using the stencil buffer, occlusion queries etc. for solving this are just workarounds for a rendering pipeline that doesn't make much physical sense.

gzotti:
Sounds interesting, but isn't this too much for the smaller systems? I mean, the only effect we want to avoid here is the gradual visibility of the star halos as long as the actual star is hidden. The non-stars should not have a glare effect added as this visually only blurs the image, and the test for covering will test the pinpoint star position (1 pixel) only. This should also work with the geometric Moon model you showed elsewhere where we could see the mountains along the Lunar edge. 

I wonder about the true cause of the current mismatch. Is this only the tessellation (inscribed polygonal edge vs. circle) mismatch or is there another reason?

10110111:
> Sounds interesting, but isn't this too much for the smaller systems?

Might be too much for Raspberry indeed.

>  I mean, the only effect we want to avoid here is the gradual visibility of the star halos as long as the actual star is hidden.

I also want to replace the fake glare of the Moon and the Sun, which become incorrect in various circumstances like e.g. upscaled the Sun or crescent Moon. This is why I prefer this solution that would _automatically_ solve these problems.

> The non-stars should not have a glare effect added as this visually only blurs the image

If done correctly, this won't blur the image. The blurring kernel is supposed to have a very low-amplitude tail, which will  only become noticeable on very bright targets (which the stars should be I think, otherwise they wouldn't look so fat).

> I wonder about the true cause of the current mismatch.

The mismatch where? In this PR? I didn't look much into it, but the tesselation should certainly introduce a mismatch between calculation against a sphere vs rendering.

gzotti:
> > Sounds interesting, but isn't this too much for the smaller systems?
> 
> Might be too much for Raspberry indeed.

If it else works everywhere (Intel?), it could also become a switchable option. (classic star halo/spikes vs. modern: glare simulation)

> > I mean, the only effect we want to avoid here is the gradual visibility of the star halos as long as the actual star is hidden.
> 
> I also want to replace the fake glare of the Moon and the Sun, which become incorrect in various circumstances like e.g. upscaled the Sun or crescent Moon. This is why I prefer this solution that would _automatically_ solve these problems.

OK. Replacing the large "halo" textures would also be fine. 
 
> > The non-stars should not have a glare effect added as this visually only blurs the image
> 
> If done correctly, this won't blur the image. The blurring kernel is supposed to have a very low-amplitude tail, which will only become noticeable on very bright targets (which the stars should be I think, otherwise they wouldn't look so fat).

OK. But consider generalisation or extensions to have some spiky pattern as well. Happy to see a technically advanced solution to this. 

> > I wonder about the true cause of the current mismatch.
> 
> The mismatch where? In this PR? I didn't look much into it, but the tesselation should certainly introduce a mismatch between calculation against a sphere vs rendering.

Yes, if you build this PR you see that stars often vanish just outside the Lunar edge. I suspect that this is just because the tessellated Moon's circumference is an n-gon inscribed into the "spherical cap" used for testing occultation that should have the same geometrical diameter, which is why this solution does not work well.

ElonVampire:
收到！

Issue #1283, opened 2020-09-25
screenshot file naming logic is bugged


axd1967:
This has been rejected as a bug in #1180 , yet I call on any developer that ever touches that code area, to review the code.

The code in question is following

https://github.com/Stellarium/stellarium/blob/23078c5912cd61129f72e1f4ba7489af38330299/src/StelMainView.cpp#L1680-L1685

This code contains internal logic errors as a result of a sloppy choice of magic numbers / hardcoded related constants.

Why limit to 10000? why not 10_0001? or 10_000_000? There is no immediate need for this limit. The full unsigned 32bit range is fine (although a time based range such as Unix time is a superior choice that would not break the existing implementation).

Meanwhile, the code is internally contradicting itself (counting to 100000 but limiting to a field width of 3 digits will cause sorting issues).
The code will forcibly (and silently) overwrite an existing screenshot once rolling past 100000 . a `while` loop over an `unsigned int` makes this problem not go away, but _far_ less likely (even if that also is Y2K reasoning).

The logic is also smelly: when the user moves some of the screenshot files, gaps are opened in the file names which will be filled by chronologically later screenshots, but alphabetically earlier (gap fillers). This will cause subtle sorting problems. _(For me an indication that the use of an index is smelly. Call it an opinion...)._

## todo

- [ ] raise the limit or get rid of the limit and continue from the largest existing number retrieved from the directory, bearing in mind that an (unlikely, yet predictable) overflow can occur. unless all numbers are in use, the routine will find the next usable file name (but again, this is a waste of processing, see Unix time instead)
- [ ] use a field width that matches the counter (10 digits when going for an unsigned 32bit solution)
- [ ] for readability, make these constants rather than literals 
- [ ] if still needed, optimize the expression `shotDir.filePath() + "/" + screenShotPrefix` as new variable (as it is invariant wrt `j`). this might probably be further optimized by including the suffix `+ "." + screenShotFormat` and using a parameter to insert the found index.

A follow-on solution is to replace the numbering with Unix time, while searching for a solution (64 bit? ) for [Epochalypse](https://en.wikipedia.org/wiki/Year_2038_problem).

axd1967:
Additionally, I discovered in https://github.com/Stellarium/stellarium/issues/1180#issuecomment-699004723 that the index suffix is always added, also when invoking the screenshot function from scripting.

An improvement is to allow the caller to skip the logic (eg a flag or new argument `generate_index = true` ?)

[...]

PR #1969, opened 2021-10-14
new filter : sats not in any group

axd1967:
### Description
I want to find satellites for which no group has been assigned. e.g. SKYMED series are - for some reason - not all assigned to the visual group, so it is handy to find which sats are not assigned.

image

### Type of change

### How Has This Been Tested?

**Test Configuration**:
KDE Plasma Version: 5.18.5
Qt Version: 5.12.8
Kernel Version: 5.4.0-88-generic
OS Type: 64-bit
Memory: 15.3 GiB of RAM

## Checklist:


plugins/Satellites/ChangeLog
```diff
@@ -1,3 +1,5 @@
+include "no group assigned" secundary filter
```
gzotti, line 1: add version/date and fix typo "secondary".

plugins/Satellites/src/Satellite.hpp
```diff
@@ -117,13 +125,13 @@ class Satellite : public StelObject
 	enum OptStatus
 	{
 		StatusOperational		= 1,
```
gzotti, line 127: I think it'd be prettier to start this list with StatusUnknown=0 and else remove the explicit numbers.

plugins/Satellites/src/Satellites.cpp
```diff
@@ -288,7 +288,7 @@ bool Satellites::backupCatalog(bool deleteOriginal)
 		return false;
 	}
 
-	QString backupPath = catalogPath + ".old";
+	QString backupPath = catalogPath + ".old";  // FIXME replace with datetime stamp as generic routine to backup files
```
gzotti, line 291: If you add a date to the old filename, also add a method to at least warn the user to manually remove the 50 old files that may appear soon and clutter your user dir. And add documentation about how to use old files.

plugins/Satellites/src/Satellites.hpp
```diff
@@ -60,8 +60,7 @@ the SGP4/SDP4 algorithms (J.L. Canales' gsat library).
 
 <i>Name and identifiers</i>
 
-Each satellite has a name. It's displayed as a label of the satellite hint and in the list of satellites. Names are not unique though, so they are used only
-for presentation purposes.
+Each satellite has a name. It's displayed as a label of the satellite hint and in the list of satellites. Names are not unique though, so they are used only for presentation purposes.
```
gzotti, line 63: Please don't use over-long lines where not really needed (e.g. some table views). See online view on Github. It is easier to review when the lines are fully visible in the box.

plugins/Satellites/src/gui/SatellitesDialog.cpp
```diff
@@ -460,7 +463,7 @@ void SatellitesDialog::updateSatelliteData()
 	Q_ASSERT(SatellitesMgr);
 	Vec3f mColor, oColor, iColor;
 
-	// set default
+	// set default // XXX oy-pasta
```
gzotti, line 466: Huh?

plugins/Satellites/src/gui/SatellitesDialog.cpp
```diff
@@ -1240,7 +1247,7 @@ void SatellitesDialog::setRightSideToROMode()
 	ui->periodLineEdit->setEnabled(false);
 	ui->periodLineEdit->setText(QString());
 
-	// set default
+	// set default // XXX copy-pasta
```
gzotti, line 1250: ?

plugins/Satellites/src/gui/SatellitesDialog.cpp
```diff
@@ -1240,7 +1247,7 @@ void SatellitesDialog::setRightSideToROMode()
...
 	buttonMarkerColor = QColor(QColor::fromRgbF(0.4, 0.4, 0.4));
```
gzotti, line 1251: It's actually possible to remove the wrapper. Just 

`buttonMarkerColor = QColor::fromRgbF(0.4, 0.4, 0.4);
`
etc.

plugins/Satellites/src/gui/SatellitesDialog.hpp
```diff
@@ -163,7 +165,7 @@ private slots:
 #endif
 
 	Ui_satellitesDialog* ui;
-	bool satelliteModified;
+	bool satelliteModified; // currently not used
```
gzotti, line 168: You can comment this away then. Also its actual (although useless) uses in SatellitesDialog::updateSatelliteData() and in the constructor.

plugins/Satellites/ChangeLog
axd1967: ### The plugin vs. catalog version string is a source of confusion

Also, there are **two** _catalog_ versions for the format: `0.12.0` and `0.12.4` ; maybe a small oversight?

The _JSON file format version_ and  _plugin_ version are not immediately related. One can change without the other changing, _sometimes_ one _does_ influence the other.

https://github.com/Stellarium/stellarium/blob/31611416a8f02163bf42b05ee80c3b670ef992c7/plugins/Satellites/CMakeLists.txt#L1

https://github.com/Stellarium/stellarium/blob/31611416a8f02163bf42b05ee80c3b670ef992c7/plugins/Satellites/src/Satellites.cpp#L85

https://raw.githubusercontent.com/Stellarium/stellarium/master/plugins/Satellites/resources/satellites.json contains _two_ version strings??

**Proposition:** 
Separate _plugin_ version shown in the plugin dialog (make it `1.1.0`) from the catalog format version ? (`0.12.4` is currently shown in the Satellite dialog, but that is the _catalog format_ version... If needed, include versions of plugin as well as catalog format ?).

The "catalog format version" would only change when the format changes (what is exactly what the value means).

plugins/Satellites/src/Satellite.hpp
axd1967: here is another point of view: by putting a zero/default value at the bottom, whenever someone inserts new enums, there will never be a need to juggle with comma's at the end of the list.

(In Python, a list can end with a comma, not in C++).

plugins/Satellites/src/Satellites.cpp
axd1967: Out of scope, this should be tackled in a separate issue, also because there are more places where this logic can be applied (eg log files); also, the presence of many log files can indeed become a burden for the unaware user. So this requires some more reflexion. 

(Also related to #1283 )

alex-w:
The comment "// FIXME replace with datetime stamp as generic routine to backup files" should be removed immediately! The current behaviour is correct and logical. Alex, I'm many time said that your proposal to create a milliones files is very bad proposal, but you propose it again, again and again.

plugins/Satellites/src/gui/SatellitesDialog.cpp
axd1967: typo, should be "copy-pasta". the color setting code is indeed copied further down the file. 
refactored.

plugins/Satellites/src/gui/SatellitesDialog.cpp
axd1967: see above :-)

axd1967:
> The comment "// FIXME replace with datetime stamp as generic routine to backup files" should be removed immediately! The current behaviour is correct and logical. Alex, I'm many time said that your proposal to create a milliones files is very bad proposal, but you propose it again, again and again.

(see above)

alex-w:
> (see above - it has happened too often that Stellarium clears the JSON file, and the user defined groups are lost. with a decent backup file, the user can restart an older version if needed.

Seriously? Why the groups are lost only on your host? Why i never lost the defined groups and the plugin is updated the TLE data every 72 hours? Maybe this is a result of messy, which you doing by hands and black magic?

> also, this should not happen a lot; only when the catalog format changes. so certainly not a "million times"...)

The data of catalog by default updated every 72 hours and for every event you propose create a separate backup file. So, this is definitely a "million times".

plugins/Satellites/ChangeLog
gzotti: I was referring to entries in exactly this file which show some version number and date. The formatting of your ChangeLog entry does not match the other entries. It can be assumed that the version referred to in the ChangeLog is the plugin version.

plugins/Satellites/src/Satellite.hpp
gzotti: I don't understand the difficulty of "juggling with commas". Is there a difficulty with formatting a list of enums?  These are simple consecutive enums, the numerical value of which don't have to be given explicitly. 

TBH the plugin is not my code and I may have overlooked something.

plugins/Satellites/src/Satellites.hpp
axd1967: Strange, see https://github.blog/2013-12-03-soft-wrapping-on-prose-diffs/ (from https://stackoverflow.com/questions/16475439/is-it-possible-to-wrap-long-lines-in-the-github-diff-viewer). If I understand correctly, this should from now on work fine. I'll test this in the next commit on this branch.

alex-w:
> and at some point in the future, format changes can be accompanied with upgrade logic rather than bluntly throw away the current file. this will give even more value to the catalog version number.

I hope you will create a special tool to correct upgrading the data for each variant of every version of satellites.json file since version 0.12.0 and I hope you will create a tool to manage all thousand of backup files. Of course with GUI and human readable options. 

The first point is already implemented by me 4 versions ago, but you just ignore this fact and propose solutions to nonexistent bug. As always.

axd1967:
As far as I can see, the **backup** (call of ` Satellites::backupCatalog()`) is done only when the format changes (is different from the expected value), the JSON file is missing, the user presses restore defaults. 

But not when the data is updated?

But I might be mistaken.

alex-w:
> As far as I can see, the **backup** (call of ` Satellites::backupCatalog()`) is done only when the format changes (is different from the expected value), the JSON file is missing, the user presses restore defaults.
> 
> But not when the data is updated?
> 
> But I might be mistaken.

Ah, yes, I've changed the behaviour for this feature a year ago or something near it. In any way your FIXME is not needed.

plugins/Satellites/src/Satellites.hpp
```diff
@@ -74,7 +74,7 @@ By convention, group names are in lowercase. The GUI translates some of the grou
 
 <b>Satellite Catalog</b>
 
-The satellite catalog is stored on disk in [JSON](http://www.json.org/) format, in a file named "satellites.json". A default copy is embedded in the plug-in at compile time. A working copy is kept in the user data directory.
+The satellite catalog is stored on disk in [JSON](http://www.json.org/) format, in a file named "satellites.json". A default copy is embedded in the plug-in at compile time. A working copy is kept in the user data directory. Test for GitHub diff, are lines now soft wrapped?
```
axd1967, line 78: @gzotti here the soft wrap does work fines in GitHub diff (split mode, but apparently not in unified mode). nevertheless IMHO we should avoid hard newlines, as most editors offer soft wrap anyway.

In fact, hard newlines are a thing of the past in the era of 4k screens and intelligent editors... GitHub should not lead to this kind of limitations.

And otherwise at some point someone will have to edit away hard line breaks because lines become too fragmented.

gzotti: I see the line upto "... in a file nam" in the box above. It is cumbersome enough to move the scroller (which is available in the box above), but most online diff views don't provide a scroll bar, making it extra nasty to move around. Yes, I have a wide screen, but GH's style only uses 1/3 of it. And others don't have wide screens. 

Line length is something that may be regarded as part of the project code style, similar to e.g. indentation or bracketing rules. It is not only for Github. Our style guide. Defined by the maintainers and forced upon the heads and fingers of all contributors before a maintainer will press "Merge". Please adhere to the code style when not instructed otherwise. Resistance is futile.

axd1967: ok!

axd1967: by the way, try shift+scroll while in the diff pane.

(and remember: life = permanent evolution, permanent re-questioning of existing values, permanent refactoring. be agile.)

plugins/Satellites/src/Satellites.hpp
```diff
@@ -74,7 +75,10 @@ By convention, group names are in lowercase. The GUI translates some of the grou
 
 <b>Satellite Catalog</b>
 
-The satellite catalog is stored on disk in [JSON](http://www.json.org/) format, in a file named "satellites.json". A default copy is embedded in the plug-in at compile time. A working copy is kept in the user data directory. Test for GitHub diff, are lines now soft wrapped?
+The satellite catalog is stored on disk in [JSON](http://www.json.org/) format,
+in a file named "satellites.json". A default copy is embedded 
+in the plug-in at compile time. A working copy is kept in the
+user data directory. Test for GitHub diff, are lines now soft wrapped?
```
gzotti, line 81: Lines are not soft-wrapped, but line length is OK now. 

This branch would have been almost done according to my understanding, only these minor issues were open. I don't know how much time it took to research the counterexamples of code that breaks the line length style rules, and I did not check all of these entries. Still, when reviewers insist on these things with new code from external contributors, there is no need to get upset on such minor details.

axd1967: that's exactly the point I wanted to make: it took me only a few minutes to find those examples (I just browsed the first PRs that came up), and **many** more could be pointed out. but I'm more than fed up with all these stupid remarks and daft counter arguments that basically stem from irrational attitudes and broken English and that bypass far more important issues such as improving software engineering practices, given for free from someone who has years of experience (and believes in and has practiced the Agile approach).

sorry guys, but that's NOT how you treat contributors. **and it's not the first time I point this out**,

PR #1971, opened 2021-10-17
GUI: focus and tabbing

gzotti:
### Description

As discussed in #1834 the Qt-based interface allows using the tab key to switch between GUI elements. So far the GUI was not designed to be used that way, and the GUI did not indicate which button had the "focus" to receive the next keypress. 

However, for a few users this seems a critically important feature to add. This branch should be used to add the focus highlight to GUI elements where it makes sense, and then to adapt the .ui files to define a "tabbing order". 

* Note that the tabbing order will then have to be checked and re-developed every time a UI file is changed. 
* Note that all UI file elements "colorrole" must be deleted after every edit of an ui file with QtCreator to make them compilable with Qt5.9. Or why do we have to do that? I don't find documentation on that.
* A new GUI element "SpaceCheckBox" can be used to modify QCheckBoxes into spacebar-toggling elements. 

Fixes # (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: all
* Graphics Card: all
## Checklist:

### DIFF
--- a/src/gui/SpacePushButton.cpp
+++ b/src/gui/SpacePushButton.cpp
@@ -25,13 +25,13 @@ SpacePushButton::SpacePushButton(QWidget* parent)
 {
 }
 
-void SpacePushButton::keyPressEvent(QKeyEvent *e)
+void SpacePushButton::keyReleaseEvent(QKeyEvent *e)
 {
 	switch (e->key()) {
 	case Qt::Key_Space:
-		animateClick();
+		click();
 		break;
 	default:
-		QPushButton::keyPressEvent(e);
+		QPushButton::keyReleaseEvent(e);
     }
 }
--- a/src/gui/SpacePushButton.hpp
+++ b/src/gui/SpacePushButton.hpp
@@ -23,7 +23,7 @@
 #include <QPushButton>
 
 //! @class SpacePushButton
-//! A QPushButton which can be triggered by pressing Space when it has focus.
+//! A QPushButton which can be triggered by pressing (actually: releasing) Space when it has focus.
 //! To use this class in the QtCreator UI designer, add a regular QPushButton to the UI,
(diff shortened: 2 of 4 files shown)


alex-w:
> Note that all UI file elements "colorrole" must be deleted after every edit of an ui file with QtCreator to make them compilable with Qt5.9. Or why do we have to do that? I don't find documentation on that.

The “colorrole” for placeholder’s text was introduced in version 5.12 and, of course, for Qt 5.9 this is unknown instruction.

gzotti:
> 
> 
> @gzotti if you like I can review other UIs via this branch

Yes. If you rebase #1834 on current master and PR to this branch, we can start with that. But note that there is only so much time to invest in this topic. Nothing will ever be 100% perfect for all users, but some improvements are possible. The new GUI element variants provide focus highlights now for buttons and checkboxes , so they can be included when they make sense. (Not sure about pages like view/markings, would anybody really not use a mouse here?)

axd1967:
> I intended to develop things around tabbing a bit further in this branch. The point is that tabbing was a non-issue, it was not bad, it was basically NOT AVAILABLE for the last 20 years of this project. Nobody cared for it, and I don't remember a single user complaining before 2021. You want to introduce it, so it must be done well. There are a few weeks to go before this could go to master.

OK. I will continue on this branch (via #1997 ), so please don't rebase it any more.

axd1967:
Just one note: please remember that this is **not** a _"critically important"_ feature; it has simply been neglected by developers and never contemplated by the vast majority of the users (my guess is that the UI is already unusual, therefore not inviting for this kind of change requests). Now is an opportunity to improve the keyboard UI because I am able to assist here to review that tabbing order (plus some other minor changes).

PR #2026, opened 2021-11-11
Add action to open Solar System Editor #2022

### LINKED ISSUE #2022: Add action to open Solar System Editor
For some reason that plugin deserves more attention (it's hard to believe that it is catalogued as a plugin); I often use `ctl-alt-S` to open the import dialog, but in fact there is a need to add an action shortcut to open the Solar System editor - from where the import dialog (which does have a shortcut action) can then be opened.

_Originally posted by @axd1967 in https://github.com/Stellarium/stellarium/issues/1045#issuecomment-964595580_

axd1967:
- add shortcut action 'Ctl+alt+shf+S' to 'open editor' (although the swap 'ctl+alt+s' could be better...)
- SSE UI: make page 2 ("Solar System") as default page
- add confirmation check

Also
- bump plugin to version 1.1.0 (1.0 + minor change)

### Description

Fixes #2022

### Type of change

### How Has This Been Tested?

**Test Configuration**:
KDE Plasma Version: 5.18.5
Qt Version: 5.12.8
Kernel Version: 5.4.0-90-generic
OS Type: 64-bit
Memory: 15.3 GiB of RAM

## Checklist:


plugins/SolarSystemEditor/src/gui/solarSystemManagerWindow.ui
```diff
@@ -132,7 +132,7 @@
    <item>
     <widget class="QTabWidget" name="tabWidget">
      <property name="currentIndex">
-      <number>0</number>
```
alex-w, line 135: First window should open by default

plugins/SolarSystemEditor/src/SolarSystemEditor.cpp
```diff
@@ -136,6 +136,7 @@ void SolarSystemEditor::init()
 	isInitialized = true;
 
 	// key bindings and other actions
+	addAction("actionShow_SSE", N_("Solar System Editor"), N_("Open Solar System editor"), mainWindow, "openMainWindow()", "Ctrl+Alt+Shift+S"); // it's a pity that such a window is so difficult to reach!
```
alex-w, line 139: Name of action and name of translatable label should be consistent

plugins/SolarSystemEditor/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 
-SET(SOLARSYSTEMEDITOR_MAJOR "0")
```
alex-w, line 2: Why 1.1.0? This definitely is 0.2.8, in better case 0.3.0, but not 1.1.0

guide/plg_interfaces.tex
```diff
@@ -638,7 +638,7 @@ \section{Solar System Editor Plugin}
 The second tab lists all currently loaded minor bodies (see figure~\ref{fig:SolarSystemEditor:SolarSystem}).  It is recommended
 to remove old entries of yesteryear's comets if you don't need them any
 longer. Just select one or more objects and press \button{Remove}.
-If you have a very weak computer, you may want to reduce the number of
```
alex-w, line 641: "very weak" is better wording here, because "slow" maybe PC with 16-cores modern CPU for subjective reasons...

alex-w (changes requested):

plugins/SolarSystemEditor/CMakeLists.txt
gzotti: Is it even meaningful to keep separate plugin version numbers? The plugins are all static. Just set to the main program version. If not, I agree, this mini change does not warrant a big version number change.

alex-w:
@axd1967 why you try added very annoying confirmation dialogs to everything?

plugins/SolarSystemEditor/CMakeLists.txt
alex-w: Syncing plugin version with version of main application is acceptable for me. The separate plugin version numbers is coming by historical reasons and we have no guarantee that in future we'll not produce dynamical plugins again.

gzotti: Sure. We could then split out the numbers again if that appears useful.

axd1967: IMHO we should get rid of plugin version info _for now_. I think it is a smokescreen, a source of confusion and triggers needless discussions and draws effort to keep it up-to-date. 

But for what?

I don't ever see _any_ of the current plugins _ever_ appear outside of Stellarium. If that ever happens, then the version can be shown again. **So I propose to hide the version info and freeze the numbers for now.**

The only advantage could be to draw the attention to the user as to what happened. Assuming the user understands SemVer...

About 1.0: the plugin is **no longer in alpha/beta phase**. It is used extensively by the end users. And 1.1.0 signifies a minor update w.r.t. the previous version. 

x= major/important/breaking change
y = minor change / improvement
z = bug fix

If we would keep 0.y.z, how else can new plugins be marked as being in alpha/beta phase? Only new plugins deserve the 0.y.z numbering, until they are released/part of a Stellarium release. Unless it is uncertain how the plugin behaves, and _end users_ can play with it, knowing it is still in _beta_.

This is mainly a matter of **courage**. Stop thinking that everything should be 0.y until everything works fine, because that will **never** happen. 

1.0 = released.

axd1967:
> @axd1967 why you try added very annoying confirmation dialogs to everything?

not to everything, but to places where the user can lose information.

plugins/SolarSystemEditor/CMakeLists.txt
alex-w: @axd1967 if you something do not known then this is not meaning that this is not exists!

plugins/SolarSystemEditor/src/gui/solarSystemManagerWindow.ui
axd1967: The main dialog is not the MPC import dialog, but the actual editor window - that's the name of the plugin. In that editor, a button allows to open the MPC import dialog, a secondary dialog.

Actually, the second tab ("Solar System") should be swapped with the first tab, because that is the **core tab** of the SSE. It also draws the attention to the MPC import button.

The "configuration file" tab is a less important and less used tab that contains (dangerous) actions that are seldom used. That's why I propose to make the "Solar System" tab open by default.

Also, SSE should receive the same treatment as other plugins, i.e. a toolbar button that opens a main window.

alex-w:
> not to everything, but to places where the user can lose information.

Do you understand that without selection users will not lost anything here? But right now you want to "ask" deleting twice.

plugins/SolarSystemEditor/src/SolarSystemEditor.cpp
```diff
@@ -136,7 +136,7 @@ void SolarSystemEditor::init()
 	isInitialized = true;
 
 	// key bindings and other actions
-	addAction("actionShow_SSE", N_("Solar System Editor"), N_("Open Solar System editor"), mainWindow, "openMainWindow()", "Ctrl+Alt+Shift+S"); // it's a pity that such a window is so difficult to reach!
+	addAction("actionShow_SSE", N_("Solar System Editor"), N_("Open Solar System editor"), mainWindow, "showSSE()", "Ctrl+Alt+Shift+S"); // it's a pity that such a window is so difficult to reach!
```
alex-w, line 139: See "Open Solar System editor" - do you see problem?

alex-w, line 139: See "Open Solar System editor" - do you see problem?

plugins/SolarSystemEditor/src/gui/solarSystemManagerWindow.ui
alex-w: This is not reasons for me. @gzotti ?

plugins/SolarSystemEditor/src/SolarSystemEditor.cpp
alex-w: See "actionShow_SSE" - do you see problem?

guide/app_hotkeys.tex
```diff
@@ -352,6 +352,7 @@ \subsection{Solar System Editor}
 \toprule
 \emph{Shortcut key}	& \emph{Description}\\\midrule
 \key{\ctrl+Alt+S}	& Import orbital elements in MPC format... \\
+\key{\ctrl+Alt+Shift+S}	& Open Solar System Editor \\
```
gzotti, line 355: I am really not sure we need more than one hotkey action to bring this dialog up. One is fine to avoid the long path through settings/plugins/configure, but how often do you (or the average user) really use it?

plugins/SolarSystemEditor/src/gui/solarSystemManagerWindow.ui
axd1967: OK. if you want, we can discuss this. (about a single change that nobody else will complain about, I'm rather sure about that. )

you ask me for my motivation, I've written it down in detail, and it sounds quite reasonable to me.

remember that the pages can be swapped in Qt Editor. I'm more than happy to do that, I already tested it, but I just need confirmation for that.

now I want to know what your motivation is.

"_it has been like that before_" or "_nobody complained about it_" are no good reasons. that's not the way software evolves. maybe I oversaw something?

guide/app_hotkeys.tex
axd1967: It's not "_more than one hotkey_" - the main dialog has _no_ hotkey, that's the point.


For me, these should be the actions:

- ctl-alt-S: open SSE
- ctl-alt-shf-S: open import window

But I guess most users are too used to the existing MPC import shortcut, and anyone can use the keyboard shortcut editor if needed.

plugins/SolarSystemEditor/src/gui/solarSystemManagerWindow.ui
alex-w: I don't want discuss anything and I don't like breakage behaviour of the plugin. Just check GUI and behaviour of all other plugins and say us where is you wrong in your "reasons".

plugins/SolarSystemEditor/src/SolarSystemEditor.cpp
alex-w: Please fix English error and please use human readable name for action.

PR #2027, opened 2021-11-11
Orbit epoch

### LINKED ISSUE #2000: Astrocal ephemeris ignores Ceres, Vesta, ...
Tested with 0.21.2 fc4c6bb

When computing the ephemeris of Ceres from 2021-10-28 to 2021-11-28, the dialog (F10) remains empty for Ceres.  this appears in b87b419af774a3cb8a2fa5b91bf2badcee0963c7.

image

I tried to update Ceres via asteroid import data; not sure the import succeeded: was Ceres finally updated?

Assuming this is an issue of outdated orbital data (but ... Ceres?) https://github.com/Stellarium/stellarium/pull/1880#discussion_r713987279 might have helped here. 

- [ ]  selecting an object should show the validity date range of that object as a new info line (for_wishlist)

gzotti:
### Description

This finally adds the epoch information to the KeplerOrbit object. 

For some time now we have avoided showing objects (minor bodies) when their orbital elements were outdated. 
However, it was always assumed that the orbital elements' epoch was equal to perihelion date, which is not necessarily the case.
This lead to unclear situations like wrong cut-away dates for asteroids with several years' orbital period. 

Orbital elements should in fact be updated regularly. I also added a warning in the InfoString when date is outside the "good" range. It may be discussed if the orbit_good evaluation or Planet::hasValidPositionalData() should be extended to allow differentiation between elements "good enough to show closed orbit" (years) and "good enough to find in telescope" (months).

Fixes #2000  (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 (irrelevant)
* Graphics Card: (irrelevant)

## Checklist:


alex-w:
Please update `comet_orbit` strings in default `ssystem_minor.ini` file according to Stellarium User Guide.

Please move "note" line to the end of list.

gzotti:
I would still leave "comet_orbit" for compatibility with earlier versions. In 0.21+ we need no explicit orbit_func at all, but understand kepler_orbit and comet_orbit. But I often run earlier versions for comparison. 

The NOTE in infostring is just below the coordinates. Do you mean I should move that to the end of the infostring?

alex-w:
> I would still leave "comet_orbit" for compatibility with earlier versions. In 0.21+ we need no explicit orbit_func at all, but understand kepler_orbit and comet_orbit. But I often run earlier versions for comparison.

But I said about default file, which will be used in newly installation only. Exist `ssystem_minor.ini` file will be used as before with `comet_orbit` string and in backward mode.

> The NOTE in infostring is just below the coordinates. Do you mean I should move that to the end of the infostring?

Yes, at end of the infosting - this line will be more visible here.

alex-w:
> does that default to sth meaningful? Do we need it at all?

Yes, we need it to some moons and probably to some asteroids/comets

gzotti:
It's an oldish comment. We could just use one revolution as default for closed orbits. I think in previous years this and orbit_good was sometimes mixed up a bit. Probably we should use this to plot comet orbit around perihelion, and report orbit_good around epoch? Or still plot orbit with this value around epoch as well? These are things to still think over a bit.

alex-w:
> It's an oldish comment. We could just use one revolution as default for closed orbits. I think in previous years this and orbit_good was sometimes mixed up a bit. Probably we should use this to plot comet orbit around perihelion, and report orbit_good around epoch? Or still plot orbit with this value around epoch as well? These are things to still think over a bit.

It’s all not for this PR

alex-w (approved):
Thank you very much!

gzotti:
Don't merge yet. I may want to check a few more things.

alex-w:
OK, please do it yourself

axd1967 (commented):
@gzotti There exists `1P/Halley (2061)` - does it fit the current algorithm? 2061 is more than 3 years (approx 1000 days) in the future. It has `orbit_good = 20000`

axd1967 (commented):
@gzotti have a look at https://github.com/axd1967/stellarium/commit/984084dc91f6c8b45a34a4bb8144882278742671

- I really would avoid `1000` but put this in a constant
- `0` and `-1` beg to be cast in enums. but i somehow have a feeling that `0` could lead to bugs

(you can fetch that branch to make work easier)

gzotti:
The element set has an epoch of 1994. 20.000 days is definitely uselessly much for assume anything accurate, but it might be good enough for plotting an orbit. 

I am just adding a distinction about whether the orbital data is "good enough for plotting an orbit" (defaults to 1/2 sidereal period or orbit_good) or "good enough to be sufficient for setting a telescope". The latter has higher requirements and uses the smaller of (epoch+/-orbit_good) and (epoch+/-1 year) to suggest updating elements after 1 year.

axd1967:
> I am just adding a distinction about whether the orbital data is "good enough for plotting an orbit" (defaults to 1/2 sidereal period or orbit_good) or "good enough to be sufficient for setting a telescope". The latter has higher requirements and uses the smaller of (epoch+/-orbit_good) and (epoch+/-1 year) to suggest updating elements after 1 year.

consider somehow coding these two requirements more explicitly? I proposed two enums to get rid of the magic `0` and `-1`, but these are probably unrelated. 
but explicitly coding these requirements will make the code so much more clear.
I'd suggest to use 3*365 rather than 1000 days - even if both values still are magic and probably not 100% justifiable, it's just a bit more human-readable.

alex-w:
The changes is good for me.

gzotti:
(According to my understanding this is a false positive. All cases in the switch are handled, and whatever is not handled by some orbit object returns false. )

axd1967:
> (According to my understanding this is a false positive. All cases in the switch are handled, and whatever is not handled by some orbit object returns false. )

one day Murphy will knock on that door...

a switch should almost always include a default, even if never used. it's part of defensive programming.

and it's cheap to add a default case; just add a warning (that should never appear; so how about as `ASSERT` then ? ;-) .

gzotti:
Do you know the clang in-editor warning that all switch cases have been fulfilled so there is no need for a default clause?

Else, show me the logical error in the current code. Which code path does not return anything?

axd1967:
I never wrote that the current situation is wrong ;-) 
so _currently_ it's OK  (especially because all enums are covered.)

alex-w:
> I never wrote that the current situation is wrong ;-)

You did it 2 hours ago

PR #2006, opened 2021-11-02
Allow compilation by disabling cmake flag WITH_QTWEBENGINE

gzotti:
### Description

On some platforms (ARM based) the QtWebEngine seems to be missing. The CMAKE flag WITH_QTWEBENGINE can now be set to false on those to disable the webbrowser view internal to the OnlineQueries plugin. The websites will instead be queried inside the system's default web browser.

* Currently this flag has to be manually set. It may be possible to set the flag as consequence of cmake being unable to find the QtWebEngine modules, and emit a warning to the developer during compilation. 

* If there is some style guide for cmake, I surely did something wrong! Advice?

Fixes #2004, #1982 

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10; Raspberry Pi 4.
* Graphics Card: irrelevant.

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #2049, opened 2021-11-19
Additional cmake fixes over #2006

gzotti:
### Description

#2006 was merged too early. We must make sure that the program can be compiled even on systems where QtWebEngine does not exist (esp. ARM platform, but also a few others.)

This here attempts to
- [x] Let cmake detect availability of QtWebEngine, or warn the user that a less functional alternative will be built
- [x] Build the QtWebEngine-dependent GUI element on the fly. 
- [x] Useful behaviour with WSL
- [x] Test on RaspberryOS
- [x] Test in CI systems
- [x] Test on MacOS

#### Note
There is a problem seen in the test system used: cmake on WSL/Ubuntu with qtwebengine5-dev not installed sees the installed Qt libraries in /mnt/c/qt (This is C:\Qt in the Windows file hierarchy and obviously the wrong place to access from WSL) when it does not find the really appropriate libraries in /usr. I *hope* this is a so radically absurd setting that CI and other ways of self-building deployment finds no problem with that. Else **I need to know a way to specify that cmake must never access the wrong Qt installations.**

The next issue on WSL is that building *with* QtWebEngine enabled results in a system where the OnlineQueries plugin fails to launch: 

```
ERROR:gl_surface_glx_qt.cpp(141)] GLX 1.3 or later is required.
ERROR:gl_ozone_glx_qt.cpp(57)] GLSurfaceGLXQt::InitializeOneOff failed.
shared memfd open() failed: Die angeforderte Funktion ist nicht implementiert

```
So it's better to detect WSL with cmake and disable using QtWebEngine in this case. 

Fixes #2006 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating systems: 
  - Windows 10 / Qt5.13 
  - WSL/Ubuntu 20.4 with Qt5.12.8
  - Raspberry OS (RPi4)
* Graphics Card: irrelevant

## Checklist:


gzotti:
https://gitlab.kitware.com/cmake/cmake/-/issues/20878 describes a similar case. It seems there is no solution in this particular case. WSL users should certainly install the correct package as documented in the build instructions. The standalone CI systems should work, hopefully. I will try RPi4 next.

alex-w (approved):

alex-w:
@gzotti do you use Qt 5.12 or 5.15 in WSL?

gzotti:
I run Ubuntu 20.04.3 LTS, where cmake finds Qt5.12.8. The problem is described above. 
How would I update to 5.15, and/or switch between two installed versions on this platform? I have seen a lot of deprecation warnings in the CI logs when 5.15 is used.

gzotti:
I think this is OK for now. Few people will run Stellarium on WSL, and compilation now creates a crash-safe application (OK, at least it does not crash over this issue...). I would merge when CI runs have completed without issues.

Issue #1982, opened 2021-10-24
Fresh build, stuck in black screen on startup


rayw000:
### Expected Behaviour
Start normally.

### Actual Behaviour
Black screen forever after splash disappearing.

P.S. Version 0.21.2 installed with `brew` works normally on my laptop.

### Steps to reproduce
1. Update code to ba80d33d4
2. Run the following commands to build
```shell
git clean -fx .
mkdir build && cd build
cmake -DCMAKE_PREFIX_PATH=/usr/local/Cellar/qt@5/5.15.2_1 ..
make -j4
# wait for build done
./src/stellarium
```
Stellarium can start, but after splash disappears, it stucks in black screen forever.

### System
* Stellarium version: 0.21.2@ba80d3
* Operating system: macOS Big Sur version 11.6
* Graphics Card: Intel Iris Plus Graphics 1536 MB
* Screen type (if applicable): Built-In Retina LCD

### Logfile 
Below is console output of command `./src/stellarium`.
```
WARNING StelFileMgr::StelFileMgr: could not find install location: "/Users/ray/repo/open-source/stellarium"  (we checked for  "/Users/ray/repo/open-source/stellarium/data/ssystem_major.ini" ).
Maybe this is AppImage or something similar? Let's check relative path...
WARNING StelFileMgr::StelFileMgr: could not find install location: "/Users/ray/repo/open-source/stellarium/stellarium/build/src/../share/stellarium"  (we checked for  "/Users/ray/repo/open-source/stellarium/stellarium/build/src/../share/stellarium/data/ssystem_major.ini" ).
Maybe this is development environment? Let's check source directory path...
 ----------------------------------------------------------------------
[ This is Stellarium 0.21.2-ba80d33 [master] - https://stellarium.org/ ]
[ Copyright (C) 2000-2021 Fabien Chereau et al.                        ]
 ----------------------------------------------------------------------
Writing log file to: "/Users/ray/Library/Application Support/Stellarium/log.txt"
File search paths:
  0 .  "/Users/ray/Library/Application Support/Stellarium"
  1 .  "/Users/ray/repo/open-source/stellarium/stellarium"
Config file is:  "/Users/ray/Library/Application Support/Stellarium/config.ini"
[...]

PR #2067, opened 2021-11-28
Fix typo

rayw000:
Fix typo

### DIFF
--- a/plugins/OnlineQueries/src/CMakeLists.txt
+++ b/plugins/OnlineQueries/src/CMakeLists.txt
@@ -30,7 +30,7 @@ FIND_PACKAGE(Qt5WebEngineWidgets QUIET)
 IF(Qt5WebEngine_FOUND AND Qt5WebEngineWidgets_FOUND)
     ADD_DEFINITIONS(-DWITH_QTWEBENGINE)
     SET(QTWEBENGINE_LIBS Qt5::WebEngine Qt5::WebEngineWidgets)
-ELSE(Q5tWebEngine_FOUND AND Qt5WebEngineWidgets_FOUND)
+ELSE(Qt5WebEngine_FOUND AND Qt5WebEngineWidgets_FOUND)
     MESSAGE(WARNING "Qt5WebEngine and/or Qt5WebEngineWidgets not found. \
                      If these packages exist for your platform, install and rerun. \
                      Else the program will call the system web browser where required.")


alex-w (approved):
Thanks!

gzotti:
Oops, thanks!

rayw000:
Thanks @alex-w 

BTW, seems `WITH_QTWEBENGINE` didn't work well for me.

I built with brew installed Qt 5.15.2. Seemed that it's shipped along with `QtWebEngine`. Line 30 evaluated true in my case.
https://github.com/Stellarium/stellarium/blob/dd006bc4095790dba6ceb7fe485284a6804a9fd4/plugins/OnlineQueries/src/CMakeLists.txt#L28-L37

I've read #2006 and #2049. `-DWITH_QTWEBENGINE=0` didn't work for me, neither `-DWITH_QTWEBENGINE=off`. There was a cmake warning saying
```
CMake Warning:
  Manually-specified variables were not used by the project:

    WITH_QTWEBENGINE
```
and issue #1982 was still.

Only fully disabling online queries plugin with option `-DUSE_PLUGIN_ONLINEQUERIES=false` worked for me.

I built on Monterey 12.0.1 with Apple M1 Max chip. I'm trying to find out why.

alex-w:
Issue #1982 happened with Qt 5.15 only and probably this is bug in Qt/macOS.

rayw000:
Oh, I thought we could walk around it by `-DWITH_QTWEBENGINE=off` after reading your reply https://github.com/Stellarium/stellarium/pull/2006#issuecomment-957986750.

gzotti:
I have made another round of corrections, seeing that my Odroid C1 SBC with Ubuntu 18LTS has QWebEngine, shows a webengine panel, but fails to work, emitting error messages to stderr. 

Now 
* We detect QWebEngine availablilty at compile time, emit a warning if not found, but compile anyhow.  
* On systems where QWebEngine is present but buggy, compile with it, but manually disable the built-in QWebEngineView and open the URL in the system webbrowser. 

Now finally the external system webbrowser also works on that (minimal, 5fps) platform. 
The User Guide (sources) includes instructions for users.

gzotti:
The point seems to be that at least on ARM platform, QWebEngine does not work properly before Qt6.2, and also not on X11 platforms with some GLX1.3 problem.

PR #2085, opened 2021-12-04
GZ atmosphere tweaks

gzotti:
This branch allows to play with the nonintuitive parameters of the default sky model. The branch has a long history, and I resurrected it to be able to compare current master and these excursions, in order to better understand what happens in PR #1469.

### Description

Stellarium is using the widespread Preetham sky model, or rather, a modified version that combines its colours with Bradley Schaefer's sky brightness model. Add some color transformation and tonemapping, and there may be some incongruences that have creeped in. This branch has an extended GUI to change atmospherical parameters. To understand, you must be familiar with the original paper. If misapplied, you can create arbitrarily colorful skies, or skies which look simply broken. Therefore only "power users" should use it to tweak some detail settings after enabling a special GUI dialog by a flag to be set manually in config.ini. 

Another part that is made user-accessible here is the tonemapping. Some parameters like screen brightness should be modifiable. PR #1469 wanted to replace it by a modified algorithm which IMHO looks bad. This solution seems more acceptable, as it allows the user to switch between both paths. We can decide later about further tonemapping improvements.

Fixes: #1469 (issue)
Necessary for: #624

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: (Windows 10)
* Graphics Card: (Geforce 960M)

## Checklist:


Atque:
Any chance this can lead to a merge if better values to the parameters are found? I have noticed that tweaking certain parameters will give somewhat better colors. Although no earth shadow and other awesome atmospheric phaenomena present in the Bruneton model.

gzotti:
This was an experimental branch where I tried to analyze the sky model which unfortunately was almost undocumented. I still don't know where the other (Stellarium's own?) values for the sky colors come, are they just empirical "to make it look better", or did Preetham or somebody else publish other values elsewhere? The branch makes sense only if you are fully familiar with the original paper and with CIE color spaces, as the parameters are far from intuitive. However, the results sometimes look interesting. (1960s Sci-Fi kind of interesting...)

We have a contributor working on the Bruneton model (to be activated on capable systems only). The result looks extremely promising. But first I need to investigate one preliminary bit of his work which IMHO is not correct, so I resurrected this, and was promptly called to other duties... We could of course merge it, but only with a dedicated chapter in the Guide.  Maybe we should guard 100% - 10 users with a manually modified entry in config.ini, though, because most values cause nonsense or even buggy skies. It is a great tool to analyze that sky model, however, and also the assumption of the tonemapper that a CRT screen has 100cd/m^2 has to be re-investigated with IPS and more modern screens running now at 250cd/m^2 and brighter.  

@alex-w is this interesting enough to add for others? Then I would start writing that chapter.

gzotti:
It may get interesting enough for general use. I still have to hide the SkylightDialog from regular users, and maybe revert a few extra excursions.

Atque:
I think letting advanced users play around with the parameters is a good idea, especially since the upcomming Brunteton model seems highly customizable.

I noticed that the Moon keeps flickering when panning around, even after resetting all values.

gzotti:
You mean the Moon itself, or its halo, or its brightening in the sky dome? 
Do you have an Intel GPU? We have several reports about problems. Else the Moon rendering itself did not change, and I don't see a problem.

Atque:
> You mean the Moon itself, or its halo, or its brightening in the sky dome?

Both its halo and (some, not all) of the sky brightening.

> Do you have an Intel GPU? We have several reports about problems. Else the Moon rendering itself did not change, and I don't see a problem.

No, I run AMD (Radeon RX 5700 XT).

alex-w (approved):
Good work, thanks!

I've played with tone-mapping (please check the English term in the GUI) settings - this is an interesting effects with some parameters. At least default settings show the sky as was implemented before.

wosher112:
@gzotti how i can make the new atmosphere for stellarium is nvidia or intel graphics i not want wasting my time

Issue #105, opened 2018-02-17
no option to remove halo around stars


alexandervlpl:
There's a handy `halo` option to render bright solar system objects without a halo. Can you please add the same option for stars?

Currently bright stars are rendered with a large halo:
![halo](https://user-images.githubusercontent.com/8945883/36340406-48d35aee-13ed-11e8-917e-012e673743c0.png)

Part of this is the "glow" effect in the star texture file (star16x16.png), which is fine. But a large area of sky around the star is lighter too. It would be nice to have an option to disable this.

gzotti:
This is by intention to highlight bright stars which else would need even larger disks or will appear too dim. What do you think is disturbing with this solution?

alex-w:
Size of halo around stars may be customized by adjusting relative scale in stars section of sky tab - why you do not use this option?

alexandervlpl:
@gzotti I know there's a purpose for this, but the effect can be distracting and unrealistic sometimes. This is probably why there is a `halo` option for solar system objects, right? Seems logical to have one for stars too.

@alex-w I've tried adjusting both relative and absolute scale in `config.ini`. To remove the halo effect completely for bright stars, the values have to be very low and many stars become way too small.

alex-w:
@alexandervlpl seems you want to get an SkyChart behaviour in Stellarium - why you just do not use SkyChart in this case? Just for understanding - Stellarium never set the goal to be a cartographic planetarium (like SkyChart/Cartes du Ciel). The main goal of Stellarium is realistic sky.

alexandervlpl:
@alex-w For me the glow actually makes the sky *less* realistic, especially at high resolutions when absolute and relative scale need to be increased. Maybe it's just my eyes.

I chose Stellarium over other tools exactly because the visuals are better and there are many options to customize them for any requirements and tastes. A star `halo` option would be a great addition to this. I'll be happy to see it in `config.ini`, no need to add it to the GUI.

gzotti:
Just compiling to see how bad it will look...

gzotti:
Looks bad, but we also allow twinkling in space so I will allow it via config.ini [star] / flag_star_halo=false. I won't bother with a GUI checkbox.

alexandervlpl:
Perfect, thanks again @gzotti !

gzotti:
Done.

PR #2097, opened 2021-12-13
Add photorealistic halos feature

maldenol:
Abstarct
In Stellarium there is no option to switch off stars' and planets' halos.
According to [this](https://github.com/Stellarium/stellarium/issues/105), [this](https://sourceforge.net/p/stellarium/discussion/278769/thread/c9d3d099/) and [this](https://sourceforge.net/p/stellarium/discussion/278769/thread/50c6aaa4/) conversations, the solution is to use flag_star_halo=false (which actually does nothing anymore) in config.ini or to change certain textures.
In my opinion, the availability of photorealistic mode would be great.
So this pull request will add astro/flag_photorealistic option in config.ini for this feature.

Added:
- [x] Add photorealistic mode.
       If enabled, planets wouldn't have halos and stars' halos would be much smaller.
       The flag is available through astro/flag_photorealistic in config.ini (default is false).
- [x] Add ability to choose whether minor planets have halos or not.
       The flag is available through <planet_section>/halo in ssystem_minor.ini (default is true).

Screenshots:
image
image
image

Tests:
 - [x] Hope.
 Actually there is nothing to break.

### DIFF
--- a/src/core/StelSkyDrawer.cpp
+++ b/src/core/StelSkyDrawer.cpp
@@ -55,6 +55,7 @@ StelSkyDrawer::StelSkyDrawer(StelCore* acore) :
 	flagForcedTwinkle(false),
 	twinkleAmount(0.0),
 	flagDrawBigStarHalo(true),
+	flagDrawPlanetHalo(true),
 	flagStarSpiky(false),
 	flagStarMagnitudeLimit(false),
 	flagNebulaMagnitudeLimit(false),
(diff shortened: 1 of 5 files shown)


gzotti:
I don't see why this is necessary. What's wrong with decreasing relative size of the stars in the GUI?
Two of the references you mentioned are from late antiquity and irrelevant as of today.

maldenol:
I agree. Maybe it would be better to leave only "planet" half (toggle only planets' halos) and rename the option appropriately?

maldenol:
> Two of the references you mentioned are from late antiquity and irrelevant as of today.

Still there is no solution.

maldenol:
Done.

gzotti:
Removing the halo patch from the planets hides them when the sphere is too small. This is just no solution. We have discussed this several times in the last decade. The star scaling also relates to the planet "halos".

maldenol:
Agree. Closing.

PR #2132, opened 2021-12-27
Allow moving StelDialogSeparate to a separate screen

gzotti:
### Description

Until now, the particular implementation of the BarFrame deliberately keeps a dialog window within the bounds of the main window. The OnlineQueries plugin provides content in a separate window. With this change, it can finally be really moved to a separate monitor. 

A StelDialogSeparate will be created if possible at the coordinates and size that was last stored. However, the screen number is not preserved, so it will be started inside the main window, and if needed rescaled to fit into this window. 

There are a few things to test: What happens if the program is ended in split-screen arrangement, and then launched with only 1 monitor? What happens switching from a large+small to one small monitor only? Is the dialog always rescalable? 

Fixes # (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
Each system should be tested with at least 2 screens (laptop+1, etc.). Screen config changes between Stellarium launches must preserve accessibility of the OnlineQueries window.
- [x] Win10/Geforce 960M
- [ ] Win10 WSL Ubuntu 20.04
- [ ] Ubuntu 20.04
- [ ] Ubuntu 20.04 MATE?
- [ ] Linux Mint?
- [ ] SuSE Linux?
- [x] MacOSX

## Checklist:

### DIFF
--- a/src/gui/Dialog.cpp
+++ b/src/gui/Dialog.cpp
@@ -42,21 +42,26 @@ void BarFrame::mouseMoveEvent(QMouseEvent *event)
 	QWidget* p = dynamic_cast<QWidget*>(QFrame::parent());
 	QPoint targetPos = p->pos() + dpos;
 	
-	// Prevent the title bar from being dragged to an unreachable position.
-	QWidget& mainWindow = StelMainView::getInstance();
-	int leftBoundX = 10 - width();
-	int rightBoundX = mainWindow.width() - 10;
-	if (targetPos.x() < leftBoundX)
-		targetPos.setX(leftBoundX);
-	else if (targetPos.x() > rightBoundX)
-		targetPos.setX(rightBoundX);
-	
-	int lowerBoundY = mainWindow.height() - height();
-	if (targetPos.y() < 0)
-		targetPos.setY(0);
-	else if (targetPos.y() > lowerBoundY)
-		targetPos.setY(lowerBoundY);
-	
+	QWidget *parent=parentWidget();
+	Q_ASSERT(parent);
+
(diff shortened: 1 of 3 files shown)


alex-w:
> There are a few things to test: What happens if the program is ended in split-screen arrangement, and then launched with only 1 monitor? What happens switching from a large+small to one small monitor only? Is the dialog always rescalable?

I can try test MacBook (retina) + external monitor (regular)

gzotti:
Very good. It must never happen that the program launches and the dialog is off-screen or the scale handle (bottom-right) is not reachable.

alex-w:
See:
* https://doc.qt.io/qt-5/qdesktopwidget.html
* https://doc.qt.io/qt-5/qguiapplication.html#screens

gzotti:
Yes, I have seen the QDesktopWidget, but it is obsolete. And docs don't provide a replacement.

alex-w (approved):
> What happens if the program is ended in split-screen arrangement, and then launched with only 1 monitor?

Nothing: OQ window doesn't support saving information about monitor - it will open on first monitor on next startup always.

gzotti:
> > What happens if the program is ended in split-screen arrangement, and then launched with only 1 monitor?
> 
> Nothing: OQ window doesn't support saving information about monitor - it will open on first monitor on next startup always.

That's OK, IMHO. If not, it'd be time to go into saving screen numbers, but a fallback (if screen has been taken away) must always start overlapping the main window. And this on 3 platforms, with different scaling paradigms, ... we must test that a bit before merge.

PR #2158, opened 2022-01-06
Fix nomenclature longitudes

gzotti:
### Description

There is a bit of confusion about the planetary coordinate systems used in the Nomenclatures. Some use planetocentric, others use planetographic, some count east positive, some count west positive, and while most count 0...360°, at least on the Moon coordinates traditionally count -180...+180°. For Mars, the century-old cartographic tradition was inverted recently, so even modern literature differs. We should clarify these details in this branch. 

Also selection should finally be fixed. When a point is displayed, it must be selectable without problems. 

Fixes # (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: (Win10, irrelevant)
* Graphics Card: irrelevant

## Checklist:


alex-w:
@gzotti I've added 2 new methods for output a "geographical coordinates" - now the coordinates of planetary features is more human readable IMHO

gzotti:
@alex-w if NomenclatureItem::setFlagLabels() is in fact only used as a global switch (switch all in an iterator loop!), not every Item needs its own fader. I would replace this one by a static variable here, ok?

alex-w:
> @alex-w if NomenclatureItem::setFlagLabels() is in fact only used as a global switch (switch all in an iterator loop!), not every Item needs its own fader. I would replace this one by a static variable here, ok?

OK, please do it

alex-w:
Interesting... the Centre point is very difficult to selection after switch to new selection algorithm (try select centres giants from Earth with and without scaling mode)

gzotti:
> Interesting... the Centre point is very difficult to selection after switch to new selection algorithm (try select centres giants from Earth with and without scaling mode)

Yes, this needs some more work. The point is that the previous behaviour provided selection of features while the planet was still too small to see anything. This must be worked upon some more. We can allow selection of non-plotted features when they are large enough to be seen, but previously I could often not select plotted features.

gzotti:
where is NomenclatureItem::update() ever called? Maybe this should be moved to the N-Mgr?

gzotti:
I think to remove some duplicate code the static vars should be moved from NI to vars in the Mgr class, and maybe a pointer to the Mgr as static var in the NI class, to call those property setters/getters. (use NMgr::init() to set that pointer...)

I won't have time until next week. There is no big rush required, though.

gzotti:
@alex-w NomenclatureItem::getAngularSize() may need some attention! Semidiameter like for the planets or really keep it as is with full diameter?

alex-w:
> @alex-w NomenclatureItem::getAngularSize() may need some attention! Semidiameter like for the planets or really keep it as is with full diameter?

I can’t imagine semidiameter of the mountains for example. The planetarynames database have a full sizes of all planetary features, so any crater have a diameter value.

gzotti:
> > @alex-w NomenclatureItem::getAngularSize() may need some attention! Semidiameter like for the planets or really keep it as is with full diameter?
> 
> I can’t imagine semidiameter of the mountains for example. The planetarynames database have a full sizes of all planetary features, so any crater have a diameter value.

The point is the method prototype documentation in StelObject::getAngularSize(): _Return the angular radius of a circle containing the object as seen from the observer with the circle center assumed to be at getJ2000EquatorialPos()._

It seems that only Planet::getAngularSize() does that correctly. Nebula returns the average of major and minor axis which may be quite off e.g. for very edge-on galaxies. I cannot say now if these sizes are axes or semiaxes? 

In StelObject::getInfoMap we have:

`double angularSize = 2.*getAngularSize(core)*M_PI_180;
`

From this the intention should be clear that this must always 1/2 the apparent angular diameter. Obviously, if NIs store full sizes in km, the return values must still be half the apparent diameter. Of course, the Infostring should show full diameter|size (if any).

The method should probably be renamed to getAngularRadius everywhere (and fixed where required, esp. Nebula) to make that clearer. But not in this branch, only after merge.

gzotti:
Of course the debug info in getInfoString() must be removed or at least hidden before merge.

alex-w (changes requested):
Special points are not having angular sizes, so, this data shouldn't be visible for it (of course we added some sizes for these point to work the selection tool). 

IMHO no need angular size ratio data.

I think angular size in angular seconds maybe not very readable for most users, esp. when user will be see hundreds of seconds. Maybe "smart" DMS format will be good here: DMS if degrees are not zero; MS if degree have zero value and angular minutes are not zero and S otherwise.

alex-w (approved):

alex-w:
Sorry, this is debug data!

gzotti:
:-) Yes, these were for finding a better selection priority criterion, not for general use. I will hide them before merge.

alex-w (approved):

PR #2179, opened 2022-01-16
Switch from Bortle class to luminance as the quantity of light pollution

10110111:
### Description

This PR addresses #1586 and is required for #624.

Bortle scale has several disadvantages:

1. It was designed as a guide for humans, not input for machines. From this follows the next disadvantage.
2. It's discrete: each point on the scale corresponds to a range of NELM values, which makes it imprecise as input for a program.
3. It's subjective: based on naked-eye limiting magnitude that depends on observer's visual acuity.
4. It isn't as widely used (and thus validated) as photometric units.

OTOH, luminance is the quantity that defines all the relevant effects of light pollution.

The first commit here makes luminance in cd/m² the input parameter, leaving for backwards compatibility the option of supplying Bortle class in Landscapes. Some changes to calculations had to be introduced, as listed below.

1. `StelSkyDrawer::getNELMFromBortleScale` was reimplemented in `StelCore::bortleScaleIndexToNELM` using a bit different representative values of NELM - to simplify the comment documenting the choice. The values are almost the same: each old value except one will be reproduced by adding 0.05 to the new ones.
2. Calculation of `B_mpsas` in `Nebula.cpp` was trying to convert from Bortle class to magnitude scale using some formula from Schaefer paper, the same paper as I used in `StelCore::luminanceToNELM`, but without explaining the steps to get the formula. The old expression form `Nebula.cpp` (which is widespread over the Internet) doesn't work near NELM==8 (see the discussion at https://astronomy.stackexchange.com/q/43313), while the one I used does. Anyway, `Nebula`'s logic is tied to luminance, so if the user supplies luminance instead of Bortle class, it'll be as intended.

For some reason, `stars/init_bortle_scale` setting defaulted to 3 while `StelLocation::DEFAULT_BORTLE_SCALE_INDEX` is 2. I've retained these preferences.
Technical note: I had to use `QVariant` instead of an `optional` type because Stellarium isn't using C++≥17 (C++17 has `std::optional`), nor anything from Boost (which has `boost::optional`). The other alternative was to make my own `optional`, but that seemed like overkill.

The second commit adds a GUI that lets one manipulate light pollution settings in the following modes:

* Automatic from location database
* Manual by moving a slider, with tooltip showing corresponding NELM
* Manually entering an SQM reading in selectable units

### Screenshots:
Screenshot - 170122 - 01:28:57

Screenshot - 170122 - 01:29:02

Screenshot - 170122 - 01:29:14

### Type of change

## Checklist:


alex-w (approved):
Thank you very much!

gzotti:
Yes, thank you. But please give me test time until the weekend, I am busy these days. First thoughts: I think we should keep Bortle scale as one of the user inputs and display it somewhere.  But I have to test what you did before saying more.

alex-w:
> First thoughts: I think we should keep Bortle scale as one of the user inputs and display it somewhere. 

But “manual” slider is a smooth Bortle’s scale by the fact.

10110111:
I think one more screenshot will help:

Screenshot - 190122 - 00:07:23

Atque:
@10110111 I notice that the light pollution is displayed without smooth steps, despite the slider. Maybe this looks better with your atmosphere model?

Also, rounding the apparent magnitude estimate might be advisable.

10110111:
> I notice that the light pollution is displayed without smooth steps, despite the slider

The sky brightness changes smoothly AFAICT. But the stars indeed appear in groups, which I suppose is due to the algorithm that computes their visibility that maybe uses Bortle index for this. I think this should be fixed in a later change, not here.

> Also, rounding the apparent magnitude estimate might be advisable

Indeed, didn't think of this. Would `X.Y` format be sufficient?

gzotti:
I assume (without checking the sources) stars and their visibility are only computed in 0.1mag bins. A NELM display rounded to 0.1mag steps should be sufficient. Of course, the µCd/m² can also rounded a bit for the display unless showing all digits makes really sense. 
Somewhere in the GUI I think you should also provide the Bortle index for the currently set circumstances, to make a transition for users used to it easier. As said, I really want to check the PR in detail later this week. Of course, these considerations and PR should also provide a section on the renewed light pollution for the User Guide, new ch. 11. For this I could provide a nomogram from my thesis which we could further annotate with the Bortle scale ranges.
image

10110111:
OK, I've tweaked the tooltip a bit (this commit will be squashed into the previous one if it looks OK):

Screenshot - 190122 - 15:13:26

10110111:
Another way, to reduce the total width of the tooltip, is to use a line break instead of the semicolon:

Screenshot - 190122 - 15:20:05

I don't push this yet, it's for the reviewers to choose.

gzotti:
Yes, that's nice. (two-line solution)

gzotti:
I just compiled and started it. Now I wonder which conversion takes place here? In all my textbooks I can read "naked eye" stars go down to mag 6.5, with a handful of observers being able to go somewhat beyond 7 (7.2?). Who can see reliably mag8 stars? Do you have references for that?

Sliding the slider I agree with Atque that this seems to go in distinct steps. May be connected to MilkyWay calling int bortle=drawer->getBortleScaleIndex(), and likewise LandscapeMgr and ZodiacalLight. These all jump in int steps.

Atque:
@gzotti The article in Sky & Telescope, which is the first appearance of the Bortle dark sky scale, cites that the naked eye limiting magnitude is about 8 at a class 1 site. 

https://skyandtelescope.org/astronomy-resources/light-pollution-and-astronomy-the-bortle-dark-sky-scale/

Also, the current tooltip in Stellarium claims the NELM is 8 at Bortle 1.

gzotti:
Yes, I asked because of the tooltip. Thanks for the link, I must re-read this. Maybe we even need a visual acuity model that influences max NELM? (Too much Voodoo here?)

10110111:
> Who can see reliably mag8 stars? Do you have references for that?

My reference [1] given [here](https://github.com/Stellarium/stellarium/pull/2179/commits/70e4210ca0f2e1abca5f95f2ae98012f2d31cd57#diff-53376f69c53ee3d4d12864eb5cc0ba116d834949563ad3130aa88b01a842f1d1R2770), in section 4 _"Visual Recovery of Halley's Comet"_ cites experimental verification of O'Meara's ability to consistently see stars of magnitude 8.4 with unaided eyes.

Besides, as Atque says, Bortle himself [2] mentions the 8.0 NELM (marking it as "with effort").

Both of the references are given in my first commit.

### References

1: Schaefer, B. E.. "Telescopic limiting magnitudes". Astronomical Society of the Pacific, Publications (ISSN 0004-6280), vol. 102, Feb. 1990, p. 212-229. Available online [here](http://adsbit.harvard.edu/cgi-bin/nph-iarticle_query?bibcode=1990PASP..102..212S)

2: Bortle, John E. (February 2001). "Gauging Light Pollution: The Bortle Dark-Sky Scale". Sky & Telescope. Sky Publishing Corporation. Available online [here](https://skyandtelescope.org/astronomy-resources/light-pollution-and-astronomy-the-bortle-dark-sky-scale/).

alex-w:
@10110111 please resolve conflicts.

gzotti:
When I set "Manual from SQL", "cd/m²", I cannot increase the value beyond 26,5, where the tooltip shows Bortle1/NELM8. 
The value is bounded to 26,5 even with mcd/m² and µcd/m². Presumably the 26,5 are meaningful only for the setting  mag/arcsec²?

Maybe the Tooltip could really show all variants: Bortle index, NELM, mcd/m², mag/arcsec² in 4 lines to make all users happy. The mcd/m² may have to be auto-detecting which of cd/mcd/µcd are best suitable. (Just check log<sub>10</sub>.) When the sun is above horizon (or cd/m² is too high), probably only cd/m² makes sense. 

I know the name O'Meara of course, but I thought he was a singularity not applicable to other humans. I must still re-read the Schaefer paper to put his data into my graph. (This is not a merge stopper!)

It would be nice if you can work out a new balancing in MilkyWay and ZodiacalLight that calls drawer->getLightpollutionLuminance() directly instead of getBortleScaleIndex(), to avoid the steps which @Atque has mentioned.

10110111:
> The value is bounded to 26,5 even with mcd/m² and µcd/m². Presumably the 26,5 are meaningful only for the setting mag/arcsec²?

My fault, the spinboxes weren't being swapped on unit change, only on mode change. I've force-pushed the fix.

> When the sun is above horizon (or cd/m² is too high), probably only cd/m² makes sense.

We are setting up light pollution, not general sky brightness, so the Sun being above horizon is irrelevant.

10110111:
> The mcd/m² may have to be auto-detecting which of cd/mcd/µcd are best suitable. (Just check log10.)

This is an entry field, not a display label. The user decides what is to be entered, the software just shows the current value in user-chosen units.

gzotti:
Thanks for the spinner fix, works nicely!

> > The mcd/m² may have to be auto-detecting which of cd/mcd/µcd are best suitable. (Just check log10.)
> 
> This is an entry field, not a display label. The user decides what is to be entered, the software just shows the current value in user-chosen units.

Wait, when the slider is visible, there is no configured unit. Or do you remember which of them was last configured? This is also OK. Then we may have 3 lines, although all 4 would also be nice.

10110111:
> Or do you remember which of them was last configured?

Exactly, see [here](https://github.com/Stellarium/stellarium/pull/2179/commits/1ebdc02b71431f28ce237d735f196e2c2fd07a32#diff-e3a970902dacbfcafa536bf7a90947f059f4156b909d4a85d8ee9f9b8f70a7a6R159).

10110111:
> Then we may have 3 lines, although all 4 would also be nice.

Isn't the tooltip already a bit overloaded? Do you really think there should be so much info in a tooltip? The numeric value of brightness is already a mode-switch away. If the user needs to set up at this level, then the slider isn't the mode to use.

gzotti:
What about a label next to the slider that permanently displays the NELM and/or Bortle value, (Similar to the sporadic meteor count.), and a tooltip that shows cd/m² and the textual description of Bortle. This would free the user from having to point and make the tooltip pop up every time he only wants to look, not change the value. I know this is a difficult and nasty question around what we think is best for the "typical" user. Keep the simple "Bortle tradition" this program has had for more than the past decade, add a likewise (or more, for some) useful NELM value, and provide instrumental values as well for those who are still deeper in the subject, while still not overloading the GUI.

10110111:
The problem with adding another widget next to the slider is that the slider is already shorter than I would like it to be. Another label will steal a big chunk from this length. Any additional widget like this label will require overflowing to the next line, which seems undesirable with the current layout.

gzotti:
Yes, it's still all on 1 line. Stupid that there is still no slider widget with built-in value display. Then for now, I would go for 3-4 lines tooltip which shows the values for all types of users while the user is sliding, and the slider position must be feedback enough. Maybe we can optimize further later. @alex-w ?

alex-w:
I see no problem to adding the second line into widget. Where is I am wrong?

gzotti:
Currently there would even be room for it, indeed. Restructuring takes place every 1-3 years or so :-)

alex-w:
Hmm... I checked the GUI again and I no need to extra data in tooltip for slider. I think the current solution is enough.

10110111:
> It would be nice if you can work out a new balancing in MilkyWay and ZodiacalLight that calls drawer->getLightpollutionLuminance() directly instead of getBortleScaleIndex(), to avoid the steps which @Atque has mentioned.

Meanwhile I tried doing this and found that the more noticeable problem is the brightness of stars that also changes in steps (the problems mentioned above I didn't even notice). Any pointers to the module responsible for this?

gzotti:
Ah yes, look in StelSkyDrawer::update: setInputScale. You are entering decade-old code here :-) bortleToInScale. Looks like Fabien did this first fine-tuning in 2008, I did something small in 2011, and Alexander in 2016. 
I am afraid there is much hand-tuning here, see the sparse documentation.

10110111:
OK, the smoothing is done. For the stars I've implemented the exponential fit from the figure.

alex-w:
I think this feature is done and ready to merge. 

@gzotti ?

gzotti:
Ah yes. I see the feature works smoothly now, although I think can see a slight difference against 0.21.3 around Bortle class 4. It may be necessary to re-calibrate something later. 
What I am missing is a discussion of these new things like NELM and SQM in the User Guide. In the optimal case the light pollution model is described in a section of chapter 11, or as extension to Appendix B. The minimum is a change in section 4.4.1.

10110111:
I've updated the 4.4.1 section to list and explain the new manipulators (didn't add the regenerated PDF, just edited the `.tex` and a screenshot).

The things like NELM and SQM had already been mentioned in each entry of appendix B, without any discussion of their meaning. Thus I think this addition should be independent of this PR.

gzotti:
Until now all the user had was the Bortle scale, with indeed unexplained other entries in App.B. To probably 99.8% of users it was probably unneeded extra info, just like the structure of out star catalogs. Now NELM and even SQM readings come into the GUI. Somebody who is as familiar as you are with this topic *right now* should add a page (or whatever it takes) of information. And more scientifically inclined users may even want to know the gory implementation details, what is "perfect" (according to whose model!), what is "shortened", or where "preliminary 'good enough' solutions" have been used for years which deviate from the models from the literature. I know from my own reverse-engineering it is not easy to identify undocumented fragments of various sky models, especially when they are so entangled as here. But it would be really good to identify and finally document those sources.

10110111:
OK, I've added a section on light pollution to chapter 11.

gzotti:
OK, that's good for now. You can of course add your name to the chapter authors! Else, approved!

10110111:
Let's just have it merged.

timegate:
@gzotti Is there an API changing NELM insted of BortleIndex?, or could you give me a rough guideline to make it?

PR #2180, opened 2022-01-16
Refactor: getAngularSize->getAngularRadius

gzotti:
It occurred to me that we need to rename this method to clarify that these values are in fact radii, not sizes
See the documentation in StelObject.h

### Description

Fixes # (issue)

### Type of change

### TODOs
- [x] Rename method
- [x] add default implementation to StelObject base class
- [x] clarify CustomObject's weird size
- [x] clarify Nebula's intermediate result. Better return majorAxisSize/2? Adapt other code accordingly.
- [x] clarify why MeteorShowers, Quasars, Pulsars, Exoplanet systems, Novae, Supernovae have non-zero but some distinct fixed small radii? Proposal: remove method override to inherit StelObject's zero size.
- [x] clarify for the Satellite class what RCS is (document the variables!!!) and how to derive radius, not diameter (as it is maybe now), from that. Adapt other places in the satellites code accordingly!

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win10 (irrelevant)
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/plugins/Exoplanets/src/Exoplanet.cpp
+++ b/plugins/Exoplanets/src/Exoplanet.cpp
@@ -557,11 +557,6 @@ bool Exoplanet::isVMagnitudeDefined() const
 	return Vmag<98.;
 }
 
-double Exoplanet::getAngularRadius(const StelCore*) const
-{
-	return 0.0001;
-}
-
 bool Exoplanet::isDiscovered(const StelCore *core)
 {
 	int year, month, day;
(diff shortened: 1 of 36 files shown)


gzotti:
@alex-w several classes have small arbitrary but distinctly non-zero sizes. The size for stars is zero. Why not also for these classes? A non-zero value implies we could enlarge it to see more than a dot.
The Nebula and Satellite I think do not follow the specification and return diameters, not radii. This should be fixed finally in this branch.

gzotti:
@alex-w I have added 2 TODO annotations in the StelObject::getInfoMap(): I don't understand the test for negative values. (Just add an assert if in doubt?) And why do we need size in radians?

alex-w:
> @alex-w I have added 2 TODO annotations in the StelObject::getInfoMap(): I don't understand the test for negative values. (Just add an assert if in doubt?) And why do we need size in radians?

I don't remember details, probably this part was added into map to backward compatibility. I think we may delete this data.

alex-w:
> @alex-w several classes have small arbitrary but distinctly non-zero sizes. The size for stars is zero. Why not also for these classes? A non-zero value implies we could enlarge it to see more than a dot.

I've refactor this stuff. The most of these non-zero values was added a long time ago (around series 0.11-0.12) and probably on some step of refactoring we just forgot the fix this points.

gzotti:
I think I caught all cases now.

alex-w (approved):
Thanks!

Issue #407, opened 2018-09-26
Calculates the eclipses


alex-w:
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1106741

This feature request from old bug tracker.

It would be good that Stellarium can show and calculates the eclipses and the planets positions up to 100 years in the future and also in the past...like Redshift can do this.

Type of eclipses:
- [x] Lunar eclipses
- [x] Solar eclipses

worachate001:
Is anyone working on this issue or planning to do it soon?

It should be in "Astronomical Calculations", right?

alex-w:
> Is anyone working on this issue or planning to do it soon?

The task in my TODO list, but not for near future
 
> It should be in "Astronomical Calculations", right?

Yes

worachate001:
I'm working on the lunar eclipse. This is an example of what I have so far. You can search for lunar eclipses and go to the moment of greatest eclipse (Note: date and time are in my zone - UTC+7).

![lunareclipseforstellarium](https://user-images.githubusercontent.com/41257965/150092252-b6d10491-a3c4-4a92-8145-f584896179bb.png)

I'm using the icon from transit because I don't know how to add a new one for lunar eclipse (and still don't have it).

Hope to create a pull request soon for review and testing. After this, I will try to create something like this for solar eclipses.

alex-w:
See for example: https://github.com/Stellarium/stellarium/blob/master/data/gui/tabEclipses.png

worachate001:
In my draft of the pull request, it says worachate001 wants to merge 7 commits into Stellarium:master from worachate001:LunarEclipseCalc. It appears that the request include previous changes years ago. I'm not sure how to get rid or manage this situation (the changes look okay with just 3 files that I have made). If I go ahead with the pull request, it shouldn't be a problem, right?

gzotti:
In the tab "Files changed" we can see that 3 files were changed. the total number of commits is usually not important when we squash all changes into one commit on merge. Only large changes are merged as sequence of commits. But we need to review your work first :-)

worachate001:
This is Solar Eclipses Tab for the AstroCalc. I believe this is enough for summary of the eclipse. I can write the code to find duration of totality/annularity, but it will take more time to calculate. Any suggestions are welcome.

![solareclipses](https://user-images.githubusercontent.com/41257965/151186510-159ffe5d-7f28-4d7c-965b-9e8694e3a711.png)

gzotti:
[...]

Issue #2050, opened 2021-11-19
Diurnal aberration


Atque:
Diurnal aberration is the aberration caused by the rotation of the body on which the observer resides. On Earth, it is quite small (about 0.3 arcseconds at the equator, compared to 20 arc seconds of the annual aberration), but I would imagine it is quite a bit larger on e.g. Jupiter with its much faster rotation period. While I understand that Stellarium's goal is not to perfectly simulate the conditions on Jupiter, I think diurnal aberration should at least be given a thought.

gzotti:
It is rare that I observe from Jupiter...

As you have seen with the stellar occultation problem, the Moon as of now not even has a dependable radius but is rendered as disco-ball where it is too small by up to a few arcseconds. 

My personal target was aiming at arcsecond accuracy, which we have reached or rather superseded with 0.21.2. This is enough for all practical purposes of finding objects. 

At this point we therefore accept working pull requests. I may go after this when an encyclopedia of other features has been done, else this has super low priority.

PR #2190, opened 2022-01-19
Add Lunar Eclipses to Astronomical calculations

worachate001:
### Description

This PR partially addresses #407 by adding the method to search for lunar eclipses during user-input time span.

It shows:
- Date of time of greatest eclipse.
- Saros series
- Type of eclipses
- Penumbral and umbral magnitudes at greatest eclipse

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: NVidia Geforce RTX

## Checklist:


alex-w:
I think the main tab should be named "Eclipses" and it should contains at least 2 "inner" tabs (similar to Graphs) - "Lunar eclipses" and "Solar eclipses". Probably "Solar eclipses" subtab in future may be splitted on 2 subtabs.

worachate001:
I got the idea, it will take some time to add solar eclipse calculations in the second tab. I think I should make it to be draft for now.

alex-w:
> I got the idea, it will take some time to add solar eclipse calculations in the second tab. I think I should make it to be draft for now.

No need implement computation of solar eclipses right now in this branch, but prepared to compute it will be good step.

alex-w:
Apparently the shortest way to preparation the support of both types of eclipses will be renaming the main tab to "Eclipses" at this step.

alex-w:
I'm checked the work of tool with Astronomical Calendars and results is very good. Thank you!

So, the cosmetic fixes:
1) The phases data should be aligned to right side for improve readability and no need 4 digits for accuracy - 3 digits is enough (we use 1 digit after dot for percentage info about phases in infostring also)
2) The phase cannot be negative, so, all negative values should replaced by "dash" char (QChar(0x2014))

alex-w (approved):
Thank you very much!

alex-w:
@gzotti ?

src/gui/AstroCalcDialog.cpp
```diff
@@ -2159,6 +2174,342 @@ void AstroCalcDialog::saveTransits()
...
+				const double uMag = (L2 - m) / (2. * mSD); // umbral magnitude
+
+				if (pMag>0.)
+				{
+					EclipseTypeStr = "Penumbral";
+					if (uMag>=1.) EclipseTypeStr = "Total";
+					if (uMag>0. && uMag<1.) EclipseTypeStr = "Partial";
+
+					// Saros series Calculation - useful to search for eclipses in the same Saros
+					// Adapted from Solar Eclipse Saros calculation in S&T magazine (seem to matching well with data from NASA)
```
gzotti, line 2360: Thanks for the documentation, but please add the issue number.

src/gui/AstroCalcDialog.hpp
```diff
@@ -159,6 +159,17 @@ class AstroCalcDialog : public StelDialog
 		GraphDeclinationVsTime      = 10
 	};
 
+	//! Defines the number and the order of the columns in the lunar eclipse table
+	//! @enum LunarEclipseColumns
+	enum LunarEclipseColumns {
+		LunarEclipseDate,            //! date and time of lunar ecipse
```
gzotti, line 165: typo: ecipse->eclipse

src/gui/AstroCalcDialog.hpp
```diff
@@ -203,6 +214,12 @@ private slots:
 	void saveTransits();
 	void setTransitCelestialBodyName();
 
+	//! Calculating lunar eclipses to fill the list.
```
gzotti, line 217: Please add any references here. "Algorithm taken from..." (or developed by yourself?)

gzotti:
@worachate001 Thank you for this very welcome addition. I noticed however that the center times of this year's eclipses are off by about 1/2 minute. Are the displayed times approximate, or should they provide the displayed seconds of accuracy? Then something is not yet finished.

worachate001:
@gzotti Thank you for the comments. Looking at the values, I think the time of greatest eclipse are accurate to within a few seconds (you have to adjust delta T for comparison and use the same ephemeris). I'll try to solve all of comments and investigate deeper.

src/gui/AstroCalcDialog.hpp
worachate001: This part is just follow the same pattern from a few lines above it. I think you mean the AstroCalcDialog.cpp?

gzotti: The header file is the place for general developer documentation (Doxygen format). Literature information (if available) should go here. Inside the cpp file of course more details can be given in further comments.

gzotti:
> @gzotti Thank you for the comments. Looking at the values, I think the time of greatest eclipse are accurate to within a few seconds (you have to adjust delta T for comparison and use the same ephemeris). I'll try to solve all of comments and investigate deeper.

Do you give DT or zone time results? It seems we are off by 1/2 minute, but DeltaT is 72 seconds, so either you are searching by some step width and don't refine enough (then you may want to round to the nearest minute), or there is something else.

worachate001:
Comparisons with https://eclipse.gsfc.nasa.gov/LEcat5/LE2001-2100.html

2021 May 26  11:19:53 TDT (dt=72) = 11:18:41 UT1 Stellarium = 11:18:40 UT1 
2021 Nov 19  09:04:06 TDT (dt=73) = 09:02:53 UT1 Stellarium = 09:02:52 UT1
2022 May 16  04:12:42 TDT (dt=73) = 04:11:29 UT1 Stellarium = 04:11:28 UT1
2022 Nov 08  11:00:22 TDT (dt=73) = 10:59:09 UT1 Stellarium = 10:59:08 UT1 
2023 May 05  17:24:05 TDT (dt=73) = 17:22:52 UT1 Stellarium = 17:22:50 UT1
2023 Oct 28  20:15:18 TDT (dt=74) = 20:14:04 UT1 Stellarium = 20:14:03 UT1

We are very accurate!

src/gui/AstroCalcDialog.hpp
worachate001: I hope I did right here. If not, please educate me. I'm still in the learning process.

alex-w:
@gzotti please review the changes. I got accuracy around 1 second for my location.

@worachate001 I think "Saros number" for name of second column will be more accurate term.

worachate001:
Thanks for suggestions. While waiting for the reviews, I have been preparing the solar eclipse calculations and almost finish, it would be ready in a few days.

src/gui/AstroCalcDialog.cpp
```diff
@@ -256,6 +259,8 @@ void AstroCalcDialog::createDialogContent()
 	ui->phenomenToDateEdit->setDateTime(currentDT.addMonths(1));
 	ui->transitFromDateEdit->setDateTime(currentDT);
 	ui->transitToDateEdit->setDateTime(currentDT.addMonths(1));
+	ui->lunareclipseFromDateEdit->setDateTime(currentDT);
```
alex-w, line 262: Maybe change default values for both fields will be good idea: the first field should contains first day of current year and second field - the last day of current year. This is cosmetics.

@worachate001 @gzotti how about this idea?

src/gui/AstroCalcDialog.cpp
```diff
@@ -2159,6 +2174,356 @@ void AstroCalcDialog::saveTransits()
 	}
 }
 
+void AstroCalcDialog::setLunarEclipseHeaderNames()
```
alex-w, line 2177: Please add the call of this method into retranslate() - this step will solve problem of translation the column names when language is changed.

src/gui/AstroCalcDialog.cpp
gzotti: Yes, indeed. A default of "search all eclipses of this year" sounds reasonable. Alternatively, from "today" to "1 year in the future".

src/gui/AstroCalcDialog.cpp
```diff
@@ -2159,6 +2174,356 @@ void AstroCalcDialog::saveTransits()
...
+	if (onEarth) // Not sure it's right thing to do but should be ok.
+	{
+		initListLunarEclipse();
+
+		const double currentJD = core->getJD();   // save current JD
+		double startJD = StelUtils::qDateTimeToJd(QDateTime(ui->lunareclipseFromDateEdit->date()));
+		double stopJD = StelUtils::qDateTimeToJd(QDateTime(ui->lunareclipseToDateEdit->date()));
+		startJD = startJD - core->getUTCOffset(startJD) / 24.;
+		stopJD = stopJD - core->getUTCOffset(stopJD) / 24.;
+		int elements = static_cast<int>((stopJD - startJD) / 29.530588853);
```
alex-w, line 2247: I think values `29.530588853` and `2451550.09765` should be defined as `const double` variables and just called in right places

gzotti:
Something is still weird. Please look here and compare with the list above.
grafik
What am I doing wrong? &Delta;T algorithm is set to default.

alex-w:
@gzotti please share your `config.ini` file

gzotti:
[config.ini.txt](https://github.com/Stellarium/stellarium/files/7920934/config.ini.txt)
(renamed for github)

alex-w:
> [config.ini.txt](https://github.com/Stellarium/stellarium/files/7920934/config.ini.txt)
> (renamed for github)

ΔT = 72.73 seconds with your config.ini file... WTF?!?

Atque:
@gzotti your config says that aberration is off. Could that be the reason?

alex-w:
> @gzotti your config says that aberration is off. Could that be the reason?

Yes, of course! After enabling aberration:
stellarium-000

gzotti:
@Atque, many thanks! Indeed. On one of my other tasks I had switched that off and was able to overlook it...

I also just checked with WSL. (Its default config.ini). It shows your values. So I welcome this addition, @worachate001 many thanks!

alex-w:
@worachate001 I'll merge it now and I'll add changes from my comments in the master after merge.

worachate001:
I just woke up in the morning here, so I missed everything. Thanks for additional improvements.

worachate001:
I have noticed that the time of greatest eclipse magnitude is always 1-2 seconds ahead of official time for every eclipses!. So it must be systematic errors -- the Moon is a little ahead of the location where it should be.

Is it possible that this may caused by the fact that aberration is not currently applied to the Moon?

Atque:
@worachate001 It could be that Stellarium currently does not support diurnal aberration. I have noticed that lunar occultations are also off by 1 or 2 seconds when comparing to e.g. Cartes du Ciel (which takes diurnal aberration into account).

alex-w:
Aberration is applied to the Moon now. See PR ;)

alex-w:
see PR #2050

worachate001:
In this window, no change of coordinates for the Moon when checked and uncheck. But for the Sun and planets, they changes. Is it normal?

aberration

gzotti:
We have discussed that in summer. See #1626 . It seems that for the Moon aberration may be included in the ephemeris already. 

I have modified the lunar position somewhat to reflect the difference between center of figure and center of gravity. 
See EphemWrapper.cpp line 418. This explains exactly those 1-2 seconds, but should improve accuracy.
 
If diurnal aberration is noticeable in stellar occultations by the moon by more than 1 second or so, it may be worth considering. But it is not in my priority list.

worachate001:
@gzotti Thank you for explanation. Modifying the Moon's position that way is really explain the differences. I have read that Astronomical Almanac is using this method but Fred Espenak decided to ignore it in his Catalog of Lunar Eclipses.

gzotti:
Usually a difference of 2 seconds should not pose a problem in long-time eclipse canons, given uncertainties in &Delta;T. We would need a couple of (or rather a few thousand) securely timed (observed, not precomputed!) lunar occultations (intros and exits) to understand the limits of the current solution in Stellarium. I do not know the real figure offset between a perfect sphere and our tessellated (triangulated) polygonal Moon. See also discussion in #1959. Again, occultations are not my primary interest, but if somebody finds a solution for #1959, this would also be welcome of course. Anybody in IOTA reading this and and want to help? Of course, we have no model for the Lunar rim, so a difference with real-world conditions will likely persist. (I don't know if the displacement map can be interpreted and rendered in terms of real geometry to make the lunar rim appear jagged with the correct mountain silhouette.)

PR #2195, opened 2022-01-20
Compactification of the GUI (Atmosphere)

alex-w:
### Description
This pull request is doing the buttons in the View dialog is more compact. Plus settings for refraction, extinction and tone mapping are moved into one dialog. 

Updating the screenshots and text in the Stellarium User Guide will be after approvement the code and before merge.

view-dialog

### Type of change

## Checklist:


gzotti:
Hmm, good and bad. It's OK to have smaller buttons, but tone mapping is a very distinct thing from atmospheric settings. These should remain separate GUI panels IMHO. If you need an icon, it should be something like for contrast settings in other programs, maybe a gamma curve running through some gray ramp. Actually, more parts of the program could be switched to real-world luminance and tonemapping, e.g. the landscapes. Currently we do our own careful balancing, but why not let color science do its miracles?

PR #2200, opened 2022-01-21
Switch to use C++17 standard

alex-w:
Switch to use C++17 standard when compiling the source code

### DIFF
--- a/.lgtm.yml
+++ b/.lgtm.yml
@@ -2,17 +2,23 @@ extraction:
   cpp:
     prepare:
       packages:
-      - "libqt5serialport5-dev"
+      - "qtbase5-dev"
       - "qtscript5-dev"
       - "libqt5svg5-dev"
-      - "qttools5-dev-tools"
       - "qttools5-dev"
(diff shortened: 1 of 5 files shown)


.lgtm.yml
```diff
@@ -2,17 +2,23 @@ extraction:
...
-      - "libqt5serialport5-dev"
+      - "qtbase5-dev"
       - "qtscript5-dev"
       - "libqt5svg5-dev"
-      - "qttools5-dev-tools"
       - "qttools5-dev"
+      - "qttools5-dev-tools"
       - "libqt5opengl5-dev"
-      - "qtpositioning5-dev"
-      - "libgps-dev"
```
gzotti, line 12: Delete libgps-dev?

alex-w: Oops... sorry for missing it

gzotti (approved):
Not sure if removing libgps-dev is correct.

Else, no further objection. I just upgraded to Qt5.15.2, it still works with MSVC17.

PR #2228, opened 2022-02-02
Moon RTS

### LINKED ISSUE #2024: Rise, transit and setting time changes during the same day
Rise, transit and set times of an object should remain constant during the same day. But in Stellarium if you fast forward of some hours, the rise, transit and set time change two or more times in the same day.

### To reproduce
Select an object (e.g. the Moon) when it's just rised over the horizon. Start forwarding time  by hitting the +1 hour button. Monitor the rise, set and transit time in the information list: they will change a couple of times during the same day.

### Discussion
While it could be argue that after an event (e.g. after an object has rised over the horizon) that information could be refered to the next event (the one from the day after), this is not a suitable behaviour in my opinion, since people would like to know when is the time of that event in that specific day. But even so, these times should not change more than one time per dey.

### System
* Stellarium version: 0.21.0
* Operating system: W11
* Graphics Card: Nvidia

worachate001:
### Description
Rise, Transit and Set times of the Moon are not stable throughout the day. I have a solution that seem to effectively fix it. The idea is recompute a few more times at the instance of RTS (no interpolation involved). The algorithms is only applied to the Moon (for now). It may be able to apply to other objects in the future.

Fixes #2024 

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -4606,6 +4606,7 @@ Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
 		{
 			core->setJD(currentJD+mt);
 			core->update(0);
+			ho = - getAngularRadius(core) * M_PI_180; // semidiameter;
 			Theta2=obsPlanet->getSiderealTime(currentJD+mt, currentJDE+mt) * (M_PI/180.) + L;  // [radians]
 			StelUtils::rectToSphe(&ra, &de, ssystem->getMoon()->getEquinoxEquatorialPos(core));
 			cosH0=(sin(ho)-sin(phi)*sin(de))/(cos(phi)*cos(de));
@@ -4618,6 +4619,7 @@ Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
 		{
 			core->setJD(currentJD+mr);
 			core->update(0);
+			ho = - getAngularRadius(core) * M_PI_180; // semidiameter;
 			Theta2=obsPlanet->getSiderealTime(currentJD+mr, currentJDE+mr) * (M_PI/180.) + L;  // [radians]
 			StelUtils::rectToSphe(&ra, &de, ssystem->getMoon()->getEquinoxEquatorialPos(core));
 			cosH0=(sin(ho)-sin(phi)*sin(de))/(cos(phi)*cos(de));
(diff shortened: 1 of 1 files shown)


gzotti:
Thank you! I have also considered refining our solution using binary search as there is no direct way when taking parallax shift into account. I am busy with other things right now but want to check this in the next days. In principle this is needed for all close and fast objects, fortunately this is only the Moon :-)

worachate001:
@gzotti Yes. My very next target are Phobos and Deimos. I see that there are large errors there.

alex-w:
If you want to fix RTS for Phobos & Deimos also, then some checks should be changed IMHO. For example: `if (getParent()==core->getCurrentPlanet())` to check if selected body is the moon of current planet.

worachate001:
Thanks @alex-w Should we push current PR before looking at other cases?

alex-w:
> Should we push current PR before looking at other cases?

Probably "yes",  because other cases is other cases and all of them do not touches specific bug.

> I think it's looking good now for the Moon.

A quick-check gives perfect data for the Moon! Thanks!

alex-w (approved):
Thanks!

@gzotti?

gzotti:
Hmm, moon looks perfect now, indeed, however, at cost of setting core->setJD() which initiates all other computations (whole SolarSystem). (using core1 is camouflage to avoid the const thing, but there is only 1 core.) My solution (IIRC) avoided that, but yes, the atmosphere and parallax are bad when we need high accuracy. 

The display in the "Files changed" tab here looks strange and confusing. @worachate001 Did you only insert the if(Earth/Moon) case, or did you also change the result of the more general planet solution? 

I think the next extension could be a variant which finds RTS when sun/moon are sitting "above" the horizon (Just invert h0!), and with center in horizon. Not sure if we need all distinct versions, or rather configure which variant of h0 should be used. 

And then finally variants of all those ideas which takes the landscape into account (start with current solution, then test with landscape->getOpacity(azAlt), increase/decrease time in minute steps to avoid holes in the texture. I envisioned to have this around v0.13... ).

worachate001:
@gzotti I just inset if ((getEnglishName()=="Moon") && (loc.planetName=="Earth")) and keep all the old codes. That means all other objects are using the previous algorithm. I have tested the new method with some normal moons like Io, Europa, ..., it seems to work well, but for fast movers like Phobos/Deimos, we need another solution.

gzotti:
OK, then we can merge this part so far. 
For Mars, we must likely do something else, when the orbital duration is less than the planet's rotation period. Then some principal assumptions fail. Maybe a binary-search for 1/2 orbit durations around "now" for rise/set events. It would be good to do that with only the planet/moon, not the full core, but I cannot work it out now, this gets quite messy pretty soon :-)

alex-w:
> The display in the "Files changed" tab here looks strange and confusing. @worachate001 Did you only insert the if(Earth/Moon) case, or did you also change the result of the more general planet solution?

It looks like EOL or formatting settings issue

alex-w:
OK, @gzotti  I'll merge it

PR #2229, opened 2022-02-02
Calendars: Astro functions

gzotti:
### Description

This adds the ~100 astro functions required for astronomically guided calendars in the second half of the book.
I have tried to use Stellarium's functions and data types (e.g. StelLocation) where applicable. 

So far I cannot test those functions. Whenever a StelCore is needed, the test functions fail. So it's unlikely that this is all error-free...

Fixes # (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win10
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/StelUtils.cpp
+++ b/src/core/StelUtils.cpp
@@ -1776,9 +1776,7 @@ double getDeltaTByReingoldDershowitz(const double jDay)
 	}
 	else if ((year >= 1800) && (year <= 1986))
 	{
-		// FIXME: This part should be check because this part gives the strange values of DeltaT (see unit tests)
 		double c = (getFixedFromGregorian(1900, 1, 1)-getFixedFromGregorian(year, 7, 1))/36525.;
-
 		if (year >= 1900) // [1900..1986]
 		{
 			deltaT = ((((((-0.212591*c + 0.677066)*c - 0.861938)*c + 0.553040)*c - 0.181133)*c + 0.025184)*c + 0.000297)*c - 0.00002;
@@ -2363,10 +2361,15 @@ float *ComputeCosSinRhoZone(const float dRho, const unsigned int segments, const
 
 int getFixedFromGregorian(const int year, const int month, const int day)
 {
+	bool leap = false;
+	if (year % 100 == 0)
+		leap = (year % 400 == 0);
+	else
+		leap = (year % 4 == 0);
 	int y = year - 1;
 	int r = 365*y + intFloorDiv(y, 4) - intFloorDiv(y, 100) + intFloorDiv(y, 400) + (367*month-362)/12 + day;
 	if (month>2)
-		r += (isLeapYear(year) ? -1 : -2);
+		r += (leap ? -1 : -2);
 
 	return r;
 }
--- a/src/core/StelUtils.cpp
+++ b/src/core/StelUtils.cpp
@@ -1760,7 +1760,7 @@ double getDeltaTByReingoldDershowitz(const double jDay)
 	{
 		// [2051..2150]
 		double x = (year-1820)/100.;
-		deltaT = (- 20 + 32*x*x + 0.5628*(2150-year));
+		deltaT = (- 20 + 32*x*x - 0.5628*(2150-year));
 	}
 	else if ((year >= 1987) && (year <= 2050))
 	{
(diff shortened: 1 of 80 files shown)


alex-w:
> So far I cannot test those functions. Whenever a StelCore is needed, the test functions fail. So it's unlikely that this is all error-free...

This is correct behaviour, because StelApp is singleton. Please check DeltaT test for R&D algorithm.

gzotti:
> > So far I cannot test those functions. Whenever a StelCore is needed, the test functions fail. So it's unlikely that this is all error-free...
> 
> This is correct behaviour, because StelApp is singleton. 

Yes, the CI tests run in debug mode and test for this, and therefore fail as well...

> Please check DeltaT test for R&D algorithm.

I have fixed one error from the latest official error list from R&D. Else, in the program, I use the core's currently used DeltaT algorithm. R&D's version is just another reformulation of Meeus AA1 (1991), and can actually be removed.

Not sure how to test and continue with these things now. Just implement further calendars on top, and see if they work in the program? (CI/tests won't work for those!)

One idea of a future benefit from implementing all this would be scripting accessibility. Just make slots from them? This could also allow at least manual/semi-automatic testing.

alex-w:
> > Please check DeltaT test for R&D algorithm.
> 
> I have fixed one error from the latest official error list from R&D. Else, in the program, I use the core's currently used DeltaT algorithm. R&D's version is just another reformulation of Meeus AA1 (1991), and can actually be removed.

But R&D solution has fits of expanded time range in comparison to AA1.

> Not sure how to test and continue with these things now. Just implement further calendars on top, and see if they work in the program? (CI/tests won't work for those!)

Probably code of methods should be reformed or omitted StelApp/StelCore-related methods from unit testing.

> One idea of a future benefit from implementing all this would be scripting accessibility. Just make slots from them? This could also allow at least manual/semi-automatic testing.

Well, at the moment this is best solution IMHO.

alex-w:
@gzotti could you check isLeap method for Gregorian calendar?

gzotti:
I am stuck with the scripting interface. Just simple QVector<int> which should be converted to JavaScript Arrays. The other useful object to make scriptable in this context would be StelLocation. The current situation is barely enough for scriptable debugging output. @alex-w any idea how to make these?

alex-w:
Oops, sorry, I missed this question. :(

> I am stuck with the scripting interface. Just simple QVector which should be converted to JavaScript Arrays. 

Why? Do you meant `QVector::toList()`? 

> The other useful object to make scriptable in this context would be StelLocation. 

See [setObserverLocation](https://stellarium.org/doc/0.21/classStelMainScriptAPI.html#aa6655a5dc00083085c3fc12646536231) in the scripting's core and both getObserverLocation methods in same "class".

> The current situation is barely enough for scriptable debugging output. @alex-w any idea how to make these?

No idea right now, but... see how Vec3d/Vec2d types was added into scripting engine

gzotti:
I played for a while with additional functions wrapping the QVector with .toList(). But now I found the right conversion/registration (see StelScriptMgr), so this part is solved, and most calendar functions are scriptable :-). 

Thanks for the hint about setObserverLocation, this may help. And some sleep before doing this ;-)

alex-w:
Maybe adding context for calendar names in the GUI will be good idea?

src/core/StelUtils.cpp
```diff
@@ -1776,7 +1776,7 @@ double getDeltaTByReingoldDershowitz(const double jDay)
 	}
 	else if ((year >= 1800) && (year <= 1986))
 	{
-		double c = (getFixedFromGregorian(1900, 1, 1)-getFixedFromGregorian(year, 7, 1))/36525.;
+		const double c = (getFixedFromGregorian(year, 7, 1)-getFixedFromGregorian(1900, 1, 1))/36525.;
```
alex-w, line 1779: Oops… it was my fail :(

gzotti:
> Maybe adding context for calendar names in the GUI will be good idea?

Where do you see necessity? I have added it in `Calendars::draw()` long ago.

src/core/StelUtils.cpp
gzotti: Can happen, but always unfortunate. Finally fixed.

gzotti:
@alex-w I think we can merge this quite large step. Some remaining errors in the astro functions are possible but will only be found as new calendars will come which depend on them.

alex-w (approved):
Thanks!

PR #2237, opened 2022-02-05
Fix bug in AstroCalc/Eclipses where type of local solar eclipses are incorrect

worachate001:
### Description
There is a bug in the AstroCalc's Local Solar Eclipses tool (in development) that must be fixed.

To reproduce
* open 'All Solar Eclipses' tab
* search for eclipses in the year 1986
* double click at 1986 Oct 3
* go to 'Local Solar Eclipses' tab
* search for eclipses in the same year

The result shows that 1986 Oct 3 was a partial eclipse as seen from the location of greatest eclipse along very narrow eclipse path (path width = 0.46 km) which is incorrect. For (unrealistically) perfect sphere Moon, this eclipse was a total eclipse with sub-second duration - 0.049 second - to be exact. In reality, it's beaded total eclipse. Both NASA web site and Stellarium tell us that it's hybrid eclipse - technically, it was total eclipse at mid point.

This PR will correctly classify this rare event as total eclipse as seen from the location and it will also fix similar situation on other dates when you are standing at the point of transition between total and annular eclipse.

Notes:
* This PR will fix the oblateness parameters of the Earth to be consistence across all related parts of Stellarium.
* As we reproduce the bug, we can see that it seems to be a slight difference of Sun & Moon positions at greatest eclipse point for some eclipses. This is still under investigations, would be addressed in a separate issue if I found a way to fix it.

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Nvidia

## Checklist:

### DIFF
--- a/src/gui/AstroCalcDialog.cpp
+++ b/src/gui/AstroCalcDialog.cpp
@@ -3018,11 +3018,15 @@ pSEparameter partialSolarEclipse() {
 	double L1 = bessel.L1;
 	double L2 = bessel.L2;
 	const double mu = bessel.mu;
-	constexpr double e2 = 0.00669398;
-	// e^2 = 0.00669398 : Earth flattening parameter
-	// f = 298.25642 : e^2 = 2f-f^2
-	// Source: IERS Conventions (2003)
+	constexpr double e2 = 0.00669438;
+	// e^2 = 0.00669438 : Earth flattening parameter
+	// Inverse flattening 1/f = 298.257223563 : e^2 = 2f-f^2
+	// Source: 1984 World Geodetic System (WGS 84)
+	// https://web.archive.org/web/20200710203711/http://www.epsg-registry.org/export.htm?gml=urn:ogc:def:ellipsoid:EPSG::7030
+	// There is a modern value from IERS Conventions (2003)
+	// 1/f = 298.25642 But seem to be not widely used
 	// https://www.iers.org/IERS/EN/Publications/TechnicalNotes/tn32.html
+	// We use older value to be comparable with literatures and consistence across Stellarium
 
 	const double rho1 = sqrt(1.- e2 * cos(d) * cos(d));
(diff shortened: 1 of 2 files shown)


alex-w (approved):
Thanks!

src/gui/AstroCalcDialog.cpp
```diff
@@ -3018,11 +3018,15 @@ pSEparameter partialSolarEclipse() {
 	double L1 = bessel.L1;
 	double L2 = bessel.L2;
 	const double mu = bessel.mu;
-	constexpr double e2 = 0.00669398;
-	// e^2 = 0.00669398 : Earth flattening parameter
-	// f = 298.25642 : e^2 = 2f-f^2
-	// Source: IERS Conventions (2003)
+	constexpr double e2 = 0.00669438;
+	// e^2 = 0.00669438 : Earth flattening parameter
+	// Inverse flattening 1/f = 298.257223563 : e^2 = 2f-f^2
```
gzotti, line 3023: We are actually using this as property of the Planet object Earth. See ssystem_major.ini entry for Earth: 
`oblateness=0.003352810664747481
`
This is f. 
The parameter is used in Planet::getOneMinusOblateness() and therefore in Planet::getRectangularCoordinates(), which was crucial to get ... accurate eclipse simulation for 2017. 

For consistency I would prefer if you don't hardcode the value here but use Planet::getOneMinusOblateness(), should we ever want to change to the IERS2003 or other values. But please keep the documentation, this is important for traceability.

gzotti:
@worachate001 Thank you for your meticulous work! It's fine to have contributors who triple-check their work, even for such incredibly rare cases :-)

src/gui/AstroCalcDialog.cpp
worachate001: Thanks, I'll add it in my to do list.

PR #2238, opened 2022-02-05
Give names to some magic constants: radii of Earth, Moon and Sun

10110111:
### Description
These particular magic numbers are scattered over the code base. I came across them when splitting `Atmosphere` into the two implementations and trying to change the code related to solar eclipse. I'd really dislike to write code in this magic-number style, so I thought I'll fix at least these constants. The alternative would be to create a local constant definition, but this seemed not much better than proceeding with hard-coded values.

While doing this, I found that `Meteor` module uses its own `EARTH_RADIUS` constant, which lacks 4 significant digits that are present elsewhere. Besides, it does all the computations in `float`. I changed it to use the global constant (cast to `float`), so the value used will be a bit different now. I expect that this doesn't matter, since calculations in `float` imply that precision is "don't care" in this module.

I decided not to touch the Sun/Earth radii ratio, whose hard-coded values are also scattered over the code base. The numeric value there would change, and I wasn't sure how sensitive the modules that have this are to such changes.

### DIFF
--- a/src/core/StelUtils.hpp
+++ b/src/core/StelUtils.hpp
@@ -38,6 +38,12 @@
 #define SPEED_OF_LIGHT 299792.458
 // Ecliptic obliquity of J2000.0, degrees
 #define EPS_0 23.4392803055555555555556
+// Equatorial radius of the Earth in km
+#define EARTH_RADIUS 6378.1366
+// Equatorial radius of the Sun in km
+#define SUN_RADIUS 696000.
+// Equatorial radius of the Moon in km
+#define MOON_RADIUS 1738.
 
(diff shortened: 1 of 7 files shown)


alex-w (approved):
Thanks!

gzotti:
I just had a similar suggestion by @worachate001 about solar eclipse code. I agree that centralized named (and documented/referenced/explained) constants are better than hardcoded magic numbers which are to be found here and there. The sizes of sun and planets should best be taken from the values which are given in ssystem_major.ini, i.e. each Planet object already knows its `getEquatorialRadius()`.

PR #2243, opened 2022-02-06
Fix solar longitude in Meteor Showers data

worachate001:
### Description
Solar longitudes in Meteor Showers data are now showing negative values. They should be positive.

### Type of change

### DIFF
--- a/plugins/MeteorShowers/src/MeteorShower.cpp
+++ b/plugins/MeteorShowers/src/MeteorShower.cpp
@@ -491,6 +491,7 @@ QString MeteorShower::getSolarLongitude(QDate date)
 	// put it in the range 0 to 360 degrees
 	l /= 360.;
 	l = (l - static_cast<int>(l)) * 360. - 1.;
+	l = StelUtils::fmodpos(l, 360.0);
 
 	return QString::number(l, 'f', 2);
 }


worachate001:
I look at these values again and I strongly believe that this implementation is not fully correct for annual meteor showers. Solar longitudes at ZHRmax are almost fixed value (epoch J2000) for each shower. The thing that changes is date and time of maximum ZHR. I will look deeper to check this again in the future.

plugins/MeteorShowers/src/MeteorShower.cpp
```diff
@@ -491,6 +491,7 @@ QString MeteorShower::getSolarLongitude(QDate date)
 	// put it in the range 0 to 360 degrees
 	l /= 360.;
 	l = (l - static_cast<int>(l)) * 360. - 1.;
```
gzotti, line 493: what does this line do? This and the previous seem to do a modulo 360, but why subtract 1?

worachate001: Feel like they tried to workaround to make it aligned with present day. Anyway, as I said above, this is not the right way to calculate annual meteor showers.

worachate001:
I will make this PR draft for now. The issue is much deeper than just fix the negative values. It's clear for me that this plugin needs more improvements.

worachate001:
The data file showers.json doesn't give anything about solar longitude. I need to add that into the file to make use of it. It means that will affect current and older version of Stellarium, right?

gzotti:
This is easily possible. I hope sources for the model have been documented... Thank you for having a look at this!

PR #2244, opened 2022-02-06
Fix planetocentric rectangular coordinates

worachate001:
### Description
As mentioned in #2237 Sun & Moon are not aligned at greatest solar eclipse. I suspect that something is wrong with observer's coordinates on Earth and it's right.

There is a bug in planetocentric rectangular coordinate calculations that give wrong values of rhoCosPhiPrime and rhoSinPhiPrime. After fixing this issue, the Sun and Moon are aligned perfectly for all central solar eclipses! This should also improve the accuracy of topocentric coordinates for all objects.

### Type of change

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -1416,16 +1416,18 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup& 
 						// distance between center point and current location
 						double distance = loc.distanceKm(lngDeg, latDeg);
 						double azimuth = loc.getAzimuthForLocation(lngDeg, latDeg);
-
 						oss << QString("%1 %2 %3 %4°<br/>").arg(
 								 q_("Shadow center point is"),
 								 QString::number(distance, 'f', 1),
 								 q_("km towards azimuth"),
 								 QString::number(azimuth, 'f', 1));
-						oss << QString("%1: %2 %3<br/>").arg(
-								 q_("Width of umbra/antumbra"),
+						if (dRatio < 1.0)
+							oss << QString(q_("Width of antumbra"));
+						else
+							oss << QString(q_("Width of umbra"));
+						oss << QString(": %1 %2<br/>").arg(
 								 QString::number(pathWidth, 'f', 1),
-								 q_("km"));
+								 qc_("km", "distance"));
 					}
 				}
 				core1->setUseTopocentricCoordinates(useTopocentric);
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -1350,7 +1350,6 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup& 
 			const double eclipseObscuration = 100.*(1.-eclObj.first);
 			if (eclipseObscuration>1.e-7) // needed to avoid false display of 1e-14 or so.
 			{
-				oss << QString("%1: %2%<br />").arg(q_("Eclipse obscuration"), QString::number(eclipseObscuration, 'f', 2));
 				PlanetP obj = eclObj.second;
 				if (onEarth && obj == ssystem->getMoon())
 				{
@@ -1361,6 +1360,7 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup& 
 							/ angularSize;
 					oss << QString("%1: %2<br />").arg(q_("Eclipse magnitude"), QString::number(eclipseMagnitude, 'f', 3));
 				}
+				oss << QString("%1: %2%<br />").arg(q_("Eclipse obscuration"), QString::number(eclipseObscuration, 'f', 2));
 			}
 
 			if (onEarth)
@@ -1385,6 +1385,28 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup& 
 
 					if (pathWidth > 0.) // only display when shadow axis is touching Earth
 					{
+						oss << QString("%1: %2 ").arg(
+								 q_("Moon/Sun diameter ratio"), // It seems magnitude of total/annular eclipses sometimes represented by this value
+								 QString::number(dRatio, 'f', 3));
+						if (dRatio < 1.0)
+							oss << QString(qc_("(annular)","type of solar eclipse"));
+						else
+							oss << QString(qc_("(total)","type of solar eclipse"));
+						oss << "<br/>";
+						double centralDuraton = abs(duration);
+						int durationMinute = int(centralDuraton);
+						int durationSecond = round((centralDuraton - durationMinute) * 60.);
+						if (durationSecond>59)
+						{
+							durationMinute += 1;
+							durationSecond = 0;
+						}
+						oss << QString("%1: %2%3 %4%5<br/>").arg(
+								 q_("Central eclipse duration"),
+								 QString::number(durationMinute),
+								 q_("m"),
+								 QString::number(durationSecond),
+								 q_("s"));
 						QString info = q_("Center of solar eclipse (Lat./Long.)");
 						if (withDecimalDegree)
 							oss << QString("%1: %2°/%3°<br />").arg(info).arg(latDeg, 5, 'f', 4).arg(lngDeg, 5, 'f', 4);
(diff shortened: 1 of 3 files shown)


src/gui/AstroCalcDialog.cpp
```diff
@@ -3107,7 +3109,10 @@ cSEparameter centralSolarEclipse(double JD) {
 	double L2 = bessel.L2;
 	const double mu = bessel.mu;
 
-	constexpr double e2 = 0.00669438;
+	static SolarSystem* ssystem = GETSTELMODULE(SolarSystem);
+	const double f = 1.0 - ssystem->getEarth()->getOneMinusOblateness(); // flattening
```
gzotti, line 3113: you could also make this and the next 2 static.

src/core/modules/Planet.cpp
```diff
@@ -916,15 +916,17 @@ class SolarEclipse
 		double L1 = z * tf1 + (0.2725076 / cos(f1));
 		double L2 = z * tf2 - (0.272281 / cos(f2));
 		double mu = gast - a / M_PI_180;
-		constexpr double e2 = 0.00669438;
+		const double f = 1. - ssystem->getEarth()->getOneMinusOblateness(); // flattening
```
gzotti, line 919: good to identify these as well. could become static, and save a multiplication with 
`e2=f*(2.-f);
`

src/core/modules/Planet.cpp
```diff
@@ -1748,7 +1750,7 @@ Vec4d Planet::getRectangularCoordinates(const double longDeg, const double latDe
 	// For unclear reasons latDeg can be nan. Safety measure:
 	const double latRad = std::isnan(latDeg) ? 0. : latDeg*M_PI_180;
 	Q_ASSERT_X(!std::isnan(latRad), "Planet.cpp", QString("NaN result for latRad. Object %1 latitude %2").arg(englishName).arg(QString::number(latDeg, 'f', 5)).toLatin1());
-	const double u = (abs(latRad) - M_PI_2 < 1e-10 ? latRad : atan( bByA * tan(latRad)) );
+	const double u = (abs(abs(latRad) - M_PI_2) < 1e-10 ? latRad : atan( bByA * tan(latRad)) );
```
gzotti, line 1753: Whoa! Thanks for that one!

src/gui/AstroCalcDialog.cpp
```diff
@@ -3018,16 +3018,19 @@ pSEparameter partialSolarEclipse() {
 	double L1 = bessel.L1;
 	double L2 = bessel.L2;
 	const double mu = bessel.mu;
-	constexpr double e2 = 0.00669438;
+
+	static SolarSystem* ssystem = GETSTELMODULE(SolarSystem);
+	const double f = 1.0 - ssystem->getEarth()->getOneMinusOblateness(); // flattening
```
gzotti, line 3023: same as above...

src/gui/AstroCalcDialog.cpp
```diff
@@ -3225,24 +3230,20 @@ localSEparameter localSolarEclipse(double JD,int contact,bool central) {
...
 	double lon = static_cast<double>(core->getCurrentLocation().longitude);
 	double elevation = static_cast<double>(core->getCurrentLocation().altitude);
-	lat = lat * M_PI_180;
 	lon = -(lon);
 	double L = 0.;
 
+	static SolarSystem* ssystem = GETSTELMODULE(SolarSystem);
+	//PlanetP earth = ssystem->getEarth();
+	Vec4d geocentricCoords = ssystem->getEarth()->getRectangularCoordinates(lon,lat,elevation);
+	const double rc = geocentricCoords[0]/(EARTH_RADIUS/AU); // rhoCosPhiPrime
```
gzotti, line 3239: not sure where EARTH_RADIUS is set here. preferrably `ssystem->getEarth->getEquatorialRadius()` somewhere?

worachate001:
@gzotti Thanks.

src/core/modules/Planet.cpp
worachate001: Yeah, take some times to realized that.

src/gui/AstroCalcDialog.cpp
worachate001: Replaced with getEquatorialRadius().

src/gui/AstroCalcDialog.cpp
```diff
@@ -3225,24 +3230,19 @@ localSEparameter localSolarEclipse(double JD,int contact,bool central) {
 	double lat = static_cast<double>(core->getCurrentLocation().latitude);
 	double lon = static_cast<double>(core->getCurrentLocation().longitude);
 	double elevation = static_cast<double>(core->getCurrentLocation().altitude);
-	lat = lat * M_PI_180;
 	lon = -(lon);
```
gzotti, line 3233: Not sure what this does. Is `lon` use further down? (It looks like in the previous solution it was required to switch from our eastern-positive location to an western-positive implementation?)

Else, very good that you can identify and clean up all these bits and pieces, thank you!

src/gui/AstroCalcDialog.cpp
```diff
@@ -3225,24 +3230,19 @@ localSEparameter localSolarEclipse(double JD,int contact,bool central) {
...
 	double L = 0.;
 
+	static SolarSystem* ssystem = GETSTELMODULE(SolarSystem);
+	Vec4d geocentricCoords = ssystem->getEarth()->getRectangularCoordinates(lon,lat,elevation);
+	static const double earthRadius = ssystem->getEarth()->getEquatorialRadius();
+	static const double rc = geocentricCoords[0]/earthRadius; // rhoCosPhiPrime
```
gzotti, line 3239: Uh, `rc `and `rs `must not be static! When you change location, the `geocentricCoords` should also change! (static here means "set at first run and keep results until program exits")

src/gui/AstroCalcDialog.cpp
worachate001: Uhmm. I think we still need to do that.

src/gui/AstroCalcDialog.cpp
worachate001: Oops. fixed. It should be okay now.

alex-w:
Part of code in Planes.cpp and AstroCalcDialog.cpp files are identical - why not reuse it?

worachate001:
> Part of code in Planes.cpp and AstroCalcDialog.cpp files are identical - why not reuse it?

Yes. Thinking about that, but I need a rest soon, it's almost midnight here.

worachate001:
Update: the revision is going well until I found a new bug in AstroCalc/Eclipses. Trying to fix it before pushing this commit.

alex-w:
Why not use official term instead of "Sun/Moon diameter ratio"?

worachate001:
> Why not use official term instead of "Sun/Moon diameter ratio"?

magnitude-ratio-tse

This is an example of total eclipse when you're standing near the limit of umbra.

We already have 'Eclipse magnitude' below 'Eclipse obscuration' - eclipse magnitude here indicates how close you are to the northern or southern limit of umbra. It will be close to 1.0 if you are just inside the umbra.

Moon/Sun diameter ratio is different value, if we use the same term, it will be contradict to Eclipse magnitude above it. I think users may be confused of having two magnitudes at the same time.

worachate001:
Example of magnitude at greatest eclipse from two sources that show different value for the same event (look at 1955 Jun 20)

NASA

and 
https://astro.ukho.gov.uk/eclipse/0311955/Bangkok_Thailand_1955Jun20.png

gzotti:
Probably we should re-arrange the lines, though:

* Eclipse magnitude
* Eclipse obscuration
* Moon/Sun diameter ratio
* Central eclipse duration
* ...

By this, we would have the "coverage" items first, then the geographic data.

src/core/modules/Planet.cpp
```diff
@@ -1427,6 +1407,25 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup&
...
+
+						oss << QString("%1 %2 %3 %4°<br/>").arg(
+								 q_("Shadow center point is"),
+								 QString::number(distance, 'f', 1),
+								 q_("km towards azimuth"),
+								 QString::number(azimuth, 'f', 1));
+						oss << QString("%1: %2 %3<br/>").arg(
+								 q_("Width of umbra/antumbra"),
+								 QString::number(pathWidth, 'f', 1),
+								 q_("km"));
```
alex-w, line 1428: I think variable `km` with context data was defined before

alex-w (approved):
Thanks for improving!

src/core/modules/Planet.cpp
gzotti: yes, please use qs_("km", distance")

alex-w: A typo: `qc_(“term”, “context”);`

src/core/modules/Planet.cpp
```diff
@@ -1275,34 +1379,53 @@ QString Planet::getInfoStringExtra(const StelCore *core, const InfoStringGroup&
...
-						oss << QString("%1: %2 ").arg(
-								 q_("Magnitude of central eclipse"),
-								 QString::number(pos[2], 'f', 3));
-						if (pos[2] < 1.0)
-							oss << QString(qc_("(annular)","type of solar eclipse"));
-						else
-							oss << QString(qc_("(total)","type of solar eclipse"));
-						oss << "<br/>";
+						oss << QString("%1: %2 %3<br/>").arg(
+								 q_("Width of umbra/antumbra"),
```
gzotti, line 1426: Can you separate and give two separate words: either "umbra" (for total eclipses or total phase of hybrid) or "antumbra" (annular/annular phase of hybrid)?

src/core/modules/Planet.cpp
gzotti: Argh, sorry. qc_...

PR #2253, opened 2022-02-08
Fix light pollution node in TUI

### LINKED ISSUE #2248: Text User Interface: light pollution
After switching from Bortle scale one node (Light pollution) in the Text User Interface plugin is broken due slot/signal is deprecated.

10110111:
This fixes #2248.


alex-w (approved):
Thanks!

PR #2254, opened 2022-02-08
Simplify reading location from CLI

### LINKED ISSUE #2239: negative longitude with decimals on the command line does not work properly
### Discussed in https://github.com/Stellarium/stellarium/discussions/2094

<sup>Originally posted by **HeinBloed32** December 12, 2021</sup>
As a newbie to stellarium i am confused about using the command line option --longitude. 
When i enter stellarium --longitude=-0.5, stellarium opens in position E 0° 30' instead of W 0°30'.
Entering --longitude=0.5 (instead of -0.5) does give the same position. 
Same error when using format -0d30'

What  is my error ?

Thanks 
</div>

10110111:
### Description
Current transformation from radians to degrees via conversion to string
and back is ridiculous. This patch changes it to multiplication by a
coefficient.

Fixes #2239

### Type of change

## Checklist:

### DIFF
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -747,12 +747,20 @@ const StelLocation StelLocationMgr::locationFromCLI() const
 {
 	StelLocation ret;
 	QSettings* conf = StelApp::getInstance().getSettings();
-	bool ok;
 	conf->beginGroup("location_run_once");
-	ret.latitude = parseAngle(StelUtils::radToDmsStr(conf->value("latitude").toDouble(), true), &ok);
-	if (!ok) ret.role = '!';
-	ret.longitude = parseAngle(StelUtils::radToDmsStr(conf->value("longitude").toDouble(), true), &ok);
-	if (!ok) ret.role = '!';
+
+	const auto latVar = conf->value("latitude");
+	if (latVar.isValid())
+		ret.latitude = 180/M_PI * latVar.toDouble();
+	else
+		ret.role = '!';
+
+	const auto lonVar = conf->value("longitude");
+	if (lonVar.isValid())
+		ret.longitude = 180/M_PI * lonVar.toDouble();
+	else
+		ret.role = '!';
+	bool ok;
 	ret.altitude = conf->value("altitude", 0).toInt(&ok);
 	ret.planetName = conf->value("home_planet", "Earth").toString();
 	ret.landscapeKey = conf->value("landscape_name", "guereins").toString();


src/core/StelLocationMgr.cpp
```diff
@@ -747,12 +747,20 @@ const StelLocation StelLocationMgr::locationFromCLI() const
 {
 	StelLocation ret;
 	QSettings* conf = StelApp::getInstance().getSettings();
-	bool ok;
 	conf->beginGroup("location_run_once");
-	ret.latitude = parseAngle(StelUtils::radToDmsStr(conf->value("latitude").toDouble(), true), &ok);
```
gzotti, line 752: wow, this was really weird.

src/core/StelLocationMgr.cpp
```diff
@@ -747,12 +747,20 @@ const StelLocation StelLocationMgr::locationFromCLI() const
...
-	if (!ok) ret.role = '!';
-	ret.longitude = parseAngle(StelUtils::radToDmsStr(conf->value("longitude").toDouble(), true), &ok);
-	if (!ok) ret.role = '!';
+
+	const auto latVar = conf->value("latitude");
+	if (latVar.isValid())
+		ret.latitude = 180/M_PI * latVar.toDouble();
```
gzotti, line 754: OK, this is the minimal change that removes this nonsense, but maybe does not yet fix the actual problem. Maybe in CLIProcessor the first translation should go to  decimal degrees already?

gzotti (commented):
OK for me, but I think it does not fix 2239. That one must be solved in CLIProcessor.

src/core/StelLocationMgr.cpp
10110111: Or do you mean by "problem" the fact that the value is saved in radians instead of degrees?

gzotti: But does it also work with DMS formatted ones?

10110111:
> OK for me, but I think it does not fix 2239. That one must be solved in CLIProcessor.

It does. I tested with the following cases:
```
stellarium --latitude -13°57\'10\" --longitude 78°33\'54\"
stellarium --latitude -13.567 --longitude 78.2356
stellarium --latitude -13.567 --longitude -78.2356
```
Everything worked as expected, up to least significant digit: `78°33\'54\"` transforms into `78°33\'55\" in the status bar, while `-78.2356` becomes `W 78.235603°` in the Location dialog. These discrepancies, I think, stem from the use of `float`.

src/core/StelLocationMgr.cpp
10110111: As commented in the main thread, yes it does. I guess you are puzzled why this change fixes the problem. The actual problem is in the [parseAngle](https://github.com/Stellarium/stellarium/blob/53299955f3448a4f17cb9338eea975cecaabd58c/src/core/StelLocationMgr.cpp#L675) function in the same file. This should indeed be fixed so that parsing of locations works correctly, but this patch is a separate change, which, incidentally, does fix #2239.

gzotti:
Ah, did you fix that regexp problem earlier? Sorry, yes, then this is at least working. I still think the the storing into config.ini as runonly... looks clumsy, though. Also the conversion to Radians and back.

gzotti (approved):

src/core/StelLocationMgr.cpp
gzotti: d'oh, OK. Then parseAngle is and remains broken and must still be fixed. Either in a new issue/PR or in this one.

10110111: Yeah, I was going to make another PR for this (don't like my diligently separated commits being squashed).

PR #2257, opened 2022-02-08
Pass Sun, Moon and current planet objects as well as location directly to Atmosphere::computeColor

10110111:
### Description

These two commits make `Atmosphere::computeColor` accept Sun, Moon and Earth as `Planet const&` instead of their positions + lunar magnitude + lunar phase. This is necessary to avoid the set of parameters growing hard when I add ShowMySky atmosphere model. The latter also wants distance between the Earth and the Moon, unextincted magnitude of the Moon, angular size of the Sun, and in the future may require geometric positions of the Sun and the Moon.

The Moon is passed as a pointer, which is null if current planet isn't the Earth.

Location is now also passed as `StelLocation const&` to reduce number of parameters.

### Type of change

### How Has This Been Tested?

I tried to check that LP:1673283 doesn't reproduce, although it was a bit hard because on Mars it's rare to have the Sun far under the horizon while the Moon is above horizon. This seemed to be OK: I didn't notice any brightening due to the Moon.

**Test Configuration**:
* Operating system: GNU/Linux (LFS)
* Graphics Card: GeForce GTX 750Ti with proprietary `nvidia` driver

## Checklist:

### DIFF
--- a/src/core/modules/Atmosphere.cpp
+++ b/src/core/modules/Atmosphere.cpp
@@ -214,8 +214,8 @@ void Atmosphere::computeColor(StelCore* core, const double JD, const Planet& cur
 
 		// Update the eclipse intensity factor to apply on atmosphere model
 		// these are for radii
-		const double sun_angular_size = atan(SUN_RADIUS/AU/sunPos.length());
-		const double moon_angular_size = atan(MOON_RADIUS/AU/moonPos.length());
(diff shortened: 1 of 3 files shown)


alex-w (approved):
I see not difference in the sky between master and this branch :)

gzotti (approved):
Looks OK for me.

PR #2261, opened 2022-02-10
add source of Messier objects (issue #11)

silas1037:
I checked those names from B500 and find some are from O'Meara's book.
Further talking in #11


alex-w (approved):
Thank you!

gzotti:
Thank you, this makes MO the primary reference for those.

PR #2263, opened 2022-02-11
Fix support for Qt<5.11

10110111:
`QSharedPointer::get()` appeared first in Qt 5.11, while Stellarium supports down to 5.9. This patch replaces this call with `QSharedPointer::data()`, which is equivalent, although non-API-compatible with standard pointers.


alex-w (approved):
Thank you!

worachate001:
Thank you. I was wondering why I can't compile it in Linux.

PR #2345, opened 2022-03-12
Fix bug in AstroCalc/Eclipses for local solar eclipses

worachate001:
### Description
This bug in AstroCalc/Eclipses is found when total/annular eclipses take place very close to the horizon.

**To reproduce**
- Set the location to North Pole
- Open 'AstroCalc/Local Solar Eclipses' tab
- Set 2015 as the starting year

The calculation shows that 2015 Mar 20 was a partial eclipse, contradicting to the magnitude at greatest eclipse of 1.013 (total eclipse). The bug happens because the total eclipse took place very close to the horizon. The same situation can be found on other dates when your location is at the start or end point of any central eclipses.

This patch will also avoid to show time of 1st contact or last contact when total or annular eclipses are in progress at Sunrise/Sunset.

### Type of change

### How Has This Been Tested?

Test cases:
Lat. 26.0
Long. 47.5
Date: 2019-12-26 (annular eclipse at Sunrise)

Lat. 47.4
Long. -19.6
Date: 2024-04-08 (total eclipse at Sunset)

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/gui/AstroCalcDialog.cpp
+++ b/src/gui/AstroCalcDialog.cpp
@@ -3499,7 +3499,7 @@ void AstroCalcDialog::generateSolarEclipsesLocal()
 							}
 							double C3altitude = eclipseData.altitude;
 
-							if (eclipseData.ce > 0. && ((C2altitude > 0.) || (C3altitude > 0.))) // Central eclipse occurs
+							if (eclipseData.ce > 0. && ((C2altitude > -.3) || (C3altitude > -.3))) // Central eclipse occurs
 							{
 								centraleclipse = true;
 								if (eclipseData.L2 < 0.)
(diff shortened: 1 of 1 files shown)


src/gui/AstroCalcDialog.cpp
```diff
@@ -3539,6 +3539,9 @@ void AstroCalcDialog::generateSolarEclipsesLocal()
 							treeItem->setData(SolarEclipseLocalDate, Qt::UserRole, JDmax);
 							treeItem->setText(SolarEclipseLocalType, eclipseTypeStr);
 							treeItem->setText(SolarEclipseLocalFirstContact, QString("%1").arg(localeMgr->getPrintableTimeLocal(JD1)));
+							if (centraleclipse)
```
alex-w, line 3542: Why not use `if (centraleclipse && JD2<JD1)`?

src/gui/AstroCalcDialog.cpp
```diff
@@ -3555,6 +3558,9 @@ void AstroCalcDialog::generateSolarEclipsesLocal()
 								treeItem->setText(SolarEclipseLocal3rdContact, dash);
 							treeItem->setToolTip(SolarEclipseLocal3rdContact, q_("The time of third contact"));
 							treeItem->setText(SolarEclipseLocalLastContact, QString("%1").arg(localeMgr->getPrintableTimeLocal(JD4)));
+							if (centraleclipse)
```
alex-w, line 3561: Why not use `if (centraleclipse && JD3>JD4)`?

src/gui/AstroCalcDialog.cpp
worachate001: Yes, I'm about to change that. Thanks.

src/gui/AstroCalcDialog.cpp
worachate001: Thanks, I have added else... after that. Decided to remove it and forgot to change it.

worachate001:
I would like to note that situations like this are sometimes difficult to handle. Not showing first or last contact maybe a bit weird. But there are cases like this when totality or annularity begins just before Sunrise and ends just after that. So first contact actually occurs but we can't see it.

alex-w (approved):
Thanks!

PR #2353, opened 2022-03-18
Fix a memory leak in interferometry_uv_coords_vector

Teemperor:
`vector` is allocated in this function but never free'd.


Teemperor:
FWIW, we might also just delete `vector` completely as it seem unused?

gzotti:
This is a problem in INDI which should be reported there. We should care for a regular update strategy, not fix their problems.

Teemperor:
Sorry, didn't realize that is actually an external library in the source tree. I checked in the libindi source and it seems that bug was fixed as a side effect of some refactoring.

Issue #2374, opened 2022-03-30
0.22.0 problems with changes in "light pollution from location database"


gzotti:
The change to the light pollution implementation is causing havoc for my local users!  The automatic from database setting used to work fine, but now simply gives an orange sky with no objects visible - not the same as previously AND not representative of reality at all, and the manual settings are virtually useless.  All of my group (Leicester Astronomical Society members) are reverting to version 21.3

_Originally posted by @PaulJRead in https://github.com/Stellarium/stellarium/discussions/2365#discussioncomment-2457161_

gzotti:
I tried a bit. All tried locations load Bortle class 9 now, this is really not plausible.

10110111:
Erm... I'm not sure what exactly changed since my commits. You seem to have played with it before accepting the PR, haven't you?

gzotti:
I am afraid I did not check the "load LP from locations". :-(

10110111:
By locations do you mean just pick a city from the Location dialog?

10110111:
OK, the problem is not mine, the database itself was broken by 4cab8b348b2a4c70f9c75749bdd4a63892ef3ba6, which happened after my commits.

Reverting the database as follows fixes the problem for me.
```
git checkout 4cab8b34^ -- data/base_locations.bin.gz
```

alex-w:
So, the tool for conversion from text to binary format (`StelLocationMgr::generateBinaryLocationFile()`) is broken now.  Within conversion tool we use method `StelLocationMgr::loadCities()` to create a list of locations, which uses `StelLocation::createFromLine()` to parse the string and that method expected light pollution luminance value now. Same method used to loading user defined list of locations.

10110111:
Looks like fixed after 597e7d125625f26c2b0f2f19c8cc410d597ce5f3.

alex-w:
Please check the latest stable version of Stellarium:
https://github.com/Stellarium/stellarium/releases/latest

10110111:
The latest release seems to be the same 0.22.0 that this issue was opened for.

alex-w:
Latest release: https://github.com/Stellarium/stellarium/releases/tag/v0.22.1

PR #2377, opened 2022-03-31
Fix bortle index input mistreated as luminance

10110111:
This addresses #2374. The database will have to be regenerated to finally fix the issue.


gzotti (approved):
Thank you, looks good (although I cannot test today)

alex-w:

PR #2379, opened 2022-03-31
Fix #1283: properly detect largest screenshot number

### LINKED ISSUE #1283: screenshot file naming logic is bugged

chrisx8:
This implements a proper mechanism to detect the largest screenshot number. Fixes #1283.

### Description
- Detect the largest screenshot number in screenshot directory, using the following logic:
  - Get a list of filenames in screenshot directory
  - Find the last filename (which should contain the largest number), and extract the number from it
  - Number + 1 should be the number for the new screenshot. A failsafe is provided (in case `stellarium-number.format` already exists for some reason).
- This replaces the current (poor) implementation for detecting next available screenshot number. See https://github.com/Stellarium/stellarium/issues/1283#issuecomment-1071992445 for details.

### Type of change

### How Has This Been Tested?

I tried generating screenshots in the following conditions:
- Blank screenshot directory
- Screenshot directory has screenshots with sequential numbers
- Screenshot directory has screenshots with non-sequential numbers
- Screenshot directory has screenshots with large numbers (more than 3 digits)
- Screenshot directory has random (non Stellarium screenshot) files

In all cases listed above, new screenshots are named properly (highest number + 1).

**Test Configuration**:
* Operating system: Fedora 36 (arm64 VM on M1 MacBook)
* Graphics Card: QEMU virtual GPU
* *Note: GPU info is likely irrelevant here, because this change only involves screenshot file naming.* 

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1717,24 +1717,34 @@ void StelMainView::doScreenshot(void)
 		// get highest-numbered file in screenshot directory
 		QDir dir(shotDir.filePath());
 		QStringList existingFiles = dir.entryList(fileNameFilters);
-		QString lastFileName = existingFiles[existingFiles.size() - 1];
 
-		// extract number from highest-numbered file name
-		QString lastShotNumString = lastFileName.replace(screenShotPrefix, "").replace("." + screenShotFormat, "");
-		// new screenshot number = start at highest number
-		int shotNum = lastShotNumString.toInt() + 1;
+		if (existingFiles.empty())
+		{
+			// empty screenshot dir, just name screenshot prefix-001.format
+			shotPath = QFileInfo(shotDir.filePath() + "/" + screenShotPrefix + "001." + screenShotFormat);
+		}
+		else
+		{
+			// already have screenshots, find largest number
+			QString lastFileName = existingFiles[existingFiles.size() - 1];
 
-		// build new screenshot path: "path/prefix-num.format"
-		// num is at least 3 characters
-		QString shotNumString = QString::number(shotNum).rightJustified(3, '0');
-		QString shotPathString = QString("%1/%2%3.%4").arg(shotDir.filePath(), screenShotPrefix, shotNumString, screenShotFormat);
-		shotPath = QFileInfo(shotPathString);
-		// validate if new screenshot number is valid (non-existent)
-		while (shotPath.exists()) {
-			shotNum++;
-			shotNumString = QString::number(shotNum).rightJustified(3, '0');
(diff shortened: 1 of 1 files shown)


alex-w:
This solution will cause crash when screenshots directory is empty

chrisx8:
> This solution will cause crash when screenshots directory is empty

I am able to reproduce the crash, which is caused by incorrectly indexing into the list of files in the screenshot directory. I added a check for this, and the first screenshot would just be numbered `1`. Its name would be something like `stellarium-001.png`.

@alex-w, would you mind trying out the fix?

gzotti:
Sorry, no time to test this week. If it works and empty directory problem is fixed, fine for me.

alex-w:

PR #2387, opened 2022-04-09
Add dso names for v0220

silas1037:
Collections from web.


gzotti (changes requested):
The abbreviation TBD="to be done" was not thought to be used as a key to actual entries in the list below, but as reminder to us developers that this list of references is not yet complete. If entries are in Wikipedia, WP is enough. In case we really want to add single entries from "just about every source", add a MISC category. But I would be happier to limit our list to meaningful sources which actually represent practical value.

ultrapre:
> The abbreviation TBD="to be done" was not thought to be used as a key to actual entries in the list below, but as reminder to us developers that this list of references is not yet complete. If entries are in Wikipedia, WP is enough. In case we really want to add single entries from "just about every source", add a MISC category. But I would be happier to limit our list to meaningful sources which actually represent practical value.

Okay, I will resolve some of them.

ultrapre:
I rarely add TBD alone as a separate source, but as a secondary source of identification that avoids tedious naming. Could it be understood that TBD does not need to be added after the entry reference?

gzotti:
If you don't want to create a new source in the source entries, add and use a MISC pseudo-source with comment entries like now seen under TBD. TBD is only a reminder to add new references where needed. When used in the list of named DSO, TBD reads like "we need to look up where this name comes from and maybe will delete it if we cannot find it."

alex-w (approved):

gzotti (approved):
Thanks, OK for me now.

alex-w:

PR #2390, opened 2022-04-09
AstroCalc with QCharts

gzotti:
### Description

The external QCustomPlot in its current form seems to conflict with Qt6. Qt has added a similar module with QtCharts, which we should preferrably use instead. This PR shall be used to replace the plots in question.

### Fixes: 
Qt6-inflicted deprecation

### Type of change

### Tasks:

- [x] Recreate existing plot
- [x] Fix bugs
- [x] Improve
- [x] Remove old version
- [x] SUG changes

### Done:

 - [x] AltVsTime
 - [x] AzVsTime
 - [x] MonthlyElevation
 - [x] Curves
 - [x] Lunar Elongation
 - [x] Planetary Calculator

### How Has This Been Tested?

**Test Configuration**:
* Qt 5.15.2

## Checklist:


gzotti:
@alex-w we should get rid of QCustomPlot in this branch (or else make sure to upgrade to a license-compatible version that works in Qt6). However I don't fully understand the various actions in AstroCalc. For some reason, drawAltVsTimeDiagram() can be called again while it is running already. To avoid this I added a Mutex which however now can lead to blocking. Overlapping calls look like waste of time in any case, though. 
Currently you can see a comparison by starting the program, calling Astrocalc, selecting the Graph tab, then selecting a star in the sky. 
When you first select an object, then call AstroCalc, it blocks on a Mutex. I don't understand why when these calls would be processed in two threads. Uncomment the Mutex and read the debug output to see the weird call sequence.

The QSplineSeries look somewhat better (smoother). Grids and colors can still be refined, of course. Changing the charts (transit and current time lines) seems difficult, and the lines currently don't show. (Not sure why.)  

It seems we will need 4 charts. These could be shown in a single QChartView or in several, just as needed. But some calls (signal/slot connections) will have to be re-arranged to avoid intertwined calls.

gzotti:
@alex-w the QCharts::QDateTimeAxis supports dates from 4714BC onwards, with Qt's interpretation of proleptic calendar dates. Does the QCustomPlot time-based axis also have such limitation? I would just show some explanatory message like "plot unavailable for the remote past".

alex-w:
> @alex-w the QCharts::QDateTimeAxis supports dates from 4714BC onwards, with Qt's interpretation of proleptic calendar dates. Does the QCustomPlot time-based axis also have such limitation? I would just show some explanatory message like "plot unavailable for the remote past".

I don't know about limitations for QCustomPlot, but probably it have same limits.

alex-w:
Just a note to memory: AltVsTime and AziVsTime graphs do not account time zone shift.

gzotti:
Yes I know, some finetuning is still required. And probably the other time graphs should adhere to tickmarks at local midnight, not "now".  I need still more insight into QDateTime and some of its internal behaviour to e.g. avoid automatic timezone adjustment by Qt's own logic... One problem may also be Qt's no-year-0 policy, or some other antiquity issues. But overall I think we will be able to move over to QtChart. 

Another issue is the AzVsTimeCurve's 360/0 wraparound. With the Spline curve this causes some tremors. Maybe I have to split this curve into two.  

What special features did you build into the graphs? Apparently in some you can set time by clicking into the graph, or read curve values via mouse-over&tooltip. I have not touched those. Maybe not everything is available.

gzotti:
@alex-w I think the Azi and Alt plots may be finished, and I could progress by removing the old plots. 

I have not yet checked the date details in the other plots. My feeling is that the vertical lines marked with calendar days should indicate local midnight, not "integer offsets from now". Also maybe some tooltips needs special handling.

The tooltips probably only display the actual data points, not the curve value in between. The vertical lines cannot be read I am afraid, unless we add e.g. an otherwise unneeded point at y=0.

For the Azi plot I had to split the line into two parts to avoid the discontinuity with the ugly spline "tremor". However, the RightAscension graphs have the same problem and need some more thinking. Either I can deal with them as QLineSeries (no spline smoothness), or find some auto-numbering scheme that can deal with "unnamed" lines (lines without named enum ID).

alex-w:
Visual style remarks: Could be legend to splitted onto 2 lines (to improve readability)? Also all newly graphs has big "borders" - could be these borders thin or none? Most newly graphs (except Graphs/Graphs) has title onto graph canvas in additional to titles onto widget - probably titles onto graphs no needed.

Architectural remarks: could be some parts of AstroCalcGraphs class compatible with Qt 5.12? I ask it because we cannot automatically testing building Stellarium (see our GH actions)?

gzotti:
> Visual style remarks: Could be legend to splitted onto 2 lines (to improve readability)? 

Must check what the legend allows.

> Also all newly graphs has big "borders" - could be these borders thin or none? 

I think this is a border in the QChartView class where the actual chart is then embedded. Any CSS fix possible?

> Most newly graphs (except Graphs/Graphs) has title onto graph canvas in additional to titles onto widget - probably titles onto graphs no needed.

On the contrary, the labels in the main dialog are not required any longer and will go away. My aim is to encapsulate as much as possible in the charts, maybe also some of the actual plotting, as the AstroCalcDialog class is far too large and complex to be "nice".

 
> Architectural remarks: could be some parts of AstroCalcGraphs class compatible with Qt 5.12? I ask it because we cannot automatically testing building Stellarium (see our GH actions)?

Not sure. QtChart has seen improvements in latest versions which I may already use, but I have not checked the details of the error messages. Or was it just missing packages in the CI system? The work towards Qt6 aims for code compatibility between Qt5.15.2 and Qt6.2. 

Who still requires Qt5.12 without a choice for an upgrade? Ubuntu 16.04? Dead for most uses. Or even 18.04? I think these users may still happily keep using 0.22.1. But we should really go forward.

alex-w:
> > Also all newly graphs has big "borders" - could be these borders thin or none?
> 
> I think this is a border in the QChartView class where the actual chart is then embedded. Any CSS fix possible?

Hmm... maybe, it's need to more investigation...

> > Most newly graphs (except Graphs/Graphs) has title onto graph canvas in additional to titles onto widget - probably titles onto graphs no needed.
> 
> On the contrary, the labels in the main dialog are not required any longer and will go away. My aim is to encapsulate as much as possible in the charts, maybe also some of the actual plotting, as the AstroCalcDialog class is far too large and complex to be "nice".

OK, good solution.

> > Architectural remarks: could be some parts of AstroCalcGraphs class compatible with Qt 5.12? I ask it because we cannot automatically testing building Stellarium (see our GH actions)?
> 
> Not sure. QtChart has seen improvements in latest versions which I may already use, but I have not checked the details of the error messages. Or was it just missing packages in the CI system? The work towards Qt6 aims for code compatibility between Qt5.15.2 and Qt6.2.

See notices for use sentences like `QColorConstants::Svg::orange` for example.

> Who still requires Qt5.12 without a choice for an upgrade? Ubuntu 16.04? Dead for most uses. Or even 18.04? I think these users may still happily keep using 0.22.1. But we should really go forward.

Ubuntu 20.04 hasn't Qt 5.15 for example (our linux-boxes in GH actions has U20.04 and this is latest used version of Ubuntu in GH infrastructure).

gzotti:
I have a unit test fail:

TestConversions::testRadToDecDegStr() 'degree==expectedValue' returned FALSE. (0.00000 radians = 0.00d (expected +0.00d))

alex-w:
OK, the unit test is fixed now

gzotti:
Yes, I just was not sure whether the plus sign is some design requirement. IMHO, we don't need it, it's just glyph noise. In the process of this branch I found a few age-old mini issues.

gzotti:
OK, the border issue is solved. 

I am afraid I cannot force the legends into a 2-line layout, but it's only 1 graph where this would matter. 

What seems also not possible: two-line layout for the QDateTime-based axis labels. At least I don't know how.

alex-w:
Probably first 3 graphs in Graphs tab and one in PC tab ready to replace...

alex-w:
It the bold text really need for axis labels?

gzotti:
> It the bold text really need for axis labels?

I think it reads better, but we can change it.  Seems to be bold by default. See QAbstractAxis::setTitleFont()

gzotti:
OK, first old plot removed. The pcChart could actually live without the legend, given two colored Y axes. @alex-w ?

alex-w:
> OK, first old plot removed. The pcChart could actually live without the legend, given two colored Y axes. @alex-w ?

Yes, the legend can be removed here.

gzotti:
Finally the mouse click time setting. Currently it's active for all charts, I think from user perspective this is a nice move. I have not tested all consequences, though.

gzotti:
TODO: Clicking in the ME chart currently sets a nonsense date, sorry.

gzotti:
@alex-w I have temporarily added points to the graphs to see the density of the curves' data. 
Else, I think the two-graph tab is also ready to finalize. 

All graphs have the ability to read curve values via mouse-over, and to set the time by clicking somewhere in the data area. (But I think clicking on the curve itself does nothing). Reading the time accurately from the curve may however not be possible. Make a plot for Ganymede and see its magnitude dips for occultations. It is not possible to read the exact time from a point.

alex-w:
I found a few issues:
- [ ] AltVsTime: Altitude axes have missing values when range is big: please select artificial satellites (e.g. ISS) or Fomalhaut and enable Sun graph for example
- [ ] AltVsTime: Graphs for Sun, Moon and Twilights have wrong "shape" when artificial satellites are selected
- [ ] ME: Altitude axes have missing values when range is big: please select the Moon
- [ ] Graphs: years splitters are missing
- [ ] Graphs: very slow rendering when many years are selected

gzotti:
You mean the clipping of top/bottom curves? This is caused by the spline curve exceeding the XY point range. I can add some extra space. 

Admittedly I must check the satellites, working without this plugin currently. 

For the graphs: The date axis should be labeled with full dates now, I see no need for year splitters. And slow rendering also applies to the old version - the computation did not really change. Of course, the double graph currently takes 2x the time. OTOH in the altVsTime graph, I inlined sun and moon. Even it density is increased, it is the core->setJD(jd); core->update(0); where most time is required. The coordinate then come for almost nothing.

alex-w:
> You mean the clipping of top/bottom curves? This is caused by the spline curve exceeding the XY point range. 

No. See an example:
stellarium-006

> I can add some extra space.

I will be helpful

gzotti:
OK, satellites fixed. More tomorrow...

gzotti:
Now there is a 5% buffer where applicable. 
I was able to thin out the point distance for the AltVsTime plot unless we are computing a satellite.

We could of course thin out the steps for the 2-curve charts, but at cost of undersampling and missing some events. One interesting example is a magnitude curve for Ganymed which shows the occasional shadow transit. There are certainly many more which are missed by the 1-day stepwidth, but increasing density of course requires dramatically more time. Maybe we should even add a GUI element for that? (Sampling density in hours, 1...72, with a warning tooltip that this may take a long time?)

Another improvement in GUI agility could be in the GUI page switching code. I understand that you call the plotting routine when activating a tab. But sometimes we may have plotted the chart already. Imagine switching rapidly between AltVsTime and AzVsTime charts. The AzVsTime chart is slow because of the requirement to keep the 360/0 transition as vertical as possible. Just test existence of the ...Chart, and force redrawing only when object or date has changed?

alex-w:
Please try select artificial satellites and open Graphs/ME graph

alex-w:
> Now there is a 5% buffer where applicable. I was able to thin out the point distance for the AltVsTime plot unless we are computing a satellite.

Very good, thanks!

> We could of course thin out the steps for the 2-curve charts, but at cost of undersampling and missing some events. One interesting example is a magnitude curve for Ganymed which shows the occasional shadow transit. There are certainly many more which are missed by the 1-day stepwidth, but increasing density of course requires dramatically more time. Maybe we should even add a GUI element for that? (Sampling density in hours, 1...72, with a warning tooltip that this may take a long time?)

May be. Introducing the dencity factor may be give interesting results. Of course the warning label required here (not a tooltip).

> Another improvement in GUI agility could be in the GUI page switching code. I understand that you call the plotting routine when activating a tab. But sometimes we may have plotted the chart already. Imagine switching rapidly between AltVsTime and AzVsTime charts. The AzVsTime chart is slow because of the requirement to keep the 360/0 transition as vertical as possible. Just test existence of the ...Chart, and force redrawing only when object or date has changed?

If you can improve the respondence of Graphs tool and flawless rendering the graphs - very good, go ahead for experiments!

gzotti:
Yes, I just played with Europe and its eclipses by Jupiter. :-( But smaller time steps are possible&better.

gzotti:
Really bad bug around QDateTime and its interpretation of time zones. Back to field 1 with finetuning :-(

alex-w:
> Really bad bug around QDateTime and its interpretation of time zones. Back to field 1 with finetuning :-(

We had used Qt4 in 2008, so, it may be not a bug initially. :-/

gzotti:
I think it approaches finalisation. Of course, cleanup, debug printouts, some preliminary comments have to be cleaned up.

This branch may have opened a Pandora's box though. The functions around JD manipulation with QDateTime looked overly messy, why construct a date string that needs to be parsed when we have date/time elements already identified? Did this circumvent some problem with invalid dates (which dates are considered invalid?). I know, QDateTime has no year 0. But what would be our solution? It seems the charts are still plotted, but with a year label just off by 1 year. This could be handled in the mouse click and tooltip handler. But Qt's assumption of a proleptic Gregorian calendar in prehistory may cause issues. It may be enough to specify this in the documentation. 

I added a flag forceUTC to StelUtils::jdToQDateTime(). The original intention of this function must have been to use UTC in all cases, however the handling was in fact Qt::LocalTime. There are a few cases where this is used and I don't know whether forcing UTC now would change anything. (GRS!?) @alex-w  any insights? Its use must be fixed to use UTC everywhere, and then the flag removed again.

alex-w:
Export an image is not working :-/

gzotti:
> Export an image is not working :-/

Huh? Works well for me. The only weirdness is 2 seconds display of a small window in the foreground after the file dialog. File created, no problem. What does it do on Linux/Mac?

alex-w:
> Huh? Works well for me. The only weirdness is 2 seconds display of a small window in the foreground after the file dialog. File created, no problem. What does it do on Linux/Mac?

Just a crash on linux

gzotti:
Uh, running on WSL/XMing: As soon as the file dialog appears, I see messages on the console:

> QOpenGLEngineShaderManager::useCorrectShaderProg() - Unsupported composition mode
> Unsupported composition mode
> QOpenGLEngineShaderManager::useCorrectShaderProg() - Unsupported composition mode
> Unsupported composition mode
> QOpenGLEngineShaderManager::useCorrectShaderProg() - Unsupported composition mode
> Unsupported composition mode
> ...
> 

The dialog shows no file area. Bad. ... Hmm, there is a file dialog behind, but unresponsive. It seems the simple QFileDialog cannot just be used here :-(

gzotti:
@alex-w can you please look into these LGTM FIXMEs and also AstroCalcDialog.cpp line 257? What was the original intent? Is it OK to work with UTC dates? (The original behaviour used Qt's local timezone by utilizing that function from StelUtils.)

alex-w:
> @alex-w can you please look into these LGTM FIXMEs and also AstroCalcDialog.cpp line 257? What was the original intent? 

These fields was original intent to entering the local dates and times.

>Is it OK to work with UTC dates? (The original behaviour used Qt's local timezone by utilizing that function from StelUtils.)

No, this is not very good behaviour, because it will be confusing the users.

gzotti:
> No, this is not very good behaviour, because it will be confusing the users.

OK. If the currently present behaviour is OK for you (it should be identical to behaviour in master), we will have to keep a second argument for StelUtils::jdToQDateTime(), but I may switch that to Qt::TimeSpec, to clearly specify Qt::UTC or Qt::LocalTime.

alex-w:
> OK. If the currently present behaviour is OK for you (it should be identical to behaviour in master), we will have to keep a second argument for StelUtils::jdToQDateTime(), but I may switch that to Qt::TimeSpec, to clearly specify Qt::UTC or Qt::LocalTime.

I see no needs to use Qt::TimeSpec at the moment (maybe I am wrong and it may be efficient in some pieces of our code?)

The result of refactoring seems ready to merge after some cleanup (hiding knots on graphs for example)

gzotti:
Some Qt::TimeSpec is used always in a QDateTime. If Qt::UTC is not set explicitly, it appears the default is Qt::LocalTime, which however is wrong for most of the cases in our code. And I think Qt::LocalTime is the system timezone, not necessarily the currently set Stellarium timezone which may lad to tricky errors. So, using UTC and "manual" timezone calculations is "safe", but for the GUI "calendar" widgets it may be OK. I just need to add enough documentation how to use it. 

If else you also think it's good now, I will start cleaning up comments (and the dots).

alex-w:
> Some Qt::TimeSpec is used always in a QDateTime. If Qt::UTC is not set explicitly, it appears the default is Qt::LocalTime, which however is wrong for most of the cases in our code. And I think Qt::LocalTime is the system timezone, not necessarily the currently set Stellarium timezone which may lad to tricky errors. So, using UTC and "manual" timezone calculations is "safe", but for the GUI "calendar" widgets it may be OK. I just need to add enough documentation how to use it.

Good additions to the development documents

> If else you also think it's good now, I will start cleaning up comments (and the dots).

At least it ready to public testing IMHO :)

gzotti:
Ouch. I did a `git pull --rebase ` as usual to integrate your ui fixes, and now have weird conflicts. Did you delete testConversions.cpp and GreatRedSpotDialog.cpp? Or did git just really mess it up this time?

alex-w:
> Ouch. I did a `git pull --rebase ` as usual to integrate your ui fixes, and now have weird conflicts. Did you delete testConversions.cpp and GreatRedSpotDialog.cpp? Or did git just really mess it up this time?

O_O I didn't touches these files!

alex-w (approved):

gzotti:
OK, seems resolved. I will cleanup a bit before doing the merge.

PR #2391, opened 2022-04-12
Update star name

ultrapre:
Is there a historical reason for the frequent out-of-bounds cases in the constellations Hydra and Crater? Same as 10 UMa.


alex-w (commented):
Where is the source for the additional designation?

alex-w (approved):

alex-w:

PR #2394, opened 2022-04-13
Removed exoplanets diagram from the GUI

alex-w:
### Description
Removed exoplanets diagram from the GUI of the plugin as an intermediate step to the Qt6 migration process.

### Type of change

## Checklist:


gzotti:
Why not replace by QtChart immediately? This could probably re-use most of the plotting code.

plugins/Exoplanets/src/gui/ExoplanetsDialog.cpp
```diff
@@ -275,7 +270,7 @@ void ExoplanetsDialog::setAboutHtml(void)
 	html += "<tr><td><strong>" + q_("Author") + ":</strong></td><td>Alexander Wolf</td></tr></table>";
 
 	html += "<p>" + QString(q_("This plugin plots the position of stars with exoplanets. Exoplanets data is derived from \"%1The Extrasolar Planets Encyclopaedia%2\"")).arg("<a href=\"http://exoplanet.eu/\">", "</a>. ");
```
gzotti, line 272: Upgrade links to https where available.

alex-w:
> Why not replace by QtChart immediately? This could probably re-use most of the plotting code.

I have no time for this work currently :(

gzotti (changes requested):
Seriously, we should re-use the existing code and move to QChart. I think it helps during migration to set a QChartView side-by-side with the old plot and develop until both look roughly the same. The major difference is that in the old version you set X values and Y values independently, while in the QChart you use QXYSeries that would have to be created on the fly from the various already existing lists.

One point I just found: year of discovery is apparently 0 for at least one entry, making the graph almost useless. Force minimum of 1995? Or better remove entry with unknown year from graph.

gzotti:
> > Why not replace by QtChart immediately? This could probably re-use most of the plotting code.
> 
> I have no time for this work currently :(

OK. Maybe I can have a look into this then, and may open my own replacement PR soon. However I just started the AstroCalc changes, these would be more important, and maybe you can share some insights there. 

It would be a pity to simply delete the in-fact interesting diagrams without proper replacement.

alex-w:
OK, let's postpone this request right now for some future review or replace

PR #2405, opened 2022-04-17
Added support directional elongation

alex-w:
### Description
Added option to switch between directional and nondirectional ecliptic elongation. See discussion #2368 for details.

### Type of change

## Checklist:


gzotti (changes requested):
For me it can be even simpler, no flag, config etc. Just write E/W instead of +/-. 

However, I just see that in the InfoMap the signed &Delta;&lambda; is missing. I think there a signed number is more useful (let the user of the data decide what to do). 

Given that the other (purely angular) elongation is always positive, can we write it without the plus sign?

alex-w:
> For me it can be even simpler, no flag, config etc. Just write E/W instead of +/-. But extra flag is OK if you prefer.

Now it uses E/W instead of +/-

> However, I just see that in the InfoMap the signed Δλ is missing. I think there a signed number is more useful (let the user of the data decide what to do).

Added, on Earth only

> Given that the other (purely angular) elongation is always positive, can we write it without the plus sign?

OK, but it some hackish

src/core/modules/Planet.cpp
```diff
@@ -782,22 +782,26 @@ QString Planet::getInfoStringEloPhase(const StelCore *core, const InfoStringGrou
 		StelUtils::rectToSphe(&ra, &de, getEquinoxEquatorialPos(core));
 		StelUtils::equToEcl(raSun, deSun, obl, &lSun, &bSun);
 		StelUtils::equToEcl(ra, de, obl, &ecLong, &ecLat);
-		double elongAlongEcliptic=StelUtils::fmodpos(ecLong-lSun, M_PI*2.);
+		double elongationDecDeg, elongAlongEcliptic = StelUtils::fmodpos(ecLong-lSun, M_PI*2.);
 		if (elongAlongEcliptic > M_PI) elongAlongEcliptic-=2.*M_PI;
 
 		QString pha, elo, dLam;
+		bool sign;
+		StelUtils::radToDecDeg(elongAlongEcliptic, sign, elongationDecDeg);
```
gzotti, line 790: The function radToDecDeg is IMHO unneeded, and I remove its use it whereever I see it. Why dissect sign into a separate variable just to apply it immediately after once and never again? Just say 

elongationDecDeg=fmod(elongAlongEcliptic*M_180_PI, 360.); 

We have this (what feels like) thousands of times in our code. 

I will push a commit with some other changes I see useful.

src/core/modules/Planet.hpp
```diff
@@ -199,12 +199,15 @@ class Planet : public StelObject
...
-	//! - elongation
-	//! - elongation-dms (formatted string)
-	//! - elongation-deg (formatted string)
+	//! - phase-angle (radians)
+	//! - phase-angle-dms (formatted string; DMS)
+	//! - phase-angle-deg (formatted string; degrees)
+	//! - elongation (radians)
+	//! - elongation-dms (formatted string; DMS)
+	//! - elongation-deg (formatted string; degrees)
+	//! - modern-elongation (elongation in ecliptic longitude or Δλ; on Earth only; radians)
```
gzotti: I fixed the "modern" name use (changed to ecl-elongation), because it is actually the more classical use.

gzotti (approved):
Else it's fine for me now.

alex-w:
> Else it's fine for me now.

Please merge it

PR #2461, opened 2022-05-24
Switch from QtScript to QJSEngine

gzotti:
### Description

We need to switch to the QJSEngine for JavaScript execution when we want to port this program to Qt6, which is necessary in the near future.

Most existing scripts run, but some of the scripting code (Vec3f/Vec3d types) has to be changed. 

### Fixes
Qt6 deprecation issues

### Requires
Qt 5.14 and later. This version introduces a way to stop running scripts.

### Differences to the previous solution
- no pause/resume functions
- Vec3f/Vec3d features require V3d/V3f/Color glue objects

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Qt 5.15.2 (Should work with 5.14 and later), Qt6.3.1
* Graphics Card: irrelevant

## Checklist:


gzotti:
It may be possible to add other JS modules, like keyboard control e.g. via https://www.npmjs.com/package/keyboardjs

alex-w:
> It may be possible to add other JS modules, like keyboard control e.g. via https://www.npmjs.com/package/keyboardjs

Possible solution to handling keys within the scripts

gzotti:
> > It may be possible to add other JS modules, like keyboard control e.g. via https://www.npmjs.com/package/keyboardjs
> 
> Possible solution to handling keys within the scripts

Sorry, false hope. This requires in-browser elements. But maybe the core object can be extended to provide temporary hotkeys/actions. Of course, existing hotkeys must not be overwritten. This is not part of this effort, though.

gzotti:
I am a bit stuck currently. All our scripts seem to work nicely, except for those by Qam1 which show a "hacked" keyboard interface (earth_[1-7].ssc). 

However, I cannot read Vec3[idf] values from script as the new interface only allows QObject-derived classes, and our custom Vec3* converters don't fit AFAICS. Maybe we have to develop QObject-based wrappers for Vec3[idf] input, or use QVector3D intermediates (I tried, but does not work yet), or if all else fails, we must use 3-component setters. In any case, the extended Vec3 support developed as late as 2020 with named (rgb) components is (as expected then) likely to go away again. Fortunately it seems not to have gained much attention. 

I am adding a test script which can be run piecewise in the script console.

gzotti:
I have tried a few things. Most promising is an intermediary class V3d(x, y, z) which can be used in JS calls. What seems to work is 

`StelMovementMgr.setViewDirectionJ2000(new V3d(0., 1., 0.).toVec3d());`

which is not as elegant as the previous 

`StelMovementMgr.setViewDirectionJ2000(new Vec3d(0., 1., 0.));` (?)

but maybe somebody can propose a better solution. 
`core.output(StelMovementMgr.getViewDirectionJ2000());
`
Looks like before. 

Also the calendars appear to work. 

Comments/suggestions/improvements welcome.

gzotti:
I have added glue classes V3d, V3f and Color which interact with the now non-scriptable Vec3d and Vec3f, and added some wrapper methods to the core scripting object. The SUG has new instructions. It appears that with a little care we can write scripts which work with both series. If some JS experts want to improve the system, please feel invited, else IMO we can use this solution.

alex-w (approved):
Looks good! I've tested few scripts - no problem.

PR #2463, opened 2022-05-25
add new names to DSO

silas1037:


alex-w:
Why name for NGC 1966 is skipped?

alex-w (approved):
Thanks!

silas1037:
> Why name for NGC 1966 is skipped?

That is a name I added but duplicate with another nebula, which is more famous with the name. The ordering regarding the precedence of duplicate names seems to be catalog-number dependent, but I would like to have some way of making the order definable, like in the order of names.dat, before adding some duplicate names.

PR #2477, opened 2022-06-01
AstroCalc: Transits of Mercury and Venus

worachate001:
### Description
I put them into the same tab with eclipses. Checking against NASA Eclipse website between the years 1601 to 2300 for Mercury and -1999 to 4000 for Venus, not a single transit is missing from our results.

Note: There is prediction of near simultaneous transits of Mercury and Venus that occur on the same date (about 16 hours apart) on 17 September 13425 (TT or Dynamical Time) publish in Meeus's book. In order to simulate/predict the event, I extend the valid range of years to -13000 to 17000. There is no such event for VSOP87 analytical theory but we can confirm the event when using JPL DE431 with accuracy to within a minute.

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10, Ubuntu 20.04
* Graphics Card: NVidia Geforce RTX

## Checklist:

### DIFF
--- a/src/gui/AstroCalcDialog.cpp
+++ b/src/gui/AstroCalcDialog.cpp
@@ -170,6 +170,7 @@ void AstroCalcDialog::retranslate()
 		setLunarEclipseHeaderNames();
 		setSolarEclipseHeaderNames();
 		setSolarEclipseLocalHeaderNames();
+		setTransitHeaderNames();
 		populateCelestialBodyList();
 		populateCelestialCategoryList();
 		populateEphemerisTimeStepsList();
@@ -341,7 +342,7 @@ void AstroCalcDialog::createDialogContent()
 	connect(ui->rtsTreeWidget, SIGNAL(doubleClicked(QModelIndex)), this, SLOT(selectCurrentRTS(QModelIndex)));
 	connect(objectMgr, SIGNAL(selectedObjectChanged(StelModule::StelModuleSelectAction)), this, SLOT(setRTSCelestialBodyName()));
 
-	// Tab: Eclipses
+	// Tab: Eclipses and Transits
 	initListLunarEclipse();
 	connect(ui->lunareclipsesCalculateButton, SIGNAL(clicked()), this, SLOT(generateLunarEclipses()));
 	connect(ui->lunareclipsesCleanupButton, SIGNAL(clicked()), this, SLOT(cleanupLunarEclipses()));
@@ -357,6 +358,11 @@ void AstroCalcDialog::createDialogContent()
 	connect(ui->solareclipseslocalCleanupButton, SIGNAL(clicked()), this, SLOT(cleanupSolarEclipsesLocal()));
 	connect(ui->solareclipseslocalSaveButton, SIGNAL(clicked()), this, SLOT(saveSolarEclipsesLocal()));
 	connect(ui->solareclipselocalTreeWidget, SIGNAL(doubleClicked(QModelIndex)), this, SLOT(selectCurrentSolarEclipseLocal(QModelIndex)));
+	initListTransit();
+	connect(ui->transitsCalculateButton, SIGNAL(clicked()), this, SLOT(generateTransits()));
+	connect(ui->transitsCleanupButton, SIGNAL(clicked()), this, SLOT(cleanupTransits()));
+	connect(ui->transitsSaveButton, SIGNAL(clicked()), this, SLOT(saveTransits()));
+	connect(ui->transitTreeWidget, SIGNAL(doubleClicked(QModelIndex)), this, SLOT(selectCurrentTransit(QModelIndex)));
 
 	// Let's use DMS and decimal degrees as acceptable values for "Maximum allowed separation" input box
 	ui->allowedSeparationSpinBox->setDisplayFormat(AngleSpinBox::DMSSymbols);
@@ -565,6 +571,7 @@ void AstroCalcDialog::createDialogContent()
 	ui->gammaNoteLabel->setStyleSheet(style);
 	ui->gammaNoteSolarEclipseLabel->setStyleSheet(style);
 	ui->UncertaintiesNoteLabel->setStyleSheet(style);
+	ui->transitNoteLabel->setStyleSheet(style);
 	style = "QCheckBox { color: rgb(238, 238, 238); }";
 	ui->sunAltitudeCheckBox->setStyleSheet(style);
 	ui->moonAltitudeCheckBox->setStyleSheet(style);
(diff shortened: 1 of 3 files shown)


src/gui/AstroCalcDialog.cpp
```diff
@@ -3749,6 +3756,311 @@ void AstroCalcDialog::saveSolarEclipsesLocal()
...
 }
 
+void AstroCalcDialog::setTransitHeaderNames()
+{
+	transitHeader.clear();
+	transitHeader << q_("Date");
+	// TRANSLATORS: The name of column in AstroCalc/Transits tool
+	transitHeader << qc_("Planet", "column name");
+	// TRANSLATORS: The name of column in AstroCalc/Transits tool
+	transitHeader << qc_("Separation", "column name");
```
alex-w, line 3766: Why the column is named 'separation'? What is mean 'separation' here?

src/gui/AstroCalcDialog.cpp
```diff
@@ -6210,7 +6522,8 @@ void AstroCalcDialog::changeEclipsesTab(int index)
 	const QMap<int, QString> headermap = {
 		{0,	q_("Table of solar eclipses")},
 		{1,	q_("Table of solar eclipses visible in current location")},
-		{2,	q_("Table of lunar eclipses")}
+		{2,	q_("Table of lunar eclipses")},
+		{3,	q_("Transits of Mercury and Venus")}
```
alex-w, line 6526: Maybe "Transits of Mercury and Venus across the Sun" will be better here?

alex-w (changes requested):
Is it possible to rename "Date" column into "Date of mid-transit" and add 2 columns - "Date of start transit" and "Date of end transit"? Maybe adding the column "Duration of transit" (difference between "start datetime" and "end datetime") will be helpful also.

src/gui/AstroCalcDialog.cpp
worachate001: It's minimum angular distance between the planet and Sun's center. I put that into tooltip (we can change it to "Angular distance" or something like that for clearer meaning).

src/gui/AstroCalcDialog.cpp
worachate001: Agree.

src/gui/AstroCalcDialog.cpp
alex-w: OK, I see your explanation and I found related part in the code :) Thanks! I think we should use clearer term here. 

I think the comment `// TRANSLATORS: The name of column in AstroCalc/Eclipses/Transits tool` will be better also.

worachate001:
> Is it possible to rename "Date" column into "Date of mid-transit" and add 2 columns - "Date of start transit" and "Date of end transit" (or "Greatest transit", "First contact" and "Last contact" respectively)? Maybe adding the column "Duration of transit" (difference between "start datetime" and "end datetime") will be helpful also.

Yes, will be more helpful. There will be two sets of contacts (geocentric and topocentric). We can provide geocentric contacts for general info and topocentric contacts for local info but we may have to separate them by adding another tab (like solar eclipses).

alex-w:
> Yes, will be more helpful. There will be two sets of contacts (geocentric and topocentric). We can provide geocentric contacts for general info and topocentric contacts for local info but we may have to separate them by adding another tab (like solar eclipses).

Just compute the topocentric contacts for table (users can compute geocentric contacts by toggle flag in the settings of whole planetarium ;) ) - no need extra columns for geocentric data.

alex-w:
In any way thank you very much for interesting feature!

worachate001:
> IMHO: The name of tab should be stay "Eclipses", but added the tooltip "Eclipses and transits" because in some languages the translation of phrase "Eclipses and transits" may be very long.

Okay. I feel the same that it's a bit too long.

worachate001:
> In any way thank you very much for interesting feature!

I though about adding this months ago but just found the way to get the position of Mercury and Venus :D

worachate001:
This is latest development (in case you may have more suggestions before another push commit). Both geocentric and topocentric information are covered.

Stellarium-Transits

alex-w:
Let's imagine we show to users geocentric or topocentric data according to flag "Tools -> Topocentric coordinates". In this case we can reduce the number of columns: _Planet, Exterior Ingress, Interior Ingress, Mid-transit, Distance, Interior Egress, Exterior Egress, Duration_. The column _Mid-transit_ should contains date and time, other _Ingress/Egress_ fields - time only, but with full date and time as tooltip. The column _Duration_ should contains the duraction of transit visibility. For geocentric computations all fields will be filled. For topocentric computations - as we can see on the screenshot. 

Some notes for topocentric computations:
- Duration can be computed with comparison with RTS data: Mercury at year 2006 or at year 2111 - apparently duration will be computed as difference between time of rising and exterior egress. Mercury at year 1970 - duration is equals to difference between Exterior Ingress and Set time.
- Double click on the line should set mid-transit date and time, but if at this time the planet is under horizon, then we should use other date and time - as example the middle of visibility. At the moment we don't have column where we can use to stroring these data as Qt::UserRole field. Maybe adding column "Vis. Cond." ("Visibility Conditions" like for Lunar eclipses) can resolve problem for storing the extra data + share human readable suggestions for observations (This column should contains dash for geocentric computations of course).

@gzotti any comments and suggestions?

src/gui/AstroCalcDialog.cpp
gzotti: I think we can really leave away "column name" in the translation string of "Planet", thus re-using/saving one translation item. 
The term "separation" is OK for me, esp. when explained in the tooltip. But distance is also OK.

alex-w: > I think we can really leave away "column name" in the translation string of "Planet", thus re-using/saving one translation item.

We don't have translatable string exactly "Planet" :)

gzotti:
WOW! Actually the layout of the screenshot is superb, and having geocentric and topocentric data on the same table is very welcome, esp. if marked with the green/yellow bars (current PR build does not show it). Else the geocentric columns should be marked "GC distance"..."GC duration". Also thanks for the long-time check! The fields in the topocentric data could differ however between "---" (not existing) and "unobservable" (because of "below horizon"). 

If user is interested only in geocentric data and has topocentric correction configured away (Tools->Topocentric correction disabled), then only create the geocentric columns.

About the second point: double click sets maximum time at location. Yes. And if the planet is below horizon at maximum/center ("unobservable"), you can set to sunrise/sunset time (maybe with the sun sitting "above" the math. horizon, not below as usual!).

src/gui/AstroCalcDialog.cpp
gzotti: Uh. Yes, we have e.g. "planet", "Planet:", "Planets", and tens of combinations. But it's such a general term that we don't have to self-limit its use as "column name". it may easily be used elsewhere.

gzotti:
Actually, you can also just set the time of mid-event for simplification, even if below the horizon. Users may have to switch off the ground if they see grass in the screen center. ;-)

alex-w:
> The fields in the topocentric data could differ however between "---" (not existing) and "unobservable" (because of "below horizon").

+1 

> If user is interested only in geocentric data and has topocentric correction configured away (Tools->Topocentric correction disabled), then only create the geocentric columns.

Yes, this is acceptable behaviour for me too.

> About the second point: double click sets maximum time at location. Yes. And if the planet is below horizon at maximum/center ("unobservable"), you can set to sunrise/sunset time (maybe with the sun sitting "above" the math. horizon, not below as usual!).

+1

worachate001:
@alex-w  @gzotti  Thank you for comments and suggestions. I believe we are very close to finish this feature. The table is more compact with fewer columns. I tried inserting the word "invisible" in the table but that makes the table hard to read, so I use --- to symbolize it, dash means there is no specific contact -- very rare case of grazing transit, 1937, for example).

Unchecking topocentric box in Tools will compute the geocentric contact times. IMHO, I think the visibility conditions are not needed, the existence/non-existence of contact times should already tell users that it's visible or invisible at their location and we may already have too many columns.

src/gui/AstroCalcDialog.cpp
worachate001: I guess changing it to q_("Planet")

alex-w:
> @alex-w @gzotti Thank you for comments and suggestions. I believe we are very close to finish this feature. The table is more compact with fewer columns. I tried inserting the word "invisible" in the table but that makes the table hard to read, so I use --- to symbolize it, dash means there is no specific contact -- very rare case of grazing transit, 1937, for example).

Did you tried use unicode symbol `U+2E3A` or `U+2E3B` instead "---" (see definition for `dash` variable)

alex-w (approved):

gzotti:
I can check this only in the evening, please wait before merging.

alex-w:
> I can check this only in the evening, please wait before merging.

OK, I'll wait your review

alex-w:
@worachate001 please update related part of Stellarium User Guide

worachate001:
> @worachate001 please update related part of Stellarium User Guide

Okay, I just found a small bug, need to fix it too. But I will be busy this evening and may not be able to finish it tonight. I will make it draft for now.

src/gui/AstroCalcDialog.cpp
```diff
@@ -3751,6 +3758,593 @@ void AstroCalcDialog::saveSolarEclipsesLocal()
...
+	result.ce = ce;
+	result.magnitude = magnitude;
+	result.altitude = altitude;
+	return result;
+}
+
+TransitBessel::TransitBessel(PlanetP object, double &besX, double &besY,
+	double &besD, double &bestf1, double &bestf2, double &besL1, double &besL2, double &besMu)
+{
+	// Besselian elements (adaped from solar eclipse)
```
gzotti, line 3893: adapted

src/gui/AstroCalcDialog.cpp
```diff
@@ -3751,6 +3758,593 @@ void AstroCalcDialog::saveSolarEclipsesLocal()
...
+	{
+		initListTransit();
+		const double currentJD = core->getJD(); // save current JD
+		const bool saveTopocentric = core->getUseTopocentricCoordinates();
+		for (int p = 0; p < 2; p++)
+		{
+			double startyear = ui->eclipseFromYearSpinBox->value();
+			double years = ui->eclipseYearsSpinBox->value();
+			double startJD, stopJD;
+			StelUtils::getJDFromDate(&startJD, startyear, 1, 1, 0, 0, 1);
```
gzotti, line 3956: could start at 0:00:00 hours

gzotti:
Hm, I must say I liked the previous solution (geocentric+topocentric) better.  :-(

If you have found a time for events and just replace it with "---" because it's unobservable, maybe putting the value in brackets would still be better.

One more: If user has configured "Use decimal degrees", the angular distance should likewise be given in decimal degrees.

worachate001:
I put them in parentheses and make them gray like this. I think it looks better than just parentheses.
Stellarium-Transits2

alex-w:
> I put them in parentheses and make them gray like this. I think it looks better than just parentheses. 

Please update note also ;)

gzotti:
> Column "Duration" - please use center or right alignment

And a tooltip should clarify that this is "observable duration". Else duration should be time(contact4)-time(contact1).

src/gui/AstroCalcDialog.cpp
worachate001: Done.

src/gui/AstroCalcDialog.cpp
worachate001: Done.

worachate001:
All above suggestions are now included in new update. Thanks for that. I have also edited the SUG (waiting for new screenshots after finalizing the 0.22.2 version - using lunar eclipses screenshot in place of planetary transits for now). Adding new content seems to create some issue with automatic arrangement of text and images surrounding it.

alex-w (changes requested):
Ingress/Egress columns had wrong sorting rules - please add time in decimal format as `Qt::UserRole` data into columns and apply sorting by these data. Similar solution already exists for other tables.

worachate001:
> Ingress/Egress columns had wrong sorting rules - please add time in decimal format as `Qt::UserRole` data into columns and apply sorting by these data. Similar solution already exists for other tables.

I tried adding this for external ingress:
treeItem->setData(TransitContact1, Qt::UserRole, StelUtils::getHoursFromJulianDay(JD1+utcShift));
It seems to make no different of sorting behavior, not sure how to do it properly?

alex-w:
In your case you can just add JD1 without converting or additions

alex-w (approved):

alex-w:
This is a cosmetic requirement and I can do it after merge

worachate001:
> This is a cosmetic requirement and I can do it after merge

The result is not exactly what I expected. I have also added explanation of contacts in tool tip.

alex-w:
> > This is a cosmetic requirement and I can do it after merge
> 
> The result is not exactly what I expected. I have also added explanation of contacts in tool tip.

Thanks for tooltips. You have missing columns in `AstroCalcDialog.hpp file` - see part `if (column == AstroCalcDialog::TransitDate)` to enable correct sorting the Ingress/Egress and Duration columns

worachate001:
> > > This is a cosmetic requirement and I can do it after merge
> > 
> > 
> > The result is not exactly what I expected. I have also added explanation of contacts in tool tip.
> 
> Thanks for tooltips. You have missing columns in `AstroCalcDialog.hpp` file - see part `if (column == AstroCalcDialog::TransitDate)` to enable correct sorting the Ingress/Egress and Duration columns

I see. I have to make a list of what to do, there are things that I keep forgetting.

gzotti:
I just tried current state, and found a transit of Mercury where I can observe only first contact. (Also 2nd contact, but this is flagged as unobservable as part of the solar disk is already below the horizon.)  Duration is given as 0, caused by the algorithm. I really think now that duration should just be based on contact times, not RTS times. The visibility clues with gray and brackets should be enough to indicate full or partial observability.

alex-w:
> I just tried current state, and found a transit of Mercury where I can observe only first contact. (Also 2nd contact, but this is flagged as unobservable as part of the solar disk is already below the horizon.) Duration is given as 0, caused by the algorithm. I really think now that duration should just be based on contact times, not RTS times. The visibility clues with gray and brackets should be enough to indicate full or partial observability.

RTS of the planet should be used for prediction the duration of observable transit. I didn't check the code for this case :(

gzotti:
OK, we disagree on this point. Can we then please have two columns: "Duration" and "Observable"? A contact-time based duration indicated how close to the Solar rim the transit occurs. (OK, there is also a separation column. But still, this is interesting.) Or maybe have another configuration flag for that.

My location is a bit south of Vienna, 
grafik

. Check 2062.
grafik

alex-w:
> OK, we disagree on this point. Can we then please have two columns: "Duration" and "Observable"? A contact-time based duration indicated how close to the Solar rim the transit occurs. (OK, there is also a separation column. But still, this is interesting.) Or maybe have another configuration flag for that.

Extra column is acceptable for me, but this is low interesting data for me (I prefer to get homogen geocentric/topocentric data).

> Check 2062.

E.I. = 20h15m21s; I.I = 20h18m58s; Set: 20h20m, so, the "duration" columns should contain around 5 minutes and I.I. should exists too (Sun set is 20h21m) --- location: Vienna, Austria

PR #2497, opened 2022-06-27
Quick way to jump to the solstice/equinox

### LINKED ISSUE #1394: Quick way to jump to the next solstice/equinox
Now that Stellarium has fixed nutation, we can accurately display the time of solstices and equinoxes. It would be great to have a button, or a key command, to jump to the next solstice or equinox point. When I press it, the time is set to 2020-12-21 13:30 UTC, and when I press it again, I go to 2021-03-21 09:37 UTC etc.

alex-w:
### Description
This pull request contain the code for compute the time of equinoxes and solstices and set of actions to jump to the solstice/equinox at current, next or previous year. The default hotkeys for actions are not defined.

Fixes #1394 (issue)

### Type of change

## Checklist:

### DIFF
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -311,6 +311,20 @@ void SolarSystem::init()
 	addAction("actionShow_Planets_EnlargeSun", displayGroup, N_("Enlarge Sun"), "flagSunScale");
 	addAction("actionShow_Skyculture_NativePlanetNames", displayGroup, N_("Native planet names (from starlore)"), "flagNativePlanetNames", "Ctrl+Shift+N");
 
+	QString timeGroup = N_("Date and Time");
+	addAction("actionCurrent_March_Equinox",      timeGroup, N_("March equinox at current year"),      this, "currentMarchEquinox()");
+	addAction("actionNext_March_Equinox",         timeGroup, N_("March equinox at next year"),         this, "nextMarchEquinox()");
+	addAction("actionPrevious_March_Equinox",     timeGroup, N_("March equinox at previous year"),     this, "previousMarchEquinox()");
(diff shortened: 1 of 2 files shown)


alex-w:
@gzotti please review this PR

P.S. I guess we can add new groups of actions - Special date and time - for equinoxes/solstices and RTS/twilight. What you think about it? Mabe with extract all related code into separate class.

Atque:
I see this only has hotkeys that can be assigned. Would it be possible to add a windows or something (perhaps an extra tab in the Date/Time menu?), so the user won't have to use designated keys?

alex-w:
Maybe later, I have some ideas for it. 

P.S. Thank you very much for the screenshot

Atque:
Hm, yes, that might be the most sane place for it.

gzotti (changes requested):
I see several problems: 
(1) You are always computing all 4 dates and then use just one. This is usually not a problem with today's fast PCs, but a principal waste of computing. 
(2) Add source! This is from Astronomical Algorithms (2nd ed.), ch.27. 
(3) It is only for -1000...+3000 and delivers 0 or does nothing outside those dates. Most users only interested in their own lifetimes will not notice. However, a more general approach should find 0/90/180/270 from ecliptical longitudes in our own algorithms, be it VSOP or DE*. The Calendars plugin has such a function. Calendar::solarLongitudeAfter(). It also uses its private version of solar longitude, to stick to its source book, but this and related functions could be used for a more general solution.  For now, you can add these functional deficiencies as code and SUG comments, and add into the Doxygen a @todo note that we don't forget to improve this function later. Or we try to solve it now for good.

alex-w:
Hmm... The (3) can be not a fast to resolve, so, let me postpone this PR to version 0.22.3

alex-w:
This PR was superseded by the [specific-time](https://github.com/Stellarium/stellarium/tree/refactoring/specific-time) branch

Issue #1328, opened 2020-10-15
shortcuts window: double-click to execute action


axd1967:
**Is your feature request related to a problem? Please describe.**
Sometimes I want to see / test what the result is of a specific action, without having to define a shortcut. The shortcuts window (F7) has a nice feature to filter for e.g. "month", which then gives all month actions:

![image](https://user-images.githubusercontent.com/3529789/96086069-fbbee200-0ec1-11eb-9a3e-930f3b639a7a.png)

**Describe the solution you'd like**
When I double-click on a shortcut, I would like that action to trigger. E.g. when click on "subtract one draconic month", Stellarium executes it.

This could be a controversial issue, because it is not the task of the shortcut window to do this kind of things. But I see no immediate use for a double-click in this dialog, although plenty of counter arguments could be given (e.g. reset the action to default).

**Describe alternatives you've considered**
Assign a shortcut, press it, reassign the shortcut to a different key, press it, etc.

**Additional context**
Note that the arrangement of a dialog having actionable entries is a very nice alternative to execute actions.

A similar approach works in HiPS:
(note that the user needs to check rather than double-click, but this could easily be added; in fact, there is currently a small ergonomy issue in this behaviour as the user is forced to click inside the checkbox)
![image](https://user-images.githubusercontent.com/3529789/96086280-61ab6980-0ec2-11eb-9ace-cddd3a79e955.png)

in Oculars:
![image](https://user-images.githubusercontent.com/3529789/96086363-81db2880-0ec2-11eb-8ef3-7cf919422949.png)

In sats:
![image](https://user-images.githubusercontent.com/3529789/96086415-9b7c7000-0ec2-11eb-9d6d-362529b6ed0c.png)

alex-w:
Double-click in shortcut dialog is enable editor of shortcut

axd1967:
> Double-click in shortcut dialog is enable editor of shortcut

Dang - you mean this

https://github.com/Stellarium/stellarium/blob/ac0bfee840ccd2ae006256784cfdaa858bb39eec/src/gui/ShortcutsDialog.cpp#L307-L308

Back to the drawing board.

alex-w:
Yes, you mark correct LOC

PR #2516, opened 2022-07-15
[refactoring] "Specific Time"

### LINKED ISSUE #1394: Quick way to jump to the next solstice/equinox

alex-w:
### Description
This pull request contain 2 parts:
- [x] The code for compute the time of equinoxes and solstices and set of actions to jump to the solstice/equinox at current, next or previous year. The default hotkeys for actions are not defined.
- [x] The code and actions for jump to the moments of rise, set and transit (include twilight) of selected objects. This code was moved from `StelObjectMgr` class.

The computation the equinoxes and solstices are divided on separate methods now and both solutions are obtained from "Astronomical Algorithms". Valid range -1000..+3000 with zero values outside of it. 

Notes from the [review](https://github.com/Stellarium/stellarium/pull/2497#pullrequestreview-1023516881) of previous solution: 
- [x] You are always computing all 4 dates and then use just one. This is usually not a problem with today's fast PCs, but a principal waste of computing (_fixed by dividing the methods_ -@alex-w)
- [x] Add source! This is from Astronomical Algorithms (2nd ed.), ch.27. (_fixed by adding the info_ -@alex-w)
- [x] It is only for -1000...+3000 and delivers 0 or does nothing outside those dates. Most users only interested in their own lifetimes will not notice. However, a more general approach should find 0/90/180/270 from ecliptical longitudes in our own algorithms, be it `VSOP` or `DE*`. The Calendars plugin has such a function. `Calendar::solarLongitudeAfter()`. It also uses its private version of solar longitude, to stick to its source book, but this and related functions could be used for a more general solution. For now, you can add these functional deficiencies as code and SUG comments, and add into the Doxygen a **todo** note that we don't forget to improve this function later. Or we try to solve it now for good. (_at the moment I've added **todo** notes_ -@alex-w)

Fixes #1394

### Type of change

## Checklist:


alex-w:
Additional features are not implemented and I want to see some comments (votes?) about the "expectations" of these features.

@gzotti @sushoff @worachate001 @Jocelyn1109 @Atque @axd1967 @johnbellora

Atque:
I for one prefer menus over hotkeys. As I said in the previous PR, I would like a menu where the equinox/solstice dates and times are given for a given year, and an arrow or button, which when clicked sets the time to the corresponding equinox/solstice. A similar window could be used for the different stages of twilight (civil, nautical, astronomical), rises/transits/sets of an object and even lunar phases, if there is code to support it.

alex-w:
> I for one prefer menus over hotkeys. As I said in the previous PR, I would like a menu where the equinox/solstice dates and times are shown for a given year, and an arrow or button, which when clicked sets the time to the corresponding equinox/solstice. A similar window could be used for the different stages of twilight (civil, nautical, astronomical), rises/transits/sets of an object and even lunar phases, if there is code to support it.

I planned add “buttons” into AstroCalc/Almanac tool. This PR is first step for the new tool. The features, which I marked as “additional” are optional and both can be implemented later or not implement. I need some feedback or voting for these features.

axd1967:
A quick forward to a specific time at startup feels unneeded to me - I think this will not be used often; also, shortcuts already make this possible. Probably this is also possible with a user-written startup script (I did not check if these new methods can be invoked via scripting; otherwise that would be very recommendable.)


Also
- I still believe that #1328 is an interesting way to extend Stellarium UI (as an alternative to menu actions), even if it will slightly change the behaviour of the F7 dialog (assuming the current - Qt? - behaviour can be tweaked)

alex-w:
Ok, I’ve removed all additional features from description of PR, because both have a low priority at the moment.

alex-w:
@gzotti please review the pull request

gzotti:
> I for one prefer menus over hotkeys. As I said in the previous PR, I would like a menu where the equinox/solstice dates and times are shown for a given year, and an arrow or button, which when clicked sets the time to the corresponding equinox/solstice. A similar window could be used for the different stages of twilight (civil, nautical, astronomical), rises/transits/sets of an object and even lunar phases, if there is code to support it.

As a minimum with only configurable actions this PR may be OK. But with @Atque 's screenshot in the other PR this could also justify another main dialog reachable from the left menu bar. (Shift-F5? F8?) It could show buttons for all those actions. Probably useful to separate this away from the huge AstroCalc dialog, to keep the screen clean. (I had actually considered adding such a quick-access page to the RemoteControl UI)

alex-w:
I had an idea for compact GUI for these actions for one subpage (AstroCalc/Almanac/Seasons) of new tool. But this is task for another PR.

gzotti:
> I had an idea for compact GUI for these actions for one subpage (AstroCalc/Almanac/Seasons) of new tool. But this is task for another PR.

Yes, you wrote that before. But I think a separate small dialog (high, narrow, small buttons, something to move to the right screen edge) would be better, as it covers less of the screen. Indeed, a topic for another PR.

gzotti (approved):
As first step towards possible GUI changes this looks OK.

gzotti:
> A quick forward to a specific time at startup feels unneeded to me - I think this will not be used often; 

We already have the possibility to start with "system date at XX:XX hours". A new option like "system date at end of civil/nautical twilight" would be fine IMHO. Sure, could also be done in a startup.ssh for "power users".

alex-w:
> > A quick forward to a specific time at startup feels unneeded to me - I think this will not be used often;
> 
> We already have the possibility to start with "system date at XX:XX hours". A new option like "system date at end of civil/nautical twilight" would be fine IMHO. Sure, could also be done in a startup.ssh for "power users".

Yes, but not for regular users.

PR #2535, opened 2022-07-26
Refactor Observability plugin to use signals for reacting to location state changes

colossatr0n:
Previous to this PR, the Observability plugin would track the location state during draw loops. This is error prone as the developer must ensure to reset its state in the appropriate places.

What this PR does is let StelLocationMgr be the sole owner of the state by using signals and slots to react to the location change event. This is the first PR in a series of PRs that are derived from a branch I'm working on. I'll try to keep these PRs as focused as possible.

There are A LOT of whitespace/formatting changes. I used the requested formatting configuration, and I guess it hasn't been run on those files for a long time.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

### How Has This Been Tested?

The changes in this PR are relatively simple. No intentional changes have been made to the calculation logic.

Testing consisted of running Stellarium, enabling the Observability plugin, clicking an object, and ensuring that a report showed up in the bottom left hand side of the screen.

Application logs were also checked to ensure that connections were made and closed at the appropriate times. They were also checked to see if new errors were introduced.

**Test Configuration**:
* Operating system: macOS Catalina v10.15.7
* Graphics Card: Intel Iris 1536 MB

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2536, opened 2022-07-26
Replace object type booleans with methods. This removes the need for …

colossatr0n:
…managing object type state and delegates it to the StelObjectMgr.

### Description

The Observability plugin manages object type state within the draw method. This is an error prone manual process that must be tracked by the developer when refactoring. This PR removes the need to manage the state and delegates it back to the StelObjectMgr by using methods that consume the object selection and return whether or not the object is the specified type.

Fixes # (issue)

### Type of change

### How Has This Been Tested?

The changes in this PR are essentially a one-to-one conversion from the previous logic, with the exception of the introduction of `isSystemPlanet`.

These changes were tested by running a pre-change and post-change stellarium instance, selecting the moon, sun, a planet, and an interstellar star, and comparing the output from each instance.

**Test Configuration**:
* Operating system: macOS Catalina v10.15.7
* Graphics Card:  Intel Iris 1536 MB

## Checklist:

  - I refrained from doing this since I ran the formatter in https://github.com/Stellarium/stellarium/pull/2535.

### DIFF
--- a/plugins/Observability/src/Observability.cpp
+++ b/plugins/Observability/src/Observability.cpp
@@ -110,9 +110,6 @@ Observability::Observability()
 	, curYear(0)
 	, nDays(0)
 	, dmyFormat(false)
-	, isStar(true)
-	, isMoon(false)
-	, isSun(false)
 	, isScreen(true)
 	, hasRisen(false)
 	, configChanged(false)
@@ -238,6 +235,7 @@ void Observability::draw(StelCore* core)
 	bool locChanged, yearChanged;
 	StelObjectP selectedObject;
 	PlanetP ssObject, parentPlanet;
+	QList<StelObjectP> objectSelection = StelApp::getInstance().getStelObjectMgr().getSelectedObject();
 
 // Only execute plugin if we are on Earth.
 	if (core->getCurrentLocation().planetName != "Earth")
@@ -261,7 +259,7 @@ void Observability::draw(StelCore* core)
 	double currLocalT = 24.*modf(currJD + GMTShift,&currJDint);
 	int auxm, auxd, auxy;
 	StelUtils::getDateFromJulianDay(currJD, &auxy, &auxm, &auxd);
-	bool isSource = StelApp::getInstance().getStelObjectMgr().getWasSelected();
+	bool objectWasSelected = StelApp::getInstance().getStelObjectMgr().getWasSelected();
 	bool show_Year = show_Best_Night || show_Good_Nights || show_AcroCos; 
 
 //////////////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


plugins/Observability/src/Observability.cpp
```diff
@@ -1845,3 +1832,32 @@ void Observability::showReport(bool b)
...
+{
+   return !objectSelection.empty() && "Moon" == objectSelection[0]->getEnglishName();
+}
+
+bool Observability::isSun(QList<StelObjectP> & objectSelection)
+{
+   return !objectSelection.empty() && "Sun" == objectSelection[0]->getEnglishName();
+}
+
+bool Observability::isInterstellar(QList<StelObjectP> & objectSelection)
```
gzotti, line 1845: This should be called isStar() or isFixed(). 

We have "Interstellar objects" now, i.e. moving bodies, and they are special types of planets. But this method actually separates planets from other types.

gzotti:
Thanks for taking up work on this plugin. It actually deserves a thorough redesign!

plugins/Observability/src/Observability.cpp
colossatr0n: Changed back to `isStar`.

plugins/Observability/src/Observability.cpp
```diff
@@ -1850,10 +1838,41 @@ void Observability::showReport(bool b)
...
+
+bool Observability::isSun(QList<StelObjectP> & objectSelection)
+{
+   return !objectSelection.empty() && "Sun" == objectSelection[0]->getEnglishName();
+}
+
+bool Observability::isStar(QList<StelObjectP> & objectSelection)
+{
+   if (objectSelection.empty()) {
+       return true; // no object selected, so the selection is the screen/space.
```
gzotti, line 1854: The method should be named isStarOrNothing() in this case!

gzotti: Maybe even isNoSSO(), as it can refer to DSO as well.

gzotti (approved):
Otherwise it seems OK. Please rebase your PRs on master so that we can merge more easily.

alex-w (approved):
Thanks!

plugins/Observability/src/Observability.cpp
colossatr0n: I'll update this. Does "SSO" stand for "Solar System Object?" What does "DSO" stand for?

gzotti: Yes, SSO=Solar System Object. DSO=Deep Sky Object (galaxy, cluster, nebula, SNR [SuperNova Remnant]...)

colossatr0n:
Rebased. Still need to test the rebased changes.

plugins/Observability/src/Observability.hpp
```diff
@@ -370,7 +390,7 @@ private slots:
 	Vec3d EquPos, LocPos;
 
 	//! Some booleans to check the kind of source selected and the kind of output to produce.
```
gzotti, line 392: Uh, sorry. this comment should be updated.

plugins/Observability/src/Observability.cpp
```diff
@@ -1850,10 +1838,41 @@ void Observability::showReport(bool b)
...
+   }
+   // This works because the Planet class only exists for in system objectSelection (including moon and sun).
+   Planet * planet = dynamic_cast<Planet *>(objectSelection[0].data());
+   return planet == Q_NULLPTR;
+}
+
+bool Observability::isSystemPlanet(QList<StelObjectP> & objectSelection)
```
gzotti, line 1861: I am afraid this is not what the original intent was: "SSO but neither sun nor moon." Apart from planets, there are moons, asteroids, comets, KBOs, ...
You should probably just test the Planet name for "Sun" or "Moon" like before.

colossatr0n: On line 1868, it checks that the planet's type is "planet." 

As far as I can tell, the only planet objects with type=planet are the planets within the solar system, e.g., https://github.com/Stellarium/stellarium/blob/51db4ee66f176d5c94f6ebe95eb0a72d33dc9306/data/ssystem_major.ini#L451.

Doesn't checking that the `Planet` type is "planet" ensure that it's not a moon, asteroid, etc.?

gzotti: Yes, but that's the problem. How to check observability for e.g. Vesta? ("moon" in this context is only "The Moon" (Earth's moon))

colossatr0n: Ah, I see. I'll change this.

PR #2593, opened 2022-08-22
Let's enable LTO support

alex-w:
Added LTO support by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,10 +301,10 @@ IF(WIN32)
           # C4996: deprecated POSIX names (used in zlib)
 	  # C5105: defines in macros
           SET(STEL_MSVC_FLAGS "/wd4244 /wd4305 /wd4351 /wd4996 /wd5105")
-          # Avoid type conflict with C++17 standard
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -259,6 +259,9 @@ ADD_DEFINITIONS(-D_USE_MATH_DEFINES)
 # Qt 5.7 requires C++11 support
 SET(STEL_GCC_C_FLAGS "-Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas")
 SET(STEL_GCC_CXX_FLAGS "-Wall -Wextra -Wno-unused-parameter -Wno-unknown-pragmas")
+# LTO support
+SET(STEL_GCC_C_FLAGS "${STEL_GCC_C_FLAGS} -flto -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing")
```
DarthGandalf, line 263: I'd leave these default ones (in CMakeLists) without Werror, but just Wodr, etc (though they might already be enabled via Wall Wextra)

Werror can be added in CI

alex-w:
Well, the LTO should be disabled by default apparently due limitation of CI images

PR #2602, opened 2022-08-25
Add option to activates Link Time Optimization

alex-w:


CMakeLists.txt
```diff
@@ -105,6 +105,7 @@ MESSAGE(STATUS "Platform: ${CMAKE_SYSTEM} (${CMAKE_SYSTEM_PROCESSOR})")
 
 ########### Main global variables ###########
 SET(ENABLE_TESTING 0 CACHE BOOL "Define whether the unit tests should be activated.")
+SET(ENABLE_LTO 1 CACHE BOOL "Define whether the Link Time Optimization should be activated.")
```
gzotti, line 108: You wrote in the PR:
>In the future apparently ENABLE_LTO should be Off by default and enabling for preparation the binary packages only

IMO this should mean it should indeed be 0 by default, and a line of comment above like: 
`# LTO has longer build times but optimizes code (smaller by XY percent)`
`# LTO fails a test for unknown reasons. We can enable it only {with clang|on MacOS|??}`

If somebody finds out what the underlying problem is, I will be happy to read about it. How much binary size becomes smaller on which system, or what other advantages we may see (faster?), should still be evaluated.

alex-w: See changes...

Issue #2551, opened 2022-08-04
Problem with Qt6 packaging on Windows


alex-w:
I'm going to test version 1.22.2-3d47d9a in my Windows 10 with no Qt6 installed. Stellarium refuses to run with these errors:

The code execution cannot proceed because Qt6OpenGL.dll was not found. Including Qt6Charts.dll, Qt6OpenGLWidgets.dll and  Qt6MultimediaWidgets.dll.

_Originally posted by @worachate001 in https://github.com/Stellarium/stellarium/issues/2548#issuecomment-1205057150_

alex-w:
@worachate001 please check artefact from here: https://ci.appveyor.com/project/alex-w/stellarium/builds/44372237/job/v7tjjc13cykqbnt1

worachate001:
> @worachate001 please check artefact from here: https://ci.appveyor.com/project/alex-w/stellarium/builds/44372237/job/v7tjjc13cykqbnt1

Looks like it failed to build.

alex-w:
@gzotti I have a bad news - Qt 6.3.1 didn't work here, because Qt6 needs Windows 10, but at work I have Windows 7. Maybe I can get the work environment in VirtualBox, but probably I will need help from you (as far as I know you have Win10).

gzotti:
Ouch. Yes, Win7 is obsolete and Qt6 does not support it. I have zero knowledge of the deployment settings, though. What to do?

gzotti:
It seems @10110111 has started to add integration steps to the CI (appveyor config). Should we complete #2604 first? I think this can be merged already. (We can still disable building with ShowMySky in case of unexpected emergency.)

gzotti:
@alex-w I must admit I never even built an installer, always ran from the build dir. I tried to run Inno with the script created by cmake, but it seems to assume that Stellarium has already been installed (with all libs?) in C:\Program Files\stellarium. There is no "sudo make install" on Windows. What is the proper process of

- deploy from within QtCreator (to c:\Program Files\stellarium? Other preferred path?
- create installer with Inno.

Is the inno script relevant for the package generation by CI, or does it use something else?

alex-w:
The short instructions:
1) `CMAKE_INSTALL_PREFIX `should be defined at cmake run to avoid mixup the binaries (e.g., `-DCMAKE_INSTALL_PREFIX=C:\stellarium-bin`)
2) Building Stellarium
3) Run target `install`
4) Run target `stellarium-installer`

[...]

PR #2604, opened 2022-08-26
Add support for ShowMySky atmosphere model

### LINKED ISSUE #624: Implementing a new atmosphere model
Moving discussion started in #623 here.

> looking forward to seeing this side-to-side with my timelapse photos!

Are they published anywhere? Would be interesting to compare the model with real-world photos.

> You could allow many user tweaks as well (with a reset to useful defaults!).

This has an issue. Namely, the precomputation stage is quite resource-hungry: on my nVidia GeForce GTX 750Ti it takes several tens of seconds to precompute the necessary textures, yielding 128MiB main texture when I set up constants for a decent resolution (otherwise deep twilight is blocky in azimuth and jumpy in time, see [this animation](https://www.youtube.com/watch?v=AsBlW9WGXMA); if you don't notice jumpiness, try watching it in a lower speed).

Moreover, until I added a couple of `glFinish` calls in [my fork of PAS](https://github.com/10110111/precomputed_atmospheric_scattering/tree/stellarium) (branch "`stellarium`"), precomputation made my X11 session unusable (frozen mouse cursor etc.), and sometimes locked up X server indefinitely. Also, sometimes the results were garbage (e.g. missing red channel — I suppose this is due to some nvidia driver bugs, but might be races in the implementation).

Also, since a decent-resolution main texture is at least 128MiB, precomputation stage takes multiple times that amount of VRAM, so this looks unfriendly to mobile devices.

So, I'm not sure if we really want the user to have an option to change parameters from within Stellarium GUI.

> But please make it switchable (old/new) in the config dialog.

Well, if you insist, OK. Currently my [WIP `atmosphere` branch of Stellarium](https://github.com/10110111/stellarium/tree/atmosphere) simply hacks into class `Atmosphere`.

> And if the new model allows (no time to go into details), think of conditions on other planets (Mars?).

This might well be possible, but for each new planed we need its own profile of atmospheric density, scattering&absorption cross-sections, Mie scattering phase functions (based on a sensible particle size distribution) etc.. With this data we can indeed try to precompute corresponding textures (maybe increasing number of orders of scattering).

10110111:
### Description
This patch adds support for ShowMySky atmosphere model.

* CI is updated to fetch current pre-releases of CalcMySky binaries, so that Stellarium can be built and, potentially, deployed.
* Actual deployment is not changed, it will need someone to actually put `libShowMySky.so` / `ShowMySky.dll` / `libShowMySky.dylib` in the correct place on installing, as well as put the Earth atmosphere model in the package.
* User Guide is not modified yet, this will be done in a subsequent PR.
* I tested 32-bit and 64-bit versions on Windows 11, but only with Qt5, because Qt6 beta appears broken due to #2551.
* ~~Error messages are shown in red, which is not quite readable (see a screenshot below); I need suggestions on how to improve this.~~ Done.
* ~~There is a quirk that the combo box for choosing between Preetham and ShowMySky models needs to be switched back and forth to retry if an error occurs during loading a ShowMySky model. I'll fix this soon.~~ Fixed.

Fixes #624

Error messages in the GUI look like this:

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11 x86_64, Linux From Scratch x86
* Graphics Card: NVidia GeForce GTX 750Ti

## Checklist:


CMakeLists.txt
```diff
@@ -542,6 +542,7 @@ FIND_PACKAGE(Qt${QT_VERSION_MAJOR} COMPONENTS Charts REQUIRED)
 IF(USE_PLUGIN_TELESCOPECONTROL OR ENABLE_GPS)
      FIND_PACKAGE(Qt${QT_VERSION_MAJOR} COMPONENTS SerialPort REQUIRED)
 ENDIF()
+FIND_PACKAGE(ShowMySky REQUIRED)
```
gzotti, line 545: As much as I will like to use it on my PC, I think this should be an optional build component. Just testing regular build on an old Odroid C1 (Ubuntu Mate 20.04LTS (upgrade to offered 22.04 failed), Qt5 only, GL ES2 on Arm SBC with Mali GPU). If anything goes wrong, we should be able to build without ShowMySky. Please add a CMAKE config variable and depending code exclusions. (Many examples in this file.)

alex-w:
> Error messages are shown in red, which is not quite readable (see a screenshot below); I need suggestions on how to improve this.

Apparently strong black will be good here

gzotti:
To make deployment easier, we can probably hold finalized built DLLs in util/CalcMySky like we do for Spout and pack those. Or installation for the user just requires copying files around, then it must be clearly documented what goes where.

10110111:
> Or installation for the user just requires copying files around, then it must be clearly documented what goes where.

I hope the user won't need to do this, just as currently there's no need to manually copy `Qt5Core.dll` anywhere. Ideally, I'd also supply the atmosphere model (textures etc.) with a preset path — so that if the user switches from Preetham to ShowMySky, it'll just work.

10110111:
@alex-w 
> Apparently strong black will be good here

Do you mean like in the screenshot below? To me it doesn't look like an error, and is not prominent enough to call for attention.

Are there any examples of already existing error messages in Stellarium?

a

gzotti:
Thin red was bad. Thick black not conspicuous. What about thick red? And maybe prepend "ERROR:" or "PROBLEM:", else some users may not comprehend this.

10110111:
Yeah, was trying exactly this just now:

b

Another options is like below (I'm currently trying to make the corners rounded to match the style of the groupboxes):

c

10110111:
So, the options with rounded label:

a

b

gzotti:
> So, the options with rounded label:

Good! I prefer the second version.

gzotti:
Missing a file?

>CMake Error at CMakeLists.txt:635 (FIND_PACKAGE):
  By not providing "FindShowMySky.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "ShowMySky", but CMake did not find one.

>Could not find a package configuration file provided by "ShowMySky" with
  any of the following names:

    ShowMySkyConfig.cmake
    showmysky-config.cmake

  >Add the installation prefix of "ShowMySky" to CMAKE_PREFIX_PATH or set
  "ShowMySky_DIR" to a directory containing one of the above files.  If
  "ShowMySky" provides a separate development package or SDK, be sure it has
  been installed.

>-- Configuring incomplete, errors occurred!

10110111:
> Missing a file?

Is it you trying to build Stellarium? You need to pass `-DCMAKE_PREFIX_PATH=<CalcMySky installation path>` so that CMake could find `ShowMySkyConfig.cmake`, see the example [here](https://github.com/10110111/Stellarium/blob/showmysky/.appveyor.yml#L77).

gzotti:
I now see https://10110111.github.io/CalcMySky/installation.html. I built CMS in QtCreator following your hints, but did not run the "install" target. Umm, how would I, in QtCreator? As you know, Windows users are GUI-affine, as the usual command processor is still a nightmare. I am tempted to use the cmake command on WSL, but that builds a WSL-compatible file...

10110111:
> Umm, how would I, in QtCreator?

I think you can add a _Deployment_ step in _Run Settings_ that would run something like `cmake --build . --target install` similar to the build process in the _Build Steps_. After that you should be able to install using *Build*→*Deploy Project* menu.

gzotti:
I cannot find it, sorry. 
Now I added CMake to the PATH and opened a command shell with admin privileges. 
The build had been completed in QtCreator.
From within the build directory in the CMD window, I gave as instructed:
`cmake --build . --config Release --target install
`
This installed CalcMySky in c:/Program Files (x86). Probably it should not be set as default as it needs admin privileges. I will provide `-DCMAKE_INSTALL_PREFIX=../install`  next time :-)

BTW, is this a 32bit app? The (x86) implies this. 

I copied the examples dir also into c:\Program Files (x86)\CalcMySky\CalcMySky.
On launch, it asks for data. Whatever I do, there is an unhappy mix of examples and shaders dir. OK, scatch it and restart, following https://10110111.github.io/CalcMySky/installation.html. 

Only on the command line:

```
cd <git>
mkdir build_CalcMySky CalcMySky_install
cd build_CalcMySky
set PATH=%PATH%;%QT_BASEDIR%\bin
cmake -G "Visual Studio 16 2019" -DCMAKE_CXX_FLAGS="/ID:/StelDev/GIT/glm" -DCMAKE_PREFIX_PATH="D:/StelDev/GIT/eigen/share/eigen3/cmake" -DCMAKE_INSTALL_PREFIX=../CalcMySky_install ../CalcMySky -DQT_VERSION=6
cmake --build . --config Release

```
This compiles, but with a warning:

>...
 Generating check version update
  Building Custom Rule D:/StelDev/GIT/CalcMySky/CMakeLists.txt
  Generating version.cpp
  Automatic MOC and UIC for target version
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241
,5): warning MSB8065: Der benutzerdefinierte Build für das Element "D:\StelDev\GIT\build_CalcMySky\CMakeFiles\3b7dff5fe
2837d27982811eb7d6a439c\check version update.rule" war erfolgreich, aber die angegebene Ausgabe "d:\steldev\git\build_c
alcmysky\check version update" wurde nicht erstellt. Dies kann dazu führen, dass der inkrementelle Build nicht ordnungs
gemäß funktioniert. [D:\StelDev\GIT\build_CalcMySky\version_autogen.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241
,5): warning MSB8064: Der benutzerdefinierte Build für das Element "D:\StelDev\GIT\build_CalcMySky\CMakeFiles\3b7dff5fe
2837d27982811eb7d6a439c\version.cpp.rule" war erfolgreich, aber die angegebene Abhängigkeit "d:\steldev\git\build_calcm
ysky\check version update" ist nicht vorhanden. Dies kann dazu führen, dass der inkrementelle Build nicht ordnungsgemäß
 funktioniert. [D:\StelDev\GIT\build_CalcMySky\version_autogen.vcxproj]
  Building Custom Rule D:/StelDev/GIT/CalcMySky/CMakeLists.txt
  version.cpp
  mocs_compilation_Release.cpp
  Code wird generiert...
  version.vcxproj -> D:\StelDev\GIT\build_CalcMySky\Release\version.lib
  Automatic MOC and UIC for target ShowMySky
...

Now, 
`cmake --build . --config Release --target install
`
Again, a warning
>  Generating version.cpp
  Automatic MOC and UIC for target version
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241
,5): warning MSB8065: Der benutzerdefinierte Build für das Element "D:\StelDev\GIT\build_CalcMySky\CMakeFiles\3b7dff5fe
2837d27982811eb7d6a439c\check version update.rule" war erfolgreich, aber die angegebene Ausgabe "d:\steldev\git\build_c
alcmysky\check version update" wurde nicht erstellt. Dies kann dazu führen, dass der inkrementelle Build nicht ordnungs
gemäß funktioniert. [D:\StelDev\GIT\build_CalcMySky\version_autogen.vcxproj]
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Microsoft\VC\v160\Microsoft.CppCommon.targets(241
,5): warning MSB8064: Der benutzerdefinierte Build für das Element "D:\StelDev\GIT\build_CalcMySky\CMakeFiles\3b7dff5fe
2837d27982811eb7d6a439c\version.cpp.rule" war erfolgreich, aber die angegebene Abhängigkeit "d:\steldev\git\build_calcm
ysky\check version update" ist nicht vorhanden. Dies kann dazu führen, dass der inkrementelle Build nicht ordnungsgemäß
 funktioniert. [D:\StelDev\GIT\build_CalcMySky\version_autogen.vcxproj]
  version.vcxproj -> D:\StelDev\GIT\build_CalcMySky\Release\version.lib
  Automatic MOC and UIC for target ShowMySky

But it seems to have finished. I also copied the examples dir into the install folder. 
However, running ShowMySky asks for a directory with a params.atmo file. Is this expected? 

Now I tried again to build the PR. 
In QtCreator, project panel, added in the CMake args line;  -DCMAKE_INSTALL_PREFIX=../CalcMySky_install/CalcMySky/lib/ShowMySky

And this seemed to have done the trick. Now building...

gzotti:
OK, it built. Started. Atmosphere settings, saw the warning. Copied ShowMySky.dll to the stellarium build directory. Started. Seems to have loaded. 

Now, where are the data? Or do I have to run CalcMySky first? (Sorry, I forgot everything from January (?)...)

When I have the library in the stellarium source dir, there is no warning at start. But when I select a directory "path to data", I see again the "cannot load library" warning. I had expected the library is loaded from one place, and data can be elsewhere?

10110111:
> This compiles, but with a warning

This warning can be ignored, it's expected on MSVC.

> However, running ShowMySky asks for a directory with a params.atmo file. Is this expected?

Yes, the examples are the input for `calcmysky`, which would take this file and compute the model textures and create the rendering shaders. This is what you would feed to `showmysky`. Alternatively, take the model from the pre-release binaries, [earth-atmosphere-model.tar.xz](https://github.com/10110111/CalcMySky/releases/download/v0.0.4/earth-atmosphere-model.tar.xz), unpack, and you can use the directory that appears.

> But when I select a directory "path to data", I see again the "cannot load library" warning. I had expected the library is loaded from one place, and data can be elsewhere?

Please show me the screenshot. The error message should tell exactly what happens. The library is loaded after a valid path has been obtained; its location is unrelated to the model data location.

gzotti:
Ah, ok. create 

`calcmysky --out-dir sample_atmo examples/sample.atmo`

Wow, what a megajob! >10 minutes...
Then moved sample_atmo in <STEL_USER_DIR>/CalcMySky and addressed this dir in the Stellarium GUI.

OK, finally it works. Horray!!

Still, IMO the error warning is misleading: if data are missing, the red panel should not warn that the library could not be loaded.

10110111:
> I cannot find it, sorry.

See the screenshot below. I did on Linux, but I think it shouldn't be too different on Windows.

a

> BTW, is this a 32bit app? The (x86) implies this.

It depends on the compiler and its default settings. I never tried to install in the default prefix on Windows, so I'm not sure if CMake takes it right. How to check whether an exe/dll is 32-bit or 64-bit is described [here](https://superuser.com/a/358459/87525).

10110111:
> Still, IMO the error warning is misleading: if data are missing, the red panel should not warn that the library could not be loaded.

That's why I asked for a screenshot. It doesn't tell anything about the library for me, when I choose a wrong directory:

b

gzotti:
> > I cannot find it, sorry.
> 
> See the screenshot below. I did on Linux, but I think it shouldn't be too different on Windows.

Ah, there! OK, next time... :-)

gzotti:
I quit Stellarium with a CMS configured and active. Move the data elsewhere. Started Stellarium. Atmo panel, it had fallen back to Preetham. Good!

grafik

The panel here only shows an empty red field. 

Selecting another path to data, again invalid, 

grafik


(it says "file not found"). OK. Selected right dir. All fine. 

In the first occurrence, the error message may have to be forced again. 

Not sure, I cannot recreate the previous observation. 

Other than that, it seems to work. :-)

gzotti (approved):
OK, fixes acknowledged.
Fine for me, but user documentation and workflow for lib/data deployment must come soon.

10110111:
BTW, I noticed that CodeQL/Analyze fails at this command:
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DBUILD_DOCS=OFF -DCATKIN_ENABLE_TESTING=OFF -DBUILD_DOCUMENTATION=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_FLAGS=-fpermissive ..
```
But I don't find this command anywhere in the YAML files in `.github/workflows`. How do I fix it?

gzotti:
I am afraid only @alex-w can answer this. However, is `-DCATKIN_ENABLE_TESTING=OFF` correct? (What is Catkin?)

gzotti:
> What about definition some default path for finding the atmosphere model (like for scripts, sceneries, etc.? maybe `INSTALL_DIR/atmosphere/`)?

Yes, this should be done. I have added a dir CalcMySky to my private User Directory where I have added the example atmosphere data directories that came packed with calcmysky.  Another name like atmosphere or maybe clearer "showmysky" (small letters only; In a vague idea that there may even be another skylight model, "atmosphere" may be too general) is possible. I think for the first iteration the path selection is OK, and default could be installDir/showmysky/sample-small-size. However in later development (until V1.1?), we could have the small example data set in the install dir, and for advanced users, private atmosphere datasets with access via StelFileMgr. This could show a combobox that shows the available datasets combined from install and user dir. The directory names could later be used as optional keys in landscape descriptions. Imagine "martian"...

alex-w (approved):
OK, let's add it to catch the issues :)

alex-w:
@10110111 would you check @github-code-scanning reports?

10110111:
@alex-w 
They are fixed now. But something is still wrong with LGTM & CodeFactor. The former seems to also use some kind of autobuild, but I failed to locate its config.

alex-w:
> They are fixed now. But something is still wrong with LGTM & CodeFactor. The former seems to also use some kind of autobuild, but I failed to locate its config.

Please ignore CodeFactor, because the test is not passed by other reasons.

For LGTM see https://lgtm.com/help/lgtm/cpp-extraction#cpp-example and https://github.com/Stellarium/stellarium/blob/master/.lgtm.yml but I fear the downloading external libraries and data are not possible here.

gzotti:
After #2551 I am now trying to locally create an installer package from this PR. I have 

-DCMAKE_INSTALL_PREFIX=../CalcMySky_install/CalcMySky/lib/ShowMySky.

Now the packaging adds D:\StelDev\GIT\CalcMySky_install\CalcMySky\lib\ShowMySky\share\Stellarium etc., but the inno packager errors with `Error on line 50 in D:\StelDev\GIT\build-stellarium-Desktop_Qt_6_3_1_MSVC2019_64bit-Release\stellarium.iss: No files found matching "D:/StelDev/GIT/CalcMySky_install/CalcMySky/lib/ShowMySky\qtstuff\*".`

Probably you should define some other variable like -DSHOWMYSKY_INSTALL=../CalcMySky_install/CalcMySky/lib/ShowMySky  and work with that so that -DCMAKE_INSTALL_PREFIX=<wherever>/Stellarium_install can be used?

I will be away for 2 days now, not sure if I can follow in these days.

10110111:
@gzotti 

Not sure I follow. What package are you configuring with `-DCMAKE_INSTALL_PREFIX=../CalcMySky_install/CalcMySky/lib/ShowMySky`? From what you're describing next it seems that you pass this option to Stellarium's cmake command, which doesn't make sense to me. Did you confuse `CMAKE_INSTALL_PREFIX` and `CMAKE_PREFIX_PATH`? The former defines where to install the package, while the latter specifies search paths for dependencies.

gzotti:
> @gzotti
> 
> Not sure I follow. What package are you configuring with `-DCMAKE_INSTALL_PREFIX=../CalcMySky_install/CalcMySky/lib/ShowMySky`? From what you're describing next it seems that you pass this option to Stellarium's cmake command, which doesn't make sense to me. Did you confuse `CMAKE_INSTALL_PREFIX` and `CMAKE_PREFIX_PATH`? The former defines where to install the package, while the latter specifies search paths for dependencies.

Indeed, now I have set CMAKE_PREFIX_PATH=D:/Qt/6.3.1/msvc2019_64;D:/StelDev/GIT/CalcMySky_install/CalcMySky/lib/ShowMySky

Building works. However, to build the installer as deploy step, you should please apply 5c839b3edda5f3c61 from master, or rebase from master, or edit src/CMakeLists.txt: replace CMAKE_INSTALL_PRERIX by CMAKE_INSTALL_PREFIX (2x).

Then we can at least build the installer *without* the SMS library. It ewould of course be better if some CMakefile entry would add this to the deploy step.

10110111:
It's rebased already and includes this commit (I did it 6 hours ago). That rebase also updated CalcMySky dependency to v0.0.5.

gzotti:
OK, rebuilt CMS/SMS and this PR. Now installer builds, but still without copying SMS library. This can probably made similar to the Spout library, src/CMakeFile.txt line 492. Do you want to do that, or shall I try to find a solution?

10110111:
@gzotti Pushed, please check.

gzotti:
The library is not yet copied over to the installer or at least to the intermediate Stellarium_install directory. Please, as example, check both CMakeList.txt and src/CMakeList.txt for occurrences of Spout, and try to develop something similar to address the library in the ISS installer script. It's split into several locations.

10110111:
> The library is not yet copied over to the installer or at least to the intermediate Stellarium_install directory

Well that's strange. If it fails to copy, the `cmake` run must fail. Does it?

gzotti:
> > The library is not yet copied over to the installer or at least to the intermediate Stellarium_install directory
> 
> Well that's strange. If it fails to copy, the `cmake` run must fail. Does it?

Actually not. The packing of this extra library is done via the ISS script, and currently there is no such copy command, so there is also no error. (This process might be simplified, but I am not familiar with this system yet which at least used to work well for years but may be optimized. @alex-w ?) I cannot say how much the process of manually compiling and deploying an install package differs from the CI process on AppVeyor. I just learned the manual process and the buttons to press in QtCreator a few days ago... As I understand it, at first the parts of the program and the required Qt libraries are placed in particular subdirs (share for Stellarium data, qtstuff for Qt libraries, bin for the actual program plus MS libraries), of the Stellarium_install folder, which by itself is not a place to run Stellarium from. From here they are packed into the installer with re-arranged directories. 

I mentioned the Spout library as example of an external optional library which is copied from a particular location to the final install package. The relevant commands are dispersed among the two mentioned CMakeLists.txt files.

10110111:
OK the Spout's way of installing is a bit confusing, I failed to find how it ends up in `@CMAKE_INSTALL_PREFIX@\bin` (this is why my previous attempt was very wrong). So I've now done the `ShowMySky.dll` installation in an easier way, unrelated to Spout.

The library is now installed along with all the other files at _install_ target activation. Please check that it works.

gzotti:
Unfortunately no ShowMySky.dll lands in bin :-(

FYI, the Spout lib is copied only in the ISS script (it is Windows-only!), in an add-on segment created as ISS_STELLARIUM_STUFF in CMakeLists.txt starting at line 825. The ISS script is itself templated from stellarium.iss.cmake in the cmake folder.

10110111:
Ah, there it is! So the library copying to `bin` is irrelevant. This makes more sense. I've altered the script to similarly take the library from its origin, please see if it works now.

CMakeLists.txt
```diff
@@ -833,6 +834,10 @@ IF(WIN32)
      ELSE()
           SET(ISS_SPOUT "; Spout support is disabled")
      ENDIF(ENABLE_SPOUT)
+     SET(SHOWMYSKY_LIBRARY "${ShowMySky_INCLUDE_DIRECTORIES}/../ShowMySky.dll")
```
alex-w, line 837: Hmm... Shouldn't be this line inside `IF(ENABLE_SHOWMYSKY)/ENDIF()` box?

gzotti: I had the same thought. If the variable is never used, it would not harm much though. 

Finally, deployment (building an installer including the ShowMySky.dll) works on private builds in QtCreator!

10110111: Yeah, this doesn't make a difference. Initially I wanted to check that the file exists, that's why I actually created this variable. But then I thought that ISS will complain anyway, so this check would be superfluous. Now the variable is not useful outside the *if*, I can move it in.

gzotti:
OK, @10110111  can you move the line in, and shall I merge this then, @alex-w ? Does the build now also work on the CI system? I must leave for a week in 3 hours...

10110111:
@gzotti Done.

alex-w:
@gzotti please merge it

gzotti:
@alex-w @10110111 then there will be the point of packing the one demo sample, probably in another PR. Should this be downloaded from a site under your control, or shall we just provide a static dataset? Should we call this folder canonically like showmysky/sample-small-size_atmo ? (My own names are now directories named like the .atmo file, e.g. with-external-spectra_atmo.) Or just like the .atmo file, but without _atmo appended?

10110111:
> Should this be downloaded from a site under your control, or shall we just provide a static dataset?

How do you do it for Spout DLL? I think it should be done the same way for ShowMySky DLL and for the atmosphere model. The fact that the CI here downloads it from the GitHub releases of CalcMySky is just a way that works, not necessarily the best way.

> Should we call this folder canonically like showmysky/sample-small-size_atmo

The `sample-small-size` is just a _sample_, as it calls itself. Stellarium will likely want to adopt this model, but it doesn't have to take this name. I suggest something like "default Earth atmosphere", for future generalization by other planets' atmospheres.

gzotti:
We actually have copied the 32/64-bit dlls from a Spout release into the source tree, which is contested by some who know better (see e.g. #1295). This has the advantage of self-containment, but disadvantage of a "dead copy". When you continue developing, we will have to replace the .dll/.so/.dylib(?) from time to time. Speaking of non-dll, how does this now work on Linux/Mac? 

I tend to keep the name of the source configuration, simply for traceability. We could however call the "small sample" just "default" as we don't also provide the corresponding .atmo file. All other datasets will have to be created by users themselves, or be provided later as downloadable/installable extras. We IMHO should just decide on a directory name that collects all datasets.  As said, how to use, fine-tune or modify this new power feature has to be thoroughly described in the User Guide...

10110111:
I totally agree that system libraries must be preferred by the build system. But when distributing a Windows installer, it'd be quite strange to expect that CalcMySky has been already installed on the user's machine—just as Qt. This is why we put `ShowMySky.dll` into the installer.

The current way ShowMySky library is added to Stellarium does look for system libraries. It's just the CI that needs to fetch the library and pass it to CMake. The question is where we should take the library from for the CI. Currently we take it from GitHub releases.

On Linux distributions the new dependency should be taken by packagers automatically, and they'll likely then also package CalcMySky.

With Macs I can't tell anything, since I don't know its ecosystem.

gzotti:
No, for the Windows Installer, it can at best be assumed that the creator of the installation package has it installed and points to it by those CMAKE variables. Users should not bother, unless they want to create their own atmospheres. 

Yes, if you offer Linux packages for apt/yum/whatever..., this looks clean to me.

alex-w:
@10110111 please write `FindShowMySky` module

10110111:
@alex-w  Why? CalcMySky provides all the necessary files like `ShowMySkyConfig.cmake` for CMake's `find_package` to work.

alex-w:
> @alex-w Why? CalcMySky provides all the necessary files like `ShowMySkyConfig.cmake` for CMake's `find_package` to work.

Please try open Stellarium project in QtCreator as a new project ;)

alex-w:
> @alex-w @10110111 then there will be the point of packing the one demo sample, probably in another PR. Should this be downloaded from a site under your control, or shall we just provide a static dataset? Should we call this folder canonically like showmysky/sample-small-size_atmo ? (My own names are now directories named like the .atmo file, e.g. with-external-spectra_atmo.) Or just like the .atmo file, but without _atmo appended?

What about `atmosphere/earth` (or `atmosphere/PLANET_NAME` in general) for default dataset (similar to models, landscapes & textures)?

gzotti:
Yes, a magic Additional CMake line is required:

-DCMAKE_PREFIX_PATH=../CalcMySky_install/CalcMySky/lib/ShowMySky -DCMAKE_INSTALL_PREFIX=D:/Stellarium_install

gzotti:
> > @alex-w @10110111 then there will be the point of packing the one demo sample, probably in another PR. Should this be downloaded from a site under your control, or shall we just provide a static dataset? Should we call this folder canonically like showmysky/sample-small-size_atmo ? (My own names are now directories named like the .atmo file, e.g. with-external-spectra_atmo.) Or just like the .atmo file, but without _atmo appended?
> 
> What about `atmosphere/earth` (or `atmosphere/PLANET_NAME` in general) for default dataset (similar to models, landscapes & textures)?

As said, somebody may develop yet another skylight module. It should therefore be more specific than "atmosphere". Or we use "atmosphere" now and whoever does the new one finds another name...

"earth" again may be too unspecific, which is why I prefer "default". With different gas mixtures, you can make dry/wet/dusty/... conditions, all for earth. OTOH a PLANET_NAME name scheme would be usable for landcapes as you described.

10110111:
If CalcMySky is installed into the system paths (like `/usr` or `/usr/local` on GNU), this works fine (just checked). If not — you'll have to provide the paths somehow anyway, no `FindShowMySky` module can read your mind.

alex-w:
Damn! The shaders from data set are not working on macOS due Stellarium uses OpenGL 2.1 & GLSL 1.2 only here.

10110111:
> on macOS due Stellarium uses OpenGL 2.1 & GLSL 1.2 only here

Does your macOS support higher OpenGL for any other software?

alex-w:
> Does your macOS support higher OpenGL for any other software?

Yes and no. Google Earth Pro and Celestia said about OpenGL 2.1, but both works without any problem.

10110111:
What error do you get from ShowMySky@Stellarium? Does the `showmysky` utility show anything?

alex-w:
> What error do you get from ShowMySky@Stellarium? 

```
Will load CalcMySky atmosphere model from "/Users/aw/Downloads/earth-atmosphere-model"
QOpenGLShader::compile(Vertex): ERROR: 0:2: '' :  version '130' is not supported

*** Problematic Vertex shader source code ***

#version 130
#define lowp
#define mediump
#define highp
#line 2
in vec4 vertex;
out vec2 texCoord;
void main()
{
	gl_Position=vertex;
	texCoord=vertex.xy*0.5+0.5;
}

***
Error while compiling ShowMySky atmosphere luminance-to-screen vertex shader: ERROR: 0:2: '' :  version '130' is not supported

ERROR: Failed to initialize ShowMySky atmosphere model: Shader compilation failed
WARNING: Falling back to the Preetham's model
```

> Does the `showmysky` utility show anything?

Of course not, because this is not a standard macOS application.

alex-w:
I didn't check, but you are really using GLSL 1.3 in the shaders?

10110111:
Yes, GLSL 1.3 made an incompatible change from 1.2, so it's not just a matter of declaration of another version.
Besides, geometry shaders that are used for rendering into 3D textures became core in OpenGL 3.2, which is a long way from 2.1.

10110111:
> Of course not, because this is not a standard macOS application.

What do you mean? It's just a usual Qt-based Mach-O app, how is it not a standard app?

alex-w:
> What do you mean? It's just a usual Qt-based Mach-O app, how is it not a standard app?

See a short explanation: https://doc.qt.io/qt-6/macos-deployment.html

alex-w:
> If CalcMySky is installed into the system paths (like `/usr` or `/usr/local` on GNU), this works fine (just checked). If not — you'll have to provide the paths somehow anyway, no `FindShowMySky` module can read your mind.

Well, it works, but... writing `FindShowMySky` module may give us:
1) Filling the standard variables for location the library and source code files - no need workarounds for finding the files
2) Standard log messages for finding files - see cmake output in the CLI of within QtCreator
3) Easy adding deployment rules - without workarounds (see point 1)

alex-w:
> > > @alex-w @10110111 then there will be the point of packing the one demo sample, probably in another PR. Should this be downloaded from a site under your control, or shall we just provide a static dataset? Should we call this folder canonically like showmysky/sample-small-size_atmo ? (My own names are now directories named like the .atmo file, e.g. with-external-spectra_atmo.) Or just like the .atmo file, but without _atmo appended?
> > 
> > 
> > What about `atmosphere/earth` (or `atmosphere/PLANET_NAME` in general) for default dataset (similar to models, landscapes & textures)?
> 
> As said, somebody may develop yet another skylight module. It should therefore be more specific than "atmosphere". Or we use "atmosphere" now and whoever does the new one finds another name...
> 
> "earth" again may be too unspecific, which is why I prefer "default". With different gas mixtures, you can make dry/wet/dusty/... conditions, all for earth. OTOH a PLANET_NAME name scheme would be usable for landcapes as you described.

OK, `skylight/default` or `atmosphere/default` + rules for installing it when module has flag ENABLE_SHOWMYSKY=1?

I prefer to use `atmosphere/default` because it enables in atmosphere settings and this is "_visual model of atmosphere_"

@10110111 @gzotti @worachate001 @Atque?

Atque:
Please forgive this rookie question, but what is the difference between `skylight` and `atmosphere`?

gzotti:
> Please forgive this rookie question, but what is the difference between `skylight` and `atmosphere`?

It's more of what users expect from these bits. From "atmosphere" I expect all things atmospheric: gas composition, scattering, refraction, extinction, reddening, .... "skylight" is rather limited to the more or less blue color. E.g. the default Preetham sky model has no notion of extinction or refraction, it just creates the colors. Extinction and refraction, and also the actual sky brightness, are solved by 3 independent, separate classes. A complete "Atmosphere" class could solve these effects in a more coherent way.

gzotti:
> > Does your macOS support higher OpenGL for any other software?
> 
> Yes and no. Google Earth Pro and Celestia said about OpenGL 2.1, but both works without any problem.

This is something we have observed a long time ago: On Windows and Linux, you request an OpenGLFormat like 2.1 and receive whatever the driver can provide, e.g. 3.3 or 4.5 or 4.6. On Mac you get what you requested, i.e. 2.1. As said, after a previous team member had complained about my stupid tests about what kind of OpenGL actually was used, I excluded this test from Mac and took his word as "Macs always work". I added the --compat33 (?) switch to force an OpenGL3.3 Compatibility profile. However, I understand that Macs either have 2.1 or "3.2 Core" only, and no compatibility profile which we however require. When I ask for a 3.2 Core profile, graphics on Windows act weird as the compatibility functions like screen blanking simply don't work any longer. Some OpenGL expert could try to remove/replace the compatibility functions from Stellarium to enable a Core profile. This may be a large job, though. 

I can only hint vaguely at what may have to be done about Macs here. Try to force a 3.2 profile and see how this looks like. Also note that specifying the GLSL language level may break running on GLES devices (our shaders don't specify GLSL level which indicates GLSL1.1 or GLSL ES1.0). Interestingly enough, my most valued GLES device, the RasPi4, meanwhile has real OpenGL2.1 support. I can test that after September 12, earliest. If it does not work, users must at least be informed gracefully, and the system switched back to the old solution.

10110111:
I suspect that current QtOpenGL might be incompatible with some quirks required on macOS to get modern OpenGL versions (see [this SO answer](https://stackoverflow.com/a/26982694/673852) to learn about the quirks).

@alex-w
Could you confirm that [this app](https://apps.apple.com/us/app/opengl-extensions-viewer/id444052073?mt=12) reports OpenGL ≥ 3.2?

Also, I've pushed a CalcMySky branch called `opengl-probe`, could you post what `calcmysky --opengl-info` from this branch prints on your Mac?

10110111:
Just checked on a friend's MacBookPro11,4, where GLView reports OpenGL 4.1 INTEL-18.8.4, and `calcmysky` gets this exact version, so Qt is not a problem.

Now it's just a matter of making Stellarium work with Core profile. But the transition itself is quite a bit of work. I tried doing it for Stellarium about 2-3 years ago, and didn't finish for this reason.

alex-w:
> Could you confirm that [this app](https://apps.apple.com/us/app/opengl-extensions-viewer/id444052073?mt=12) reports OpenGL ≥ 3.2?

I have a MacBook Pro M1, so, Apple M1 renderer support OpenGL up to version 4.1 according to GLView summary & OpenGL reports.

10110111:
I was trying out the current weekly build, and I expected to find a default atmosphere model (shaders, textures etc.) after installation. But I haven't found it. Is it supposed to be there?

alex-w:
> I was trying out the current weekly build, and I expected to find a default atmosphere model (shaders, textures etc.) after installation. But I haven't found it. Is it supposed to be there?

I'm work for this task (the testing build is compiling right now)

alex-w:
@10110111 please see PR #2634

Atque:
@alex-w I saw you had disabled ShowMySky on macOS. Why not enable the model and let the user test if it crashes?

alex-w:
> @alex-w I saw you had disabled ShowMySky on macOS. Why not enable the model and let the user test if it crashes?

OK

alex-w:
@10110111 Qt6-based package has ShowMySky library & data

t4saha:
getting this error :  `params.atmo:1: unsupported format version 5: supported is 6`.

10110111:
@t4saha yeah, this is because you're using a recent version of CalcMySky package and older atmosphere dataset. You can either regenerate the dataset, or temporarily downgrade to `v0.0.5` tag.

I'm in the process of making a new release which will also provide a new dataset.

t4saha:
Thanks @10110111

PR #2616, opened 2022-09-01
Screen Blanking - GPS - Angle Measure Snap - Oculars Format Label

### LINKED ISSUE #2166: Display goes blank frequently in v0.21.3
My display keeps going blank in v0.21.3.  Although the has been a known issue for several previous versions I have been using ANGLE mode to launch Stellarium with no problems.
Now, in this latest version, the screen blanking has returned even in ANGLE mode.  Same thing for MESA mode.  It does not happen when I mouse around on the regular star screen or move the sky view around.  It does seem to occur once I click on any of the options buttons on the left side of the window.  Trying to do some things in the windows that open seems to induce the blanking.

Running Windows 11 on an HP laptop with:
CPU: Intel(R) Core(TM) i7-8565U CPU @ 1.80GHz, 8 Gb RAM, Graphics: Intel(r) UHD Graphics 620, driver v27.20.100.8935
Log file is attached
[log.txt](https://github.com/Stellarium/stellarium/files/7856982/log.txt)
.

### LINKED ISSUE #2472: Borderless fullscreen windowed mode
It's not really addressing any specific problem, I would categorize this instead as a matter of ease of use, especially when switching between Stellarium and other windows on the system.

I would like to be able to run Stellarium in a borderless fullscreen windowed mode, which visually would be presented as is the default fullscreen mode, covering the task bar, etc., but allowing to seamlessly ALT+TAB between windows without entering and exiting the exclusive fullscreen mode, as is the case with the latter. This would resemble the behavior that happens when using ANGLE, but in this case it would not require ANGLE for that purpose.

None, beside using the software in ANGLE mode, or in standard window mode with the downside that it has the title bar and doesn't cover the whole screen resolution (ie: taskbar). You can manipulate the windows with tools such as *Borderless Window Utility*, to force the window to become borderless, but it still will not be able to cover the taskbar. Regardless, it would be great if it could be done natively

**Example of the current window maximized**
stellarium_kd7obQVjv2

**Forced borderless window (still cannot cover taskbar)**
stellarium_L2R8s8x65J

### LINKED ISSUE #2484: Screen goes black for a brief moment then image reappears, flickering
### Steps to reproduce

Start any Stellarium executable, click on any object, scrollin in a bit and screen flickers after that.

### System
* Stellarium version: 0.22.1
* Operating system: Win 10 Pro, 21H2, build 19044.1706
* Graphics Card: Nvidia gtx 1080ti, driver 512.15, no overclocking.
* Screen type (if applicable): LG C9 TV, resolution 1920:1080, HDMI, 120Hz, 12 bpc, RGB signal, no HDR,  scaling 150% (changed scaling - no effect)

### Logfile 
File attached
[log.txt](https://github.com/Stellarium/stellarium/files/8853144/log.txt)

uwes-ufo:
### Description

1) Avoid screen blanking on Intel UHD (#2166 #2472 #2484)
Screen flickering on Intel UHD Graphics is a long-known issue. The backlight turns off and on multiple times! The current workaround is to switch to Angle mode. But in Qt6 there is no more angle mode...
Perhaps the user should make their choice based on a configuration setting value like:
... if (confSettings->value("video/SwapBehaviorSingleBuffer", false).toBool()) ...

2) GPS location managed by Windows
On Windows systems, built-in GPS sensors are managed by the operating system. There is no serial port or NMEA protocol. But Qt can handle it.

3) AngleMeasure ~> Align to selected object
You can pan the screen and left-click to select an object as usual, and a right-click will snap the end of the protractor line (closest to the mouse pointer) to the selected object. Double-click with the right mouse button to pick up the end of the line. Another double click with the right mouse button removes the angle measuring line.

4) Oculars ~> format label on HDPI monitors
Makes the label readable on HDPI screens.

Fixes #2166
Fixes #2472
Fixes #2484

### Type of change
- [?] This change requires a documentation update

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Intel UHD 620

## Checklist:

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>stellarium</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
--- a/plugins/AngleMeasure/src/AngleMeasure.cpp
+++ b/plugins/AngleMeasure/src/AngleMeasure.cpp
@@ -17,6 +17,9 @@
  * Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
  */
 
+#define PAN_AND_ZOOM // This keeps the left mouse button for panning and selecting objects - only the right mouse button is used for angle measurement
+
+#include "StelObjectMgr.hpp"
 #include "StelTranslator.hpp"
 #include "StelUtils.hpp"
 #include "StelProjector.hpp"
@@ -257,8 +260,10 @@ void AngleMeasure::drawOne(StelCore *core, const StelCore::FrameType frameType, 
 		int y  = static_cast<int>(120*ppx);
 		int ls = static_cast<int>(ppx*painter.getFontMetrics().lineSpacing());
 		painter.drawText(x, y, messageEnabled);
+#ifndef PAN_AND_ZOOM
 		y -= ls;
 		painter.drawText(x, y, messageLeftButton);
+#endif
 		y -= ls;
 		painter.drawText(x, y, messageRightButton);
 	}
@@ -324,6 +329,7 @@ void AngleMeasure::handleMouseClicks(class QMouseEvent* event)
 		return;
 	}
 
+#ifndef PAN_AND_ZOOM
 	if (event->type()==QEvent::MouseButtonPress && event->button()==Qt::LeftButton)
 	{
 		const StelProjectorP prj = StelApp::getInstance().getCore()->getProjection(StelCore::FrameEquinoxEqu);
@@ -351,7 +357,6 @@ void AngleMeasure::handleMouseClicks(class QMouseEvent* event)
 #else
 		prjHor->unProject(event->x(),event->y(),startPointHor);
 #endif
-
 		// first click reset the line... only draw it after we've dragged a little.
 		if (!dragging)
 		{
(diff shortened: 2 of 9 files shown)


.project
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gzotti, line 1: What do we need this file for?

plugins/AngleMeasure/src/AngleMeasure.cpp
```diff
@@ -17,6 +17,9 @@
  * Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
  */
 
+#define PAN_AND_ZOOM
```
gzotti, line 20: Don't just define it. Comment what your intent for this is.

gzotti:
These are 4 very different issues, please create 4 pull requests. 

The GPS query was developed on Windows and Linux using an external GPS mouse which I no longer have access to. I think it also worked with some BlueTooth connection from my smartphone. Both used NMEA on WIndows, and gpslib on Linux. I am unaware of builtin GPS sensors but will try if the BT connection still works.

plugins/AngleMeasure/src/AngleMeasure.cpp
uwes-ufo: this definition switches to the "new" behavior - otherwise the "old" behavior is activated

.project
uwes-ufo: This is the standard project file from eclipse. So you can open Stellarium as eclipse project.

uwes-ufo:
> These are 4 very different issues, please create 4 pull requests.
done

alex-w:
> > These are 4 very different issues, please create 4 pull requests.
> 
> done

Could you done it by separate branches (one feature per branch), because currently we have 5 almost identical pull request (differences in the descriptions, but not in the code).

In any way thank you for contribution!

uwes-ufo:
I cleaned up the commits.
Let's try again.
Thank you for your patience!

alex-w:
OK, I'm closed this PR as superseded by the pull requests #2619, #2620, #2621 and #2622

plugins/AngleMeasure/src/AngleMeasure.cpp
gzotti: Will the next developer know this two years from now?

PR #2627, opened 2022-09-04
Adding a new Stellarium Sky Culture: Samoan

alex-w:
### Description
A new Stellarium Sky Culture - Samoan - from Jake Fitisemanu and Dan Smale with small changes from me to respect latest changes in the code of SC support.

### Type of change


skycultures/samoan/description.en.utf8
```diff
@@ -0,0 +1,122 @@
+<h2>Samoan</h2>
+<p>Formerly known as the <i>Navigator Islands</i>, the Samoan archipelago has been the launch site for countless ocean voyages since its settlement 3,000 years ago by the ancestors of today’s Polynesians. 
+Like other Pacific Islander societies, the Samoans possessed an incredible body of astronomical knowledge, cultivated through generations of scientific observational study (tofāmamao). 
+Over time, indigenous wayfinding science has been almost completely replaced by digital navigation technologies, but the names and stories attached to some celestial features have persisted 
+through oral traditions and colonial era publications. Beside their value as nautical aids, the moon, stars, and planets also marked the lunar months, annual socioreligious festivals, and the 
+timing of agricultural cycles. </p>
+
+<h3>General Terminology</h3>
+<ol>
```
gzotti, line 9: I would prefer a dl here and below. Key should be the Samoan term in both cases, this looks more harmonic IMHO.

alex-w: See the text later - not all terms can be good converted into dl-list

gzotti: Sorry, but which line cannot be converted? Most have "samoan"="explanation", or "explanation"="samoan". Few lines have no equal sign, but I think a term can be isolated from them for use as dt key.

alex-w: Done

gzotti (approved):
Looks good!

@sushoff please next time, try to stick to editing files only in the PR (or branch) in question, so we have a less convoluted history.

PR #2631, opened 2022-09-05
Unify method of finding ShowMySky library

10110111:
No need to "find" anything, nor hardcoding relative paths (as I did it before): the `ShowMySky::ShowMySky` target already has the `LOCATION` property.

I have checked that the location is correct on Windows and Mac, but please check that the installer and whatever deployment is done on Mac work correctly.


alex-w (approved):
Thanks!

PR #2647, opened 2022-09-13
Start vectorization of GUI controls: checkboxes, radio buttons, up/down arrows

10110111:
### Description
As there are now more and more high-DPI monitors around, we have the problem of scaling of the GUI. Qt supports this, scaling font sizes, resizing icons and controls, but does it in a very ugly way: the fonts become blurred (though at least not pixelated), and icons and controls are pixelated _and_ blurred.

This PR starts vectorization from the following elements:

* Checkboxes,
* Radio buttons,
* Up and down arrows on spinboxes, comboboxes and other controls that used `uieSpin*.png` icons.

The arrows appeared to be a bit tricky because, while PNG images for them were rendered at their native pixel sizes, the SVG replacements get scaled to the whole control. This made it necessary to add padding to the SVG canvas, and this, in turn, required to split the arrows for spinboxes&comboboxes from those for scrollbars. I haven't tested any other controls with arrows, because I couldn't find them in the Stellarium UI. Also, I didn't change the left and right arrows.

I tried hard to make the SVG versions at 1× scale factor look as closely as possible to the PNG versions of the same images. Exact pixel-wise perfection wasn't reached, but I think they are fairly close, see the screenshots.

### Screenshots:

Original Stellarium GUI at 1× scale:

1a

Vectorized Stellarium GUI at 1× scale:

1b

Original Stellarium GUI at 1.5× scale:

2a

Vectorized Stellarium GUI at 1.5× scale:

2b


alex-w (approved):
Thanks!

t4saha:
Nice job. Now it's nice to have other ui icons as svg too, specially markers. They looks blurry when zoomed. @10110111

10110111:
> Now it's nice to have other ui icons as svg too, specially markers.

Not sure what markers you mean. I did try to start working on the icons like the ones in GUI tabs. But each item takes quite a bit of time for me, since I'm not an artist.

So, I'm looking for ways to automatically trace these PNGs. In particular, if you know what free program can *successfully* trace colorless icons like [`tabDSO.png`](https://github.com/Stellarium/stellarium/blob/master/data/gui/tabDSO.png), I'll be grateful. So far, `potrace` failed (I got empty black square instead of the pattern).

alex-w:
See https://github.com/Stellarium/stellarium/blob/master/data/gui/icons.svg

gzotti:
Maybe @martinber can help with SVG icons?

10110111:
Argh! I lost so much time while everything has already been there!

10110111:
OK, so the question now is to simply cut this "texture atlas". I think this PR should be closed then.

alex-w:
Why? If your solution is working then we should apply it

alex-w:
At least you may extract individual icons into separate files

10110111:
> Why? If your solution is working then we should apply it

It was achieved via reverse-engineering of the raster images, and this is imperfect. I don't think it's good enough to continue with it when we have the vector sources.

Anyway, there has been a good effect from this PR: I've learned that I can achieve the desired result **much** faster :D

Though, the SVG atlas renders rather slowly in my browser, using these icons directly might make Stellarium GUI a bit sluggish. I have to try.

martinber:
Hello! I'm not active on this problem but if you push me a bit I can help in anything you may need.

I couldn't see the difference of the screenshots but I understand that you want to vectorize all icons to make them look well in HiDPI. You may already know some of these, but these are the things I'm thinking about now:

- I'm sorry that you didnt see my SVGs before, I don't know where this should be documented for next people working in GUI

- It can be a problem to use SVG directly in the program as you do because I don't know how good is QT rendering SVG. There can be problems with the shadows and especially the glow effects. This would be a very big problem if later we realize that the icons are rendered differently in different OS. That said, I didnt try to see if actually QT has problems with SVG, maybe it just looks right for all icons.

- It is a possibility to continue doing the things I was doing them, and start rendering icons to .png in a hiDPI and a low DPI version?

- I render the icons one by one to .png by hand with inkscape. It would be nice to have one SVG per icon and render with a script but I only trust inkscape to render properly the shadows, etc. The way I do it right now would take like 30min of manual work to render all icons one by one...

10110111:
Hi @martinber !

I just opened the SVG with Inkscape, and was amazed to see documentation there :) Never even thought that artwork could be documented! Thanks for this!

> I'm sorry that you didnt see my SVGs before, I don't know where this should be documented for next people working in GUI

I think it's my fault that I didn't ask for sources. I just expected that if an icon has a vector source, it would be for this single icon, and not having seen anything with a similar name to a PNG, I assumed the sources were lost.

> It is a possibility to continue doing the things I was doing them, and start rendering icons to .png in a hiDPI and a low DPI version?

Not with Qt. This was what I tried first. At least Windows 11 has multiple scale levels: 100%, 125%, 150%, 175%, And I don't know what other platforms need.

Providing a single hi-DPI version will make Qt resample the images. And Qt does it, as it seems from the results, via nearest-neighbor interpolation, which looks ugly (see the third screenshot above).

Providing all the scale levels might require too many icons.

> I render the icons one by one to .png by hand with inkscape

I seem to remember that Inkscape had some scripting capabilities. This might help with rasterizing single icons.

> There can be problems with the shadows and especially the glow effects.

Hmm, I should have expected that there were glows. My approximation in this PR was a flat thick repetition of the tick in the checkbox (and yeah, I guessed the line cap wrong), see yours vs mine:

exact approximated

Maybe, if Qt renders the glows/shadows too slowly or incorrectly, I'll have to flatten them.

10110111:
So, @martinber , you were right: Qt doesn't support feGaussianBlur, so the glows don't look right when supplied as SVG.

But the problems don't end at this. In particular, we simply can't supply raster icons for scale factors of 125%, 150% and 175% that I see on a Windows machine in screen settings. Instead Qt only accepts icons with integral scale factors, named like `uieCheckbox-checked@2x.png` for 200% scale factor. And when we provide 2x while the actual system scale factor is 150%, we get ugly downscaling: compare PNG vs SVG below. (All pairs of screenshots here have PNG on the LHS, SVG on the RHS.)

1 5x-png 1 5x-svg

OTOH, when supplying 3x and setting `QT_SCALE_FACTOR` to 2.5, scaling is OK:

2 5x-png 2 5x-svg

As for glows, we can see in the following 3x cases that indeed, Qt renders SVG incorrectly:

3x-png 3x-svg

---

So I'm not sure what we should do. Is there an easy way to convert these Gaussian blurs into linear gradients (possibly with multiple color stops to approximate Gaussian profile)?

t4saha:
Screenshot from 2022-09-15 11-46-31

10110111:
@t4saha  This looks like a separate problem unrelated to HiDPI scaling. Also, I don't have this large jagged rounded rectangle here, even if I enable all objects and markers.

martinber:
I think we should check first if there is no some configuration problem when using PNGs as icons, it seems weird to me that only integer scales look well, it would mean that Qt is quite broken in my opinion. Maybe there is some kind of [configuration error I dont understand like this](https://stackoverflow.com/questions/51717443/high-dpi-scaling-in-qt-application-make-icons-smoothed-out)

If SVGs are used as icons, then we should compromise in quality of shadows and glows. If we manage to make the first row of icons [here](https://github.com/Stellarium/stellarium/blob/master/data/gui/icons.svg) look good, then we can continue with easier icons like checkboxes. Maybe also Qt allows to use plain SVGs and add glows and shadows in Qt code.

gzotti:
Yes, the icons used for DSO are fixed-size PNGs. I have made a few versions, but using scalable SVG icons would be nice to have in the upcoming year or so.

10110111:
> [configuration error I dont understand like this](https://stackoverflow.com/questions/51717443/high-dpi-scaling-in-qt-application-make-icons-smoothed-out)

This is quite an easy to understand problem. The whole Qt support for high DPI, at least in its current state, is a kludge, intended to be backwards compatible with old code base that wasn't born high-DPI-aware. Just look at the name: "scaling", not "high-DPI styling". Even the name hints at a half-measure.

What Qt (apparently) does is simulate 96 DPI screen when rendering icons while actually stretching everything by `QT_SCALE_FACTOR`; fonts are also undergoing something fishy, because even though we can see no upsampling artifacts in the text, and it looks great at 2× and other integral scale factors, it still is blurry at 1.5× scale.

SVG rendering, similarly to fonts, also behaves strangely: you can notice how the fourth screenshot in the OP here has artifacts at the bottom of the spin-down arrows in the spinboxes on the RHS (see e.g. "Absolute scale").

Finally, when I tried to vectorize the close button, I found that SVG renders exactly as the original PNG at 1.5×: as if rendered in low resolution and coarsely upscaled.

So, unfortunately, Qt support for high DPI seems broken, I expect to have to report some bugs to Qt if these haven't already been filed. Though, what we may see as crazy, Qt might say "works as intended", like in [this bug report](https://bugreports.qt.io/browse/QTBUG-79692). I now see how much premature this PR was.

10110111:
> Yes, the icons used for DSO are fixed-size PNGs. I have made a few versions, but using scalable SVG icons would be nice to have in the upcoming year or so.

I think what @t4saha complains about is that `neb_diff_xlrg.png` is aliased, while all other `neb*.png` are OK. This indeed looks like an issue to be reported. But definitely not in this thread ;)

PR #2658, opened 2022-09-18
Increase resolution of some PNGs to improve high-DPI look

10110111:
This is a second attempt at #2647. Now I don't try to go the vectorization route, but instead just increase resolution of the raster images.

5× increase was chosen because at least in Windows, it seems to be the maximum scaling factor supported. In any case, I think it should suffice for at least a decade.

Not all images have been changed (far from this, in fact):

* Spinbox arrows will require more attention due to the problems I had with SVGs on them in the previous attempt.
* Dialog Close button doesn't fit the image inside the widget due to the use of `background-image` CSS property, and switching to anything else like `qproperty-icon` will lose hover effect. So more work is required than a simple change of PNGs.
* Side/bottom/top panel buttons are based on `QGraphicsPixmapItem`, which bases its size and bounding rectangle on resolution of the image, and even if we fix this by `setScale` or `QPixmap::setDevicePixelRatio` (and a bit of other actions), the result of downsampling of the high-resolution images is ugly.

# Screenshots

At scale factor 1 nothing changes: first image how it was, second how it's now:

## View dialog
1a1

2a1

## Configuration dialog

1b1

2b1

Now, scale factor 1.5:

## View dialog

1a1 5

2a1 5

## Configuration dialog

1b1 5

2b1 5


gzotti (approved):
I see it looks "nicer", at cost of a few kB. Fine for me.

alex-w (approved):

PR #2660, opened 2022-09-18
Add support for up to 500% scaling of dialog close button

10110111:
The close button is now implemented as a special widget class, `StelCloseButton`, which changes icon of its `QPushButton` superclass on mouse hover to achieve the same effect as before but in a way compatible with high-DPI scaling.

The border of the button has been made transparent to make it look better, but it retained 1px size. (I assume the comment about Qt 5.15 regression implies the border shouldn't ideally be there.)

# Screenshots

## Scaling factor 100%

Before:

1a

After:

2a

## Scaling factor 150%

Before:

1b

After:

2b


alex-w (approved):
Thanks!

10110111:
Just found an issue with this button in a Qt-5.12.8 build. For some reason the icon is stretched too much. I hope to find a fix in several days. What is the expected date of release?

close-button-issue

alex-w:
25 September

PR #2668, opened 2022-09-19
Paint version on the fly on the splash screen instead of rasterizing it into a pixmap

10110111:
This makes high-DPI splash screen a bit nicer. See the screenshots below.

## Scaling factor 100%

Before
1a

After
2a

## Scaling factor 150%

Before
1b

After
2b

### DIFF
--- a/src/StelSplashScreen.hpp
+++ b/src/StelSplashScreen.hpp
@@ -27,16 +27,10 @@ class SplashScreen
 {
 	class SplashScreenWidget : public QSplashScreen
 	{
+		QFont splashFont;
 		bool painted=false;
-		void paintEvent(QPaintEvent* e) override
-		{
-			QSplashScreen::paintEvent(e);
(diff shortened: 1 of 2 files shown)


src/StelSplashScreen.hpp
```diff
@@ -27,16 +27,10 @@ class SplashScreen
 {
 	class SplashScreenWidget : public QSplashScreen
 	{
+        QFont splashFont;
```
gzotti, line 30: Please fix indentation. (use the code style file!)

gzotti:
Looks good otherwise, thank you. 

Please PM me for a little surprise!

src/StelSplashScreen.hpp
10110111: Always forget to switch editor settings :/
Fixed now.

alex-w (approved):
Thanks!

PR #2673, opened 2022-09-20
Make it easy to disable search for Qt6

10110111:
Otherwise, to make a Qt5 build on a system with Qt6, one has to edit `CMakeLists.txt`, which is not a friendly interface, and doesn't support simultaneous build directories with different options.


alex-w (approved):
This is good addition, thanks!

PR #2674, opened 2022-09-20
Work around too much stretching of close button icon on Qt 5.12.8

10110111:
In Qt 5.12 after #2660 the close button's icon appears to be overly stretched, like in the screenshot below. This happens in both scaled and unscaled UI.

This patch works around this by requesting a downsampled 16×16 pixmap from the original icon and rendering from it—only on Qt<5.13. This works fine for unscaled UI, but still looks ugly for scaled one:

close-bad-remains

I think this might be acceptable if we assume that Qt 5.12 is only going to be used on low-end systems which will not have high-DPI monitors anyway.

### DIFF
--- a/src/gui/StelCloseButton.cpp
+++ b/src/gui/StelCloseButton.cpp
@@ -27,6 +27,15 @@ StelCloseButton::StelCloseButton(QWidget* parent)
 {
 	// Qt 5.15 regression: QBushButton without borders does not respond to clicks
 	setStyleSheet("border: 1px solid rgba(0, 0, 0, 0); background: none;");
+
+	// In Qt 5.12.8 the icon stretches too much for some reason, so let's
+	// simply make it the expected size. This will break high-DPI scaled mode,
+	// but I think such high-end systems should use newer Qt anyway.
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
+	iconNormal = QIcon(iconNormal.pixmap(16,16));
+	iconHover  = QIcon(iconHover.pixmap(16,16));
+#endif
+
 	setIcon(iconNormal);
 }
 


alex-w (approved):

10110111:
> AppImage for linux is based on Qt 5.12.8 too

Is there a good reason for this?

alex-w:
> > AppImage for linux is based on Qt 5.12.8 too
> 
> Is there a good reason for this?

Ubuntu 20.04 have Qt 5.12.8 in standard repos

10110111:
> Ubuntu 20.04 have Qt 5.12.8 in standard repos

Isn't the point of AppImage to be independent from the repos and self-contained?

alex-w:
> > Ubuntu 20.04 have Qt 5.12.8 in standard repos
> 
> Isn't the point of AppImage to be independent from the repos and self-contained?

Yes, but it's depend on libc, so, if AppImage built on Ubuntu 22.04 then if cannot be run on Ubuntu 20.04 for example. Ubuntu 18.04 is good choice, but, cmake and Qt are too old and we cannot use it...  

P.S. Of course I can try get the AppImage on Ubuntu 18.04 to run the package on RHEL/CentOS

10110111:
On Ubuntu 20.04 you can use Qt 5.15 from [`ppa:beineri/opt-qt-5.15.4-focal`](https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.15.4-focal). I've been using this series of Qt5 PPAs for years.

alex-w:
> On Ubuntu 20.04 you can use Qt 5.15 from [`ppa:beineri/opt-qt-5.15.4-focal`](https://launchpad.net/~beineri/+archive/ubuntu/opt-qt-5.15.4-focal). I've been using this series of Qt5 PPAs for years.

Is there fixed a huge RAM usage for lconvert/lupdate tools?

PR #2676, opened 2022-09-20
Enjoy ShowMySky on HDPI GPU with limited graphics performance.

uwes-ufo:
### Description
A slightly higher frame rate would be desirable on an high-DPI GPU with limited graphics performance.

- To achieve this, the new configuration parameter "landscape/atmosphere_resolution_reduction" allows reducing the resolution.

- The new configuration switch "landscape/flag_atmosphere_dynamic_resolution" makes it possible to use the reduced resolution only during panning, zooming, dimming or in time-lapse display. In the case of real-time display, on the other hand, the full resolution is retained.

atmosphere_resolution_reduction=1 ~> full resolution (default)
atmosphere_resolution_reduction=2 ~> half resolution
atmosphere_resolution_reduction=4 ~> quarter resolution (recommended on Intel UHD 620)

flag_atmosphere_dynamic_resolution=false ~> static resolution (default)
flag_atmosphere_dynamic_resolution=true ~> dynamic resolution (recommended on Intel UHD 620)

### Type of change

### How Has This Been Tested?
* Graphics Card: Intel UHD 620

## Checklist:


src/core/modules/AtmosphereShowMySky.cpp
```diff
@@ -868,6 +902,7 @@ bool AtmosphereShowMySky::isLoading()
 
 bool AtmosphereShowMySky::isReadyToRender()
 {
+	prevWidth_=0;
```
10110111, line 905: Won't this lead to resize on each frame, even with `ppxatmo==1 && !flagDynamicResolution`?

src/core/modules/AtmosphereShowMySky.cpp
```diff
@@ -724,12 +730,40 @@ void AtmosphereShowMySky::computeColor(StelCore* core, const double JD, const Pl
...
+		dynResTimer=0;
 	}
 
 	auto sunPos  =  sun.getAltAzPosAuto(core);
 	if (std::isnan(sunPos.length()))
 		sunPos.set(0.,0.,-1.*AU);
 
+	if (flagDynamicResolution)
+	{
+		float f1=prj->getFov(), i1=fader.getInterstate();
```
10110111, line 742: I'd use more descriptive names than `<letter><digit>`.

src/core/modules/AtmosphereShowMySky.cpp
10110111: Besides, this function is supposed to be a query (and should have been made `const`, I missed it), so it doesn't seem like a good idea to alter state here.

uwes-ufo: > Won't this lead to resize on each frame, even with `ppxatmo==1 && !flagDynamicResolution`?

it's supposed to trigger a resize when switching from "Preetham" to "ShowMySky".

10110111:
> HDPI GPU

Did you mean high-DPI monitor?

Also, please show some screenshots of the Venus belt (its center, its intersection with the horizon) with the settings that you find best for your machine. I'm interested in preserving its quality as much as possible.

src/core/modules/AtmosphereShowMySky.cpp
```diff
@@ -724,12 +730,40 @@ void AtmosphereShowMySky::computeColor(StelCore* core, const double JD, const Pl
...
+
+		ppxatmo=dynResTimer<0?ppxmax:1;
+		if (prevPxa!=ppxatmo)
+			resizeRenderTarget(width, height);
+
+		bool verbose=qApp->property("verbose").toBool();
+		if (verbose)
+			qDebug() << "dynResTimer" << dynResTimer << "ppxatmo" << ppxatmo << "Fov" << df << "Fad" << di << "Pos" << dp << "Sun" << ds;
+
+		dynResTimer=dynResTimer<0?-2:-18;
```
10110111, line 759: These magic numbers need some explanation, like what each of them means, and why they are negative.

src/core/modules/AtmosphereShowMySky.cpp
10110111: A switch from Preetham to ShowMySky involves construction of a new instance of `AtmosphereShowMySky`, so this assignment should be unnecessary.

uwes-ufo:
> Also, please show some screenshots of the Venus belt (its center, its intersection with the horizon) with the settings that you find best for your machine. I'm interested in preserving its quality as much as possible.

The flag_dynamic_resolution=true causes the full resolution.

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: For example, panning reduces the resolution. when you stop panning, the "-2" causes an early redraw at full resolution.
On the other hand, only every 18th frame is drawn in real-time display (-18).

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: > A switch from Preetham to ShowMySky involves construction of a new instance of `AtmosphereShowMySky`, so this assignment should be unnecessary.

I just tested it again, without this line I get a wrong display with reduced static resolution.

10110111:
OK, I've tried it with dynamic resolution, and I'd like to suggest the following patch to make reduction of resolution less apparent:
```diff
diff --git a/src/core/modules/AtmosphereShowMySky.cpp b/src/core/modules/AtmosphereShowMySky.cpp
index cf5946e1fb..4617f30d6a 100644
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -880,6 +880,7 @@ void AtmosphereShowMySky::draw(StelCore* core)
     auto& gl=glfuncs();
     GL(gl.glActiveTexture(GL_TEXTURE0));
     GL(gl.glBindTexture(GL_TEXTURE_2D, renderer_->getLuminanceTexture()));
+    GL(gl.glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MAG_FILTER, GL_LINEAR));
     GL(luminanceToScreenProgram_->setUniformValue(shaderAttribLocations.luminanceTexture, 0));
 
     GL(gl.glActiveTexture(GL_TEXTURE1));
```

10110111:
Now, testing your branch I noticed a problem: when I open _Date and Time_ dialog and scroll very fast the hour spinbox from 1 to 5, I sometimes get resolution change flickering indefinitely, and sometimes the lower resolution gets statically displayed.

An easier to reproduce problem is when I speed up time twice (press <kbd>L</kbd><kbd>L</kbd>): I get frequent flicker between high and low resolution.

uwes-ufo:
> OK, I've tried it with dynamic resolution, and I'd like to suggest the following patch to make reduction of resolution less apparent:

Tested it immediately - perfect!

src/core/modules/AtmosphereShowMySky.cpp
10110111: Strange, it behaves identically on my machine. Could you describe your test scenario in detail?

uwes-ufo:
> An easier to reproduce problem is when I speed up time twice (press LL): I get frequent flicker between high and low resolution.

I can confirm.

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: - ppxatmo=4
- flag_dynamic_resolution=false

Start Stellarium with "ShowMySky" already activated ~> no problem
Switch to "Preetham" and back to "ShowMySky" ~> Problem ~> Sky graphic is twisted...

10110111: OK, I can reproduce. In my case I also get segfault because of wrong size of texture obtained in `getMeanPixelValue()`.

What's happening here is that initially, when `AtmosphereShowMySky` is constructed, `resizeRenderTarget()` is called with `ppxatmo=1`, value set from initializer list. Then, when `computeColor()` is called, `ppxatmo` assumes a different value, which is what should trigger a new resize.

Instead, you force the resize on status query, which is not semantically correct.

uwes-ufo: I just moved the reading of the config.ini to the initializer. The problem remains - So I'm back with another suspicious workaround :).

I noticed a second call to "resizeRenderTarget" when starting Stellarium. The first call uses default values for "width" and "height".
When switching from "Preetham" to "ShowMySky" the second call does not happen, the first call already has the actual "width" and "height".

10110111: OK I found the actual problem, not sure what side to blame. Anyway, `ShowMySky::AtmosphereRenderer` calls `resizeEvent()` using current viewport (`GL_VIEWPORT`) after all data have been loaded (see `stepDataLoading` and its callees).
One fix would be to do the resize after data loading has been done, but it seems kludgy to me. I need some time to check whether this resize at the end is actually needed in the renderer.

10110111:
Display grid appears to be broken in low-resolution mode, see screenshot below. Givent that we only change resolution of the render target, such geometric change like coarsening of the horizon in fish-eye mode shouldn't happen.

Screenshot - 210922 - 18:52:06

uwes-ufo:
just tried this...
`gridMaxX*=2; gridMaxY*=2;`

stellarium-20220921-182055894

uwes-ufo:
playing around with extremes...
`int tryit=qRound(sqrt(ppxatmo)); gridMaxX*=tryit; gridMaxY*=tryit;`

- `flag_dynamic_resolution=false`
- `ppxatmo=16`

stellarium-20220921-19035064

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: BTW

> `sunPos.set(0.,0.,-1.*AU);`

> `// If we have no moon, just put it into nadir at 1 AU distance`
> `Vec3d moonPos{0.,0.,-1.*AU};`
`

Aren't we already scaled to multiples of AU?

10110111: Hmm, indeed there's an inconsistency in units here. It just doesn't make much difference since these are cases where the Sun or the Moon aren't visible.

src/core/modules/AtmosphereShowMySky.cpp
```diff
@@ -517,6 +501,13 @@ void AtmosphereShowMySky::regenerateGrid()
 //	ppxmax = ppxatmo = conf->value("landscape/ppxatmo", 1).toInt();
 	gridMaxY = conf->value("landscape/atmosphereybin", 44).toInt();
 	gridMaxX = std::floor(0.5+gridMaxY*(0.5*std::sqrt(3.0))*width/height);
+	const auto rppx=qRound(sqrt(ppxatmo));
```
10110111, line 504: Although this indeed does improve the looks as seen in your screenshot, I'm still not convinced it fixes the problem at its origin. Why did the grid even change before this patch?

uwes-ufo: It seems that the size of the grid stays the same, but we have a lot fewer 
- 1/ppxatmo^2 (without the patch)
- 1/ppxatmo (with the patch)

elements due to the smaller render target.

The final display is then simply scaled up with ppxatmo.

Try this representation of triangles for fun...


```diff
diff --git a/src/core/modules/AtmosphereShowMySky.cpp b/src/core/modules/AtmosphereShowMySky.cpp
index 4f74c1e..508bcce 100644
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -446,7 +446,7 @@
 				std::size_t shift=0;
 				for (int y=0;y<gridMaxY;++y)
 				{
-					GL(gl.glDrawElements(GL_TRIANGLE_STRIP, (gridMaxX+1)*2, GL_UNSIGNED_SHORT,
+					GL(gl.glDrawElements(GL_TRIANGLES, (gridMaxX+1)*2, GL_UNSIGNED_SHORT,
 										 reinterpret_cast<void*>(shift)));
 					shift += (gridMaxX+1)*2*2;
 				}
```

10110111: Aha, so that's the problem. I suspected it, but misled myself during thinking :) .

The grid is generated assuming `width` and `height` representing the full-resolution viewport. What your change should be is  as follows:
```diff
@@ -511,7 +511,8 @@ AtmosphereShowMySky::~AtmosphereShowMySky()
 
 void AtmosphereShowMySky::regenerateGrid()
 {
-    const float width=viewport[2], height=viewport[3];
+    const float width = viewport[2]/ppxatmo, height = viewport[3]/ppxatmo;
     QSettings* conf = StelApp::getInstance().getSettings();
 //  flagDynamicResolution = conf->value("landscape/flag_dynamic_resolution", false).toBool();
 //  ppxmax = ppxatmo = conf->value("landscape/ppxatmo", 1).toInt();
```
Just be sure to regeneration the grid on each switch between full and reduced resolutions.

src/core/modules/AtmosphereShowMySky.cpp
```diff
@@ -772,13 +838,12 @@ void AtmosphereShowMySky::computeColor(StelCore* core, const double JD, const Pl
...
 	}
 
-
 	// FIXME: ignoring the "additional luminance" like star background etc.; see AtmospherePreetham for all potentially needed terms
 	const auto numViewRayGridPoints=(1+gridMaxX)*(1+gridMaxY);
 	for (int i=0; i<numViewRayGridPoints; ++i)
 	{
 		Vec3d point(1, 0, 0);
-		prj->unProject(posGrid[i][0],posGrid[i][1],point);
+		prj->unProject(posGrid[i][0]*(double)ppxatmo,posGrid[i][1]*(double)ppxatmo,point);
```
10110111, line 846: This is imprecise, because `posGrid` points were generated assuming width&height divided by `ppxatmo` **and rounded** down. Multiplying back we won't get the full width&height.
Instead, we need to multiply these two coordinates by `double(fullWidth)/reducedWidth` and `double(fullHeight)/reducedHeight` respectively.

src/core/modules/AtmosphereShowMySky.cpp
10110111: Funny, this multiplication by `AU` dates back to 2009: 9d6b8ada86505bebb95350b134596a1d8db4cabc.

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: You are absolutely right!

src/core/modules/AtmosphereShowMySky.cpp
uwes-ufo: I just changed it - looks good!

10110111:
What do ppx and pxa mean? I think these should be renamed to something immediately understandable — both in names of variables and in the config parameters.

Also, `fov1`, `fad1` and company seem to mean "current Fov" etc., while the number seems to imply that here are `fov0` or `fov2` counterparts. I'd suggest to rename them too into something like `currFov`.

10110111:
I suggest replacing `atmosphere_resolution` with `atmosphere_resolution_reduction`, since the higher the value, the lower the resolution.

Also, the commented out code like in `regenerateGrid()` should be removed.


Finally, _"suspicious workaround"_ will be more usefully commented as _"force resize on first render because the renderer may change framebuffer size after data loading"_.

Other than that, LGTM.


@alex-w, @gzotti — Your comments?

uwes-ufo:
> Also, the commented out code like in regenerateGrid() should be removed.

In the meantime I moved reading the configuration back to `regenerateGrid()`. So I could easily test some resolution values. I've noticed that integer powers of two (2, 4, 8 even 16) work as expected, but values in between result in more or less pronounced darkening of the sky.

10110111:
> values in between result in more or less pronounced darkening of the sky

Is it on Windows+Intel? I had a problem with Intel on Linux, where mipmaps for non-power-of-two textures are generated in poor quality. This could explain it.

I have a fix in mind, but it's a kludge. Initially I thought it was for Mesa only, but apparently Intel@Windows also has this problem.

uwes-ufo:
> Is it on Windows+Intel?

Yes. Could it be related to `getMeanPixelValue()`?

10110111:
> Could it be related to `getMeanPixelValue()`?

Exactly. If you don't reproduce it on e.g. NVIDIA, then it's OK for now.

gzotti:
I don't see a difference, but maybe moving around during solar eclipse is faster now. So if it's OK for @10110111, I won't object.

However. the ususal suspects: documentation!

* please document what the various variables do (I agree, names like pxa are incomprehensible!)
* Add documentation to the User Guide (guide directory) where it is best applied. Presumably chapter 11. Don't bother too much about formatting if you are not familiar with LaTeX, concentrate on content, I'll do the rest. 
 
I will also rename the flag landscape/flag_dynamic_resolution. This implies landscape rendering is influenced. Must become landscape/flag_atmosphere_dynamic_resolution.

Finally, there should be GUI elements in the ShowMySky settings panel. However, we can add them as well right after merge.

uwes-ufo:
Typo or intentional?
```diff
diff --git a/src/core/modules/AtmosphereShowMySky.cpp b/src/core/modules/AtmosphereShowMySky.cpp
index 22f2cc9..ae496d4 100644
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -224,7 +224,7 @@
 	// Shader program that converts XYZW texture to sRGB image
 	{
 		static constexpr char vShaderSrc[]=R"(
-#version 130
+#version 330
 in vec4 vertex;
 out vec2 texCoord;
 void main()
```

10110111:
> Typo or intentional?

Neither. Just a relic of the very early versions. Will update soon.

alex-w (approved):

src/core/modules/AtmosphereShowMySky.hpp
```diff
@@ -66,7 +66,8 @@ class AtmosphereShowMySky : public Atmosphere, public QObject
 #ifdef ENABLE_SHOWMYSKY
 	QLibrary showMySkyLib;
 	Vec4i viewport;
-	int gridMaxY,gridMaxX;
+	int gridMaxY,gridMaxX,atmoRes;
```
gzotti, line 69: I would still like to see some documentation for these variables as long as your memory is fresh! It is much harder to reverse-doc that next year.
@10110111 please also add some class doc and more background info after merge. Consider a new developer wants to start with this in 4 years.

gzotti (approved):
If @10110111 approves, I won't object. But both, please provide more source documentation for the uninitiated.

10110111:
@gzotti What exactly do you want me to document? The ShowMySky API is documented [here](https://10110111.github.io/CalcMySky/showmysky-api.html). The code here just follows it, using Preetham atmosphere class as the starting point.

gzotti:
Why not write at least exactly this, with a link to your project, into AtmosphereShowMySky.hpp?

src/core/modules/AtmosphereShowMySky.hpp
uwes-ufo: Something like that?
```diff
diff --git a/src/core/modules/AtmosphereShowMySky.hpp b/src/core/modules/AtmosphereShowMySky.hpp
index 6eb68ce..fa396fc 100644
--- a/src/core/modules/AtmosphereShowMySky.hpp
+++ b/src/core/modules/AtmosphereShowMySky.hpp
@@ -66,7 +66,17 @@
 #ifdef ENABLE_SHOWMYSKY
 	QLibrary showMySkyLib;
 	Vec4i viewport;
-	int gridMaxY,gridMaxX,atmoRes;
+	int gridMaxY,gridMaxX;
+
+	// atmosphere_resolution_reduction
+	// 1 ~> full resolution (default)
+	// 2 ~> half resolution
+	// 4 ~> quarter resolution
+	int atmoRes;
+
+	// flag_atmosphere_dynamic_resolution
+	// false ~> static resolution (default)
+	// dynamic resolution
 	bool flagDynamicResolution;
 
 	QVector<Vec2f> posGrid;
```

gzotti: Yes, good start! If you mark the comment with exclamation mark like 

//! dynamic resolution reduces resolution of the skylight texture while view is moved. 
//! This increases frame rate esp. on slow systems.

or similar, it will even be included in the Doxygen online docs and describes the motivation for this feature. (Note that the description of the PR and this whole discussion is usually not available while studying the source code at some later date.)

uwes-ufo: so let's try again...
I'm afraid the poor little header might get overloaded with all that stuff. :)
```diff
diff --git a/src/core/modules/AtmosphereShowMySky.hpp b/src/core/modules/AtmosphereShowMySky.hpp
index 6eb68ce..3483efe 100644
--- a/src/core/modules/AtmosphereShowMySky.hpp
+++ b/src/core/modules/AtmosphereShowMySky.hpp
@@ -66,7 +66,33 @@
 #ifdef ENABLE_SHOWMYSKY
 	QLibrary showMySkyLib;
 	Vec4i viewport;
-	int gridMaxY,gridMaxX,atmoRes;
+	int gridMaxY,gridMaxX;
+
+	//! In order to achieve higher frame rates on slow systems, the configuration parameter:
+	//!	config.ini
+	//!	[landscape]
+	//!	atmosphere_resolution_reduction
+	//! allows reducing the resolution of the skylight texture.
+	//! Preferred values are:
+	//!	1 ~> full resolution (default)
+	//!	2 ~> half resolution
+	//!	4 ~> quarter resolution
+	int atmoRes;
+
+	//! The configuration switch:
+	//!	config.ini
+	//!	[landscape]
+	//!	flag_atmosphere_dynamic_resolution
+	//! allows to use the reduced resolution only while moving the view,
+	//! when panning, zooming, dimming or in time-lapse mode.
+	//! With the real-time display, on the other hand, the full resolution is retained.
+	//! Possible values are:
+	//!	false ~> static resolution (default)
+	//!	true ~> dynamic resolution
+	//! A notice:
+	//! In dynamic resolution mode, a motion detector selects either full or reduced resolution.
+	//! The change in resolution could be particularly visible in close proximity to the Sun.
+	//! Especially at full resolution, frames will be skipped depending on the speed of movement.
 	bool flagDynamicResolution;
 
 	QVector<Vec2f> posGrid;
```

gzotti: Good! The header file can sustain such augmentations without great pain.

10110111: In fact, Doxygen also supports an easier syntax, using `/*! some docs, even multiple lines */`. So no need to do this `//!` on each line.

uwes-ufo: With this syntax everything can be hidden again. A feast for the eyes for code purists. :)

Screenshot 2022-09-23 145511

gzotti: @10110111 is technically right, but we have our long-used style. @alex-w ?

alex-w: We used both styles

10110111:
This PR generates a new warning:
```
In file included from /home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.cpp:22:
/home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.hpp: In constructor ‘AtmosphereShowMySky::AtmosphereShowMySky()’:
/home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.hpp:104:40: warning: ‘AtmosphereShowMySky::luminanceToScreenProgram_’ will be initialized after [-Wreorder]
  104 |  std::unique_ptr<QOpenGLShaderProgram> luminanceToScreenProgram_;
      |                                        ^~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.cpp:22:
/home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.hpp:81:6: warning:   ‘int AtmosphereShowMySky::reducedResolution’ [-Wreorder]
   81 |  int reducedResolution;
      |      ^~~~~~~~~~~~~~~~~
/home/ruslan/src/stellarium-git/src/core/modules/AtmosphereShowMySky.cpp:406:1: warning:   when initialized here [-Wreorder]
  406 | AtmosphereShowMySky::AtmosphereShowMySky()
      | ^~~~~~~~~~~~~~~~~~~
```

uwes-ufo:
This way?
```diff
diff --git a/src/core/modules/AtmosphereShowMySky.cpp b/src/core/modules/AtmosphereShowMySky.cpp
index 89f6c8f..41b3538 100644
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -411,12 +411,12 @@
 	, viewport(0,0,0,0)
 	, gridMaxY(44)
 	, gridMaxX(44)
+	, reducedResolution(1)
+	, flagDynamicResolution(false)
 	, posGridBuffer(QOpenGLBuffer::VertexBuffer)
 	, indexBuffer(QOpenGLBuffer::IndexBuffer)
 	, viewRayGridBuffer(QOpenGLBuffer::VertexBuffer)
 	, luminanceToScreenProgram_(new QOpenGLShaderProgram())
-	, reducedResolution(1)
-	, flagDynamicResolution(false)
 {
 	indexBuffer.setUsagePattern(QOpenGLBuffer::StaticDraw);
 	indexBuffer.create();

```

10110111:
Yes, no warnings now.

So this PR LGTM now.

uwes-ufo:
What about this CodeFactor issue?

gzotti:
> What about this CodeFactor issue?

We agree with CodeFactor that some methods are complex but usually ignore this warning. Some things can hardly be improved, others could be broken up into shorter methods, if there is a real benefit (e.g. move copy-pasted constructs to new methods and call them). I like to comment my source code so that I understand while reading 3 years later what I have done or thought when I wrote it. Even better, I like to see comments in code not written by myself about what others have thought or intended when they wrote it.

gzotti:
Thank you!

uwes-ufo:
Thank you!

uwes-ufo:
> We agree with CodeFactor that some methods are complex but usually ignore this warning.

I found CodeFactor's error message quite helpful, and pretty early on I moved my code changes that I had initially made in the existing `computeColor()` routine to a new separate `dynamicResolution()` routine, which indeed improved readability.

However, CodeFactor keeps repeating the same error message and then aborts the analysis.

It seems to me that CodeFactor is trying to work through the subcommits step by step instead of analyzing the final subcommit merge. As a result, CodeFactor cannot see the corrections that have already been made.

gzotti:
Yes, it is helpful in general, and where possible we should try to minimize issues. Still sometimes it just doesn't like our long methods.
Interesting observation of yours, I have not thought about that.

alex-w:

PR #2677, opened 2022-09-21
Add capability to create KML map of solar eclipses in AstroCalc

worachate001:
### Description

A button to save KML (formerly Keyhole Markup Language) file is added. Description of lines on the map, including a figure, are also inserted in Stellarium User Guide.

Click "save KML of eclipse" button will create KML file to draw region of visibility on world map (open in mapping apps like Google Earth and Marble). Different colors are used for path of central eclipse (red = total, blue = annular, purple = hybrid).

I have been creating eclipse maps for local astronomical society for years and I would like to bring it into Stellarium. I admit that the saved KML is imperfect, there are gaps due to limitation of algorithms but I think the feature is good enough to be included.

External applications like Marble or Google Earth is required to open the KML (there must be a way to draw world map inside Stellarium, like in the location window, maybe in the not-so-far future?). With the KML, we have good visual information -- detailed map in respective applications of where on Earth we can see all the eclipses in the past, present, and future.

I note that there is built-in eclipse map in Marble. The map in Marble looks pretty good but comparisons show that there are errors in northern/southern limits of umbra and penumbra. Stellarium calculations are more accurate than Marble in this area.

solareclipse-map

### Type of change

### How Has This Been Tested?
**Test Configuration**:
* Operating system: Manjaro 22.0.0
* Graphics Card: -

## Checklist:


alex-w:
Great addition, but the UI in AstroCalc/Eclipses is too wide now. I think we can shrink it - for example rename "Calculate solar eclipses" into "Calculate eclipses"; "Export circumstances of eclipse..." -> "Export circumstances..."; etc. Of course buttons should be ranamed in all eclipses tabs.

gzotti:
Simply: WOW! Awesome job! 
In a future step, can you think of connecting to GE in a secondary window to show the results directly? (Maybe all the eclipse/occultation computations should go into its own plugin?) A GE-based plugin to extend/replace the location panel would be great. (It cannot fully replace the location panel, because of off-line use!)

alex-w (approved):
Great job!

gzotti:
Small issue: During storing of KML (which does take a few seconds!) the mouse cursor should be set to busy, and reset at end of method. Something like

QGuiApplication::setOverrideCursor(QCursor(Qt::WaitCursor))). 
writeFile(eclipse.kml)
QGuiApplication::restoreOverrideCursor();

I was too fast loading the KML to GE and triggered an error there. A busy cursor would have indicated this.

gzotti:
> Will be enough “Export KML” or “Save KML” button with tooltip?

Yes, short labels are preferrable.

worachate001:
@gzotti @alex-w Thanks for suggestions. Changed from save to export KML and I hope I did it right for the cursor issue.

alex-w:
It’s fine for me, thanks!

@gzotti?

gzotti:
The lines in the KML have a few missing segments, e.g. where the "First contact" marker is. but also the northern/southern limit lines seem to miss a last segment. Is this just a 1-off error (then please fix now) or something which is more difficult to fix? In the latter case, we can include the current solution in 1.0 but I hope for finetuning in 1.1. (Some users will sure complain about this :-) BTW the fist contact and max. markers are a very nice addition. Maybe add UT timestamps in addition? Also to the eclipse shadow ovals! I really enjoy this new feature!

worachate001:
> The lines in the KML have a few missing segments, e.g. where the "First contact" marker is. but also the northern/southern limit lines seem to miss a last segment. Is this just a 1-off error (then please fix now) or something which is more difficult to fix?

It will take some times. It looks easy by just draw a line between that marker and beginning or the end of that line. But I have to think that what is the correct and best way to do it. I'm not sure that can be completed in a few days. However, I agree that it should be fixed in the near future.

gzotti:
OK. Then we take this for 1.0 and will be happy to see what else you can do! This is a great addition, really!

gzotti:
Just to be sure, is the figure 4.42 your own or is it public domain? Else we need license clearance or even cannot use it.

worachate001:
> Just to be sure, is the figure 4.42 your own or is it public domain? Else we need license clearance or even cannot use it.

It's my own. Created with Python/Matplotlib/Basemap and Inkscape.

alex-w:
> I've just opened _Astronomical calculations_ window on the Eclipses tab, and what made me wonder is that the buttons are active but don't do anything. Then, after I calculated some eclipses to fill in the tables, the buttons actually started doing something.
> 
> For better UX the buttons should be disabled if they won't do anything sensible.

Good idea, I'll add it in next few days (I hope before release 1.0)

alex-w:

PR #2691, opened 2022-09-25
Make StelGui buttons high-DPI ready

10110111:
### Description

This PR makes the side panel buttons aware of high-DPI scaling. This is a continuation of #2658 and #2660.

Some minor visual changes at 100% scale do exist, but I think they are not critical, and even make the results a bit nicer.

This works both on Qt6 (tested on 6.3.1). and Qt5 (tested on 5.12.8 and 5.15.2).

This change affects a lot of plugins, so I'd like it to be tested before I merge it to `master`.

All images for the side, top, and bottom panels are now saved in 500% size. In particular, it means that export from the SVG sources (`icons.svg`) should now be done at 480 DPI instead of the original 96 DPI. @martinber might have some comments on this (in case I've missed something).

Not all images are rescaled yet. In particular, icons in the buttons in the usual `QWidget`-based GUI are still at 100% resolution.

#### Left, bottom and top-right panels @ 100%

Old:
a1

New:
a2

#### Left, bottom and top-right panels @ 150%

Old:
b1

New:
b2

#### Bottom panel button glow on hover, at 100% scale

Old:
c1

New:
c2

### How Has This Been Tested?

**Test Configuration**:
* Linux From Scratch, Qt 6.3.1, Qt 5.15.2 on NVIDIA GeForce GTX 750Ti
* Windows 11 21H2 on Intel HD Graphics 520

## Checklist:


martinber:
For me it is goof, it will have some artifacts because of the scaling but it wont be that bad, at least this should be the easiest and more robust way to do it.

Right now I'm travelling but I can test it by the end of this week. Especially I have no idea about how the Qt code should be done (I see you multiply by the devicePixelRatio)

alex-w:
@10110111 did you check the behaviour of new buttons with scaling factor 1.1 or 1.3 (110% or 130%)?

10110111:
> @10110111 did you check the behaviour of new buttons with scaling factor 1.1 or 1.3 (110% or 130%)?

Yes. They do work, but the background is not always continuous:

Screenshot - 280922 - 12:16:15

I'm not sure if this can be fixed at `QGraphicsItem` level.

10110111:
> I'm not sure if this can be fixed at QGraphicsItem level.

Indeed, it can't. The problem here is that `BottomStelBar` buttons are positioned in units of scaled pixels. This means that, if we want to align the rendering to physical pixel grid to avoid blurring, we need to round these positions in the physical pixel scale. But this will necessarily lead to either overlaps, or gaps, or both.

The fix would require reacting to changes in current device pixel ratio (which may happen e.g. after moving Stellarium window from one screen to another, like from laptop monitor to a projector). The reaction should be a recreation of the bottom button panel, positioning the buttons according to the device pixels instead of the scaled ones.

So, this is not the concern of this PR.

10110111:
OK, I think this PR is stable now.

If you do merge it, please don't squash the commits: I'd like them to remain separated, so it's easier to checkout each of them if anything goes wrong.

alex-w (approved):
It’s fine for me, thanks!

gzotti (approved):
Looks good, thanks!

alex-w:

PR #2694, opened 2022-09-27
Fix/observing lists gz

### LINKED ISSUE #2680: 0.22.2-992d4a7-win64: opening Observing Lists deselects object
Is the same as [https://github.com/Stellarium/stellarium/issues/1941](url) which is closed but issue is still there. Using the Weekly Snapshot stellarium-1.22.2-992d4a7-win64

I select an object (say M81) but as soon as I open the Observing List, the object gets deselected.
When I select M81 again and then use Edit List, M81 is deselected again.

Expected: M81 stays selected when Opening and/or Editing the Observing List.

gzotti:
### Description

This should originally fix the lost selection issue. However I found numerous other small and larger issues that IMHO needs these further changes:

- dialog names (so sizes/positions are remembered) 
- added 4 Properties for optional entries to store/retrieve
- no global landscape for a list
- location stored completely as 1-line encoding, retrieved still from the other formats. Check if decoded location is valid. 
- bookmarks import with restored coordinates for their dates
- bookmarks really only processed once, not on every run
- restored usual close button
- reformat code to follow our style guide
- simplified code

Fixes #2680

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/gui/ObsListCreateEditDialog.cpp
+++ b/src/gui/ObsListCreateEditDialog.cpp
@@ -589,8 +589,17 @@ void ObsListCreateEditDialog::obsListSaveButtonPressed()
 	QString listName = ui->nameOfListLineEdit->text().trimmed();
 	qDebug() << "ObsListCreateEditDialog::obsListSaveButtonPressed(): list:" << listName;
 
-	bool isListAlreadyExists = !this->listNames_.isEmpty() && this->listNames_.contains(listName) &&
-			(isCreationMode || (listName.compare(currentListName) != 0 && !isCreationMode));
+//	bool isListAlreadyExists = !this->listNames_.isEmpty() && this->listNames_.contains(listName) &&
+//			(isCreationMode || (listName.compare(currentListName) != 0 && !isCreationMode));
+	// JG 2022-09-28 19:16.
+	bool isListAlreadyExists = false;
+	    if(!this->listNames_.isEmpty() && this->listNames_.contains(listName) && isCreationMode){
+		isListAlreadyExists = true;
+	 }
+// Will be simplified
+//	bool isListAlreadyExists = (!this->listNames_.isEmpty() && this->listNames_.contains(listName) && isCreationMode);
+
+
 
 	if (isListAlreadyExists) {
 		qWarning() << "[ObservingList Creation/Edition] Error: a list with the name " << ui->nameOfListLineEdit->text()
--- a/src/gui/ObsListCreateEditDialog.cpp
+++ b/src/gui/ObsListCreateEditDialog.cpp
@@ -616,7 +616,7 @@ void ObsListCreateEditDialog::obsListSaveButtonPressed()
 		qDebug() << "ObsListCreateEditDialog::obsListSaveButtonPressed(): isSaveAs:" << isSaveAs;
 		saveObservedObjectsInJsonFile();
 		this->close();
-		// emit exitButtonClicked();
+		emit exitButtonClicked();
 	}
 }
 
@@ -628,6 +628,8 @@ void ObsListCreateEditDialog::obsListExitButtonPressed()
 	ui->obsListErrorMessage->setHidden(true);
 	ui->obsListErrorMessage->clear();
 	this->close();
+	emit this->exitButtonClicked();
+
 }
 
 /*
@@ -636,7 +638,7 @@ void ObsListCreateEditDialog::obsListExitButtonPressed()
 void ObsListCreateEditDialog::close()
 {
 	this->setVisible(false);
-	//emit this->exitButtonClicked();
+	//emit this->exitButtonClicked(); // Disabled says JC.
 }
 
 /*
--- a/src/gui/ObsListCreateEditDialog.cpp
+++ b/src/gui/ObsListCreateEditDialog.cpp
@@ -17,16 +17,11 @@
  * 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
  */
 
-#include <QDateTime>
 #include <QDir>
 #include <QFileDialog>
 #include <QUuid>
 #include <StelTranslator.hpp>
-#include <utility>
-#include <Planet.hpp>
-#include <StelOBJ.hpp>
 #include <QMessageBox>
-#include <QRegularExpression>
 
 #include "NebulaMgr.hpp"
 #include "StelCore.hpp"
@@ -44,14 +39,14 @@
 
 ObsListCreateEditDialog *ObsListCreateEditDialog::m_instance = nullptr;
 
-ObsListCreateEditDialog::ObsListCreateEditDialog(std::string listUuid) :
+ObsListCreateEditDialog::ObsListCreateEditDialog(const QString &listUuid) :
 	StelDialog("ObservingListCreateEdit"),
 	ui(new Ui_obsListCreateEditDialogForm()),
 	isCreationMode(false),
 	isSaveAs(false),
 	obsListListModel(new QStandardItemModel(0, ColumnCount)),
 	core(StelApp::getInstance().getCore()),
-	listOlud_(std::move(listUuid)),
+	listOlud_(listUuid),
 	sorting("")
 {
 	objectMgr = GETSTELMODULE(StelObjectMgr);
@@ -73,10 +68,10 @@ ObsListCreateEditDialog::~ObsListCreateEditDialog() {
 /**
  * Get singleton instance of class
  */
-ObsListCreateEditDialog * ObsListCreateEditDialog::Instance(std::string listUuid)
+ObsListCreateEditDialog * ObsListCreateEditDialog::Instance(const QString &listUuid)
 {
 	if (!m_instance)
-		m_instance = new ObsListCreateEditDialog(std::move(listUuid));
+		m_instance = new ObsListCreateEditDialog(listUuid);
 	return m_instance;
 }
 
@@ -141,7 +136,7 @@ void ObsListCreateEditDialog::createDialogContent()
 	ui->nameOfListLineEdit->setText(QString(ui->nameOfListLineEdit->text()).trimmed());
 	ui->obsListSaveButton->setEnabled(!(ui->nameOfListLineEdit->text().isEmpty()));
 
-	if (listOlud_.empty())
+	if (listOlud_.isEmpty())
 	{
 		// case of creation mode
 		isCreationMode = true;
@@ -406,7 +401,7 @@ void ObsListCreateEditDialog::saveObservedObjectsInJsonFile()
 	try {
 		QVariantMap mapFromJsonFile; // Maps the whole JSON file
 		QVariantMap allListsMap;     // Extracts just the observingLists, not the global metadata of the JSON file
-		const QString oludQs = QString::fromStdString(this->listOlud_);
+		const QString oludQs = this->listOlud_;
 		const QVariantMap currentList = allListsMap.value(oludQs).toMap();
 		if (jsonFile.size() > 0)
 		{
@@ -460,7 +455,7 @@ void ObsListCreateEditDialog::saveObservedObjectsInJsonFile()
 		observingListDataList.insert(QString(KEY_SORTING), (sorting.isEmpty() ? existingSorting : sorting));
 
 		// Olud
-		QString oblListOlud = (isCreationMode || isSaveAs) ? QUuid::createUuid().toString() : QString::fromStdString(listOlud_);
+		QString oblListOlud = (isCreationMode || isSaveAs) ? QUuid::createUuid().toString() : listOlud_;
 
 		// Default list
 		if (ui->obsListDefaultListCheckBox->isChecked())
@@ -471,7 +466,7 @@ void ObsListCreateEditDialog::saveObservedObjectsInJsonFile()
 			if (defaultListUuid.isEmpty())
 				mapFromJsonFile.insert(KEY_DEFAULT_LIST_OLUD, "");
 			else
-				if (QString::compare(defaultListUuid, QString::fromStdString(listOlud_), Qt::CaseSensitive) == 0)
+				if (QString::compare(defaultListUuid, listOlud_, Qt::CaseSensitive) == 0)
 					mapFromJsonFile.insert(KEY_DEFAULT_LIST_OLUD, "");
 		}
 
@@ -538,7 +533,7 @@ void ObsListCreateEditDialog::obsListImportListButtonPresssed()
 				QVariantMap observingListMap = map.value(QString(KEY_OBSERVING_LISTS)).toMap();
 				if (!observingListMap.isEmpty() && observingListMap.size() == 1)
 					//listOlud_ = observingListMap.keys().at(0).toStdString();
-					listOlud_ = observingListMap.firstKey().toStdString();
+					listOlud_ = observingListMap.firstKey();
 				else
 				{
 					qWarning() << "[ObservingList Creation/Edition import] there is no list or more than one list:" << observingListMap.size();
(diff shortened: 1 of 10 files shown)


src/gui/ObsListCreateEditDialog.hpp
```diff
@@ -31,117 +31,118 @@
 class Ui_obsListCreateEditDialogForm;
 
 class ObsListCreateEditDialog : public StelDialog {
-Q_OBJECT
+	Q_OBJECT
 
 public:
-    static auto Instance(std::string listUuid) -> ObsListCreateEditDialog *;
+	static ObsListCreateEditDialog * Instance(std::string listUuid);
```
Jocelyn1109, line 37: If you put that you have this warning:
"Clang-Tidy: Use a trailing return type for this function". This why I put this code:
static auto Instance(std::string listUuid) -> ObsListCreateEditDialog *;

In my code I tried to correct a lot of Clang-Tidy warning.

src/gui/ObservingListCommon.hpp
```diff
@@ -21,42 +21,44 @@
...
+	ColumnConstellation, //! Constellation in which the object is located
+	ColumnDate,          //! Date
+	ColumnLocation,      //! Location where the object is observed
+	ColumnLandscapeID,   //! LandscapeID linked to object observation
+	ColumnCount          //! Total number of columns
 };
 
 static constexpr char const *JSON_FILE_NAME = "observingList.json";
-static constexpr char const *FILE_VERSION = "2.0";
+static constexpr char const *FILE_VERSION = "2.1";
```
Jocelyn1109, line 61: I'm not sure it's necessary to increment the version. This is the second version of the observation lists.

src/gui/ObsListCreateEditDialog.hpp
gzotti: QtCreator also has clang-tidy and clazy warning systems. This helped a lot getting rid of many issues, suggesting trailing return types being none of them. Probably it can be fine-tuned somehow on your system. I really dislike the new c++ syntax with `auto function(args) -> type` which does IMHO nothing to increase readability. Esp. starting to add this after 20 years of project history.

src/gui/ObservingListCommon.hpp
gzotti: OK, reverted.

src/gui/ObservingListCommon.hpp
```diff
@@ -121,39 +129,35 @@ class ObservingListUtil {
...
-            }
-        } else {
-            objectMagnitudeStr = QString::number(objectMagnitude, 'f', 2);
-        }
-
-        return objectMagnitudeStr;
+    //! Get the magnitude from selected object (or a dash if unavailable)
+    QString getMagnitude(const QList<StelObjectP> &selectedObject, StelCore *core) {
+
+	if (!core)
```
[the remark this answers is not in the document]
gzotti: We don't use Java-style trailing braces but a full line with a new brace. For 1-line clauses I however prefer brace-free code.

src/gui/ObsListCreateEditDialog.hpp
Jocelyn1109: @gzotti 
"I really dislike the new c++ syntax with auto function(args) -> type which does IMHO nothing to increase readability" -> I don't like too ;)

alex-w:
DSO wihtout proper names and with designations are added as "unnamed objects" now 
O_o

Jocelyn1109:
@alex-w  where ???

alex-w:
> @alex-w where ???

stellarium-001

alex-w:
The issue for wrong DSO names and wrong table data in ObsListDialog are not exists in RC2, but the crash is here.

Jocelyn1109:
@alex-w  for the crash I did a fix on my branch but I don't know if @gzotti aaded it on his branch.
It happens if two people work at the same time on the same functionality. I admit to being a little lost in your process.

src/gui/ObsListCreateEditDialog.hpp
gzotti: So we will not use this syntax for now.

gzotti:
The problem was the urgency of these fixes. I will invest most of today for further action. So far I solved the original observation (which was a stupid beast with a stored "const QList<StelObjectP>" being non-constant!) and that the save list button did not want to become enabled because (during my first round of style adaptations) I messed the argument of the connect-ed function. I would prefer feature freezes 2-3 weeks before release, and not a ton of last-week improvements.

Jocelyn1109:
Ok @gzotti, should I continue to work on the observation lists?

gzotti:
> Ok @gzotti, should I continue to work on the observation lists?

Given my massive reformatting, I think it's better if we don't receive a PR in your original format. (Either you must rebase with lots of editing, or make a new branch based on this branch here. Then you would have to compare your old version with what I have reformatted, and give reviewer comments in the PR here.

I solved the lost selection issue independently, so did not took your solution. 

But as said, please if you find my errors in the bad table formatting, add PR comments here. I will look into your emails from yesterday about your crash solution now.

gzotti:
@Jocelyn1109 just to calm my mind: what does OLUD mean?

alex-w:
> OLUD

Observing List Unique Identifier in French language

gzotti:
> Observing List Unique Identifier in French language

Is that Unique Designation then? (Would that not be Designation Unique in French word order?) Why not use generally preferred programming comment language (english) and OLUI?

OK, we keep OLUD but I start writing a comment section with some aspects of the architecture of ObsLists.

Jocelyn1109:
@gzotti @alex-w 
"Would that not be Designation Unique in French word order?" -> yes, that's right ;)

gzotti:
@Jocelyn1109 observation: When adding a new object, the table in the edit panel gets updated, but not that in the main panel. Did that work originally? (Maybe some signal/slot connection is missing/broken?)

I still don't know why the column1 in the list view is not hidden and the headings missing. The problem reported by Alexander is also not solved yet.  Else I made good progress so far.

Jocelyn1109:
@gzotti when you add and object in edit mode and then you save the list, normally the list in main dialog must be updated.
During my tests I didn't see any problem but I will check this evening on my branch.

"I still don't know why the column1 in the list view is not hidden and the headings missing." -> I'll take a look and if I find the problem I'll email you the solution.

gzotti:
@alex-w the type issue should be solved. This was caused by crossed variable names and lack of item docs.

Table layout mess: no solution yet.

Jocelyn1109:
@gzotti I sent you two emails to correct the problem of the edit mode and that of the column names.

gzotti:
About DSO names: Currently I see the name also in the designation column. Should we not store and display a catalog number instead? This removes this duplicity and makes for clear identification.

gzotti:
Just tracing down the crash. Pretty weird, it's about a signal emitted when I restore the previously selected object. That is then processed by the SolarSystem, and something seems to go wrong here.

gzotti:
@Jocelyn1109 , @alex-w I am stuck. It is about the original issue: selection lost during manipulation of ObsLists.

To keep the selected object, we have to store it somehow and retrieve it afterwards. It seems some objects make problems on re-selecting. If I only store the englishName and findAndSelect by that, some objects cannot be recovered.  

Critical question: How to properly preserve and restore  a selection?

See the 3 places in ObsListDialog.cpp where we need to prevent loss of selected object:

void ObsListDialog::obsListHighLightAllButtonPressed()
void ObsListDialog::loadSelectedObservingListFromJsonFile
void ObsListDialog::loadBookmarksInObservingList()

To test, select a star before calling the bookmarks button. To prepare the crash, activate the line which restores preselectedObject instead of searching for the preselectedName. What about retrieval of dim stars? A StarWrapper should be preserved in the preselectedObject. Apparently it is not! (?)


Another question: Even when the old bookmarks have been imported once, I have the impression the bookmarks file is still imported anew on every new load? It should be easy to prevent that by just testing whether the jsonFile already exists: then we can assume the import has taken place already.

alex-w:
> @Jocelyn1109 , @alex-w I am stuck. It is about the original issue: selection lost during manipulation of ObsLists.
> 
> To keep the selected object, we have to store it somehow and retrieve it afterwards. It seems some objects make problems on re-selecting. If I only store the englishName and findAndSelect by that, some objects cannot be recovered.
> 
> Critical question: How to properly preserve and restore a selection?

Best way - use `getSelectedObject`/`setSelectedObject `methods from `StelObjectMgr `class for original issue and `findAndSelect(objname, objtype)` or `findAndSelect(Vec3d position)` methods for objects from lists

> Another question: Even when the old bookmarks have been imported once, I have the impression the bookmarks file is still imported anew on every new load? It should be easy to prevent that by just testing whether the jsonFile already exists: then we can assume the import has taken place already.

Why it's happening? The import process should be going once time - open file, import data, close file, insert data into list, save list.

Jocelyn1109:
Hi @gzotti @alex-w,

Georg,
importing old bookmarks only needs to be done once.
I don't think I have this problem on my branch.
I'm going to take a look this afternoon at what could be a problem in your code.

I will also look at the problem you have with edit mode and data not updating. I had tested my fix on my branch and it worked but I have to look for this is not the case in your code :)

Jocelyn

gzotti:
The bookmarks file is loaded in createDialogContent simply if it exists. There is (AFAICS) no check about the pre-existence of a list called "bookmarks". 
A simple way out would be the assumption that if a bookmarks.json file exists, but no observingList.json file, then this is run for the first time. and only then the bookmarks will be imported. The cleaner option would be importing the observingList.json first, then test if there is a "bookmarks" list, and only if there is no such list, a bookmarks.json would be imported if it exists. The first path is very simple to adjust, not sure about the second.

Jocelyn1109:
@ I will see at lunchtime and I will try to help you.

gzotti:
About crash around missing selection, I may have found a solution. Will commit ASAP.

gzotti:
I have currently no more crash after this commit. Two more observations/issues to fix:

1) After adding/removing an object and pressing save&close, the underlying list view is not updated. Only hiding/redisplaying the window will show them. Maybe some signal that should trigger a reload is lost?
2) I have tried to keep the original time when loading. The restoring of data necessarily has to change time and do core-updates etc. After loading, I restore original time. To make up for lost time during loading I try to apply a tweak we have applied elsewhere to overcome long method runtimes. This currently causes problems, but I will see to solve that later. 

For V1.1, I think we should remove the secondary dialog which shows the very same list again. It should be possible to add/remove objects in the currently shown list. Maybe protect it with an "edit" switch that enables add/delete buttons, but that should be all.

gzotti:
About the loading of old bookmarks: This is definitely called on every start. It seems that the old bookmark entries are re-loaded into the bookmarks list, but new entries are left intact. It's just a waste of processing time, just load the existing observingList.json, check for existence of a bookmarks list, and don't attempt to re-load the bookmarks in this case.

Jocelyn1109:
> For V1.1, I think we should remove the secondary dialog which shows the very same list again. It should be possible to add/remove objects in the currently shown list. Maybe protect it with an "edit" switch that enables add/delete buttons, but that should be all.

I'm not sure because it involves redoing everything and risks overloading the main dialog window with features and buttons.

src/gui/ObsListCreateEditDialog.cpp
```diff
@@ -43,1064 +44,922 @@
...
  * Save selected object into the list of observed objects.
  */
-void ObsListCreateEditDialog::obsListAddObjectButtonPressed() {
-    initErrorMessage();
-    const QList<StelObjectP> &selectedObject = objectMgr->getSelectedObject();
-
-    if (!selectedObject.isEmpty()) {
-
-        // No duplicate item in the same list
-        bool is_already_in_list = false;
```
Jocelyn1109, line 263: You can delete this line, it is not useful because the variable is always false.

src/gui/ObsListCreateEditDialog.cpp
```diff
@@ -43,1064 +44,922 @@
...
-        QHash<QString, observingListItem>::iterator i;
-        for (i = observingListItemCollection.begin(); i != observingListItemCollection.end(); i++) {
-            if (i.value().name.compare(selectedObject[0]->getEnglishName()) == 0) {
-                is_already_in_list = true;
-                break;
-            }
-        }
-
-        if (!is_already_in_list) {
```
Jocelyn1109, line 272: You can delete this check to (see my comment above).

gzotti:
> I did a test with my code and your code but I don't have the problem.
> Could you explain me what you do and send me your bookmarks file and observing lists file.

Ah, now I see the logic:

If bookmark file exists, load bookmarks, make them compatible to the new format, and attempt to store in obslist file. If bookmark list exists, don't save the newly processed data.

This is still a waste of time, as we still load and process (restore data etc) the bookmark file on every createDialogContent(). We just don't store, but throw away the processed data.

Jocelyn1109:
@gzotti 
> This is still a waste of time, as we still load and process (restore data etc) the bookmark file on every createDialogContent(). We just don't store, but throw away the processed data.

Yes I am agree this why that's why I planned to do a refactor.
For me the goal was to deliver a first version of the observation lists in order to correct all the details that were not in the specification (and also clarify).

For the import of the bookmark file I don't have a solution at the moment. We could rename it once the import is done or even delete it but we would still have the problem if the user copies/pasts the file again and manually.

I think, for the moment we can keep the initial behavior.

alex-w:
@gzotti @Jocelyn1109 important note: I've added couple important fixes and I found one possible issue - please create a few lists, select any non-default list and press button "Edit list" - what happening?

Jocelyn1109:
@alex-w 
> please create a few lists, select any non-default list and press button "Edit list" - what happening?

No problem on my side. The list is open in edit mode
Is there a bug ?

alex-w:
> @alex-w
> 
> > please create a few lists, select any non-default list and press button "Edit list" - what happening?
> 
> No problem on my side. The list is open in edit mode. Is there a bug ?

Which list is opened to edit - default or selected? What about edition one list and later try to edit other list.

Jocelyn1109:
@alex-w 
> Which list is opened to edit - default or selected? What about edition one list and later try to edit other list.

The selected list, I thinks it is a normal behavior :)
The default list is selected when you open the observing list dialog.

src/gui/ObsListCreateEditDialog.cpp
```diff
@@ -43,1064 +44,921 @@
...
+		qWarning() << "[ObservingList Creation/Edition] Error: the list name is empty.";
+		displayErrorMessage(q_("Error: the list name is empty."));
+	}
+	else
+	{
+		isSaveAs = listName.compare(currentListName) != 0 && !isCreationMode;
+		qDebug() << "ObsListCreateEditDialog::obsListSaveButtonPressed(): isSaveAs:" << isSaveAs;
+		saveObservedObjectsInJsonFile();
+		this->close();
+		// emit exitButtonClicked();
```
Jocelyn1109, line 619: Don't comment this line :)

src/gui/ObsListCreateEditDialog.cpp
```diff
@@ -43,1064 +44,921 @@
...
  */
-void ObsListCreateEditDialog::obsListExitButtonPressed() {
-    ui->obsListErrorMessage->setHidden(true);
-    ui->obsListErrorMessage->clear();
-    this->close();
+void ObsListCreateEditDialog::obsListExitButtonPressed()
+{
+	ui->obsListErrorMessage->setHidden(true);
+	ui->obsListErrorMessage->clear();
+	this->close();
```
Jocelyn1109, line 630: Add emit this->exitButtonClicked(); after close.

gzotti:
> > Critical question: How to properly preserve and restore a selection?
> 
> Best way - use `getSelectedObject`/`setSelectedObject `methods from `StelObjectMgr `class for original issue and `findAndSelect(objname, objtype)` or `findAndSelect(Vec3d position)` methods for objects from lists

Sure. However, one must make an additional copy of the selection result into a new list. The result of `getSelectedObject` is volatile even if assigned to a const variable!

src/gui/ObsListCreateEditDialog.cpp
gzotti: I had tried to make the test more comprehensive, however it fails. OK, we accept duplicate entries for now. Maybe for 1.1...

alex-w (approved):
It's OK for version 1.0, thanks!

alex-w:

PR #2708, opened 2022-10-01
Make Date&Time dialog more usable for typing date

### LINKED ISSUE #2272: Are values of 0 and 13 for month spinbox in DateTime dialog useful?
If you try to type a date like `2017-08-21` into the DateTime dialog (pressing <kbd>Tab</kbd> instead of typing <kbd>-</kbd>), you'll find that on typing `0` in the month field so as to get `08` the value suddenly changes to `12`, which makes you select all and re-type it without the `0`.

After some investigation I found that the ui file, `dateTimeDialogGui.ui`, has strange limits set for this spinbox: 0 to 13. This is just as the natural limits of 1 to 12, but with a margin of 1. I wonder, what is the use of these insensible values?

One use I suppose is to allow scrolling through the years like December to next January or January to previous December — with mouse wheel or Up/Down arrow keys. But is this the only use? Is the _typing_ behavior described above actually intended?

If the typing behavior is unintended, I might come up with a better scroll-through implementation, to improve the typing UX.

10110111:
### Description

Dates are commonly represented in a format where month and day are two-digit, and if less than 10, they have leading zero. Same goes for time. If we were to type such a date or time into the current Date&Time dialog, we'll get an annoying result of decreasing value of the spinbox to the left of the current one and getting maximum value in the current spinbox.

This is due to the way carry from lower-order date-time elements (e.g. days) to the higher-order elements (e.g. months) is implemented: by using "overflow" values to catch a step and then fixing it up.

This patch makes the behavior closer to the expected one by catching the step event in a more natural way.

Fixes #2272

### Type of change

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -363,6 +363,7 @@ IF(STELLARIUM_GUI_MODE STREQUAL "Standard")
 	  gui/ContributorsList.hpp
           gui/StelCloseButton.hpp
           gui/StelCloseButton.cpp
+		  gui/ExternalStepSpinBox.hpp
      )
 
      ################# compiles .ui files ############
--- a/src/gui/DateTimeDialog.cpp
+++ b/src/gui/DateTimeDialog.cpp
@@ -75,6 +75,13 @@ void DateTimeDialog::createDialogContent()
 	ui->dateDelimiterLabel2->setText(delimiter);
 
 	connectSpinnerEvents();
+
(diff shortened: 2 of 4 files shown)


alex-w:
The year section is very wide - is it possible to make the D&T dialog more compact?

10110111:
It didn't change with this PR. In fact, it's always been wider than others. So this is an issue for another fix.

alex-w (approved):
Thanks

PR #2720, opened 2022-10-04
New rules for packaging

alex-w:
### Description
This pull request contains the new rules for packaging Windows binaries and an AppImage for linux. Our Windows binaries are too big now and we catches some limits from CI/CD environment now. The new rules should improve situation.

The current CI/CD limits:
* The AppVeyor has limits for downloading artifacts in 1GB/day, so, we cannot downloads and signings all Windows installers (3 files with average size 360MB)
* SignPath.io has limit 4GB/month for downloading and signings files, so, we cannot prepare signed weekly builds now

Now:
* The branch `stellarium-next` should be using for prepare unsigned Windows packages and AppImages (weekly builds)
* The branch `stellarium-stable` should be using for prepare AppImages and Windows packages, where Qt6-based package should be signed also (replacement for `release` branch)
* The branch `stellarium-oldstable` should be using for prepare Windows packages, where Qt5-based package should be signed also (replacement for `release-qt5` branch)
* The word `[publish]` in commit message will be cause prepare AppImage and Windows packages

All these branches should be updating by rebase command.

Probably since version 1.1 Qt5-based packages for Windows (Tier 2) should be prepared in second day and at stellarium.org should be link for Qt6-based only package for Windows (Tier 1).


gzotti:
Looks good. In addition, this PR description should go to MAINTAINER_BUSINESS.md or some other place where it is not overlooked!

gzotti:
Actually 2 updates not in the PR range: 

- Update Qt6 to 6.4. I am surprised apparently nobody except myself stumbled over the black window issue in 1.22.3 with Intel GPU. The other issue was reported that build fails on Arch with its Qt6.4 (Arch keeps the Qt5-based version for now!)
- line 7: clone depth: Why 5? Does it ever need to be more than 1 for a one-time build oriented checkout?

alex-w:
> In addition, this PR description should go to MAINTAINER_BUSINESS.md 

Why? The changes in our infrastructure for preparing weekly snapshots and tarballs does not touch people, who will build the binary package from source code.

alex-w:
> * Update Qt6 to 6.4. 

Well, this is acceptable, but I guess the process of building will slow again... I hope Qt 6.4 at AppVeyor not a beta already.

> I am surprised apparently nobody except myself stumbled over the black window issue in 1.22.3 with Intel GPU. 

No testers - no reports :(

> The other issue was reported that build fails on Arch with its Qt6.4 (Arch keeps the Qt5-based version for now!)

Qt 6.4 released September 29, the Stellarium released October 1. Releasing with new framework without any testing is very bad idea IMHO. 

> * line 7: clone depth: Why 5? Does it ever need to be more than 1 for a one-time build oriented checkout?

We may reduce the depth to 1, but it will cause increasing false "broken" reports due fails get the source code.

gzotti:
I agree, more extensive testing before release with all the latest features should be mandatory. This implies a feature freeze 2 weeks before release and 2 RCs with a week each. Circumstances often work differently. The error that Qt6.3 has with my home PC were quite unexpected.

OK, here are logs from 3 builds of 02feef10d62d8fdfae5d0 with 3 versions of Qt6. This is an i5-4570 (Intel HD Graphics 4600), Win10home-21H2. This represents "average mid-aged home PCs". 

[log_Qt6.2.4.txt](https://github.com/Stellarium/stellarium/files/9711126/log_Qt6.2.4.txt)
[log_Qt6.3.2.txt](https://github.com/Stellarium/stellarium/files/9711127/log_Qt6.3.2.txt)
[log_Qt6.4.0.txt](https://github.com/Stellarium/stellarium/files/9711128/log_Qt6.4.0.txt)

Compare to V1.22.3
[log_1.22.3.txt](https://github.com/Stellarium/stellarium/files/9711138/log_1.22.3.txt)

I see shader errors in both 6.3-based versions which have vanished in Qt6.4. However, for whatever reason I also no longer see a black window with 1.0, but it seems to work. This explains why we did not see more reports. I cannot say why I had the black window earlier.

alex-w:
Well, some sketeton for checklist/releasing process is needed now, you are right.

Is PR acceptable?

gzotti:
I checked the Qt6.4 build (artifact stellarium-1.22.3-5ed6ed0-win64.exe from appveyor). Something missing around QtMultimedia? The program fails to launch. Self-built on Qt6.4 works. 
[log.txt](https://github.com/Stellarium/stellarium/files/9714041/log.txt)

alex-w:
> I checked the Qt6.4 build (artifact stellarium-1.22.3-5ed6ed0-win64.exe from appveyor). Something missing around QtMultimedia? The program fails to launch. Self-built on Qt6.4 works. [log.txt](https://github.com/Stellarium/stellarium/files/9714041/log.txt)

O_O I didn't touch any ISS rules - please check Qt6.3-based artifact from previous build

gzotti:
Aargh -- `{"message":"Artifacts download limit (1024 MB/day) exceeded."}`

10110111:
If you want to have downloads for testing, it may be a good idea to split often-changing objects like EXEs and DLLs into separate artifacts. This will reduce the download size and thus make it possible to download much more times a day.

alex-w:
> If you want to have downloads for testing, it may be a good idea to split often-changing objects like EXEs and DLLs into separate artifacts. This will reduce the download size and thus make it possible to download much more times a day.

For `stellarium-next` branch & `[publish]` builds uploading stellarium.exe in additional to installer are acceptable for advanced users and developers. But (!) we are catch the problem with "installer" for Qt6.4-based Stellarium.

alex-w:
@gzotti did you found what is missing in Qt6 package?

worachate001:
I tried stellarium-1.22.3-177ad25-win64, having the same scenario.
```
could not load multimedia backend ""
QtMultimedia is not currently supported on this platform or compiler.
```
May be related to this - https://bugreports.qt.io/browse/QTBUG-105984

alex-w:
OK, it should work now

gzotti:
{"message":"Artifacts download limit (1024 MB/day) exceeded."}
Somebody was faster...

worachate001:
The stellarium-1.22.3-73c51e1-win64.exe installer is working. 👍

alex-w:
OK, the PR is ready to merge

@10110111 I guess with some modifications of _the patch installer_ you can distribute updates for testing showmysky/atmopshere (by separate PR though)

worachate001:
Is the 11 MB patch (stellarium-1.22.3-9de7ebc-patch-win64.exe) supposed to be used with 1.0 release?

I got this error after applying it (I installed version 1.0 in Stellarium-beta folder).

Patch-error

alex-w:
> Is the 11 MB patch (stellarium-1.22.3-9de7ebc-patch-win64.exe) supposed to be used with 1.0 release?

Yes, but Qt should same version for both packages. For release we using Qt 6.3.1 and for patch I used Qt 6.4.0, so, it shouldn't be work. But it should be work for case an installation the patch as an ugrading the previous beta.

P.S. I tested for release + patch on my Win7 at work (Qt 5.12.6 in both cases)

worachate001:
Okay. Testing with previous Qt 6.4-based version. It works as expected.

gzotti:
Finally I was able to download the stellarium-1.22.3-c58f3e4-win64.exe. Install works, logfile has no bad signs. Patch installation also works, taking the last used installation path. Approve from my side!

worachate001 (approved):

PR #2723, opened 2022-10-05
Optional support XLSX files

alex-w:
### Description
This pull request introduced new cmake option - `ENABLE_XLSX ` - to allow users an optional support the XLSX files and the optional dependency for QXlsx library respectively (enabled by default). As an additional feature the support XLSX files in AstroCalc was deep refactored and the LOC for XLSX was dramatically reduced. 

### Type of change

## Checklist:


plugins/Satellites/src/gui/SatellitesDialog.cpp
```diff
@@ -1624,6 +1624,7 @@ void SatellitesDialog::savePredictedIridiumFlares()
 	}
 	else
 	{
+		#ifdef ENABLE_XLSX
```
10110111, line 1627: The `*.xlsx` filter should be removed in this case, otherwise we'll have "do nothing" behavior while the user expects something useful.

gzotti: In principle yes. But this is actually disabled code. It can be reactivated if Iridium-like satellites will be ever launched again.

10110111: If we are adding this `#ifdef`, we must fix the surrounding code. Otherwise, if it's ever enabled, it will work in ways unexpected by the user.

alex-w: I've added some cosmetics for this disabled code, but in the future we should enable this code for flared satellites (after modification to support other flared sats)

CMakeLists.txt
```diff
@@ -413,6 +413,12 @@ SET(ENABLE_GPS 1 CACHE BOOL "Define whether GPS queries should be supported.")
 # Activate support for ShowMySky atmosphere
 SET(ENABLE_SHOWMYSKY 1 CACHE BOOL "Define whether to support ShowMySky atmosphere model if installed")
 
+# Activate support for XLSX files
+SET(ENABLE_XLSX 1 CACHE BOOL "Define whether to support XLSX files")
```
gzotti, line 417: Better: "Define whether to support XLSX (Excel) files"

plugins/Satellites/src/gui/SatellitesDialog.cpp
```diff
@@ -1586,6 +1587,16 @@ void SatellitesDialog::savePredictedIridiumFlares()
 							QDir::homePath() + "/iridium_flares.xlsx",
 							filter,
 							&defaultFilter);
+	#else
```
gzotti, line 1590: This should not be an exclusive else: If XLSX is supported, allow export of XLSX and also CSV. 
(OK, I know, this is unimportant, currently parked code...)

alex-w:
OK, the PR is ready to merge

gzotti (approved):
Looks good, although I have not tested everything. Fine to also have the table copies refactored to one!

PR #2729, opened 2022-10-07
Add support for OpenGL 3.3 Core profile

### LINKED ISSUE #2707: Support OpenGL 3.3 Core Profile
This is mostly a demand caused by Apple's lack of supporting an OpenGL 3.3 Compatibility Profile.

The ShowMySky feature which provides the wonderful twilight colors requires OpenGL 3.3. Most computers' OpenGL subsystems nowadays deliver 4.* Compatibility Profiles. Apple traditionally axes compatibility features and as far as we can see provides only OpenGL Core profiles. 

Stellarium still uses some OpenGL functions which are no longer available in a 3.3 Core profile. 

I don't have an Apple computer, so I personally don't care for and cannot develop this, but there are certainly Apple developers out there who are able and willing to do that.

* Initialisation on Apple computers shall deliver an OpenGL 3 3 core profile at start if possible (easy)
* Stellarium must run completely without graphical errors caused by attempting to run OpenGL functions which have been removed from the 3.3 Core profile. 

On other platforms, existing functionality must not change. OpenGL 2.1 and OpenGL ES 2.0 systems must continue to work. Where OpenGL 3.3 functions are not available, there is no ShowMySky, that's OK.
 
Currently there is no ShowMySky on Apple computers.

**Additional context**
Stellarium's ShowMySky is awesome. It would be nice to have this also on the Apple platform!

10110111:
### Description

This set of changes should make it possible to run Stellarium with ShowMySky model on macOS and other systems where Compatibility profile is not supported.
Core profile is now made the default, an option is added to request compatibility profile: `--opengl-compat`.

Fixes #2707 

@alex-w Please test it on macOS and tell me if you find problems. They may appear anywhere: in ShowMySky, in rendering of various objects (planets, stars, nebulae, ground, etc.).
@gzotti Please test it on the minor systems that are not capable of OpenGL 3.3. It may work, but it also might fail if the systems don't support VAO.

### To do

- [x] `GL_LUMINANCE` is not a supported texture format in Core profile. Needs to be replaced somehow.
- [x] Core profile requires all draw commands to be use a non-zero VAO.
- [x] In Core profile macOS doesn't support any GLSL version except the one guaranteed by the requested OpenGL version. Needs some wrappers for compatibility with ANGLE/GL2 systems.
- [x] VAO use needs to be optional for the case of ANGLE/GL2 systems.
- [x] `glLineWidth` doesn't support argument>1 in Core profile, so lines should be rendered in terms of quads.
- [x] `GL_INVALID_FRAMEBUFFER_OPERATION` appears after loading ShowMySky model on macOS. Doesn't seem to influence the operation, but leaves errors in the log.
- [x] Documentation needs update to reflect the changes to default GL version and command-line options that control it.

### DIFF
--- a/guide/ch_advanced_use.tex
+++ b/guide/ch_advanced_use.tex
@@ -505,7 +505,7 @@ For example, using the option \texttt{-c\ my\_config.ini} would resolve to the f
 -\/-mesa-mode or -m  & {[}none{]} & Use MESA as software OpenGL rendering engine.\footnotemark[1]\\
 -\/-safe-mode or -s  & {[}none{]} & Synonymous to -\/-mesa-mode.\footnotemark[1]\\
 -\/-single-buffer    & {[}none{]} & Use single-buffering. This reportedly avoids screen blanking problems with some Intel GPUs. \\
--\/-compat33 or -C   & {[}none{]} & Request OpenGL 3.3 Compatibility Profile. May help for certain driver configurations. Mac?\\
+-\/-opengl-compat or -C & {[}none{]} & Request OpenGL 3.3 Compatibility Profile. Might fix graphics problems in some driver configurations.\\
 -\/-fix-text or -t   & {[}none{]} & Alternative way of creating the Info text, required on some systems.\footnote{E.g., Raspberry Pi 2/3 with Raspbian Jessie and VC4 drivers from Mesa 11 (2016).}\\\bottomrule
 \end{longtable}
 
@@ -579,9 +579,9 @@ problems.
 \item[STEL\_USERDIR] may contain the path to a user data directory
   deviating from the default (see section \ref{sec:Directories}).
 \item[QT\_OPENGL]\footnote{Windows only} May be one of \texttt{desktop} (native OpenGL for your GPU, recommended),
-  \texttt{angle}\footnote{Stellarium 0.* series only} or \texttt{software}. The last activates pure software rendering using the MESA OpenGL library.
+  \texttt{angle}\footnote{Stellarium 0.* series only} or \texttt{software}. The last activates pure software rendering using the MESA OpenGL library. Note that command line options take precedence over this environment variable.
 \item[QT\_ANGLE\_PLATFORM]\footnote{Windows and Stellarium 0.* series only} May be one of \texttt{d3d9} (DirectX~9) or \texttt{d3d11} (DirectX~11),
-  or \texttt{warp} for another software-only solution.
+  or \texttt{warp} for another software-only solution. Note that command line options take precedence over this environment variable.
 \end{description}
 
 
--- a/guide/ch_getting_started.tex
+++ b/guide/ch_getting_started.tex
@@ -28,11 +28,11 @@ The most important hardware requirement is a contemporary graphics subsystem.
 \subsection{Minimum}
 \begin{itemize}
 \item Linux/Unix; Windows 7 and later; macOS 10.15 and later\footnote{Windows 10 and macOS 11.0 is minimal for Qt6-based Stellarium.}
-\item 3D graphics capabilities which support OpenGL 3.0 and GLSL 1.3 (2008
+\item 3D graphics capabilities which support OpenGL 3.0 (2008
   GeForce 8xxx and later, ATI/AMD Radeon HD-2xxx and later; Intel HD
-  graphics (Core-i 2xxx and later)) or OpenGL ES 2.0 and GLSL ES 1.0
+  graphics (Core-i 2xxx and later)) or OpenGL ES 2.0
   (e.g., ARM SBCs like Raspberry Pi~2/3/4). On Windows, some older cards
-  may be supported via ANGLE when they support DirectX10.
+  may be supported via ANGLE when they support DirectX9.
 \item Screen resolution $1024\times768$\footnote{On Linux, an $800\times600$ screen can still be used by scaling the desktop e.g.\ to $1200\times900$:
   \begin{commands}
     xrandr --output HDMI-1 --scale 1.5x1.5
@@ -49,7 +49,7 @@ The most important hardware requirement is a contemporary graphics subsystem.
 \subsection{Recommended}
 \begin{itemize}
 \item Linux/Unix; Windows 10 and later; macOS 11.0 and later
-\item 3D graphics card which supports OpenGL 3.3 Compatibility Profile and above and GLSL1.3 and later
+\item 3D graphics card which supports OpenGL 3.3 or higher
 \item FullHD ($1920\times1080$ or $1920\times1200$) or larger screen.\footnote{HiDPI screens may work, but show occasional platform-dependent issues.}
 \item 1 GB RAM or more
 \item 1.5 GB free on disk (About 3GB extra required for the optional DE430/DE431 files).
--- a/guide/ch_interface.tex
+++ b/guide/ch_interface.tex
@@ -607,11 +607,9 @@ There are two visual models for the atmosphere available:
   \item[Preetham] This is the legacy model (see section~\ref{sec:skylight:Preetham}), fallback for the cases
     when the other one doesn't work.
   \item[ShowMySky] This model is the more realistic visual model of
-    the atmosphere colors (see section~\ref{sec:skylight:Bruneton}). It
+    the atmosphere colors (see section~\ref{sec:skylight:ShowMySky}). It
     relies on a precomputed dataset that can be chosen in the user
     interface after the \program{ShowMySky} model is enabled.
-    \footnote{Due to compatibility issues, this mode is currently not
-      available on the Apple Mac platforms.}
 \end{description}
 
 \subsection{The Solar System Objects (SSO) Tab}
--- a/guide/ch_skylight.tex
+++ b/guide/ch_skylight.tex
@@ -71,8 +71,8 @@ disk and the solar glare. It makes a slight difference whether the
 Sun's disk is plotted before or after the glare, and whether the solar
 sphere is rendered after the atmosphere. 
 
-\subsection{Advanced Mode: The Bruneton Skylight Model}
-\label{sec:skylight:Bruneton}
+\subsection{Advanced Mode: The ShowMySky Skylight Model}
+\label{sec:skylight:ShowMySky}
 
 A much more advanced skylight model has been presented by
 \citet{Bruneton:2008}. An implementation of this has been developed by
@@ -80,12 +80,12 @@ Ruslan Kabatsayev just in time for inclusion in version 1.0. Switching
 between the models is described in
 section~\ref{sec:gui:view:sky:atmosphere}.
 
-It requires graphics hardware which uses OpenGL~3 or better and
+It requires graphics hardware which uses OpenGL~3.3 or better and
 consists of two components.  Atmospheric data (a description of gas
 composition) is processed by an auxiliary program,
-\program{CalcMySky}, which creates a lot of textures and shader
-programs. For final display, a software component (the
-\program{ShowMySky} library) is used and processes these data in
+\program{CalcMySky}, which creates a lot of textures and shaders.
+For final display, a software component (the
+\program{ShowMySky} library) is used that processes these data in
 real-time to compute the final sky colors. This model and its default
 atmosphere data especially can deliver stunning reality of twilight
 colors in a dry atmosphere, and also provides special modes for the


alex-w (changes requested):
The running Stellarium is fails on macOS due incompatible shaders
[log.txt](https://github.com/Stellarium/stellarium/files/9730433/log.txt)

gzotti:
The #version must be added only in OpenGL mode. For GL ES, it should remain version-less!

10110111:
Please retest with the new commits.

alex-w:
> Please retest with the new commits.

Not all shaders are fixed yet

[log.txt](https://github.com/Stellarium/stellarium/files/9738301/log.txt)

10110111:
Please retest.

alex-w:
> Please retest.

It can be compiled in run now, but without rendering the sky, see 
[log.txt](https://github.com/Stellarium/stellarium/files/9738994/log.txt)

10110111:
Please try now.

alex-w:
> Please try now.

OK, but ShowMySky shaders should be fixed too, see [log.txt](https://github.com/Stellarium/stellarium/files/9739125/log.txt)

At least standard mode is works now on macOS

10110111:
That was unexpected :) OK, is it fixed now?

10110111:
The second shader should be fixed now.

> `qt.qpa.openglcontext: Could not create NSOpenGLContext with shared context, falling back to unshared context.`

This is strange, especially if it works fine when signed... I'll try some things after we get the signed version to work.

alex-w:
It crashes now when ShowMySky atmosphere is loading...  see [log.txt](https://github.com/Stellarium/stellarium/files/9739271/log.txt)

10110111:
No crash info got into the log. Is there any kind of backtrace or something?

alex-w:
> No crash info got into the log. Is there any kind of backtrace or something?

Yes, of course: [log.txt](https://github.com/Stellarium/stellarium/files/9739293/log.txt)

10110111:
Looks like an exception thrown from libShowMySky failed to be caught in Stellarium binary—although there does exist a `try` statement down the call stack. Does macOS need special compiler settings to enable exception propagation between modules? E.g. I had to inlude `/EHs` on Windows to enable stack unwinding.

alex-w:
> Looks like an exception thrown from libShowMySky failed to be caught in Stellarium binary—although there does exist a `try` statement down the call stack. Does macOS need special compiler settings to enable exception propagation between modules? E.g. I had to include `/EHs` on Windows to enable stack unwinding.

Well, macOS using AppleClang compiler by default, so, all flags for clang should works here too.

10110111:
> P.S. Should I try build master of ShowMySky to test?

You'd need to patch Stellarium for this. Besides, there're no relevant fixes there, so no need to.

10110111:
Could you build CalcMySky with the following patch and get a new log from Stellarium with this libShowMySky?

```diff
diff --git a/ShowMySky/AtmosphereRenderer.cpp b/ShowMySky/AtmosphereRenderer.cpp
index 9b232b0..5efeb67 100644
--- a/ShowMySky/AtmosphereRenderer.cpp
+++ b/ShowMySky/AtmosphereRenderer.cpp
@@ -2229,6 +2229,7 @@ void AtmosphereRenderer::drawSurface(QOpenGLShaderProgram& prog)
 
 void AtmosphereRenderer::resizeEvent(const int width, const int height)
 {
+    qDebug().nospace() << "AtmosphereRenderer::resizeEvent(" << width << ", " << height << ")";
     OGL_TRACE();
 
     viewportSize_=QSize(width,height);
diff --git a/common/util.cpp b/common/util.cpp
index 21cfe3e..61f03ee 100644
--- a/common/util.cpp
+++ b/common/util.cpp
@@ -44,6 +44,7 @@ void checkFramebufferStatus(QOpenGLFunctions_3_3_Core& gl, const char*const fboD
         errorDescription=QObject::tr("unknown error 0x%1").arg(status, 0, 16);
         break;
     }
+    qDebug().noquote() << QObject::tr("%1 is incomplete: %2").arg(fboDescription).arg(errorDescription);
     throw OpenGLError{QObject::tr("%1 is incomplete: %2").arg(fboDescription).arg(errorDescription)};
 }
 
```

alex-w:
See [dump.txt](https://github.com/Stellarium/stellarium/files/9739579/dump.txt) and [log.txt](https://github.com/Stellarium/stellarium/files/9739580/log.txt)

alex-w:
See [dump.txt](https://github.com/Stellarium/stellarium/files/9739579/dump.txt) and [log.txt](https://github.com/Stellarium/stellarium/files/9739580/log.txt)

10110111:
What if you use this patch instead?
```diff
diff --git a/ShowMySky/AtmosphereRenderer.cpp b/ShowMySky/AtmosphereRenderer.cpp
index 4edb68f..80d53bc 100644
--- a/ShowMySky/AtmosphereRenderer.cpp
+++ b/ShowMySky/AtmosphereRenderer.cpp
@@ -2227,10 +2227,18 @@ void AtmosphereRenderer::drawSurface(QOpenGLShaderProgram& prog)
     drawSurfaceCallback(prog);
 }
 
-void AtmosphereRenderer::resizeEvent(const int width, const int height)
+void AtmosphereRenderer::resizeEvent(int width, int height)
 {
+    qDebug().nospace() << "AtmosphereRenderer::resizeEvent(" << width << ", " << height << ")";
     OGL_TRACE();
 
+    if(width==0 || height==0)
+    {
+        qWarning().nospace() << "AtmosphereRenderer::resizeEvent(" << width << ", " << height << "): zero area framebuffer specified";
+        width  = std::max(1, width);
+        height = std::max(1, height);
+    }
+
     viewportSize_=QSize(width,height);
     if(!luminanceRadianceFBO_) return;
 
diff --git a/common/util.cpp b/common/util.cpp
index 21cfe3e..61f03ee 100644
--- a/common/util.cpp
+++ b/common/util.cpp
@@ -44,6 +44,7 @@ void checkFramebufferStatus(QOpenGLFunctions_3_3_Core& gl, const char*const fboD
         errorDescription=QObject::tr("unknown error 0x%1").arg(status, 0, 16);
         break;
     }
+    qDebug().noquote() << QObject::tr("%1 is incomplete: %2").arg(fboDescription).arg(errorDescription);
     throw OpenGLError{QObject::tr("%1 is incomplete: %2").arg(fboDescription).arg(errorDescription)};
 }
 
```

alex-w:
WOW, it works!

alex-w:
> WOW, it works!

Damn! Switching Preetam->ShowMySky caused crash! See [log.txt](https://github.com/Stellarium/stellarium/files/9739827/log.txt) and [dump.txt](https://github.com/Stellarium/stellarium/files/9739829/dump.txt)

Same time and location (Preetam + ShowMySky):
stellarium-002
stellarium-003

Rendering ShowMySky is so slow... :(

alex-w:
In the log: `Unsupported atmosphere model "Притам"`

Please mark items of dropdown list as untranslatable or store and use English names as `Qt::UserRole`

10110111:
> Rendering ShowMySky is so slow

Don't forget about `atmosphere_resolution_reduction` introduced in #2676: it should improve the experience.

> In the log: Unsupported atmosphere model "Притам"

Yeah, I noticed it in previous logs.

alex-w:
> Don't forget about `atmosphere_resolution_reduction` introduced in #2676: it should improve the experience.

What about adding couple option near selector of atmosphere model as ShowMySky options?

Atque:
Shouldn’t a MacBook Pro cope with ShowMySky pretty flawlessly?

10110111:
> What about adding couple option near selector of atmosphere model as ShowMySky options?

I'm OK with this. Feel free to do it :) I'm currently busy with OpenGL.

10110111:
> Shouldn’t a MacBook Pro cope with ShowMySky pretty flawlessly?

Judging by a screenshot above, it's about 14 FPS at 2880×1800 resolution. For comparison, on my GeForce GTX 750Ti at 1920×1080 I get 23 FPS. So the Mac performs usably IMO.

10110111:
> Same time and location (Preetam + ShowMySky):

Does it really look as banded as the screenshots show? Could you post a PNG version instead of the artifacted JPEG one?

10110111:
@alex-w Now regarding broken exceptions, [this post](https://github.com/ocaml/ocaml/issues/9026#issuecomment-539641762) says that we indeed need a special build flag. Could you try building vanilla CalcMySky v0.1.0 passing `-DCMAKE_SHARED_LINKER_FLAGS=-Wl,-keep_dwarf_unwind`, and then try reproducing your original crash at startup?

alex-w:
> > Same time and location (Preetam + ShowMySky):
> 
> Does it really look as banded as the screenshots show? Could you post a PNG version instead of the artifacted JPEG one?

`atmosphere_resolution_reduction = 4`

stellarium-001

10110111:
> atmosphere_resolution_reduction = 4

Funny, now the frame rate with ShowMySky is higher than with Preetham atmosphere :D

The banding is still there though. Did you disable dithering?

alex-w:
> @alex-w Now regarding broken exceptions, [this post](https://github.com/ocaml/ocaml/issues/9026#issuecomment-539641762) says that we indeed need a special build flag. Could you try building vanilla CalcMySky v0.1.0 passing `-DCMAKE_SHARED_LINKER_FLAGS=-Wl,-keep_dwarf_unwind`, and then try reproducing your original crash at startup?

No lucky, see [dump.txt](https://github.com/Stellarium/stellarium/files/9740070/dump.txt)

10110111:
What happens if you run `showmysky` from the CalcMySky package, passing an invalid path to it? Does it crash, or do you get a message box with a readable error?

alex-w:
> What happens if you run `showmysky` from the CalcMySky package, passing an invalid path to it? Does it crash, or do you get a message box with a readable error?

This package will crash always on the macOS due 2 reasons: a) this is GUI application and b) this application is not packed by Apple’s rules

alex-w:
@10110111 please rebase the branch to apply fix for bundling the versioned version of ShowMySky library on macOS

10110111:
> This package will crash always on the macOS due 2 reasons: a) this is GUI application and b) this application is not packed by Apple’s rules

a) Stellarium is also a GUI application
b) When I tried to build CalcMySky with HomeBrew-based dependencies on a friend's (Intel-based) Mac, it worked without any packaging. See the screenshot below.

photo_5240370232021859785_y

alex-w:
> a) Stellarium is also a GUI application 

Stellarium is bundled :)

> b) When I tried to build CalcMySky with HomeBrew-based dependencies on a friend's (Intel-based) Mac, it worked without any packaging. See the screenshot below.

Fun, let me try it again

alex-w:
OK, I can build and run `showmysky` tool via CLI when it builded with Qt5 from homebrew. But it say "Failed to load ShowMySky library" after selection `atmosphere/default` directory.

10110111:
> But it say "Failed to load ShowMySky library" after selection atmosphere/default directory.

If you simply run it from the build directory, you may need to use `DYLD_LIBRARY_PATH` environment variable to point to the directory that contains the library. E.g. if you are in the build directory, and there's the library `./ShowMySky/libShowMySky-Qt*.dylib`, then you can run `showmysky` as follows:
```
DYLD_LIBRARY_PATH=./ShowMySky ./ShowMySky/showmysky ~/path/to/atmosphere/default
```

alex-w:
Oops... I forget about `DYLD_LIBRARY_PATH` on macOS :(

OK, the `showmysky` tool is works now via CLI

10110111:
> OK, the showmysky tool is works now via CLI

Great. Now, does it report errors correctly (e.g. when passed a non-existing directory), or does it crash?

alex-w:
> Great. Now, does it report errors correctly (e.g. when passed a non-existing directory), or does it crash?

Crash... see [dump.txt](https://github.com/Stellarium/stellarium/files/9741515/dump.txt)

gzotti:
Testing RPi4:

So far I could only compile without SMS because SMS building failed (did not continue, this was expected.)
Builds and runs with Ubuntu Mate 22.04 and Qt6. 
The Broadcom VC4 driver support is OpenGL 2.1 with GLSL1.2 For reasons I cannot remember we may have increased GLSL minimum to 1.3 and log a warning that something may fail. However, all seems to work, planet shaders, Satun's ring shadows, and even Scenery3D works with some logged issue which (I cannot remember!) may have been here before (caused by switching on PCSS shadow?)
The play about configs is interesting: Now our default format as requested is 3.3 Core, but the driver delivers 2.1 with deprecated functions. (It cannot provide more.)

Here are the logs.
[log_RPi4-U22.04-Scenery3D-OGL3branch.txt](https://github.com/Stellarium/stellarium/files/9741702/log_RPi4-U22.04-Scenery3D-OGL3branch.txt)
[log_RPi4-U22.04-OGL3branch.txt](https://github.com/Stellarium/stellarium/files/9741703/log_RPi4-U22.04-OGL3branch.txt)

I should still test the official Raspberry OS but need to install that first, need to buy a fresh SDcard for that...
RPi3 shall come next, a pure GL ES2.0 system.

10110111:
> need to buy a fresh SDcard for that

You can just dump the current image from the card to a file, install other OSes and play with them, and then restore the original image from the dump. On Unices the dump can be done using `ddrescue` tool. I read it's also usable on Windows (via Cygwin).

gzotti:
> > need to buy a fresh SDcard for that
> 
> You can just dump the current image from the card to a file, install other OSes and play with them, and then restore the original image from the dump. On Unices the dump can be done using `ddrescue` tool. I read it's also usable on Windows (via Cygwin).

I know. Just wanted to have that second OS anyway. I feel I am copying, updating and installing too many things too often...

gzotti:
RPi 3B+, Raspbian bullseye:
The OS'es repo has V0.20.4, running at about 16fps (1280x1024 screen).
This branch: builds (with Qt5), runs at launch, but it seems we have increasing issues with textures. All seems OK (although only at 5fps or so), includng planet shaders. but loading nebula textures leads to a segfault. I have seen this previously with HiPS/DSS on this platform, it seems the GPU driver has issues with texture memory handling. I have just found it would support ETC1 texture compression, but how to activate this?
I  built with the OPENGL_DEBUG_LOGGING  set, but this fails at start.

Regarding low fps, this is also quite bad, but I must build master to see if this slowdown is related to this branch.

[log_RPi3-bullseye_OGL3branch.txt](https://github.com/Stellarium/stellarium/files/9742246/log_RPi3-bullseye_OGL3branch.txt)
[log_RPi3-bullseye_0.20.4.txt](https://github.com/Stellarium/stellarium/files/9742247/log_RPi3-bullseye_0.20.4.txt)

gzotti:
Intel Core i5-4570 (Intel HD 4600, OpenGL 4.3):
I tried to build CalcMySky first, but face problems. 
grafik
grafik
grafik

The line before the red error ("command was ended with code 9009") says "command 'true' misspelled or not found". Any ideas?

10110111:
> command 'true' misspelled or not found

That's something funny. Looks like it's executing `version.sh`, which is a Bourne shell script. `true` is a shell builtin, which makes it strange to not be found.

Are you using `sh` from WSL? I remember having some problems with it. For AppVeyor I explicitly pointed the CI script to use bash that comes with git instead of the WSL one.

gzotti:
> Are you using `sh` from WSL? I remember having some problems with it. For AppVeyor I explicitly pointed the CI script to use bash that comes with git instead of the WSL one.

Ah, stupid. Actually i had taken notes on my main PC, but not exactly which shell. I remember I had to mix as required. 
But the cmake command was now given in the Windows-own cmd shell. Cmake is from QT's tools directory. Do I have to call a bash installed in my non-WSL cmd environmeint?

10110111:
> Do I have to call a bash installed in my non-WSL cmd environmeint?

Well you don't _call_ it manually, do you? I _think_ you can just make the non-WSL path closer to the beginning in `%PATH%`. In this case the correct bash would be run.

10110111:
@alex-w Please check whether the latest `master` CalcMySky has fixed the crashes.

gzotti:
Core i5-4570: I will continue the CMC build tomorrow (hopefully). For now, I have built without. It runs, but after activating a scenery3D, it shows weird artifacts (thin lines in the sky) which also remain when I switch off the 3D scene view. 

grafik

The log has an GL_INVALID_OPERATION
[log_i5-4570_OGL3branch.txt](https://github.com/Stellarium/stellarium/files/9742656/log_i5-4570_OGL3branch.txt)

10110111:
> it shows weird artifacts (thin lines in the sky)

> The log has an GL_INVALID_OPERATION

And none of this happens with `master`, right? And what if you pass `--opengl-compat` option?

alex-w:
> @alex-w Please check whether the latest `master` CalcMySky has fixed the crashes.

Crash at switching to ShowMySky model - see [log.txt](https://github.com/Stellarium/stellarium/files/9742793/log.txt) and [dump.txt](https://github.com/Stellarium/stellarium/files/9742794/dump.txt)

alex-w:
Loading ShowMySky model at startup, enable Scenery3D model - crash: [dump.txt](https://github.com/Stellarium/stellarium/files/9742828/dump.txt), [log.txt](https://github.com/Stellarium/stellarium/files/9742831/log.txt)

No crash and no artifacts from Scenery3D when Preetham model is used

10110111:
> Crash at switching to ShowMySky model

Please retest. A message should appear instead of crash. The logs should contain some info on the error.

alex-w:
> > Crash at switching to ShowMySky model
> 
> Please retest. A message should appear instead of crash. The logs should contain some info on the error.

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9745402/log.txt)

alex-w:
According to the latest log I can share one observation, which can be important - when Stellarium have 2 opened dialogs and I do screenshot then on the screenshot I see one of dialog as black rectangle. Probably something wrong in the buffers.

10110111:
Please retest with CalcMySky branch `more-gl-error-checks` built in `Debug` mode.

gzotti:
Sony Vaio VPCZ13V9E: i7-M860 from 2010. This is a pre-optimus with manually switchable Geforce 330M now dysfunctional by Sony's negligence in Win10 drivers, i.e. has only IntelHD. 
This will probably never show ShowMySky, but just run Series 0.* until 2025/Win10-EOL and could still drive a telescope. 

Qt6/master: mesa/black
Qt5.15.2 master runs OK
Qr5.15.2 this branch: black sky. 

[log_i7-M640_master_qt5-angle.txt](https://github.com/Stellarium/stellarium/files/9748628/log_i7-M640_master_qt5-angle.txt)
[log_i7-M640_master-qt6-mesa.txt](https://github.com/Stellarium/stellarium/files/9748629/log_i7-M640_master-qt6-mesa.txt)
[log_i7-M640_OGL3branch_qt5-angle-skyblack.txt](https://github.com/Stellarium/stellarium/files/9748630/log_i7-M640_OGL3branch_qt5-angle-skyblack.txt)

10110111:
> stel.OpenGLArray: Vertex Array Objects are not supported on your hardware (this is not an error)

This is ugly. Even my netbook with GMA950 (OpenGL 1.4) supports VAO (on Ubuntu). And with ANGLE we have to do double work just because it's so limited.

alex-w:
> Please retest with CalcMySky branch `more-gl-error-checks` built in `Debug` mode.

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9749550/log.txt)

10110111:
Looks like the error happens somewhere _before_ `AtmosphereShowMySky` is created. Could you build Stellarium in debug mode with the following patch applied?

```diff
diff --git a/src/core/modules/AtmosphereShowMySky.cpp b/src/core/modules/AtmosphereShowMySky.cpp
index df33afdf02..3999f3323f 100644
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -425,6 +425,7 @@ AtmosphereShowMySky::AtmosphereShowMySky()
     , viewRayGridBuffer(QOpenGLBuffer::VertexBuffer)
     , luminanceToScreenProgram_(new QOpenGLShaderProgram())
 {
+    StelOpenGL::checkGLErrors(__FILE__,__LINE__);
     indexBuffer.setUsagePattern(QOpenGLBuffer::StaticDraw);
     indexBuffer.create();
     viewRayGridBuffer.setUsagePattern(QOpenGLBuffer::DynamicDraw);
```

alex-w:
Debug mode: it just works (see [log.txt](https://github.com/Stellarium/stellarium/files/9751071/log.txt))

Release mode: crashes (see [log.txt](https://github.com/Stellarium/stellarium/files/9751097/log.txt))

10110111:
> Loading texture from "/Applications/Stellarium.app/Contents/Resources/atmosphere/default/light-pollution-xyzw.f32"... dimensions from header: 128×64... done
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model
> ShowMySky atmosphere model crashed: GL error on entry to virtual void AtmosphereRenderer::resizeEvent(int, int): Invalid value
> Loading Preetham model

This seems to happen due to dynamic resolution being on. Could you turn it off (and maybe even resolution reduction too)?

10110111:
> This seems to happen due to dynamic resolution being on.

Actually no, this is not the reason. The reason is the atmosphere switch animation. I'll have to redo it. Anyway, it'll still require several retests to debug.

alex-w:
```
[landscape]
atmosphere_resolution_reduction        = 4
flag_atmosphere_dynamic_resolution  = false
```

Release mode with proposed patch: error (see [log.txt](https://github.com/Stellarium/stellarium/files/9754009/log.txt))

alex-w:
```
[landscape]
atmosphere_resolution_reduction     = 4
flag_atmosphere_dynamic_resolution  = true
```

ShowMySky (master; release mode) + Stellarium (release mode) = SMS crash ([log.txt](https://github.com/Stellarium/stellarium/files/9754590/log.txt))

ShowMySky (master; release mode) + Stellarium (debug mode) =  Stellarium crash ([log.txt](https://github.com/Stellarium/stellarium/files/9754670/log.txt) + [dump.txt](https://github.com/Stellarium/stellarium/files/9754674/dump.txt))

ShowMySky (master; debug mode) + Stellarium (release mode) = SMS crash ([log.txt](https://github.com/Stellarium/stellarium/files/9754737/log.txt))

ShowMySky (master; debug mode) + Stellarium (debug mode) = all OK ([log.txt](https://github.com/Stellarium/stellarium/files/9754792/log.txt))

10110111:
Could you build release mode versions as `RelWithDebInfo` instead of `Release`?

alex-w:
> Could you build release mode versions as `RelWithDebInfo` instead of `Release`?

Yes, of course! Both compiled in `RelWithDebInfo` mode: [log.txt](https://github.com/Stellarium/stellarium/files/9754965/log.txt)

10110111:
Let's reduce the amount of test configurations for now:

* CalcMySky should be taken from `more-gl-error-checks` branch and compiled in Debug mode
* Stellarium should be compiled in Debug and RelWithDebInfo variants
* dynamic resolution should be off
* Scenery3d plugin should be disabled

Now, please try the current `gl33-core` Stellarium branch in these two configurations.

alex-w:
* [x] CalcMySky should be taken from `more-gl-error-checks` branch and compiled in _Debug_ mode
* [x] dynamic resolution should be off
* [x] Scenery3d plugin should be disabled
* [x] Stellarium should be compiled in _Debug_ variant - [log.txt](https://github.com/Stellarium/stellarium/files/9756479/log.txt)
* [x] Stellarium should be compiled in _RelWithDebInfo_ variant - [log.txt](https://github.com/Stellarium/stellarium/files/9756609/log.txt)

10110111:
So, do I understand it correctly: although we get lots of `GL_INVALID_VALUE` errors in the log, Stellarium appears to work fine?

alex-w:
> So, do I understand it correctly: although we get lots of `GL_INVALID_VALUE` errors in the log, Stellarium appears to work fine?

`RelWithDebInfo`: [log.txt](https://github.com/Stellarium/stellarium/files/9756908/log.txt)

stellarium-010

10110111:
What if you pass `-DCMAKE_CXX_FLAGS=-DQT_NO_DEBUG` and recompile Stellarium in `Debug` mode? This hybrid mode might shed some light on the reason why we don't get any errors in usual Debug mode while we do in Release modes.

alex-w:
> What if you pass `-DCMAKE_CXX_FLAGS=-DQT_NO_DEBUG` and recompile Stellarium in `Debug` mode? This hybrid mode might shed some light on the reason why we don't get any errors in usual Debug mode while we do in Release modes.

Please see results - first run (selecting ShowMySky model) ([log.txt](https://github.com/Stellarium/stellarium/files/9761470/log.txt)); second run (just startup with defined ShowMySky mode in config.ini) ([log.txt](https://github.com/Stellarium/stellarium/files/9761488/log.txt))

Screenshot from second run:
stellarium-011

alex-w:
Switching between models is OK now

10110111:
Aha, so this is why we don't get any errors in debug mode: they are explicitly hidden in two places. Please rebuild with the following patch, in complete `Debug` mode without `-DQT_NO_DEBUG`:
```diff
diff --git a/src/StelMainView.cpp b/src/StelMainView.cpp
index 4e58814d47..0357842a44 100644
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -368,11 +368,6 @@ protected:
         //important to call this, or Qt may have invalid state after we have drawn (wrong textures, etc...)
         painter->beginNativePainting();
 
-        //fix for bug LP:1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
-        StelOpenGL::clearGLErrors();
-#endif
-
         //update and draw
         StelApp& app = StelApp::getInstance();
         app.update(dt); // may also issue GL calls
diff --git a/src/core/StelPainter.cpp b/src/core/StelPainter.cpp
index c6550a02b2..5163ade32c 100644
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -808,11 +808,6 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
         //important to call this before GL state restore
         painter.end();
 
-        //fix for bug 1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
-        StelOpenGL::clearGLErrors();
-#endif
-
         //QPainter messes up some GL state, begin/endNativePainting or save/restore does not help
         glState.apply();
     }
```

10110111:
This doesn't make any sense to me.
OK, what does the following command print if you run it from the build directory, where `CMakeCache.txt` resides?
```
grep ^CMAKE_CXX_FLAGS CMakeCache.txt
```

alex-w:
```
CMAKE_CXX_FLAGS:STRING=
CMAKE_CXX_FLAGS_DEBUG:STRING=-g
CMAKE_CXX_FLAGS_MINSIZEREL:STRING=-Os -DNDEBUG
CMAKE_CXX_FLAGS_RELEASE:STRING=-O3 -DNDEBUG
CMAKE_CXX_FLAGS_RELWITHDEBINFO:STRING=-O2 -g -DNDEBUG
CMAKE_CXX_FLAGS-ADVANCED:INTERNAL=1
CMAKE_CXX_FLAGS_DEBUG-ADVANCED:INTERNAL=1
CMAKE_CXX_FLAGS_MINSIZEREL-ADVANCED:INTERNAL=1
CMAKE_CXX_FLAGS_RELEASE-ADVANCED:INTERNAL=1
CMAKE_CXX_FLAGS_RELWITHDEBINFO-ADVANCED:INTERNAL=1
```

alex-w:
```
+#define NDEBUG
 #ifndef NDEBUG
```

@10110111  Maybe you meant `#ifdef NDEBUG`?

10110111:
> Maybe you meant #ifdef NDEBUG?

No, this is explicitly a simulation of Release mode definitions in this particular file, while the rest is compiled in normal Debug mode. The ifndef should remain as it is.

alex-w:
Crash 
[log.txt](https://github.com/Stellarium/stellarium/files/9765084/log.txt)
[dump.txt](https://github.com/Stellarium/stellarium/files/9765093/dump.txt)

10110111:
So, looks like the difference in behavior is due to the difference in placement of `clearGLErrors()` calls in Debug and Release modes. Please try the following patch in Release mode:
```diff
diff --git a/src/StelMainView.cpp b/src/StelMainView.cpp
index 4e58814d47..d88ef3c7e0 100644
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -354,6 +354,7 @@ public:
 protected:
     virtual void paint(QPainter *painter, const QStyleOptionGraphicsItem *option, QWidget *widget) Q_DECL_OVERRIDE
     {
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         Q_UNUSED(option)
         Q_UNUSED(widget)
 
@@ -368,10 +369,9 @@ protected:
         //important to call this, or Qt may have invalid state after we have drawn (wrong textures, etc...)
         painter->beginNativePainting();
 
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         //fix for bug LP:1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
         StelOpenGL::clearGLErrors();
-#endif
 
         //update and draw
         StelApp& app = StelApp::getInstance();
diff --git a/src/core/StelPainter.cpp b/src/core/StelPainter.cpp
index c6550a02b2..52cdc900e8 100644
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -44,6 +44,7 @@
 
 static const int TEX_CACHE_LIMIT = 7000000;
 
+#define NDEBUG
 #ifndef NDEBUG
 QMutex* StelPainter::globalMutex = new QMutex();
 #endif
@@ -767,6 +768,8 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
     }
     else
     {
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
+
         QOpenGLPaintDevice device;
         device.setSize(QSize(prj->getViewportWidth(), prj->getViewportHeight()));
         // This doesn't seem to work correctly, so implement the hack below instead.
@@ -808,10 +811,9 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
         //important to call this before GL state restore
         painter.end();
 
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         //fix for bug 1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
         StelOpenGL::clearGLErrors();
-#endif
 
         //QPainter messes up some GL state, begin/endNativePainting or save/restore does not help
         glState.apply();
```

alex-w:
It works in Release mode [log.txt](https://github.com/Stellarium/stellarium/files/9770040/log.txt)

10110111:
OK, lots of errors are visible now. What if you try the same patch on `master`?

alex-w:
> OK, lots of errors are visible now. What if you try the same patch on `master`?

No errors, except ShowMySky: [log.txt](https://github.com/Stellarium/stellarium/files/9776047/log.txt)

10110111:
Let's see which modules generate these errors. Please try the following patch in Release mode on `gl33-core` branch.
```diff
diff --git a/src/StelMainView.cpp b/src/StelMainView.cpp
index 4e58814d47..d88ef3c7e0 100644
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -354,6 +354,7 @@ public:
 protected:
     virtual void paint(QPainter *painter, const QStyleOptionGraphicsItem *option, QWidget *widget) Q_DECL_OVERRIDE
     {
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         Q_UNUSED(option)
         Q_UNUSED(widget)
 
@@ -368,10 +369,9 @@ protected:
         //important to call this, or Qt may have invalid state after we have drawn (wrong textures, etc...)
         painter->beginNativePainting();
 
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         //fix for bug LP:1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
         StelOpenGL::clearGLErrors();
-#endif
 
         //update and draw
         StelApp& app = StelApp::getInstance();
diff --git a/src/core/StelApp.cpp b/src/core/StelApp.cpp
index 455212f07b..5b507f0cb1 100644
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -809,6 +809,7 @@ void StelApp::draw()
     if (!initialized)
         return;
 
+    StelOpenGL::checkGLErrors(__FILE__,__LINE__);
     //find out which framebuffer is the current one
     //this is usually NOT the "zero" FBO, but one provided by QOpenGLWidget
     GLint drawFbo;
@@ -822,7 +823,10 @@ void StelApp::draw()
     const QList<StelModule*> modules = moduleMgr->getCallOrders(StelModule::ActionDraw);
     for (auto* module : modules)
     {
+        qDebug().nospace().noquote() << "\n" << module->objectName() << "::draw()";
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         module->draw(core);
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
     }
     core->postDraw();
 #ifdef ENABLE_SPOUT
@@ -831,6 +835,8 @@ void StelApp::draw()
         spoutSender->captureAndSendFrame(static_cast<GLuint>(drawFbo));
 #endif
     applyRenderBuffer(static_cast<GLuint>(drawFbo));
+
+    StelOpenGL::checkGLErrors(__FILE__,__LINE__);
 }
 
 /*************************************************************************
diff --git a/src/core/StelPainter.cpp b/src/core/StelPainter.cpp
index c6550a02b2..92b9f77eae 100644
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -767,6 +767,8 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
     }
     else
     {
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
+
         QOpenGLPaintDevice device;
         device.setSize(QSize(prj->getViewportWidth(), prj->getViewportHeight()));
         // This doesn't seem to work correctly, so implement the hack below instead.
@@ -808,10 +810,9 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
         //important to call this before GL state restore
         painter.end();
 
+        StelOpenGL::checkGLErrors(__FILE__,__LINE__);
         //fix for bug 1628072 caused by QTBUG-56798
-#ifndef QT_NO_DEBUG
         StelOpenGL::clearGLErrors();
-#endif
 
         //QPainter messes up some GL state, begin/endNativePainting or save/restore does not help
         glState.apply();
```

alex-w:
Weird results: [log.txt](https://github.com/Stellarium/stellarium/files/9777582/log.txt)

10110111:
Nothing weird, quite fine. I've pushed a patch to look deeper, please try it in Release mode.

alex-w:
> Nothing weird, quite fine. I've pushed a patch to look deeper, please try it in Release mode.

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9778215/log.txt)

10110111:
We are getting close. I've pushed more checks, please retest.

gzotti:
Great to read the progress. I am currently not available. Hopefully testing my computer zoo again on Sunday.

10110111:
@gzotti your testing is not required yet. I've seen from one your report that VAO support may be unavailable on the target platforms, and this problem is still not addressed. What's currently being done is just a chase for OpenGL errors on macOS.

alex-w:
> We are getting close. I've pushed more checks, please retest.

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9782211/log.txt)

10110111:
Funny, it didn't clarify anything. Apparently, the problem is elsewhere. I've pushed another patch, please retest.

10110111:
BTW, do constellation lines look as they should?

alex-w:
> BTW, do constellation lines look as they should?

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9784883/log.txt)

stellarium-012

10110111:
OK, I see. `glLineWidth` with argument >1 is unsupported in Core profile. Do you get wider lines with `--opengl-compat` option?

10110111:
I've stubbed the offending call, it shouldn't generate these `GL_INVALID_VALUE` errors now. Please retest and in particular try atmosphere switching. I'll try to chase down the origin of the framebuffer errors.

alex-w:
Model switching is OK: [log.txt](https://github.com/Stellarium/stellarium/files/9788472/log.txt)

10110111:
Please retest model switching with the new commit.

alex-w:
> OK, I see. `glLineWidth` with argument >1 is unsupported in Core profile.

This is bad news. This feature is used for demonstrating constellations and asterisms via projectors - how we can change width of line in Core profile? Or can we replace `glLineWidth`?

10110111:
> how we can change width of line in Core profile? Or can we replace glLineWidth?

The general recommendation is to draw quads instead of lines. This makes it a bit harder to make them smooth (must rely on multisampling, which is currently disabled by default in Stellarium on all platforms). OTOH, if we enable it by default, using a moderate number of samples per pixel like 8, this will make the results a bit more correct when lines intersect—compared to Wu-like algorithms, which don't care about already present primitives while rendering a new one.

10110111:
> Where you found limitations for glLineWidth?

OpenGL 3.3 Core specification, chapter E.2.1.

gzotti:
> > Where you found limitations for glLineWidth?
> 
> OpenGL 3.3 Core specification, chapter E.2.1.

Yes, I remember also having read that, and it may have been one of the reasons not to invest much time going for Core profile just because it was deemed "cleaner". Also GLES (and with it, Qt5/ANGLE) does not have it, but these are IMHO indeed systems of less importance. On "genuine" OpenGL systems I would still like to see this feature preserved, i.e. Compatibility profile must remain at least as option on systems where it is principally available. (Macs differ, as always...) Or the feature implemented as thin quads, indeed. (As method in StelPainter, with simple to adapt code! This could then bring back dotted lines etc...)

alex-w:
> Please retest model switching with the new commit.

First run (switching to ShowMySky model): [log.txt](https://github.com/Stellarium/stellarium/files/9791319/log.txt)

Second run (use ShowMySky model at startup): [log.txt](https://github.com/Stellarium/stellarium/files/9791326/log.txt)

P.S. Both are OK

10110111:
Please re-run with the updated checks.

10110111:
> On "genuine" OpenGL systems I would still like to see this feature preserved

What is it used, aside from widening the lines for high-DPI scaling?

> This could then bring back dotted lines etc.

When were they in use and for what? Why were they dropped?

gzotti:
> > On "genuine" OpenGL systems I would still like to see this feature preserved
> 
> What is it used, aside from widening the lines for high-DPI scaling?

Not necessarily for high-DPI. I like to have wider lines (2-3) on FullHD. 

> 
> > This could then bring back dotted lines etc.
> 
> When were they in use and for what? Why were they dropped?

I think we tried them for borders or asterisms, but line style is also not available on ANGLE (OpenGL ES2.0), so it was no use. I read somewhere that thick lines or other line style has to be implemented in a shader, and did not continue investigating as "line thickness" seemed good enough, and other things were more important problems to solve.

10110111:
@gzotti How do you get ANGLE to work? Do you use the dedicated shortcut? What happens if you run the "normal", non-ANGLE, shortcut on your Sony laptop?

Why I'm asking is because the mechanism that's supposed to enable ANGLE doesn't work for me via `--angle-mode` option and its siblings, because they set the `QT_OPENGL` environment variable _after_ `QGuiApplication` is created, while it's this class that reads this environment variable. Setting `QT_OPENGL` manually does work.

gzotti:
> @gzotti How do you get ANGLE to work (on `master`)? Do you use the dedicated shortcut? What happens if you run the "normal", non-ANGLE, shortcut on your Sony laptop?

Without the environment variable set to ANGLE (and setting it to desktop), the Sony just reports insufficient OpenGL level. Not sure if it just quits then, can try tomorrow. The bad thing its, it was crippled by Sony's unwillingness to support their products when still technically OK. The laptop was officially not supported with Win10, however, somebody found a way to keep the Intel/Geforce hybrid system working with manually patched driver .inf files. Just the switch did not work any longer, and it needed a reboot to actually switch GPUs. I could live with that. But some Win10 version update broke the Geforce driver again, and that driver hack website has unfortunately vanished. OK, bad luck. But still, this is a Core-i7 of the first generation and else a nice little system to use, and I am pretty sure there are still millions of those CPUs around. This is a GPU that was blacklisted by Qt to immediately initialize to ANGLE with Qt5. On Qt6, it runs Mesa mode, i.e. noticeably much lower fps, and also just OpenGL3.0. 

 
> Why I'm asking is because the mechanism that's supposed to enable ANGLE doesn't work for me via `--angle-mode` option and its siblings, because they set the `QT_OPENGL` environment variable _after_ `QGuiApplication` is created, while it's this class that reads this environment variable. Setting `QT_OPENGL` manually does work.

Yes, the setting of the env vars has worked earlier (when I introduced them years ago), but at some point something changed, and the --angle-mode etc no longer work. See wiki entries about "Common Problems" on Github.

alex-w:
> Please re-run with the updated checks.

Please see [log.txt](https://github.com/Stellarium/stellarium/files/9793830/log.txt)

alex-w:
> > On "genuine" OpenGL systems I would still like to see this feature preserved
> 
> What is it used, aside from widening the lines for high-DPI scaling?

Not only HDPI systems - for projectors it's necessary too.

> 
> > This could then bring back dotted lines etc.
> 
> When were they in use and for what? Why were they dropped?

See screenshot from version 0.11.4:
stellarium-000

In series 0.12.x rendering engine was seriously updated and we "lost" rendering the stylized lines.

10110111:
Please retest with updated CalcMySky `master`.

alex-w:
> Please retest with updated CalcMySky `master`.

First run (switching to ShowMySky model): [log.txt](https://github.com/Stellarium/stellarium/files/9794290/log.txt)

Second run (use ShowMySky model at startup): [log.txt](https://github.com/Stellarium/stellarium/files/9794292/log.txt)

P.S. Both are compiled in `Release` mode and both are OK

10110111:
Great. So the remaining problems now are the `glLineWidth` and VAO@ANGLE.

@gzotti Am I right that in ANGLE mode dithering doesn't work? I've just tried it via Wine and was surprised to see banded image.

gzotti:
Hmm, dithering had ugly banding on ANGLE (but not on my GLES2 Odroid (Mali)!) before you changed it to the new algorithm. Must check again, just wanted to dig out my Vaio again.... BTW the i5-4570 also had no VAO I think.

10110111:
> BTW the i5-4570 also had no VAO I think.

Not so simple: if you judged from the logs, this may be misleading because Stellarium used to have Intel explicitly blacklisted for VAO (which makes no sense for Macs of course).

gzotti:
Here are logs from the Vaio. I built without SMS as we won't get it anyhow. Sky looks good, but we have errors in the log. 

[log_i7-M640_OGL3branch1016_qt5-angle-skyOK-but-OGLerrors.txt](https://github.com/Stellarium/stellarium/files/9795265/log_i7-M640_OGL3branch1016_qt5-angle-skyOK-but-OGLerrors.txt)
[log_i7-M640_master1016_qt5-angleDX11.txt](https://github.com/Stellarium/stellarium/files/9795266/log_i7-M640_master1016_qt5-angleDX11.txt)
[log_i7-M640_0.22.3-AngleDX9.txt](https://github.com/Stellarium/stellarium/files/9795267/log_i7-M640_0.22.3-AngleDX9.txt)
[log_i7-M640_0.21.3-AngleDX11.txt](https://github.com/Stellarium/stellarium/files/9795268/log_i7-M640_0.21.3-AngleDX11.txt)
[log_i7-M640_0.22.3-AngleDX11.txt](https://github.com/Stellarium/stellarium/files/9795269/log_i7-M640_0.22.3-AngleDX11.txt)

10110111:
@gzotti Now both in `master` and in this branch the options like `--mesa-mode` etc. should work.

gzotti:
Yes, the early setting of angle envvars works again! OTOH, on the Valo, I now see neither daylight sky nor stars! Moon, planets MilkyWay, DSOs are OK, 
[log_i7-M640_OGL3branch1016_qt5-angleOK_butNoSkyNoStars.txt](https://github.com/Stellarium/stellarium/files/9795809/log_i7-M640_OGL3branch1016_qt5-angleOK_butNoSkyNoStars.txt)

10110111:
> I now see neither daylight sky nor stars

Yup, this is expected. It's due to the unconditional use of VAO.

gzotti:
> > I now see neither daylight sky nor stars
> 
> Yup, this is expected. It's due to the unconditional use of VAO.

Note that running "pure" software GL (MESA) on Windows (OpenGL 3.0) currently has fully black screen.

[log_master_MESA.txt](https://github.com/Stellarium/stellarium/files/9795913/log_master_MESA.txt)

10110111:
> Note that running "pure" software GL (MESA) on Windows (OpenGL 3.0) currently has fully black screen.

I wonder why Qt provides such an ancient version of Mesa: 11.2.2 was released in 2016 and the project has since grown to [support OpenGL 4.6](https://mesamatrix.net/#OpenGL) in its llvmpipe renderer.

> I now see neither daylight sky nor stars.

I've pushed a fix for stars :)
Night mode and atmosphere still have to be fixed.

gzotti:
+1 on ancient Mesa version. We should probably build a newer version from the llvmpipe site.

10110111:
OK, so now it should work in all cases:
* Windows, Linux, macOS with ShowMySky
* Windows+ANGLE without ShowMySky
* Windows+Mesa without ShowMySky (due to very old Mesa version)

What I didn't test is all this with Qt6.
What remains unsolved is wide lines.

Dithering should work universally well (even on `master`, I've pushed the fix there too). If the sky is banded by default in some configuration, it's a bug.

gzotti:
OK, works on my old Vaio (Qt5/Angle). Dithering is very good now.

gzotti:
> +1 on ancient Mesa version. We should probably build a newer version from the llvmpipe site.

This may be useful:
https://wiki.qt.io/MesaLlvmpipe

else
https://github.com/pal1000/mesa-dist-win
https://docs.mesa3d.org/drivers/llvmpipe.html

10110111:
So now all the problems have been solved. Wide lines in the Core profile are implemented using a geometry shader.

Now this PR needs thorough testing on all the configurations that Stellarium intends to support.

alex-w:
macOS/Qt6 - all works OK (ShowMySky + line width)

gzotti:
Main system (Win10/Geforce 960M) starts with OpenGL3.3Core. Thank you for the wide lines, but they now look aliased. Where did we have the multisampling, any manual entry in config.ini? 
Also, it is now possible to extend the wide lines to OpenGL ES2 systems? (I did not check, but see shader version #330. 

I switched on Scenery3D. Big issue: Stars are missing. Planets are still here.

Another big issue: After once switching on a scene in Scenery3D, the sky sometimes shows "slices" in the dome. Interestingly enough, this is only in live view. A screenshot has them covered. A screenshot with another program shows it (around the moon):
grafik

These slices also remain when switching back to Preetham. 

[log_Geforce960M_OpenGL3.txt](https://github.com/Stellarium/stellarium/files/9808506/log_Geforce960M_OpenGL3.txt)

alex-w:
macOS/Qt5 - fails - [log.txt](https://github.com/Stellarium/stellarium/files/9808543/log.txt)

10110111:
> Where did we have the multisampling, any manual entry in config.ini?

`[video]/multisampling`

> Also, it is now possible to extend the wide lines to OpenGL ES2 systems?

Not without more work. [`GL_EXT_geometry_shader`](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_geometry_shader.txt) requires GL ES 3.1. If not via geometry shader, this will require pre-processing of the geometry on the CPU.

> I switched on Scenery3D. Big issue: Stars are missing.

So the stars were there before you enabled Scenery3d, and then disappeared?

Was all this in Core profile desktop GL? What if you use `--opengl-compat` option?

gzotti:
> > Also, it is now possible to extend the wide lines to OpenGL ES2 systems?
> 
> Not without more work. [`GL_EXT_geometry_shader`](https://registry.khronos.org/OpenGL/extensions/EXT/EXT_geometry_shader.txt) requires GL ES 3.1. If not via geometry shader, this will require pre-processing of the geometry on the CPU.

OK, not critically important. 
 
> > I switched on Scenery3D. Big issue: Stars are missing.
> 
> So the stars were there before you enabled Scenery3d, and then disappeared?

Yes
 
> Was all this in Core profile desktop GL? What if you use `--opengl-compat` option?

Same behaviour unfortunately. It starts with SMS sky and stars, but when I activate Scenery3D, stars vanish. And the sky shows these strange cracks.

[log_Geforce960M_OpenGLcompat.txt](https://github.com/Stellarium/stellarium/files/9809899/log_Geforce960M_OpenGLcompat.txt)

10110111:
Do the problems with Scenery3d happen if you start with Preetham atmosphere and never switch to ShowMySky?

gzotti:
> Do the problems with Scenery3d happen if you start with Preetham atmosphere and never switch to ShowMySky?

Uh, yes! Tested in Compat mode and (new) standard Core3.3. The stars remain off even when scenery3D is then again disabled.

10110111:
> Uh, yes! Tested in Compat mode and (new) standard Core3.3.

OK, then it'd be very helpful if you could run `git bisect` testing between 441061577482a17cda6849cfb2515da8ebc0e170 and 649de9cd9e7ed825a802b67e804d2200cadce8de to determine which commit broke it. This will require about 3 rebuilds. You'll need to pass `--opengl-compat` to stellarium each time, because only the last commit covers all requirements of the Core profile.

(If you don't know how to use git-bisect, see [this Wine tutorial](https://wiki.winehq.org/Regression_Testing#Running_the_bisection) on regression testing.)

gzotti:
Thanks for the tutorial, I was really not aware of this command (but used a similar strategy manually earlier).

grafik

gzotti:
Patch over the bad commit or over current head?

gzotti:
[log_M960_3.3core_patched.txt](https://github.com/Stellarium/stellarium/files/9814312/log_M960_3.3core_patched.txt)
[log_M960_4.6compat_patched.txt](https://github.com/Stellarium/stellarium/files/9814313/log_M960_4.6compat_patched.txt)
stellarium-028

10110111:
Does anything change visually if you apply this (run in compat mode)?
```diff
diff --git a/plugins/Scenery3d/src/S3DRenderer.cpp b/plugins/Scenery3d/src/S3DRenderer.cpp
index fe23ad131b..76a84f4ff2 100644
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -1205,7 +1205,6 @@ void S3DRenderer::drawFromCubeMap()
     //transform vertices on CPU side - maybe we could do this multithreaded, kicked off at the beginning of the frame?
     altAzProjector->project(cubeVertices.count(),cubeVertices.constData(),transformedCubeVertices.data());
 
-    vao.bind();
     //setup shader params
     projectionMatrix = altAzProjector->getProjectionMatrix().convertToQMatrix();
     cubeShader->setUniformValue(shaderManager.uniformLocation(cubeShader,ShaderMgr::UNIFORM_MAT_PROJECTION), projectionMatrix);
```

gzotti:
Yay!
stellarium-030
[log.txt](https://github.com/Stellarium/stellarium/files/9814866/log.txt)

10110111:
That's actually bad :)
This line was necessary for Core profile. I wonder why it breaks things on your machine, because I failed to reproduce it on my machines, including an NVIDIA ION-based one (EEEPC 1015PN).

10110111:
Could you post your `config.ini`? Maybe some things are enabled that trigger this problem...

gzotti:
[config - M960.ini.txt](https://github.com/Stellarium/stellarium/files/9815112/config.-.M960.ini.txt)

10110111:
@alex-w 
> macOS/Qt5 - fails

Does `showmysky` tool also fail in a similar way?

10110111:
@gzotti Please try the following patch in Core mode.
```diff
diff --git a/plugins/Scenery3d/src/S3DRenderer.cpp b/plugins/Scenery3d/src/S3DRenderer.cpp
index fe23ad131b..892169d9fb 100644
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -1257,11 +1257,11 @@ void S3DRenderer::drawFromCubeMap()
     }
     cubeIndexBuffer.release();
 
-    vao.release();
-
     cubeShader->disableAttributeArray(StelOpenGLArray::ATTLOC_TEXCOORD);
     cubeShader->disableAttributeArray(StelOpenGLArray::ATTLOC_VERTEX);
 
+    vao.release();
+
     glDisable(GL_CULL_FACE);
     glDisable(GL_DEPTH_TEST);
     glDisable(GL_BLEND);
```

gzotti:
No, doesn't work. If I only change these lines, the stars are gone. If  also comment away the vao.bind() above, stars are here but the scenery is not visible. 

Sorry, I must quit for today.  Good night.

10110111:
I'm pretty sure that it's a driver bug, and I'm having a hard time trying to work around it. Could you try upgrading the driver to version 522.25 that nvidia.com provides for GeForce 960M?

gzotti:
New drivers, same problem. No stars after switching on Scenery3D. I have not applied any patches to current branch head. 
[log.txt](https://github.com/Stellarium/stellarium/files/9819451/log.txt)
[log_522_Compat46.txt](https://github.com/Stellarium/stellarium/files/9819570/log_522_Compat46.txt)

I will test on an older Geforce 580M which used to be a "great" system.

gzotti:
Bad news. Geforce 580M, latest drivers. 
[log_580M_3.3core.txt](https://github.com/Stellarium/stellarium/files/9820164/log_580M_3.3core.txt)
[log_580M_4.6compat.txt](https://github.com/Stellarium/stellarium/files/9820165/log_580M_4.6compat.txt)

Both have vanishing stars on activation of scenery3D, and show the "cracks" in the sky.
 
stellarium-001

10110111:
Interesting, here we do have the scenery, although the stars become stripes. Do you observe this problem in Mesa mode?

10110111:
Also, what happens if you zoom into these stripes? How do they look when magnified?

alex-w:
> Does `showmysky` tool also fail in a similar way?

No, `showmysky` tool works as expected

gzotti:
These stripes are not related to the stars AFAICS. These are the strange cracks I was mentioning. 
The are fast to appear and vanish, I cannot zoom in on them, they look like unclosed gaps. However, I see none of our  coordinate grid that they would follow. 

https://user-images.githubusercontent.com/18099873/196703029-309e0e90-656a-400b-80ce-2eb400c790ae.mp4

Given that Scenery3D makes a huge difference between Perspective and the other projections, I tried with perspective. Even here the cracks show. When I move the scene with the mouse, the cracks move a bit, then vanish and appear elsewhere.

gzotti:
New test on my Intel i5-4570. Core3.3 profile. The same behaviour: stars vanish on activating Scenery3D. And the cracks. Here I move the scene and then time.

https://user-images.githubusercontent.com/18099873/196707312-a206fa09-8f2b-4427-ac07-f84d075d5aa5.mp4

The compat run shows just a black screen :-( Log says 3.3 compat. The hardware earlier (V1.22.3) reported 4.3 compat. 

[log_i5-4570_OGLcompat_black.txt](https://github.com/Stellarium/stellarium/files/9821016/log_i5-4570_OGLcompat_black.txt)
[log_i5-4570_3.3core.txt](https://github.com/Stellarium/stellarium/files/9821018/log_i5-4570_3.3core.txt)

10110111:
I've finally reproduced it on an Intel GPU. This problem only happens in Windows Qt6 version but not Qt5 (tested 32-bit), and these stripes are actually stars: they disappear when I press <kbd>S</kbd> and reappear on a repeated press. Don't know the reason yet, still investigating.

gzotti:
Uh, indeed. No problem on 960M with Qt5/3.3Core (64bit).  The cracks indeed appear in the stars pass, but they are too regular, geometrically speaking, to be stars.

10110111:
From what I've studied, `glGenVertexArrays` appears to return the same name twice in a row. Between these two calls (one in `StelSkyDrawer` as called from `StelApp::init`, another in `S3DRenderer` as called from `StelApp::initPlugIns`) there is a switch of OpenGL context (via `wglMakeCurrent`).

Why this happens is still to be determined. I suspect the manipulations of fullscreen etc. state between these two initializations lead to this context switch. (`StelMainView::init()` is a mess.) To check that I'll need another CI recompilation which takes 40 minutes, so going AFK for the night.

10110111:
Now the disappearing stars should be fixed. This commit, d90b5b10a50c8b235d1f48ef65b3006707a5ea5e, is not a nice solution, but it will work for now.

If done properly, it probably would affect #2755 and maybe something else (there are comments about previously applied kludges for other issues).

alex-w:
> Now the disappearing stars should be fixed. This commit, [d90b5b1](https://github.com/Stellarium/stellarium/commit/d90b5b10a50c8b235d1f48ef65b3006707a5ea5e), is not a nice solution, but it will work for now.
> 
> If done properly, it probably would affect #2755 and maybe something else (there are comments about previously applied kludges for other issues).

Notification in the log:
```
Dubious result: Landscape  "Guereins"  not calibrated. Opacity test represents mathematical horizon only.
Initializing planets GL shaders... 
"StelPainter: Warnings while linking transformShaderProgram shader program:\nWARNING: Could not find vertex shader attribute 'texCoord' to match BindAttributeLocation request.\nWARNING: Could not find vertex shader attribute 'normalIn' to match BindAttributeLocation request.\n"
/Users/aw/stellarium/src/core/modules/AtmosphereShowMySky.cpp:606: OpenGL error: 1282 (GL_INVALID_OPERATION)
```

10110111:
> `/Users/aw/stellarium/src/core/modules/AtmosphereShowMySky.cpp:606: OpenGL error: 1282 (GL_INVALID_OPERATION)`

It would be extremely useful if you could build a patched debug version of Qt, with the following patch, and then run Stellarium with it to reproduce this error. This way we'll know the exact stack trace of where the error originates.

Otherwise we'll have to once again locate the origin via manual ping-pong of posts here.

```diff
diff --git a/qtbase/src/gui/opengl/qopenglfunctions.h b/qtbase/src/gui/opengl/qopenglfunctions.h
index c288df53b5..7bb317fd68 100644
--- a/qtbase/src/gui/opengl/qopenglfunctions.h
+++ b/qtbase/src/gui/opengl/qopenglfunctions.h
@@ -54,7 +54,7 @@
 #include <QtGui/qopengl.h>
 #include <QtGui/qopenglcontext.h>
 
-//#define Q_ENABLE_OPENGL_FUNCTIONS_DEBUG
+#define Q_ENABLE_OPENGL_FUNCTIONS_DEBUG
 
 #if QT_CONFIG(opengles2)
 typedef double GLdouble;
@@ -68,7 +68,7 @@ typedef double GLdouble;
         unsigned clamped = qMin(unsigned(error - GL_INVALID_ENUM), 4U); \
         const char *errors[] = { "GL_INVALID_ENUM", "GL_INVALID_VALUE", "GL_INVALID_OPERATION", "Unknown" }; \
         printf("GL error at %s:%d: %s\n", __FILE__, __LINE__, errors[clamped]); \
-        int *value = nullptr; \
+        int *volatile value = nullptr; \
         *value = 0; \
     }
 #else
```

gzotti:
Works on my oldest laptop, with Geforce 9800M GTX, Ubuntu 22.04.1. 
[log.txt](https://github.com/Stellarium/stellarium/files/9834389/log.txt)

For whatever reasons, building with Media support fails. Some egl functions cannot be found by the linker. Maybe I am just missing a package.   This is unrelated to the PR, though.

gzotti:
New test on my Intel i5-4570. Qt6.4, Core3.3 profile. Yes, works!
[log.txt](https://github.com/Stellarium/stellarium/files/9834753/log.txt)

There is still an GL_INVALID_OPERATION in the logfile.

gzotti:
Looks good (also with SMS) on 580M. 
[log_580M_3.3core_SMS.txt](https://github.com/Stellarium/stellarium/files/9838078/log_580M_3.3core_SMS.txt)

10110111:
> There is still an GL_INVALID_OPERATION in the logfile.

Does it happen in `master`?

10110111:
Invalid operation on Intel is now fixed (both in `master` and here). It was due to a mistake that was introduced at the very inception of `StelOpenGLArray` class, but was unnoticed due to zero VAO being a "default VAO" in Compatibility profile.

@alex-w I don't know whether it'll fix the error you see on macOS. If not, this will require some debugging, ideally using the patch for Qt that I posted above.

alex-w:
The error is gone

10110111:
OK, so this PR seems done now, including doc update. Is there any testing left to do, or shall I merge it?

gzotti:
I just wanted to test 960M and then startup my Intel 4570 one more time. And yes, I must test the Raspberries 3+4 for any change/break. Sorry, please don't merge yet.

gzotti:
Trying again to buils SMS on the Intel 4570. Your previous instructions were:

```
cmake -G "Visual Studio 16 2019" -DCMAKE_CXX_FLAGS="/ID:/StelDev/GIT/glm" -DCMAKE_PREFIX_PATH="D:/StelDev/GIT/eigen/share/eigen3/cmake" -DCMAKE_INSTALL_PREFIX=../CalcMySky_install ../CalcMySky -DQT_VERSION=6
cmake --build . --config Release
cmake --build . --config Release --target install
cp -rv ../CalcMySky/examples ../install/CalcMySky
```

I used the usual Windows command line here. 
The first `cmake --build` fails (see screenshots far above). The file `version_autogen.vcxproj` includes four lines with "true" statements. After deleting the lines, build succeeds with just the previous warnings (yellow lines in console.)

10110111:
Yup, broken `/bin/sh`, as we discussed at that point.

gzotti:
Found again on Raspberry 3 (see above, 12 days ago): 

GL_DOUBLE is `#define`d to GL_FLOAT in StelOpenGL.hpp. 
In StelPainter, please add this `#if` bracket. 

```
size_t StelPainter::ArrayDesc::vertexSizeInBytes() const
...
#if GL_DOUBLE != GL_FLOAT
	case GL_DOUBLE:
		elemSize = sizeof(GLdouble);
		break;
#endif

```

gzotti:
Now I have built SMS on my Intel i5-4570. Slow but pretty. 
[log_i5-4570_33core_OK.txt](https://github.com/Stellarium/stellarium/files/9842913/log_i5-4570_33core_OK.txt)

The compat mode still fails, though.
[log_i5-4570_compat_black.txt](https://github.com/Stellarium/stellarium/files/9842914/log_i5-4570_compat_black.txt)

This is probably not a blocker as long as the (now-default!) Core profile works reliably.

10110111:
> The compat mode still fails, though

Fails in what way? Only the errors that we see in the log, or some more noticeable problems?

gzotti:
Yes, black screen...

10110111:
Well, you can try running it via `apitrace` and see what triggers these errors. the `GL_INVALID_ENUM` is quite a puzzling one, especially when it appears in Compatibility profile but not in Core.

gzotti:
Raspi 3 is OK, just that it severely lacks texture memory. Is there a way to reduce the new button icons on loading for GLES systems (or with some other criterion?)? Maybe not this PR, but I think the pretty buttons waste memory here.

10110111:
> Is there a way to reduce the new button icons on loading for GLES systems (or with some other criterion?)?

Maybe. I feel all this will finally lead to a split of the graphics engine into two: one modern and the other for legacy/small systems.

10110111:
> it severely lacks texture memory

How did you diagnose it? `GL_OUT_OF_MEMORY` errors?

gzotti:
It behaves weird, program fails after a while, and dmesg messages indicates something that Google resolves to graphics memory issues. Earlier this happened with DSS or HiPS when zooming in, but now also when just loading more DSO images. Of course, also RPi is not free from error, but if it's just that little texture memory, we should just keep its use low on this platform. No DSS/HiPS or 3D sceneries is probably OK, and maybe some DSO images could also be reduced here.

gzotti:
RPi4 looks OK.
[log.txt](https://github.com/Stellarium/stellarium/files/9843130/log.txt)

gzotti:
I have no more objections. @alex-w ?

alex-w (approved):
ShowMySky works on macOS

@10110111 thank you for great work!

10110111:
Well, GitHub, that was stupid.

PR #2742, opened 2022-10-09
Replace world map with a higher-resolution version

10110111:
### Description
The current world map in the _Location_ dialog has very poor resolution by today's standards: 500×250 pixels for the whole globe. This makes it useless to resize the _Location_ dialog to see the map closer.

This patch replaces the old map with a 5400×2700 version, taken from NASA [Blue Marble Next Generation dataset for July 2004](https://visibleearth.nasa.gov/images/74092/july-blue-marble-next-generation). As I understand, NASA imagery is in public domain, so copyright issues shouldn't arise.

The water here is marked with a darker color, but I think it's not critical.

### Screenshots

Old map:

old

New map:

new


alex-w (changes requested):
Please update section Graphics in CREDITS.md file

alex-w (approved):

gzotti:
While I agree on having a higher-res texture in principle, this will probably not work on systems with limited texture size. I have not tried yet, but OpenGL ES systems, esp. SBC like RPi3, may be limited to 2048x2048px (not sure about ANGLE). See my RPi3 remarks in #2729 : Probably even the larger button icons cause an issue by filling texture memory, so there is no more left for sky textures. I cannot say what the solution is, having some tool to reduce all textures after installation? Or adding code into the texture loader to reduce texture sizes on such systems?

10110111:
> this will probably not work on systems with limited texture size

`miscWorldMap.*` is only used via `QPixmap::scaled` and `QPainter`, so it shouldn't lead to any problems with texture size limit. Until we really find problems, it's just speculation.

PR #2754, opened 2022-10-17
Add abort telescope slew signals to LX200 and NexStar telescopes

gzotti:
### Description

I found that most telescopes are missing the AbortSlew implementation. This is severely bad. This PR should add the signals at least for LX200 and NexStar. 

However, I cannot test it, so this needs external help. I unassigned myself for now in the hope one of our active observers steps in. It may just take a rainy weekend or two. Of course, support for a Big Red Button (emergency switch, separate hardware) would also be nice!

Fixes # (issue)

- [x] Add AbortSlew to LX200
- [x] Add AbortSlew to NexStar
- [ ] Add graceful recovery (put back marker) to LX200
- [ ] Add graceful recovery (put back marker) to NexStar

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 or any other
* Graphics Card: irrelevant

- [ ] Tested AbortSlew for LX200
- [ ] Tested AbortSlew for NexStar

## Checklist:

### DIFF
--- a/plugins/TelescopeControl/src/common/Socket.cpp
+++ b/plugins/TelescopeControl/src/common/Socket.cpp
@@ -40,7 +40,7 @@ long long int GetNow(void)
 		qint64 t;
 	} tmp;
 	GetSystemTimeAsFileTime(&tmp.file_time);
-	t = (tmp.t/10) - 86400000000LL*(369*365+89);
+	t = (tmp.t/10) - 86400000000LL*(369*365+89); // TODO: Explain this magic please!
 #else
 	struct timeval tv;
 	gettimeofday(&tv, 0);
--- a/plugins/TelescopeControl/src/common/Socket.hpp
+++ b/plugins/TelescopeControl/src/common/Socket.hpp
@@ -77,6 +77,8 @@ static inline int SetNonblocking(int s)
 
 #endif //Q_OS_WIN
 
+// retrieve system time in milliseconds
+// TODO: Specify epoch?
 long long int GetNow(void);
 
 class Server;
--- a/plugins/TelescopeControl/src/ASCOM/TelescopeClientASCOM.hpp
+++ b/plugins/TelescopeControl/src/ASCOM/TelescopeClientASCOM.hpp
@@ -37,6 +37,7 @@ public:
 	void telescopeSync(const Vec3d& j2000Pos, StelObjectP selectObject) override;
 	bool isTelescopeSyncSupported() const override {return true;}
 	void telescopeAbortSlew() override;
+	bool isAbortSlewSupported() const override {return true;}
 	bool isConnected() const override;
 	bool hasKnownPosition() const override;
 	static bool useJNow(ASCOMDevice::ASCOMEquatorialCoordinateType coordinateType, bool ascomUseDeviceEqCoordType, TelescopeControl::Equinox equinox);
--- a/plugins/TelescopeControl/src/Lx200/Lx200Connection.cpp
+++ b/plugins/TelescopeControl/src/Lx200/Lx200Connection.cpp
@@ -350,3 +350,18 @@ void Lx200Connection::sendCommand(Lx200Command *command)
 	}
 }
 
+void Lx200Connection::sendAbort()
+{
+		#ifdef DEBUG4
+		*log_file << Now()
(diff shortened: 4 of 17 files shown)


alex-w:
You should read the target coordinates from the mount after stopping slewing and place marker in these coordinates

gzotti:
> You should read the target coordinates from the mount after stopping slewing and place marker in these coordinates

If I knew how this positioning interpolation stuff works I could maybe do that. But while I can estimate that I should test with hasKnownPosition() and can likely retrieve pos with getJ2000EquatorialPos(), I know zilch how to set that. How do the InterpolatedPositions work? Is that a ring buffer? What is the purpose of Position.{server_micros, client_micros,status}? Dev docs are so annoyingly often missing or useless in this plugin! IDC where some struct had been defined 15 years ago and "now" moved over. I wish somebody could reverse-engineer a design document for this plugin!

alex-w:
InterpolatedPositions is own implementation of interpolation and this is does not touches NexStar/LX200 implementation - it's similar, not equal. LX200/NexStar protocols have methods for getting current coordinates from the mount and this parts of protocols are not implemented.

alex-w:
The Telescope Control plugin should be fully redesigned and rewritten and it should support asynchronous for connection to the devices. I full agree - we need experts for it.

gzotti:
I see that the TelescopeClientDirectLx200 has an interpolatedPosition that does something. But I don't dare touching this without fitting device and high enough personal interest in this plugin. Anyone out there?

gzotti:
Almost two years, and almost no response. Does nobody have and LX200 or NexStar telescope to test this? Or is this feature of no interest after all?

gzotti:
> You should read the target coordinates from the mount after stopping slewing and place marker in these coordinates

@alex-w It seems there is regular communication from LX200/NexStar to retrieve instrument position in update(). 

This plugin is so amazingly entangled, so underdocumented, it is a pain to go through without having an actual telescope to try out... I see no input from any testers with instruments, and I guess we can wait for 5 more years before anyone would show up. We could just put that in 24.3 and wait for response by users. 

And while we are in this, I could disable ASCOM support for Qt6. I can confirm it is still broken with the ASCOM simulator. Just task-switch to other programs, and debugger finds us in a library which we cannot change. 
The problem is in function Qt6OpenGL!QOpenGL2PaintEngineEx::setState()
grafik

Then we should move on from ASCOM (Win only) to Alpaca (all platforms, TCP/IP communication), maybe as early as 24.4.

alex-w (approved):

plugins/TelescopeControl/src/common/Socket.cpp
```diff
@@ -40,7 +40,7 @@ long long int GetNow(void)
 		qint64 t;
 	} tmp;
 	GetSystemTimeAsFileTime(&tmp.file_time);
-	t = (tmp.t/10) - 86400000000LL*(369*365+89);
+	t = (tmp.t/10) - 86400000000LL*(369*365+89); // TODO: Explain this magic please!
```
10110111, line 43: The constant is used to convert from NT/NTFS time epoch (1601-01-01 0:00:00 UTC) in μs to UNIX epoch (1970-01-01 0:00:00 UTC) in μs. The difference is:

24 hour/day × 3600 s/hour × ((369×365+89) day).

Here 369=1970-1601, and 89 is the number of leap years in the interval. You can verify that 369×365+89=134774 is equal to the output of this:

```
python -c "from datetime import date as d; print (d(1970,1,1) - d(1601,1,1)).days"
```
The division by 10 in the minuend is to convert from 100-nanosecond intervals to microseconds.

gzotti: I suspected something around this, but was unaware of the 1601 epoch for NTFS. 89 leap years, right, thanks!

plugins/TelescopeControl/src/common/Socket.hpp
```diff
@@ -77,6 +77,8 @@ static inline int SetNonblocking(int s)
 
 #endif //Q_OS_WIN
 
+// retrieve system time in milliseconds
+// TODO: Specify epoch?
```
10110111, line 81: UNIX epoch. Microseconds, not milliseconds.

PR #2780, opened 2022-10-26
Fix and increase resolution of normal map of the Moon

10110111:
With the old normal map, near the poles we had dark-shaded parts of craters that should actually be brighter — as if the Sun is not where it is, and this behavior inconsistently depended on longitude... This new texture fixes behavior of shadows.

Now the resolution of the normal map is the half that of the reflectance texture, 2048×1024. Further increase leads to excessive noise due to lack of proper mip mapping (and its non-triviality for normal maps), so this looks like a good compromise between the shapes of the craters and background noise.

One unwanted but non-trivial to fix side effect is that now more faces look lit behind the terminator. This is because their normal is a bit in the direction of the Sun, but their being in the shadow even relative to spherical Moon is not detected (this is natural for bump mapping).

### View from the Earth

Old:

from-earth-0ld

New:

from-earth-new

### View from "Earth Observer" (to look at a pole)

Old:

screenshot-1k-0ld

New:

screenshot-2k256s

If you want to compare only the fix of shadows without the effect of resolution increase, here's a variant (not included in this commit) with a smaller texture that still includes the fix:

screenshot-1k-new


alex-w (approved):
Thank you!

gzotti:
Nah! You exchanged a better polar view with a bad terminator now? I dislike that, sorry. The view from Earth is seriously more important! I can accept a user-configurable switch between two Normal Maps, or even automatism (observer location on earth/observer elsewhere).

10110111:
> The view from Earth is seriously more important!

This affects view form the Earth too (aside from the terminator).

> I can accept a user-configurable switch between two Normal Maps, or even automatism

No, this would make no sense. The old normal map is incorrect, even though it makes the terminator nicer.

Anyway, feel free to revert, since this PR has already been merged.

Atque:
Is the terminator only defined by the normal map, or can it be tweaked elsewhere?

10110111:
> Is the terminator only defined by the normal map, or can it be tweaked elsewhere?

Without the normal map (i.e. if all the normals pointed to zenith) we'd simply have a smooth line separating light and dark. With the normal map we have some highlights and shadows on both sides of this line.

Atque:
Would it be possible to use a (low-res) DEM? Or would that be too slow?

10110111:
I generated the normal map from an elevation model ([this one](https://svs.gsfc.nasa.gov/cgi-bin/details.cgi?aid=4720)). Using it directly would imply making a vertex model, which I'm not sure how slow it would be.

Besides, in the absence of shadows we'll still likely have the same problems around the terminator.

gzotti:
You mean a real geometrical tessellation of the Moon? I am afraid for the full globe it would be overkill, requiring millions of triangles instead of just the current simple sphere. It would however be required mostly around the rim, to see crater effects for occultations and Solar eclipses. Somebody could play with this idea, I'd also like to see it but have seriously not enough time for everything. However, I very much dislike wrongly lit mountain peaks along the terminator. The polar areas as seen from Earth are much less visually important.

gzotti:
> > It would however be required mostly around the rim, to see crater effects for occultations and Solar eclipses.
> 
> I also thought of it, but I didn't find any way to do it _nicely_, so that there were a sensible transition from the rim to closer parts of the surface. Also, if we look at the bottom of the image [here](https://en.wikipedia.org/wiki/File:FullMoon2010.jpg), we'll notice that the mountains that stick from the circle are not on a single line, so this would need making a mesh for at least a _band_ around the rim.

Yes, sure, a band that covers the full amplitude of all librations. Time to move refraction to shader code?

10110111:
> Time to move refraction to shader code?

And projection... all 10 projections.

10110111:
OK, I've reverted this commit.

gzotti:
Thank you! Just to see the craters along the rim: https://www.youtube.com/watch?v=aFZR_UbYQfE

10110111:
> Just to see the craters along the rim

I've actually been able to notice them in my own photo at this eclipse, despite having taken it from behind a window with all the resulting reflections:

photo_5384248428097291216_x

PR #2781, opened 2022-10-27
Fix normals map of the Moon, increase its resolution and implement shadows

10110111:
This is an improvement on #2780, which was rejected basically due to the lack of shadows, which the bump mapping technique can't provide.

In this commit, in addition to fixing the normals map, a new kind of texture is added: horizon map. This texture contains elevation of the horizon in four directions: North, East, South, West. These elevations take terrain (mountains, craters etc.) surrounding each point of the map into account.
These data are used in the fragment shader to determine whether the Sun is visible or is hidden behind the terrain.

Even if we later decide that we want a geometric model of the Moon to make non-circularity of its rim visible during solar eclipses, this horizon map will still be useful, because it'll save us from the shadow mapping step, which itself has lots of pitfalls fixing which would likely reduce performance.

The data for the normals and horizon maps were generated using the elevation data in the [CGI Moon Kit](https://svs.gsfc.nasa.gov/cgi-bin/details.cgi?aid=4720).
The generation was done using [map-bumper](https://github.com/10110111/map-bumper) with the following commands:
```
heightmap2normalmap 1727.4 0.5e-3 ~/Downloads/ldem_16_uint.tif moon_normals.png 1024 256
heightmap2horizonmap 1727.4 0.5e-3 ~/Downloads/ldem_16_uint.tif moon_horizon.png 1024
```
Moon radius is taken here as 1727.4 instead of 1737.4 to take into account that the uint16 version has a global shift of values by 10 km.  Normal map was generated with 256-fold supersampling.

### View from the Earth

Old:

from-earth-0ld

New:

2now

For comparison see the previous attempt that lacked shadows (the screenshot [here](https://user-images.githubusercontent.com/6376882/198159135-d0375ba0-18fc-46c9-b36a-110cd9b5796d.png)).

A real-life photo at the same location, date and time is below. Although it's blurry, you can see some craters being shaded both in the new screenshot above and in this photo.

photo

### View from "Earth Observer" (to look at a pole)

Old:

screenshot-1k-0ld

New:

eo-2now

For comparison see the previous attempt that lacked shadows (the screenshot [here](https://user-images.githubusercontent.com/6376882/198159229-f7c6a1c3-8c83-4090-8527-a05fddbf5964.png)).


gzotti:
Much better! @Atque, do you have comparisons? Now we'd need real comparison pictures! I try to find something with a date.

Atque:
Indeed much better!

> Much better! @Atque, do you have comparisons? Now we'd need real comparison pictures! I try to find something with a date.

Sadly, no new photos. Here are two that I know the date and location of. The first was taken at 2020-02-12 23:11 UTC+1, and the second at 2020-03-05 17:54 UTC+1, both from 58,3 N, 11,4 E. I took them using a 300 mm lens of cheap quality.

IMG_0951

IMG_1005

gzotti:
This was a Saturn occultation, observed in Vienna Observatory with a historical Zeiss telescope. 
stellarium-048
_MG_0150_with_insert_800_50kB

gzotti:
This was a great year for occultations... This was through thin clouds, else also good. Same instrument.
stellarium-049
2007-06-18-17-51-40_Venusbedeckung

Atque:
Seems like it matches reality pretty good. I can recommend this website for validation too:
https://svs.gsfc.nasa.gov/4955

Keep in mind to use geocentric point of view.

gzotti:
OK, I think this is good. What we need now may be some more finetuning (again) for the Lunar brightness. This is not easy and not part of this PR.

10110111:
> I can recommend this website for validation too: https://svs.gsfc.nasa.gov/4955

But do note that its images are simulated, not the ground truth.

Atque:
> > I can recommend this website for validation too: https://svs.gsfc.nasa.gov/4955
> 
> But do note that its images are simulated, not the ground truth.

True, but I would trust NASA on this. I haven't compared the details though.

gzotti:
Just to be sure: The shader has nothing critical for weak systems, right? I see another sampler2D, but I am not aware of limitations.

10110111:
> The shader has nothing critical for weak systems, right?

The shader doesn't. Or shouldn't :) But note that we have two textures now, with total size 9.3× old single texture. In other words, if the original texture took 1.5 MiB of VRAM, these two will take together 14 MiB.

gzotti:
Uh, right. While this should still not pose a problem for PCs, this may be the end of Raspi3 support. Or we can finally provide a guideline to downsample all textures. May be in another PR, I will likely have no dev time for about two weeks now.

gzotti (approved):
This looks very good now, thank you!

alex-w (approved):

Issue #2811, opened 2022-11-02
Upgrade of Stellarium from 1.0 to 1.1 on raspberry makes all graphics below horizon disappear including menus.


mgrouch:
Raspberry Pi Bullseye,
stellarium_1.1-upstream1.2~ubuntu20.04.1_arm64.deb

started using script:

```

#!/bin/bash

# hack to control Stellarium azimuth by true north heading from signalK

MESA_GL_VERSION_OVERRIDE=3.0 MESA_GLSL_VERSION_OVERRIDE=130 onlyone stellarium "$@"  &
sleep 12

i=0
while [[ -n "$(pidof stellarium)" ]]
do
  # magnetic variation
  if [[ $((i%20)) == 0 ]]; then
    MV=$(curl -s http://localhost:3000/signalk/v1/api/vessels/self/navigation/magneticVariation/value)
    i=0
  fi
  # magnetic heading
  MH=$(curl -s http://localhost:3000/signalk/v1/api/vessels/self/navigation/headingMagnetic/value)
  if [ -n "$MH" ] && [ -n "$MV" ]; then
    # true heading
    HT=$(echo "$MH" + "$MV" - 3.1415926535 | bc)
    # control Stellarium azimuth
    curl -X POST -d "az=$HT" 'http://localhost:8090/api/main/view'
  fi
  # control Stellarium location
  POS="$(curl -s http://localhost:3000/signalk/v1/api/vessels/self/navigation/position/ | jq -M -jr '"latitude=",.value.latitude,"&longitude=",.value.longitude')"
  if [ -n "$POS" ] && [ "latitude=null&longitude=null" != "$POS" ]; then
    curl -X POST -d "altitude=2&${POS}&name=Current" 'http://localhost:8090/api/location/setlocationfields'
  fi
  sleep 3
  i=$((i+1))
done

```

mgrouch:
Even if it is started like this:
`

MESA_GL_VERSION_OVERRIDE=3.0 MESA_GLSL_VERSION_OVERRIDE=130 stellarium
`

alex-w:
Please try `MESA_GL_VERSION_OVERRIDE=3.0 MESA_GLSL_VERSION_OVERRIDE=130 stellarium --opengl-compat`

mgrouch:
Still the same issue.

alex-w:
Please enable OpenGL debugging and share log from Stellarium

mgrouch:
[...]

Issue #2803, opened 2022-11-01
stellarium-1.1.1-qt6-win64.exe show only blank / black screen


tflyer-rgb:
Expected behavior

Show normal sky view after starting the program and not only a blank / black screen

Actual behavior

The application show only a blank / black screen after startup.

Steps to reproduce

Upgrade from version  stellarium-1.22.3-win64.exe  to stellarium-1.1.1-qt6-win64.exe

I try to start the application in normal, MESA and ANGLE mode always with the same result. The same result was with version 1.1, 1.0 works fine.

Stellarium version: stellarium-1.1.1-qt6-win64.exe
Operating system: Windows 10 21H2 (Build 19044.2130)
Graphics card: AMD Radeon HD 6700M
Screen type: Notebook LCD resolution 1366 x 768 
[log.txt](https://github.com/Stellarium/stellarium/files/9909663/log.txt)

10110111:
Does pressing <kbd>F11</kbd> twice make it work?

tflyer-rgb:
Pressing F11 twice works, with and without startup option --opengl-compat

gzotti:
It seems the Radeon HD is blacklisted from OpenGL use by Qt6, and falls back to Mesa. Use a Qt5-based build (version 0.22.4) for better performance.

10110111:
> It seems the Radeon HD is blacklisted from OpenGL use by Qt6

This would be strange: it isn't present in [the blacklist](https://code.qt.io/cgit/qt/qtbase.git/tree/src/plugins/platforms/windows/openglblacklists/default.json).

gzotti:
There must be a reason why the logfile says Mesa/OpenGL3.0. And I agree, we should build a more recent Mesa now that could offer 3.3CoreProfile.

zeklh:
I have the same problem. F11 works. But there is another problem. When I zoom into the moon, planets or deep sky objects which are extended like galaxies they disappear at a certain zoom level. There is only the background to be seen. This is  independent whether I activate atmosphere or not.

10110111:
> When I zoom into the moon, planets or deep sky objects which are extended like galaxies they disappear at a certain zoom level.

This is a separate problem, to be treated in another issue.

zeklh:
Created a new issue

alex-w:
Please check the fresh version (development snapshot) of Stellarium:
https://github.com/Stellarium/stellarium-data/releases/tag/weekly-snapshot

[...]

PR #2832, opened 2022-11-06
Move initial window resizing logic out of StelGLWidget::initializeGL()

10110111:
### Description
This commit cleans up the logic of initial window resizing, removing the hack of reinstating the OpenGL context that changes after resize inside `initializeGL()` (this happens not only on Windows as we observed previously, but also on Linux/X11 with Qt 6.4). This might resolve some issues: #2803, #2811, #2755.

Please check that the following cases work correctly on all interesting hardware:

* Startup with default fullscreen mode
* Startup with default windowed mode
* Placement on the correct screen at startup

### Type of change

### How Has This Been Tested?
**Test Configuration**:
* Operating system: Linux From Scratch with Qt 5.15.2, 6.4.0
* Ubuntu 20.04 with Wine64 (Windows Qt6 binary of Stellarium)

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -122,7 +122,7 @@ public:
 		Q_ASSERT(ctx == QOpenGLContext::currentContext());
 		StelOpenGL::mainContext = ctx; //throw an error when StelOpenGL functions are executed in another context
 
-		qDebug() << "initializeGL";
+		qDebug().nospace() << "initializeGL(windowWidth = " << width() << ", windowHeight = " << height() << ")";
 		qDebug() << "OpenGL supported version: " << QString(reinterpret_cast<const char*>(ctx->functions()->glGetString(GL_VERSION)));
 		qDebug() << "Current Format: " << this->format();
 
@@ -577,15 +577,19 @@ private:
 class StelGuiItem : public QGraphicsWidget
 {
 public:
-	StelGuiItem(QGraphicsItem* parent = Q_NULLPTR)
+	StelGuiItem(const QSize& size, QGraphicsItem* parent = Q_NULLPTR)
 		: QGraphicsWidget(parent)
 	{
+		resize(size);
 		StelApp::getInstance().getGui()->init(this);
+		inited=true;
 	}
 
 protected:
 	void resizeEvent(QGraphicsSceneResizeEvent* event) Q_DECL_OVERRIDE
 	{
+		if(!inited) return;
+
 		Q_UNUSED(event)
 		//widget->setGeometry(0, 0, size().width(), size().height());
 		StelApp::getInstance().getGui()->forceRefreshGui();
@@ -593,6 +597,7 @@ protected:
 private:
 	//QGraphicsWidget *widget;
 	// void onSizeChanged();
+	bool inited = false; // guards resize during construction
 };
 
 StelMainView::StelMainView(QSettings* settings)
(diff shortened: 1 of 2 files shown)


alex-w (approved):
Tested with Qt 5.12.12, 5.15.6, 6.4.0 in macOS 13.0

Qt 5.15.6 & Qt 6.4.0 - everything OK

Qt 5.12.12 - a "black window" (see [log.txt](https://github.com/Stellarium/stellarium/files/9961106/log.txt))

gzotti:
Win10 22H2 (native), Geforce 960M, Qt6.4:
Tested windowed, fullscreen. Save settings. Next start fullscreen. OK. 
Added second screen. Starts fullscreen on primary. OK. Windowed mode, move to secondary, fullscreen, save settings. 
Next launch fullscreen on secondary. OK. 
Removed secondary screen. Startup in fullscreen on primary. OK. 
Switched to secondary only. Startup is fullscreen on secondary. OK. (A warning "screen 1 not found" is visible in the logfile)

[log.txt](https://github.com/Stellarium/stellarium/files/9989883/log.txt)

10110111:
> A warning "screen 1 not found" is visible in the logfile

That's expected and unchanged: your sole secondary is now primary.

gzotti:
Yes, sure, expected, and the message should remain. So on Win10/Qt6.4 it seems to work as intended. Just compiling on RPi4...

gzotti:
RPi4, Ubuntu Mate 22.04.01, Qt6: all OK with this change. It complains about GLSL1.20 support not being enough (Mesa 22.0.5, OpenGL 2.1, not ES2.0!). But where do we really need GLSL1.30?

10110111:
Dunno, nowhere does `#version 130` show up. But [your own comment](https://github.com/Stellarium/stellarium/blob/master/src/StelMainView.cpp#L1236) from 2014 says:

>Although our shaders are only GLSL1.10, there are frequent problems with systems just at this level of programmable shaders.
> If GLSL version is less than 1.30 or GLSL ES 1.00, Stellarium usually does run properly on Windows or various Linux flavours.

I'm not sure whether "does" was meant to be "doesn't", but together with "less than" and the subsequent version checks it seems to be a typo that should mean "doesn't".

gzotti:
Ah yes. I think I had an old Intel PC to test "graphics weak/borderline system" at this time. This may have reported GLSL1.20 and had errors, so I may have chosen 1.30 here. The RPi were on GLES2.0/GLSL ES1.0. Now on Ubuntu Mate they are configured to run OpenGL 2.1 and report GLSL1.20, but seem to work flawlessly. Maybe we just revert to test for 1.20.

10110111:
Anyway, the version check is unrelated to this PR.

10110111:
So, is there anything left to test, or is this PR free to go in?

gzotti:
I think it's fine for me. Anything missing on Apple? What about the black window? I did not check on Win/Qt5, though. Did you?

10110111:
> Anything missing on Apple? What about the black window?

It was with Qt 5.12, which I don't think is of any interest for the macOS version. The Qt libraries are supplied in the package, aren't they?

> I did not check on Win/Qt5, though. Did you?

No, only Qt6 version, and only on Wine.

gzotti:
Win10/Qt5.12.8: Seems to work. However, Scenery3D is extremely slow on ANGLE (1.5fps vs 37fps) and MESA (OpenGL 3.0; 0.5 fps), basically unusable. This seems to have happened earlier, though, as it is also that slow in master.

10110111:
> However, Scenery3D is extremely slow on ANGLE (1.5fps vs 37fps) and MESA (OpenGL 3.0; 0.5 fps), basically unusable.

Does it happen without this commit?..

PR #2843, opened 2022-11-12
Cmake with precompiled headers

gzotti:
### Description

After a previous (unsuccessful) PR, cmake has introduced direct support for precompiled headers.
This should improve build time.

I have followed a tutorial at https://www.youtube.com/watch?v=a6PdfwVE9hA plus a tool (Python script to analyze header use and optimize PCH builds) described there and tried on WSL. 

However, I came to the point where building fails when cpp headers shall be included to C files, described here: https://stackoverflow.com/questions/68392702/cant-include-standard-libs-in-cmake-pch-h

In this branch, we should therefore separate C/h files out from the stelMain sources. Maybe build a separate library (stelEphem?) from the C files. The possible massive cut in build times should make this still worthwhile.

Fixes # (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 10 / WSL Ubuntu 22.04.01 / Qt5
* Graphics Card: N./A.

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -516,6 +516,7 @@ ENDIF()
 # Main executable/library setup
 IF(GENERATE_STELMAINLIB)
      ADD_LIBRARY(stelMain SHARED ${stellarium_lib_SRCS} ${stellarium_RES_CXX})
+     target_precompile_headers(stelMain PUBLIC core/VecMath.hpp core/StelUtils.h external/SpoutLibrary.h core/*.hpp core/planetsephems/*.h core/planetsephems/*.hpp core/modules/*.hpp gui/*.hpp)
      TARGET_LINK_LIBRARIES(stelMain ${STELMAIN_DEPS})
      INSTALL(TARGETS stelMain DESTINATION lib)
      # WIN32 creates a windows app without console, no effect on non-win platforms
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -261,6 +261,12 @@ SET(stellarium_lib_SRCS
      translations_regions.h
 )
 
+IF(ENABLE_PCH)
+     SET(stellarium_lib_SRCS ${stellarium_lib_SRCS}
+         stelMain_pch.h
+     )
+ENDIF(ENABLE_PCH)
+
 IF(ENABLE_SCRIPTING)
      SET(stellarium_lib_SRCS ${stellarium_lib_SRCS}
           scripting/StelScriptOutput.hpp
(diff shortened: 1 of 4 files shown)


10110111:
Does it actually speed up build for you? Is it any better improvement than just using CCache?

gzotti:
As described: I am not yet at a point where it works. This is the first time I am attempting this, but please see the video for another Qt-based project. Including just a few Qt class headers seems to do almost magic.  Our current mix of C and C++ files does not work, so we have to separate those out.  I see C files in ephemerides and healpix.c. Probably ephemerides should become another library of its own, and healpix may become part of StelHiPS. 

There is no CCache on Windows/MSVC, so I would like to get at least PCH.

gzotti:
May I ask others (@alex-w @DarthGandalf @10110111 )  for comments/experience reports? Shall we attempt to make a stelEphem library for the pure C programs so that PCH can be applied for the C++ files? It requires some restructuring. Or is it possible to keep the current structure and specify somehow with cmake that PCH should only be applied to cpp files?

src/CMakeLists.txt
```diff
@@ -516,6 +516,7 @@ ENDIF()
 # Main executable/library setup
 IF(GENERATE_STELMAINLIB)
      ADD_LIBRARY(stelMain SHARED ${stellarium_lib_SRCS} ${stellarium_RES_CXX})
+     target_precompile_headers(stelMain PUBLIC core/VecMath.hpp core/StelUtils.h external/SpoutLibrary.h core/*.hpp core/planetsephems/*.h core/planetsephems/*.hpp core/modules/*.hpp gui/*.hpp)
```
alex-w: This code will cause problem when ENABLE_SPOUT==Off

DarthGandalf:
How feasible is it to switch to C++20 already?
Modules improve build time by a lot

пн, 14 нояб. 2022 г., 14:23 Georg Zotti ***@***.***>:

src/CMakeLists.txt
gzotti: This was a detail in the first commit. Ignore. Please see current state. We should have a few Qt classes included, and maybe VecMath.hpp, StelUtils.hpp and StelCore.hpp, but the critical issue seems to be to remove all C files from the stelMain target as (precompiled) c++ headers must not be included in them. Before that, building fails.

gzotti:
> How feasible is it to switch to C++20 already? Modules improve build time by a lot 

Not much to find, and I am not familiar with details of C++ versions. E.g. https://forum.qt.io/topic/137925/qt-with-c-20-modules

Would building with Qt5 work here? We will need Qt5 for a few more years. The current attempt would work with our minimum required cmake version.

DarthGandalf:
Re C vs C++, take a look at `COMPILE_LANGUAGE` near the end of https://cmake.org/cmake/help/latest/command/target_precompile_headers.html - I believe it should make it work without reorganizing C/C++

As a short term, pch will be fine if works, we can switch to C++20 modules later.

C++20 modules would be much nicer than precompiled headers, but it does require new enough version of compiler and of cmake. Cmake which support them isn't even released yet, and I don't want to add to many hacks here trying to make it work on older version.

DarthGandalf:
What I mean is don't spend too much effort on pch :) Same effort would better be spent on modules, when they will be available.

But if it's easy to start using pch and gain some benefits from it already, great!

alex-w (approved):

PR #2846, opened 2022-11-15
Scaling stars on a HiDPI devices

### LINKED ISSUE #2840: Scaling for stars
When running Stellarium on a high-DPI monitor, stars tend to look very small. Of course this is solvable by increasing the relative scale, but it would look better if the stars also scaled, so that relative scale = 1 and absolute scale = 1 looks about the same no matter what monitor is used.

alex-w:
### Description
This PR introducing configurable option to enabling scaling stars on a HiDPI devices to better visualisation of a faint stars

Fixes #2840 (issue)

Before:
stellarium-005

After:
stellarium-006

### Type of change

## Checklist:

### DIFF
--- a/src/core/StelSkyDrawer.cpp
+++ b/src/core/StelSkyDrawer.cpp
@@ -478,7 +478,7 @@ bool StelSkyDrawer::drawPointSource(StelPainter* sPainter, const Vec3d& v, const
 	if (!(checkInScreen ? sPainter->getProjector()->projectCheck(v, win) : sPainter->getProjector()->project(v, win)))
 		return false;
 
-	const float radius = rcMag.radius;
+	const float radius = rcMag.radius*(static_cast<float>(sPainter->getProjector()->getDevicePixelsPerPixel())*StelApp::getInstance().getGlobalScalingRatio());
 #if QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
 	const float frand=StelApp::getInstance().getRandF();
 #else
--- a/src/core/StelSkyDrawer.cpp
+++ b/src/core/StelSkyDrawer.cpp
@@ -60,6 +60,7 @@ StelSkyDrawer::StelSkyDrawer(StelCore* acore) :
 	twinkleAmount(0.0),
 	flagDrawBigStarHalo(true),
 	flagStarSpiky(false),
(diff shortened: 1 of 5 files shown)


alex-w:
@gzotti @10110111 please review the PR

10110111:
Why do we even need a flag? Shouldn't it be always-on?

gzotti:
I cannot test it. Have a 3880x1440 monitor which is not regarded HiDPI. If this is a hardware property which is either 1 or 2, I agree with @10110111 that this should be set automatically.

10110111:
> If this is a hardware property which is either 1 or 2

No, it's rather an OS property which can be anything from 100% to 500% or something like that.

gzotti:
OK, I was reading about QScreen::devicePixelRatio(). Is this what you set (on Windows) as gui scaling ("100-500, 500 not recommended")? Checking always requires a logout/login for me, so I don't like to even test it.

10110111:
> Is this what you set (on Windows) as gui scaling ("100-500, 500 not recommended")?

Yes, exactly this setting.

> Checking always requires a logout/login for me, so I don't like to even test it.

That's strange, at least on Windows 11 it's applied on the fly (Stellarium needs a relaunch).

gzotti:
I remember it worked on the fly with my previous screen.  It may depend on what screen is attached, HDMI or Thunderbolt. Only the latter has the full resolution, though.

alex-w:
OK, it's always on now

src/core/StelSkyDrawer.cpp
```diff
@@ -478,7 +478,7 @@ bool StelSkyDrawer::drawPointSource(StelPainter* sPainter, const Vec3d& v, const
 	if (!(checkInScreen ? sPainter->getProjector()->projectCheck(v, win) : sPainter->getProjector()->project(v, win)))
 		return false;
 
-	const float radius = rcMag.radius;
+	const float radius = rcMag.radius*(static_cast<float>(sPainter->getProjector()->getDevicePixelsPerPixel()));
```
10110111, line 481: Too many parentheses here: it's basically `x * (func())`, while `x * func()` is sufficient.

gzotti (approved):
Fine for me.

PR #2847, opened 2022-11-16
Add support for projection computations on GPU, make MilkyWay and ZodiacalLight use this

10110111:
### Description
This PR makes it possible to perform projection on the GPU rather than CPU. This functionality will be needed in the future to render dense meshes like e.g. lunar surface.

An additional benefit is that some modules like MilkyWay and ZodiacalLight can be rendered without broken polygons at the borders of the projection area, and without the mesh effect near the horizon (see screenshots below). Landscape should also be converted, but hasn't been yet.

#### Zodiacal light, extinction

Old:

New:

#### ZL & MW, borders of projection area (Miller cylindrical for example)

Old:
9a

New:
9b

#### MilkyWay, positioning of the texels in highly-stretched regions (stereographic for example)

Old:
2a

New:
2b

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Linux From Scratch
* Graphics Card: NVIDIA GeForce GTX 750Ti, `nvidia` driver
* Qt6 in desktop OpenGL and GLES2 modes

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -856,6 +856,7 @@ void StelMainView::init()
 	const auto format = glInfo.mainContext->format();
 	glInfo.supportsLuminanceTextures = format.profile() == QSurfaceFormat::CompatibilityProfile ||
 									   format.majorVersion() < 3;
+	glInfo.isGLES = format.renderableType()==QSurfaceFormat::OpenGLES;
 	qDebug().nospace() << "Luminance textures are " << (glInfo.supportsLuminanceTextures ? "" : "not ") << "supported";
 	glInfo.isCoreProfile = format.profile() == QSurfaceFormat::CoreProfile;
 
--- a/src/StelMainView.hpp
+++ b/src/StelMainView.hpp
@@ -74,6 +74,7 @@ public:
 		QOpenGLFunctions* functions;
 		bool supportsLuminanceTextures = false;
 		bool isCoreProfile = false;
+		bool isGLES = false;
 	};
 
 	StelMainView(QSettings* settings);
--- a/src/core/RefractionExtinction.cpp
+++ b/src/core/RefractionExtinction.cpp
@@ -353,11 +353,95 @@ vec3 modelViewForwardTransform(vec3 v)
    .replace("@TRANSITION_WIDTH_GEO_DEG@", std::to_string(TRANSITION_WIDTH_GEO_DEG).c_str());
 }
 
+QByteArray Refraction::getBackwardTransformShader() const
+{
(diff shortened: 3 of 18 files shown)


alex-w:
Good performance in macOS

gzotti:
Awesome -- many thanks!
What do you mean by "mesh effect near horizon"? You did the extinction/refraction properly AFAICS, so this is another great step.

gzotti:
Ah, I think I see the mesh effect in your screenshot now. Good!

10110111:
> What do you mean by "mesh effect near horizon"?

This is the artifact that you can see in the very first ("old") screenshot in the OP near the horizon. The edges of the triangles (that, I suppose, make up the sphere) are clearly darker than their interior. The second screenshot shows that this effect is now absent, which is an improvement.

10110111:
BTW, why I made this a draft PR is because of 07c6c0a042a77fb999b4a5693b188c7e424356b1. Just wanted to pay your attention to it in case someone might object. It just is really inconvenient to always supply precision qualifiers for each variable/uniform, especially when the usual testing mode is desktop. So I thought setting `mediump` as default would simplify development.

Also, if you do merge this, please don't squash! These commits have been carefully separated to keep the history easy to examine in case something goes wrong.

gzotti:
Just built on my RPi4 (OpenGL 2.1, Mesa 22): Seems to not impact framerate. (But it's not GLES!) So the impact of this `mediump` default can be seen mostly on ANGLE.

10110111:
> So the impact of this mediump default can be seen mostly on ANGLE.

How does the framerate compare between ANGLE and native OpenGL on the same system?

gzotti:
On a certain configuration, Win10/Geforce960M, Qt5.15.2, FullHD, I tried with less side tasks:
OpenGL 32-33fps
ANGLE 25-26fps
MESA 7.5-8.6fps
ANGLE with lowp: 25 fps. 

OK, this mediump/lowp question does not influence ANGLE performance noticeably. However, IIRC it used to be not significantly slower than OpenGL 1/2 year ago. Something else must have brought the slowdown, and this PR should be good to go. I may search/bisect for other changes at some later date.

gzotti:
Or maybe my memory is wrong. 0.21.3 has 40fps with OpenGL, 27fps with ANGLE/DX9. Still, note the slowdown; This was with Preetham sky in all cases.

10110111:
OK, so shall I merge it? (If you do, please don't squash!)

alex-w (approved):

PR #2838, opened 2022-11-10
Simplify logic of finding QXlsx

DarthGandalf:
### Description
Simplify logic of finding QXlsx, and prepare for a future release of it.

Fixes # (issue)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

### How Has This Been Tested?
By building it locally and via CI

**Test Configuration**:
* Operating system: <Name, version number>
* Graphics Card: <Manufacturer (likely Intel, NVidia, AMD?), Model (HD, Geforce, Radeon..., with model number), driver version?>

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2849, opened 2022-11-17
Make QXlsx use the correct Qt

DarthGandalf:
### Description

Fixes issue reported in #2838

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,9 @@ ELSE()
          SET(QT_VERSION_MAJOR 5)
      ENDIF()
 ENDIF()
+# This will tell QXlsx to use the same Qt version (5 vs 6) as we do.
+# https://github.com/Stellarium/stellarium/pull/2849
+LIST(PREPEND CMAKE_PREFIX_PATH ${Qt${QT_VERSION_MAJOR}_DIR})
 # TODO: Check real minimal supported version of GPSD API
 SET(MINIMAL_GPS_API_VERSION "6.0")
 


DarthGandalf:
@10110111 is this fixes your build, I'll send the patch to QXlsx upstream, and fix the URL here

10110111:
Yes, this fixes the errors.

DarthGandalf:
@10110111 he suggested a different solution at https://github.com/QtExcel/QXlsx/pull/258
Does the updated PR work for you?

10110111:
If you mean `CMAKE_PREFIX_PATH`, then how is that supposed to work if both Qt versions are installed in `/usr`?

10110111:
Still sounds like a kludge. Why have both `-DENABLE_QT6` and `-DCMAKE_PREFIX_PATH`? Maybe one should be dropped then, if the other works universally?

10110111:
So, is your idea to replace `CMAKE_PREFIX_PATH` from inside `CMakeLists.txt`? This sounds hacky.

DarthGandalf:
> So, is your idea to replace CMAKE_PREFIX_PATH from inside CMakeLists.txt? This sounds hacky.

It's already in this PR :) and it's not replace, but change. It's possible user provided their CMAKE_PREFIX_PATH already, we shouldn't break that use case.

DarthGandalf:
@10110111 https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html says that this variable is supposed to be set by the project, which I read as set by CMakeLists.txt

@alex-w of course! that's exactly what I'm trying to do here, which broke in #2838

10110111:
> It's already in this PR :)

Oh, didn't notice you changed it. OK, this works now. I think there should be a comment explaining this hack.

DarthGandalf:
> I think there should be a comment explaining this hack.

Done.

10110111 (approved):

alex-w (approved):

PR #2856, opened 2022-11-18
Projection: plate carree

### LINKED ISSUE #1120: Add equirectangular projection
The request for adding support for [equirectangular projection](https://en.wikipedia.org/wiki/Equirectangular_projection) in Stellarium was published in small_planetarium mail list by Paul Bourke to improve using Stellarium by small planetariums.

gzotti:
### Description

Ruslan's shader implementation of Milky Way and ZL makes this finally interesting: A window-filling all-sky map. I can imagine an application using Spout where the result goes into another program, as live texture for some object, or just a printed all-sky map. 

Fixes #1120 

What currently still looks bad at the seam are of course all meshes which are projected before being sent to the GPU: 
- landscapes
- constellation artwork
- DSO textures
- DSS
- HiPS
- sky images (from scripting)
- An axis-aligned grid has flickering parts on the window edges.
- Points like NCP flicker as they should be plotted on the edge but may be projected to the other side by rounding.

Fixing the meshes is not part of this PR. For the time being the application is limited to mostly MW/ZL all-sky maps. 

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Geforce 960M

## Checklist:

### DIFF
--- a/src/core/StelProjector.cpp
+++ b/src/core/StelProjector.cpp
@@ -152,7 +152,7 @@ void StelProjector::init(const StelProjectorParams& params)
 	zNear = params.zNear;
 	oneOverZNearMinusZFar = 1./(zNear-params.zFar);
 	viewportCenterOffset = params.viewportCenterOffset;
-	viewportXywh = params.viewportXywh * static_cast<int>(devicePixelsPerPixel);
+	viewportXywh = params.viewportXywh * devicePixelsPerPixel;
 	viewportCenter = params.viewportCenter * devicePixelsPerPixel;
 	gravityLabels = params.gravityLabels;
 	defaultAngleForGravityText = params.defaultAngleForGravityText;
--- a/src/core/StelProjector.cpp
+++ b/src/core/StelProjector.cpp
@@ -152,19 +152,14 @@ void StelProjector::init(const StelProjectorParams& params)
 	zNear = params.zNear;
 	oneOverZNearMinusZFar = 1./(zNear-params.zFar);
 	viewportCenterOffset = params.viewportCenterOffset;
-	viewportXywh = params.viewportXywh;
-	viewportXywh[0] *= devicePixelsPerPixel;
-	viewportXywh[1] *= devicePixelsPerPixel;
-	viewportXywh[2] *= devicePixelsPerPixel;
-	viewportXywh[3] *= devicePixelsPerPixel;
-	viewportCenter = params.viewportCenter;
-	viewportCenter *= devicePixelsPerPixel;
+	viewportXywh = params.viewportXywh * static_cast<int>(devicePixelsPerPixel);
+	viewportCenter = params.viewportCenter * devicePixelsPerPixel;
 	gravityLabels = params.gravityLabels;
 	defaultAngleForGravityText = params.defaultAngleForGravityText;
 	flipHorz = params.flipHorz ? -1.f : 1.f;
 	flipVert = params.flipVert ? -1.f : 1.f;
 	viewportFovDiameter = params.viewportFovDiameter * devicePixelsPerPixel;
-	pixelPerRad = 0.5f * static_cast<float>(viewportFovDiameter) / fovToViewScalingFactor(params.fov*(static_cast<float>(M_PI)/360.f));
+	pixelPerRad = 0.5f * static_cast<float>(viewportFovDiameter) / fovToViewScalingFactor(params.fov*(static_cast<float>(M_PI/360.)));
 	widthStretch = params.widthStretch;
 	computeBoundingCap();
 }
--- a/src/core/StelProjector.cpp
+++ b/src/core/StelProjector.cpp
@@ -152,8 +152,13 @@ void StelProjector::init(const StelProjectorParams& params)
 	zNear = params.zNear;
 	oneOverZNearMinusZFar = 1./(zNear-params.zFar);
 	viewportCenterOffset = params.viewportCenterOffset;
-	viewportXywh = params.viewportXywh * devicePixelsPerPixel;
-	viewportCenter = params.viewportCenter * devicePixelsPerPixel;
+	viewportXywh = params.viewportXywh;
+	viewportXywh[0] *= devicePixelsPerPixel;
+	viewportXywh[1] *= devicePixelsPerPixel;
+	viewportXywh[2] *= devicePixelsPerPixel;
+	viewportXywh[3] *= devicePixelsPerPixel;
+	viewportCenter = params.viewportCenter;
+	viewportCenter *= devicePixelsPerPixel;
 	gravityLabels = params.gravityLabels;
 	defaultAngleForGravityText = params.defaultAngleForGravityText;
 	flipHorz = params.flipHorz ? -1.f : 1.f;
--- a/src/core/StelProjectorClasses.cpp
+++ b/src/core/StelProjectorClasses.cpp
@@ -57,7 +57,7 @@ bool StelProjectorPerspective::forward(Vec3f &v) const
 
 bool StelProjectorPerspective::backward(Vec3d &v) const
 {
-	v[0] /= static_cast<double>(widthStretch);
+	v[0] /= widthStretch;
 	v[2] = std::sqrt(1.0/(1.0+v[0]*v[0]+v[1]*v[1]));
 	v[0] *= v[2];
 	v[1] *= v[2];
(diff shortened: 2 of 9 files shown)


src/core/StelProjector.cpp
```diff
@@ -152,19 +152,14 @@ void StelProjector::init(const StelProjectorParams& params)
...
 	oneOverZNearMinusZFar = 1./(zNear-params.zFar);
 	viewportCenterOffset = params.viewportCenterOffset;
-	viewportXywh = params.viewportXywh;
-	viewportXywh[0] *= devicePixelsPerPixel;
-	viewportXywh[1] *= devicePixelsPerPixel;
-	viewportXywh[2] *= devicePixelsPerPixel;
-	viewportXywh[3] *= devicePixelsPerPixel;
-	viewportCenter = params.viewportCenter;
-	viewportCenter *= devicePixelsPerPixel;
+	viewportXywh = params.viewportXywh * static_cast<int>(devicePixelsPerPixel);
```
10110111, line 155: Scaling factor must not be rounded, rounding will break scales like 150% or 125%, which are often the default or recommended OS settings.

src/core/StelProjectorClasses.cpp
```diff
@@ -653,6 +603,42 @@ vec3 projectorBackwardTransform(vec3 v, out bool ok)
...
+	return q_("Equirectangular");
+}
+
+QString StelProjectorCylinderFill::getDescriptionI18() const
+{
+	return q_("The full name of this variant of the Cylindrical projection mode is <i>plate carr&eacute;e</i> (French, for 'flat square'). With this projection all parallels are equally spaced. "
+		  "The view is stretched to always show a 360x180° field of view in a fixed view direction. It is provided for specialized setups.");
+}
+
+bool StelProjectorCylinderFill::forward(Vec3f &v) const
```
10110111, line 618: Why do you override these methods with code identical to that of the base class (and don't override corresponding shader methods)?

gzotti: Right. Again, done in 2020. In fact it seems only the automated stretch and blocking of zooming was necessary.

gzotti:
I think it's ready for discussion. What I see:
* great for All-sky maps including just MilkyWay/ZL. Will become also great for horizontal maps when @10110111 (?) can implement shader solutions for the landscapes and other large textured regions. (Not part of this PR.)
* With the respective grids activated, grids flicker at the screen edge. This could be remedied by plotting all 4 specific straight lines if this projection is active. Necessary?
- [ ] Draw extra gridlines along screen edges
- [ ] Ignore

* NCP flickers at the edge in alt-azimuthal mode. Solutions: 
- [ ] Repeat drawing with position set manually to the left edge. 
- [ ] Ignore

10110111:
Current state of your branch looks like below on my machine. Screen scaling factor is 1.5, if it matters.

Screenshot_2022-11-21_13-48-22

gzotti:
OK, yes, scale obviously matters...
Uh, did landscape drop below horizon as well?

10110111:
> Uh, did landscape drop below horizon as well?

They appear to be at different scales: as I pan around, the horizon moves as if with a parallax relative to the sky dome.

10110111:
Actually, no, it's the atmosphere horizon that moves. The stars move synchronously with the landscape.

gzotti:
Grmpf... I have now set 125% scaling on Windows (works on the fly only with built-in screen), and see no issues (apart from obvious screen space issues).

10110111:
> works on the fly only with built-in screen

Change on the fly is done purely by the OS. Qt doesn't get a chance to adapt for some reason. Restart of the app should make Qt aware of the change and in charge of it.

gzotti:
Yes, of course I quit Stellarium, changed resolution, and restarted Stellarium. It comes up with the new resolution/bigger GUI but still fills the window. So this is an issue visible only if running on Linux. Getting this screen scaling right on all platforms is a nightmare.

gzotti:
@10110111 just tried on Ubuntu 22.04.1/Geforce M 580. I set scaling to 125%, and now I see scaling issues in fullscreen, but for all projections.  Our view frame is confined to the upper left quadrant. These issues are also present in master!

Is the problem that you see above limited to the new projection?

10110111:
Master works perfectly fine for me (aside from failure to compile due to GregorianCalendar includes being commented out in JKV calendars).

How do you set the scaling factor? I just use the following environment variables:
```
QT_FONT_DPI=96
QT_SCREEN_SCALE_FACTORS=1.25
```
Similarly fine it is when I don't use any variables, while the screen DPI is set to 167 (in this case Qt sets scale factor 167/96, i.e. 1.73958...).

I'm testing with Qt 6.4.0 on Ubuntu 20.04 and Intel UHD Graphics 620.

gzotti:
Ah no, I set scaling in the OS's Settings GUI. (Ubuntu 22.04.1) First, enable Fractional Scaling, then set to 125%, then (to be sure all programs get it) logout/login. U22.04 includes Qt6.2.4. 
I can try with the QT env. variables as keep the OS scaling at 100%, but it is probably not what most users would do.

10110111:
> Ah no, I set scaling in the OS's Settings GUI.

Oh well, I don't know what tricks they use. Could you post output of `xrandr --verbose` after having set fractional scaling?

gzotti:
[xrandr_100pct.txt](https://github.com/Stellarium/stellarium/files/10195544/xrandr_100pct.txt)
[xrandr_125pct.txt](https://github.com/Stellarium/stellarium/files/10195545/xrandr_125pct.txt)

This is without the QT... variables set.

gzotti:
> AFAICT, what you are setting is a scaling, but not _the high-DPI_ scaling. Your setting only enlarges the desktop, downsampling the resulting raster to fit on the screen. Globally, regardless of app awareness, in a raster-downsample mode! This is definitely not something we can support well.

OK, but I have no HiDPI device. This is what the user sees.

> Still, I have tried this on my Intel-based machine, and it works without confining the viewport.
> 
> From the output names I figure that you are using `nvidia` driver, right? Generally this is the correct choice, but historically, nvidia's implementation of XRandR was not the best, so the viewport confinement issue you're seeing might be due to a bug there.

Ah, OK. Yes, this is the NVidia driver for Ubuntu. 
 
> Finally, you should be able to reproduce the problem of this branch being messed up while `master` looks OK by using the environment variable approach (resetting XRandR scaling factor to 100%).
> 
> > Is the problem that you see above limited to the new projection?
> 
> Just retested, the problem equally affects all projections in this branch.

Interesting. Changes other than the new projection (variant) were not so many. Let's see...

gzotti:
Interesting. With these variables and without setting scaling via Windows system settings, I see the issue on Windows. Is this what users are supposed to configure, or is this a weird hack which should look broken by design?
...Errm, now they don't show up any longer. Nothing changed in the code.

10110111:
> Is this what users are supposed to configure, or is this a weird hack which should look broken by design?

These variables have two use cases:

* for the developers to test the GUI on different scaling factors,
* for the users to override wrong system configuration (e.g. wrong EDID reported by a monitor).

They definitely aren't supposed to look broken.

10110111:
e624ccdf2c32ccc38722727ddb9d7737743df1a5 indeed makes things work better.

To make this simplification work, the `Vector4<T>::operator*(T)` should have a custom template parameter to take right-hand operand as is, without coercing it to `T`. (Ideally, all these overloaded operators should not force `T` for scalars.)

gzotti:
Yes, I have concluded the same. But it's not needed for this PR. So, I think this should be ready to merge. 

The next big thing would be moving some drawing code for the Landscapes into shader code to avoid the visible gaps.  For now, this new mode works best for all-sky maps without labels or lines along the edge. 

About HiDPI, as they are becoming more common, we should add a section to the User Guide, e.g. 2.5.1.

10110111:
> The next big thing would be moving some drawing code for the Landscapes into shader code to avoid the visible gaps.

I have this in progress (spherical is done, fisheye almost done, old style not started yet), but I suppose it won't get into 1.2, since I'm getting a new job next week and won't have as much time then.

gzotti:
Great to read, and of course this can wait until 23.1 or later.  Congrats on the new job! 
@alex-w any comments/objections? Or can we merge this?

alex-w (approved):

PR #2873, opened 2022-11-25
Add keyboard accelerators in the Location dialog

10110111:
### Description
This PR makes it possible to focus controls in the Location dialog by <kbd>Alt</kbd>+<kbd>SomeKey</kbd> combos. There's a problem that to actually focus one needs to press the combo twice. I don't know why this is so (maybe some even filter is interfering), but this PR seems correct in principle (and double pressing does switch focus as expected).

Screenshot_2022-11-25_20-07-26

### How Has This Been Tested?

**Test Configuration**:

### DIFF
--- a/src/gui/locationDialogGui.ui
+++ b/src/gui/locationDialogGui.ui
@@ -350,7 +350,7 @@
             <item>
              <widget class="QPushButton" name="resetListPushButton">
               <property name="text">
-               <string>Reset Location List</string>
+               <string>Reset Locat&amp;ion List</string>
               </property>
              </widget>
             </item>
@@ -445,7 +445,7 @@
                <enum>Qt::NoFocus</enum>
               </property>
               <property name="text">
-               <string>Add to list</string>
+               <string>&amp;Add to list</string>
               </property>
              </widget>
             </item>
@@ -464,7 +464,7 @@
                <enum>Qt::NoFocus</enum>
               </property>
               <property name="text">
-               <string>Delete from list</string>
+               <string>Delete fro&amp;m list</string>
               </property>
              </widget>
             </item>
@@ -477,7 +477,7 @@
                <enum>Qt::NoFocus</enum>
               </property>
               <property name="text">
-               <string>Return to default location</string>
+               <string>Re&amp;turn to default location</string>
               </property>
              </widget>
             </item>
@@ -523,7 +523,10 @@
             <item row="0" column="0">
              <widget class="QLabel" name="label_5">
               <property name="text">
-               <string>Latitude:</string>
+               <string>&amp;Latitude:</string>
+              </property>
+              <property name="buddy">
+               <cstring>latitudeSpinBox</cstring>
               </property>
              </widget>
             </item>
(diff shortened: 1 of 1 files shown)


alex-w:
The main problem with accelerators in Qt-based UI is correct working in English language only. For other languages translators should add hooks, like “Долгота (&L)”

10110111:
> The main problem with accelerators in Qt-based UI is correct working in English language only. For other languages translators should add hooks, like “Долгота (&L)”

No, there shouldn't be any problem! If you have Russian keyboard layout as the default, things like "&Долгота" will trigger via <kbd>Alt</kbd>+<kbd>д</kbd> perfectly well.
If the Russian layout is not the default, there may be problems. At least I remember having them with Qt4, not sure whether this is fixed now.

10110111:
In any case, "Долгота (&L)" is an extremely ugly kludge, this shouldn't ever be used. Instead, Qt should be fixed. I had a patch for Qt4 somewhere, if it's still relevant, I might be able to find it and adapt to modern Qt versions.

alex-w:
Stop! Why we need accelerators here and why we need accelerators for text labels?

10110111:
Labels are buddies of corresponding spinboxes etc. Press an accelerator like <kbd>Alt</kbd>+<kbd>o</kbd> (twice in Stellarium), and you activate the control to the right of the label.
As to why we need it, it's easier and faster to use accelerators than to point and click, especially when doing some repeated jumps over the Universe as I do during development work.

gzotti:
Will any of the "regular" users ever use them? Especially if they seem not to work when pressed (once)?

Another question is the translatability (using potentially different keys in each language) and mix with our Action keys. Are the keyboard accelerators limited to whatever is visible on the screen? What happens if I have an action key bound to the same key combination as is currently visible? 

There is another PR running (or better: sleeping) around Tab sequence and highlighting of the currently "active" GUI element (which may be pressed by Space). Maybe it is more meaningful to complete that one, and remove all "underlined" key combos. I started with subclasses that would show a decent frame when "active", but maybe regular CSS can do the same if somebody found out the relevant CSS key.

10110111:
> Will any of the "regular" users ever use them?

This is a common functionality known to all power users of at least Windows.

> Especially if they seem not to work when pressed (once)?

This is a bug that's located elsewhere, I intend to try to fix it some time later.

> Another question is the translatability (using potentially different keys in each language)

This is not a problem in principle: see the menu bar of a Russian-localized Windows Explorer in [this screenshot](https://ru.wikipedia.org/wiki/%D0%9F%D1%80%D0%BE%D0%B2%D0%BE%D0%B4%D0%BD%D0%B8%D0%BA_Windows#/media/%D0%A4%D0%B0%D0%B9%D0%BB:WinXPexplorer.png): it's translated, and there's no latin-letter kludge there.
A possible problem with Qt is discussed in my [earlier comment](https://github.com/Stellarium/stellarium/pull/2873#issuecomment-1327558042), but doesn't seem critical.

> and mix with our Action keys

This needs some testing. But our "main" keys like <kbd>a</kbd> for atmosphere already don't work when e.g. Location dialog is open.

> Are the keyboard accelerators limited to whatever is visible on the screen?

Yes.

> What happens if I have an action key bound to the same key combination as is currently visible?

Needs testing, not sure how to bind action keys.

> There is another PR running (or better: sleeping) around Tab sequence

This is a poor replacement for accelerators, it should be an addition instead.  Tapping <kbd>Tab</kbd> potentially dozens of times to reach a single option is not very productive in case of repeated actions.

alex-w:
The accelerator and shortcut `Alt+A` should conflicted, because key combination is correct for both of them.

alex-w (changes requested):
> The accelerator and shortcut `Alt+A` should conflicted, because key combination is correct for both of them.

Yes, I see conflict - `Alt+Z` is used as shortcut for Satellites plugin and as accelerator for Location dialog. When location dialog is opened then both - shortcut and accelerator - do not work. Well, if we want use accelerators then we should disable shortcuts when dialogs are opened and enable shortcuts when dialogs are closed. Plus additional work for translators. So, the current implementation isn't acceptable IMHO.

gzotti:
Seriously, I am afraid this mix of user-configurable Action keys and actually Translator-configurable GUI shortcuts leads to more problems than is worth bothering.

10110111:
OK, withdrawing the PR.

PR #2912, opened 2022-12-13
Sort UTC-based time zones in numeric order

10110111:
Before this patch the UTC-based time zones in the location dialog had an unnatural order:

+00:00
+01:00
...
+14:00
-00:00
-01:00
...
-14:00

After this patch they go in numeric order, which makes it easier to find the one the user needs:

-14:00
...
-01:00
-00:00
+00:00
+01:00
...
+14:00

**Test Configuration**:

## Checklist:


alex-w (approved):

gzotti:
Looks better, indeed. 
There are also timezones `Etc/GMT[+-]HH`, at least on Windows. Not sure who needs them, but sorting them would likewise look better.

10110111:
> There are also timezones `Etc/GMT[+-]HH`

Do you mean EST, not Etc? Also, can you show a screenshot of how exactly they (both it and GMT) look in the list? I don't have them on Linux.

gzotti:
Ah, are they really Win-only? The QtDocs say we must expect differences... I have them on Win10 and now 11. 

stellarium-026
stellarium-027
stellarium-028

10110111:
Oh, so are literally `Etc/GMT`! I thought it's just two sets of time zones. Are there any explicit ±0 offsets, or maybe implied (without ±)? Should we put `Etc/UTC` in the place of (or next to) the supposed `Etc/GMT+0` time zone?

gzotti:
Thinking further: If you do in fact edit/re-sort the list, I think it would be best to just remove all those "Etc" zones! They do nothing that cannot be done with UTC+/-XX:00.

10110111:
> I think it would be best to just remove all those "Etc" zones

Are the UTC±x also present?

gzotti:
Yes, and they (and your re-sorting) are OK.

10110111:
OK, added code to remove the `Etc/GMT` zones. BTW, do we need the `UTC` and `UTC-0` when we have `UTC+0` already? Maybe these should also be removed?

gzotti:
I don't know who uses UTC-00:00. This can probably also be removed. UTC (without extras) looks nicer than UTC+00:00  in some context, others may want to see the insistence on 00:00 offset, so I would leave those both. @alex-w ?

alex-w:
> I don't know who uses UTC-00:00. This can probably also be removed. UTC (without extras) looks nicer than UTC+00:00 in some context, others may want to see the insistence on 00:00 offset, so I would leave those both. @alex-w ?

This is acceptable, but after reading about inner implementation of TZ support in Qt I think we should add external library for managing original IANA data. Of course this is task not for current pull request.

gzotti:
One strength of the Qt implementation is the minute-offset like UTC+04:23 needed in some context (calendars...) and otherwise useful behaviour. Suppression of (apparently) Windows-only Etc/GMT... zone names and our private name adaptation QMap now work, what else would we need? There may be more important tasks ahead.

alex-w:
> One strength of the Qt implementation is the minute-offset like UTC+04:23 needed in some context (calendars...) and otherwise useful behaviour. Suppression of (apparently) Windows-only Etc/GMT... zone names and our private name adaptation QMap now work, what else would we need? There may be more important tasks ahead.

Around 20% of historical time zones on Windows platform are removed or has obsolete rules.

gzotti:
Hmm, this would mean we must use libtimezone, and pack tz updates privately in each release?

alex-w:
> Hmm, this would mean we must use libtimezone, and pack tz updates privately in each release?

Maybe actual copy of tzdata + tool for updating tzdata

gzotti (approved):
There are zones Etc/UTC and UTC-00:00 which could still be removed for sake of clarity. Else this is a nice clean-up!

10110111:
OK, let's be conservative and leave them in for now.

Issue #2775, opened 2022-10-23
Outer ring of Uranus has a gap


10110111:
Maybe I don't know something about the rings of Uranus, but it seems quite strange that the outer ring as rendered by Stellarium has a gap that rotates as time passes. See the screenshot:

![Screenshot - 231022 - 23:37:00](https://user-images.githubusercontent.com/6376882/197417046-55604e37-7ac5-431d-9dcd-8df5c8eef9f2.png)

I have long seen this behavior in Stellarium, only now report it. It reproduces on Intel UHD Graphics 620, NVIDIA GeForce GTX 750Ti, NVIDIA ION etc., so I suppose it's a problem in Stellarium itself—or in my knowledge about Uranus.

gzotti:
I think the rings are somehow described as incomplete. Must check again.

alex-w:
This is correct view, because outer ring of Uranus is not full circle

10110111:
> This is correct view, because outer ring of Uranus is not full circle

Any reference to support this?

gzotti:
I am no longer sure. Yes, I remember having read about ringlets and incomplete arcs,  by gravity interaction with moons.  See https://en.wikipedia.org/wiki/Rings_of_Uranus and references therein.
The &epsilon; ring is described as shepherded by Cordelia and Ophelia. Our model does not do that. Maybe somebody should check the distribution of non-empty pixels in the Uranus ring texture again. 
What surprises me however is that the texture is just a 2-px wide line. This is enough to paint a textured circle, but indeed, that the outermost circle is not closed is weird.

10110111:
The only part that talks about incomplete arc is this:

> The ε ring seems to consist of a number of narrow and optically dense ringlets, some of which may have incomplete arcs.

So the _ringlets_, not the whole ring, and _may_ have incomplete arcs, not do have. This in no way makes the whole ring have abrupt start and end.

Finally, just look at the image from Hubble as posted by the JPL [here](https://www.jpl.nasa.gov/images/pia01282-hubble-observes-the-planet-uranus), and for comparison the Stellarium render of this configuration (Aug 14 1994).

![](https://d2pn8kiwq2w21t.cloudfront.net/original_images/jpegPIA01282.jpg)

![Screenshot - 241022 - 01:07:53](https://user-images.githubusercontent.com/6376882/197420441-8ae27c00-e086-44fd-b25c-beacc3863b79.png)

10110111:
[...]

Issue #435, opened 2018-09-26
Saturn's rings wobble , as seen from Pan


alex-w:
_Original report by Pierluigi Panunzi:_ https://bugs.launchpad.net/stellarium/+bug/1559657

Hi!
I'm working with Saturn and it's famous rings...
I have created a data packet for its satellite PAN, whose orbit lies inside Enke's Division.
Collecting data from wikipedia, I organized the relevant data for Pan :

```
[pan]
albedo=0.5
color=1., 0.9, 0.75
coord_func=ell_orbit
halo=true
lighting=true
name=Pan
orbit_AscendingNode=263.0
orbit_Eccentricity=0.000035
orbit_Epoch=2453177.5
orbit_Inclination=0.0
orbit_LongOfPericenter=209.0
orbit_MeanLongitude=106.645
orbit_Period=0.57505
orbit_SemiMajorAxis=133584.0
orbit_visualization_period=0.57505
parent=Saturn
radius=15
rot_equator_ascending_node=213.7
rot_obliquity=0.0
rot_periode=13.8012
rot_rotation_offset=287
tex_halo=star16x16.png
tex_map=lune.png
type=moon
```
practically organizing data as other satellites with "ell_orbit" coord_func value: I started from Jupiter's Adrastea and modified values accordingly.
Please note the null values for "orbit_Inclination" and "rot_obliquity".
Now insert these data in ssystem.ini file and follow my instructions! You can witness a really awesome sight!

- choose Pan as Location (F6) and set coordinates to +40°00'00", +100°00'00", in order to see a magnificent view of Saturn
- with the mouse move horizontally a little in order to see rings edge-on
- choose a FOV of about 80°, just to see a smaller Saturn...
Awesome view!!!

- press 3 times the L key, in order to increase the speed of time: you can see the magnificent show of Saturn crossed by the rings' shadows, but you can see also that there's something strange and wrong with the rings, that "flash" cyclically at that rate...

If you return to the normal speed of time (K) then you can see that rings are not always edge-on as you can expect: please remember that both inclinations are put to streight 0.0

What is wrong with my Pan's setup? Or is there any kind of issue in the "ell_orbit" algorithm?

If you try the same, starting now from Enceladus (for which we have "coord_func=enceladus_special") you can see that rings remain always edge-on without any wobbling.
[...]

PR #2917, opened 2022-12-14
Improve quality of rings rendering

10110111:
### Description
Current code that renders planetary rings has some issues:

* ~Outermost layer has a gap, as noted in #2775, and it becomes obvious even on Saturn if we reduce number of "stacks" in the call to `sRing()` to 3 or 2;~
* Rings of Saturn are obviously polygonal when zoomed in to screen size (see left and right sides of the screenshot below);
* ~Additionally, ring shadow is not quite smooth: sort of aliased.~

This PR redoes the rings in the following way:

* ~Instead of making a polygonal model of the rings, the whole model becomes just a quad, and~ texture coordinates are computed in the fragment shader to yield an annulus ~instead of the square~.
* ~Mip mapping is enabled for ring texture.~
* ~Shadow is made behave correctly with mip mapping, finally becoming smooth.~

#### Old Saturn (viewed from Jupiter):

stellarium-003

#### New Saturn (viewed from Jupiter)

stellarium-004

#### Old Uranus

stellarium-001

#### New Uranus

stellarium-002

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/data/shaders/planet.frag
+++ b/data/shaders/planet.frag
@@ -349,7 +349,24 @@ void main()
     //apply texture-colored rimlight
     //litColor.xyz = clamp( litColor.xyz + vec3(outgas), 0.0, 1.0);
 
-    lowp vec4 texColor = texture2D(tex, texc);
+    lowp vec4 texColor;
+#ifdef RINGS_SUPPORT
+    if(isRing)
+    {
+        float radius = length(texc);
+        float s = (radius - innerRadius) / (outerRadius - innerRadius);
+        vec2 texCoord = vec2(s, 0.5);
+        texColor = texture2D(tex, texCoord);
+        // Guard against poor quality mipmap filtering (e.g. Mesa with NPOT textures).
+        if(radius > outerRadius)
+            texColor = vec4(0);
+    }
+    else
+#endif
+    {
+        texColor = texture2D(tex, texc);
+    }
+
 #ifdef IS_MOON
     // Undo the extraneous gamma encoded in the texture.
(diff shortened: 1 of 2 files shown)


alex-w (approved):

10110111:
But this is already much better than what we have in current `master`:

#### This PR:

stellarium-007

#### Master:

stellarium-008

gzotti:
Yes, the problem around #435 is that Pan is inside the rings, so the tessellation in all but Perspective projection makes an ugly mess. Perspective is somewhat bearable. It seems that in your solution the observer is now not surrounded by rings. 
And look into the strange break in the rings also when observed from Saturn's surface. This works well in master.

gzotti:
@10110111 , do you have a working solution at least for this view from Saturn? Else we should push it to 23.1 to not block 1.2.

10110111:
No, let this go to a next release.

10110111:
So, I have undone the simplification of the ring mesh to a square, and fixed #2775. Only roundedness of the rings (as seen from faraway observers) remains to be the topic of this PR, which shouldn't break the look from inside the rings more than it already is.

alex-w:
> So, I have undone the simplification of the ring mesh to a square, and fixed #2775. Only roundedness of the rings (as seen from faraway observers) remains to be the topic of this PR, which shouldn't break the look from inside the rings more than it already is.

I accept it and approve the PR

gzotti (approved):
Very good now, thank you! 
I also consider #435 solved by this solving of the indexing. The ring is  a bit coarse, but no mess.

PR #2925, opened 2022-12-21
Optimize rendering of spherical landscapes by avoiding texture sampling when possible

10110111:
This change should improve performance a bit when rendering spherical landscapes. This should only affect the case when the projection doesn't take full screen (e.g. fisheye at largest FoV).

I'm not sure how much it will improve on various GPUs, so please check if it makes any difference (be it positive or negative) on the machines you have.

### DIFF
--- a/src/core/modules/Landscape.cpp
+++ b/src/core/modules/Landscape.cpp
@@ -1527,14 +1527,8 @@ void main(void)
 											 dot(modelPos, modelPos);
 	float texTdx = dFdx(texc.t);
 	float texTdy = dFdy(texc.t);
-	vec2 texDx = vec2(gradLongitude.s/(2.*PI), texTdx);
-	vec2 texDy = vec2(gradLongitude.t/(2.*PI), texTdy);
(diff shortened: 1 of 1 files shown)


gzotti:
Grossmugl landscape (Spherical)

Win11/Geforce 3070Ti: grid, daylight, SMS sky to decrease fps, Qt6.4, large screen.
- master: 30fps
- this PR: 30fps

RPi3: small screen. Qt5.15. 
- master: 3.85fps &pm; 0.05fps
- this PR: 3.82fps &pm; 0.05fps

Seems not to help much, maybe even to the worse. Do you see a difference on any platform?

10110111:
> SMS sky to decrease fps

This is counter-productive. To see a difference such small as this you should disable everything except landscape, AND disable vsync.

> Do you see a difference on any platform?

It was inconclusive. The only reason why I think it might help is theoretical, that it might be better on _some_ platform. If it isn't, well, I'll just close this PR.

PR #2954, opened 2022-12-29
Fix: unified ObservingLists dialog

### LINKED ISSUE #2802: ObserveList and IC1848 Soul Nebula
### Expected Behaviour
Both IC1848 and name Soul Nebula.  (actually finnish term Sielunsumu as Soul Nebula)

### Actual Behaviour
Observe list have for example IC1848 as Soul Nebula, but there is not IC1848. I want to see both code and name. Maybe this is not bug of Stellarium, but bug of data what it uses. There are many other item in observe list which have not both data code and name. If so, it is large task to make data valid. But if it is bug of Stellarium, it can be fixed. If those are question of translation (to finnish language), I can do something to that.

###.  Steps to reproduce
1. Search (F3) IC1848
2. Add it to observeList
3. Look observeList. You see only name Soul Nebula / Sielunsumu, but not IC1848

### System
* Stellarium version: <Stellarium-1.1-RC3-x86_64.AppImage>
* Operating system: <Kubuntu, 22.04>
* Graphics Card: <MSI GeForce GT 1030 2GH LF OC>
* Screen type (if applicable): 3440x1440 LG 34GK950F

### Logfile 
In this case that is not very important.

gzotti:
### Description

The state of ObservingLists in 1.0 is functionally OK, but IMHO still not user friendly. 
- Editing opens a secondary dialog. I fail to see the point. 

It is also not maintainer friendly
- needlessly long variable names that mostly start with the same name.
- inconsistent variable names
- lots of duplicate code
- The handling of the lists requires repeated loading and looping through the lists. 

It should be enough to load the lists once and manipulate the copy kept in memory. Just occasionally, we need to store the list. Maybe even only write on exit?  Probably a copy of the existing list should be added on loading, in case of program failure. 

Fixes #2802 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11, Qt6.4
* Graphics Card: (irrelevant)

## Checklist:


gzotti:
@alex-w there is a mismatch in ObsListDialog LL 454ff vs. LL600ff when assigning type and objtype. I am not sure what the type/objtype now really should resemble. It is really problematic to cross-assign as is now done in LL454f. I also don't know why we store these in the JSON in the first place. 

This may also relate to #2802. The ObsList has entries "designation" and "nameI18n". I would never store an internationalized name in a list that we also redesigned so that users can exchange lists across (language) borders. 

Shall we develop format version 2.1 or (3.0): store "designation" and "name" (i.e., englishName), and find an i18n on the fly just for displaying? Yet another issue is star and DSO names: these are sky culture dependent. So, it may be OK to just use designation, and deal with a "name" entry for the human reader of the JSON file, but load/overwrite the name during loading from the designation.

gzotti:
@alex-w can you please look into these questions?

alex-w:
@gzotti sorry for late answer.

type/objtype is important to storing into JSON for correct finding of the object - for example _B 10_ can be deep-sky object and star. Localized name can be removed from JSON and we can storing into JSON designations  of objects only (probably except the DSO without designations). English and localized names may be found on the fly.

gzotti:
Yes, but I think it is inconsistent between importing old bookmarks and storing new objects! Or is it working correctly as it is now? 
I would store (one most common) englishName into the JSON for reference (for the people editing the JSON manually!), but on loading we better retrieve the name from the currently active list.

alex-w:
> Yes, but I think it is inconsistent between importing old bookmarks and storing new objects! Or is it working correctly as it is now? I would store (one most common) englishName into the JSON for reference (for the people editing the JSON manually!), but on loading we better retrieve the name from the currently active list.

Yeah, but in any way current observing lists are not compatible with old bookmarks

gzotti:
Yes, the JSON formats are not the same, I know. I just saw that when loading a list there is a flip between objType and type. See ObsListDialog lines 452ff and compare to lines 607ff when importing bookmarks. Is this correct? It at least requires a note in the source code that whoever developed this originally knew what they were doing. My "Caveat" comment is that of a code reviewer indicating a possible source of error.

alex-w:
> Yes, the JSON formats are not the same, I know. I just saw that when loading a list there is a flip between objType and type. See ObsListDialog lines 452ff and compare to lines 607ff when importing bookmarks. Is this correct? It at least requires a note in the source code that whoever developed this originally knew what they were doing. My "Caveat" comment is that of a code reviewer indicating a possible source of error.

Well, we should 2 important features:
a) the tool for import/exports observing lists with some specific extension - `.solist` for example (**S**tellarium **O**bserving **List**)
b) the tool for import bookmarks - a tool for converting bookmarks into observing lists (remember - the bookmarks has extension `.json`)

The current OL feature has a partially implemented of both tools.

gzotti:
Currently both list formats have the too generic json extension, but importing the old format and adding missing data works well. However, this does not answer my question about the type/objType issue. The new format (without I18n names) could be named .ol3. 

The major task of this PR was to remove the separate edit dialog and the repeated need to load the JSON file, and making the code more readable. I think the current state is feature-equivalent w.r.t. the JSON files, i.e. has the same errors in data conversion. Now we should try to remove the errors in the conversions and file format. If this warrants a new format (version 3) with e.g. .ol3 ending, it's OK for me.

alex-w:
> Currently both list formats have the too generic json extension, but importing the old format and adding missing data works well. However, this does not answer my question about the type/objType issue. The new format (without I18n names) could be named .ol3.

The `type` option was introduced in first implementation of OL and this is my fault - it was not complete, because it has stored wrong data by the fact. To resolve the problem and for backward compatibility the `objtype` was introduced later. Of course the type of object should be found on the fly. I think we should store one parameter for definition of type of object and this parameter should had data for the correct finding of the object (see second parameter for StelObjectMgr::searchByName() method)

> 
> The major task of this PR was to remove the separate edit dialog and the repeated need to load the JSON file, and making the code more readable. I think the current state is feature-equivalent w.r.t. the JSON files, i.e. has the same errors in data conversion. Now we should try to remove the errors in the conversions and file format. If this warrants a new format (version 3) with e.g. .ol3 ending, it's OK for me.

.ol3 is OK for me too

gzotti:
The current import can read old bookmarks.json format and new obslist.json formats. We can change or limit this to .ol3 files (new format only), but this is another PR. I am not sure after the month of waiting, but I think you can also import another obslist file (with several obslists) and import all lists. 

This still does not solve the type issue/bug. The PR is a branch in the main repo. Do you have time to take over and do what has to be done? I think the type/objType should be fixed first. If this requires a different file format again, this should be a well-designed one-last-time step, and then should not be changed. But I think it is not required to remain compatible with the 1.0-1.2 format 2.0 which effectively will have been used only for 6 months.

gzotti:
@alex-w please review this. As far as I can see it is feature equivalent to the old version, but has one dialog less. However, the type/objType mess is still here to be cleared up. Further, I18N names should be loaded on the fly, and never stored to (or at least never read from) JSON. New file endings can be introduced later.

alex-w:
> @alex-w please review this. As far as I can see it is feature equivalent to the old version, but has one dialog less. However, the type/objType mess is still here to be cleared up. Further, I18N names should be loaded on the fly, and never stored to (or at least never read from) JSON. New file endings can be introduced later.

I tried run updated OL without old data - neither bookmarks or lists - and I got the crash. But I did not reproduced the crash later. 

After adding few objects into list I see something like:
```
                {
                    "constellation": "Sct",
                    "dec": "-6°16'18\"",
                    "designation": "Wild Duck Cluster",
                    "fov": 0,
                    "isVisibleMarker": true,
                    "jd": 0,
                    "landscapeID": "",
                    "location": "",
                    "magnitude": "5.80",
                    "nameI18n": "Скопление Дикая утка",
                    "objtype": "рассеянное звёздное скопление",
                    "ra": "18h51m04.4s",
                    "type": "Nebula"
                },
                {
                    "constellation": "Aql",
                    "dec": "-0°49'25\"",
                    "designation": "HIP 99473",
                    "fov": 0,
                    "isVisibleMarker": true,
                    "jd": 0,
                    "landscapeID": "",
                    "location": "",
                    "magnitude": "3.20",
                    "nameI18n": "Альмизан III",
                    "objtype": "двойная звезда",
                    "ra": "20h11m17.3s",
                    "type": "Star"
                },
```

I agree - no need saving `nameI18n` field into JSON - it should be obtain on-the-fly. Probably similar behaviour should be for `objtype` field or, at least, it should be English only. I see some inconsistense for DSO objects - why not use designations for it?
Like:
```
QString dsoName = object->getDSODesignation();
if (dsoName.isEmpty())
	dsoName = object->getDSODesignationWIC()
if (dsoName.isEmpty())
	dsoName = object->getNameI18n();
```

gzotti:
The `objectMap.value(KEY_TYPE)` should actually be the object's class name, so that it can be retrieved in the objectMgr.findAndSelect(item.designation, type) call. It can have values like "Planet", "Star", "Nebula", ... 
This can be stored in item.type (and I would prefer to rename this to item.class for clarity), but display of this column can be suppressed just like display of objectUUID.  

The item.objtype on the other hand could be a temporal variable for display purposes only. Here we may show the various subtypes like "Moon", "Cubewano", "Dark Nebula", "Spiral Galaxy", ... These have to be retrieved on-the-fly when loading the list into the TreeView. It may be allowed to store it into the JSON list (for reference when manually editing data lists later), even in translated form (because it is "write only"), however, it must never be evaluated during loading. 

@alex-w do you agree? 

And can you please check whether this interpretation of the items is correct? What is the purpose of the isVisibleMarker entry? When is it evaluated? Can it be toggled by the user? Why is that not just an entry with type=Marker? 

```
struct observingListItem
	{
	public:
		QString designation;   //!< Relates to designation in the JSON file and englishName in most classes. DSO 
		QString nameI18n;      //!< Relates to name in the JSON file.
		QString type;          //!< type oriented on Stellarium's object class: Star, Planet (also for sun and moons!), Nebula, ...
		QString objtype;       //!< More physical type description: star (also Sun!), planet, moon, open cluster, galaxy, region in the sky, ...
		QString ra;            //!< Optional, position at data jd; useful for moving objects
		QString dec;           //!< Optional, position at data jd; useful for moving objects
		QString magnitude;     //!< NOT a redundant bit of information! In case of moving objects, it spares computing details on loading.
		QString constellation; //!< NOT a redundant bit of information! In case of moving objects, it spares computing positions on loading.
		double jd;             //!< optional: stores date of observation
		QString location;      //!< optional: should be a full location encoded with StelLocation::serializeToLine()
		QString landscapeID;   //!< optional: landscapeID of landscape at moment of item creation.
		double fov;            //!< optional: Field of view
		bool isVisibleMarker;  //!< WHAT IS THE PURPOSE OF THIS? 
```
I had hoped to get that solved by January 5. To fix that finally still for 23.1, I suggest we concentrate on this one now for the weekend.

alex-w:
> The `objectMap.value(KEY_TYPE)` should actually be the object's class name, so that it can be retrieved in the objectMgr.findAndSelect(item.designation, type) call. It can have values like "Planet", "Star", "Nebula", ... This can be stored in item.type (and I would prefer to rename this to item.class for clarity), but display of this column can be suppressed just like display of objectUUID.
> 
> The item.objtype on the other hand could be a temporal variable for display purposes only. Here we may show the various subtypes like "Moon", "Cubewano", "Dark Nebula", "Spiral Galaxy", ... These have to be retrieved on-the-fly when loading the list into the TreeView. It may be allowed to store it into the JSON list (for reference when manually editing data lists later), even in translated form (because it is "write only"), however, it must never be evaluated during loading.
> 
> @alex-w do you agree?

Yes, I agree.

> And can you please check whether this interpretation of the items is correct? What is the purpose of the isVisibleMarker entry? When is it evaluated? Can it be toggled by the user? Why is that not just an entry with?

Let imagine that you find some object via SIMBAD and this object is not in Stellarium database - it will be added as custom object (type=Marker by the fact, but without visible marker) which you can add into OL (type=Marker, isVisibleMarker=false). Of course you can put markers on the sky and add some of them (or all) into OL (type=Marker, isVisibleMarker=true).

> 
> ```
> struct observingListItem
> 	{
> 	public:
> 		QString designation;   //!< Relates to designation in the JSON file and englishName in most classes. DSO 
> 		QString nameI18n;      //!< Relates to name in the JSON file.
> 		QString type;          //!< type oriented on Stellarium's object class: Star, Planet (also for sun and moons!), Nebula, ...
> 		QString objtype;       //!< More physical type description: star (also Sun!), planet, moon, open cluster, galaxy, region in the sky, ...
> 		QString ra;            //!< Optional, position at data jd; useful for moving objects
> 		QString dec;           //!< Optional, position at data jd; useful for moving objects
> 		QString magnitude;     //!< NOT a redundant bit of information! In case of moving objects, it spares computing details on loading.
> 		QString constellation; //!< NOT a redundant bit of information! In case of moving objects, it spares computing positions on loading.
> 		double jd;             //!< optional: stores date of observation
> 		QString location;      //!< optional: should be a full location encoded with StelLocation::serializeToLine()
> 		QString landscapeID;   //!< optional: landscapeID of landscape at moment of item creation.
> 		double fov;            //!< optional: Field of view
> 		bool isVisibleMarker;  //!< WHAT IS THE PURPOSE OF THIS? 
> ```
> 
> I had hoped to get that solved by January 5. To fix that finally still for 23.1, I suggest we concentrate on this one now for the weekend.

gzotti:
@alex-w can you repeat your tests please?
I have not touched the isVisibleMarker AFAIK, but my workflows are often different from yours, so don't know if it works different from 1.2.

The current solution allows editing in the same dialog by reconfiguring a few buttons. The currently edited list is stored (replaces the list state of before editing) when 'Save list' is pressed.  The list JSON file is written on exiting Stellarium. 
When exiting Stellarium during list editing, the current edits are not saved. 

I hope this works finally free from further major surprises...

alex-w (approved):
Everything what I tested are works as expected

PR #2966, opened 2023-01-03
Switch old-style landscape to fragment-shader-based direction computation

10110111:
This is the hardest kind of landscape to switch to fragment-shader-based computation of direction. First of all, because it has lots of parameters, which have to be tested to make sure they weren't broken.

But the most problematic part is performance: the default Guereins landscape has (8 sides)×2 + ground + fog, which means that we have to run the fragment shader 18 times, and this really hard to make fast. In particular, with the default Guereins landscape the original Stellarium before this PR runs at 34 FPS on my machine, while with fragment-shader-based computations it runs at 22 FPS. I pushed a second, experimental, commit that stitches all 8 sides of this landscape into one, so that we have only 4 fragment shader invocations, and this results in 29 FPS, which is faster but still not as fast as before this PR.

I didn't modify the other landscapes, I need your feedback about the current state (with and without the modification of Guereins): whether the performance is acceptable, any bugs you find with custom landscapes, etc.

#### Original

stellarium-013

#### After this PR

stellarium-014

#### Original

stellarium-001

#### After this PR

stellarium-002

### DIFF
--- a/src/core/modules/Landscape.cpp
+++ b/src/core/modules/Landscape.cpp
@@ -515,12 +515,20 @@ void LandscapeOldStyle::load(const QSettings& landscapeIni, const QString& lands
 	// Load sides textures
 	nbSideTexs = static_cast<unsigned short>(landscapeIni.value("landscape/nbsidetex", 0).toUInt());
 	sideTexs = new StelTextureSP[static_cast<size_t>(nbSideTexs)*2]; // 0.14: allow upper half for light textures!
+
+	// Avoid mip mapping when we can't supply custom dFdx/dFdy to texture sampler.
+	// The noise that may appear due to linear filter is must less distracting than
+	// the seam that will appear due to wrong mip level computed from
+	// default-computed derivatives.
+	const bool textureGradSupported = StelOpenGL::globalShaderPrefix(StelOpenGL::FRAGMENT_SHADER).contains("textureGrad_SUPPORTED");
+	const auto texParams = StelTexture::StelTextureParams(textureGradSupported, GL_LINEAR, GL_CLAMP_TO_EDGE, textureGradSupported);
+
 	for (unsigned int i=0; i<nbSideTexs; ++i)
 	{
 		QString textureKey = QString("landscape/tex%1").arg(i);
 		QString textureName = landscapeIni.value(textureKey).toString();
 		const QString texturePath = getTexturePath(textureName, landscapeId);
-		sideTexs[i] = StelApp::getInstance().getTextureManager().createTexture(texturePath);
+		sideTexs[i] = StelApp::getInstance().getTextureManager().createTexture(texturePath, texParams);
 		// GZ: To query the textures, also keep an array of QImage*, but only
 		// if that query is not going to be prevented by the polygon that already has been loaded at that point...
 		if ( (!horizonPolygon) && calibrated ) { // for uncalibrated landscapes the texture is currently never queried, so no need to store.
@@ -538,7 +546,7 @@ void LandscapeOldStyle::load(const QSettings& landscapeIni, const QString& lands
 		if (textureName.length())
 		{
 			const QString lightTexturePath = getTexturePath(textureName, landscapeId);
-			sideTexs[nbSideTexs+i] = StelApp::getInstance().getTextureManager().createTexture(lightTexturePath);
+			sideTexs[nbSideTexs+i] = StelApp::getInstance().getTextureManager().createTexture(lightTexturePath, texParams);
 			if(sideTexs[nbSideTexs+i])
 				memorySize+=sideTexs[nbSideTexs+i]->getGlSize();
 		}
@@ -581,7 +589,7 @@ void LandscapeOldStyle::load(const QSettings& landscapeIni, const QString& lands
 	}
 	const QString groundTexName = landscapeIni.value("landscape/groundtex").toString();
 	const QString groundTexPath = getTexturePath(groundTexName, landscapeId);
-	groundTex = StelApp::getInstance().getTextureManager().createTexture(groundTexPath, StelTexture::StelTextureParams(true));
+	groundTex = StelApp::getInstance().getTextureManager().createTexture(groundTexPath, texParams);
 	if (groundTex)
 		memorySize+=groundTex->getGlSize();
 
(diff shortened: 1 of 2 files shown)


gzotti:
The result looks fabulous, indeed. I was under the apparently false assumption that "pushing everything onto the GPU" was the way to speed-up things, so the speed penalty is sad to see.  

I agree this is the most complicated landscape style, AFAIK it was the original (V0.6.1/2004?). The advantage is the possible high resolution with multiple textures exceeding the (old 2kx2k textures, nowadays some systems like GLES2 still have 4k limit), which is only advantageous in zoomed-in views, i.e. the problems of wide angle are less important. So this is almost contradictory. 

I can imagine it will complicate things further, but is there a way to use the "classic" path in non-wideangle views and the shader-based rendering only in wide-angle views? Basically make something like a spherical landscape with 4k texture on the fly and switch when hfov exceeds 4096 pixels?   Or make it manually switchable to allow using on weaker systems?

10110111:
> I can imagine it will complicate things further, but is there a way to use the "classic" path in non-wideangle views and the shader-based rendering only in wide-angle views?

The classic render is distorted due to linear interpolation of texture coordinates over low-polygon mesh, so we'll have jumps in the image shape when zooming in and out of the scale when the switch occurs.

I guess it would work as a last resort option, but I let's try my other test. Could you (both @gzotti and @alex-w) test my [landscape-fragment-shader-projectors--use-stencil-buffer](https://github.com/10110111/stellarium/tree/landscape-fragment-shader-projectors--use-stencil-buffer) branch on various machines, with NVIDIA on Windows and Linux, Intel on Windows, Intel/M1/M2 on macOS? I wonder whether you find any improvement in this stencil-based approach vs this PR. On my Intel@Linux system I got reduction of performance, but this is only a single data point, and it appears to behave strangely performance-wise, so I suppose it might be a driver issue.

10110111:
Just tested on a remote machine (via VNC) with NVIDIA GeForce GTX 750Ti, and the landscape performance appears to be not as problematic. So indeed, maybe Mesa/Intel/Linux is an outlier. Please do check on your machines.

gzotti:
To make testing easier with my usual PR workflow, could you please make a second PR for your second branch? Else I need help with git again.

10110111:
I removed the side-stitching workaround to make it easier to compare the two PRs.

gzotti:
First data point:
Win11/Geforce M 3070Ti with a few gridlines to bring framerate below 60fps. Guereins landscape, Preetham sky (SMS not compiled), 3440x1440 screen. Qt6.4.1. 

- Master: 53fps (fluctuating somewhat 50..53.7, but mostly on the fast side)
- PR #2966: 54fps. (53-55)
- PR #2969: 53-55fps. 

So, it seems on this admittedly high-end system there is no perceiveable performance difference, or even a mimimal improvement. Projection looks great. However, I admit I still have to produce (or find that one I had?) a high-res old_style with more than 16k width...

gzotti:
Next data point: Intel NUC i3-5010U (Intel HD5500), Ubuntu 22.04.1 kernel driver i915, Qt6.2.4, Mesa 22.0.5, OpenGL 4.6 Core Profile, 1280x1024 screen, Preetham atmosphere, Equatorial and zenith grids on to bring fps below 60, fov114°.
- Master: 44fps
- PR #2966: 42fps
- PR #2969: 38fps

Seems here is a slight penalty.

gzotti:
Next: Win11/Core i7-12700H, forced to use Intel Iris Xe (disabled main.cpp L70/71). OpenGL 3.30 CoreProfile. Again large screen.
- Master: 26fps
- PR #2966: 23.4-24.4fps
- PR#2969: 23.7 (+/- 1) fps

Minimal performance loss, in line with the NUC (previous Intel). I cannot say which of the new PRs is faster. 
I have meanwhile found a 32-tile old_style LS, the Hawaiian Star Compass from our landscape archive. It works well. Overall texture size is small though.

10110111:
@alex-w could you check this on Macs?

alex-w:
Data point: macOS 13.1/MBP/M1, Qt 6.4.1, Guereins landscape, Preetham sky, screen 2560x1600

* Master: 17-70 fps (FOV 90°)
* PR #2966: 17-48 fps (FOV 90°)
* PR #2969: 17-80 fps (FOV 90°) - no visible landscape and no errors in the log

gzotti:
Next data point: Intel i5-4570 (Intel HD4600), Win10home, Qt6.4.0, OpenGL 3.3 Core Profile, 3440x1440 screen, Preetham atmosphere, Equator and ecliptic on, fov163°.

- Master: 13fps
- PR #2966: 9.7 fps
- PR #2969: 6.8...7.6 fps

Seems here is a noticeable penalty for 2969 over 2966, and both are costly compared to the old solution.

@alex-w I usually move the screen around to force redraw and not fall into energy-saver mode. Your 17fps lower end looks like that.

10110111:
> I usually move the screen around to force redraw and not fall into energy-saver mode.

Best of all is to set a large minimum and maximum frame rate, e.g. 999 FPS, when testing performance.

gzotti:
Next data point:
Win11/Core i7-12700H, Geforce M 3070Ti with a few gridlines to bring framerate below 60fps. Guereins landscape, Preetham sky (SMS not compiled), 3440x1440 screen, 129° fov. This time with Qt5.12.12.

- Master: 46-48fps 
- PR #2966: 44-45fps
- PR #2969: 45-47fps

Insignificant change, hard to say what is the fastest.

These are not directly comparable to the values above, where I did not record fov.

ANGLE
- Master: 44-50fps
- PR #2966: 44-46fps
- PR #2969: ERROR (no ground visible)
 QOpenGLShader::compile(Fragment): ERROR: 0:30: 'gl_FragDepth' : undeclared identifier 
[log_ANGLE-pr2969.txt](https://github.com/Stellarium/stellarium/files/10369773/log_ANGLE-pr2969.txt) (This is really pr/2969. The branch ID fails to update for small updates that don't force a cmake run. )

MESA
- Master: 11.6fps
- PR #2966: 2.6fps Sorry, that's no fun anymore :-(
- PR #2969: ERROR (no ground visible)
QOpenGLShader::compile(Fragment): 0:72(5): error: could not implicitly convert operands to relational operator
[log_MESA_pr2969.txt](https://github.com/Stellarium/stellarium/files/10369778/log_MESA_pr2969.txt)

10110111:
Do the textures for the sides have to have the same dimensions for each side? Can I rely on it and fail if this is not true?

gzotti:
I think this is not guaranteed. Yes, see the starcompass "landscape". (Stellarium.org: Landscapes->Australasia)

Ah right, I still wanted to check those with the Raspi3.

PR #2982, opened 2023-01-07
Make it possible to restrict texture size of GUI panel button pixmaps

10110111:
A few times @gzotti complained that the new high-resolution pixmaps used for GUI panel buttons make performance worse on low-spec systems like Raspberry or something like that. This patch introduces a config file setting, `gui/pixmaps_scale`, which can vary from 1 to 5, to reduce the pixmaps loaded from resources to corresponding scale factor.

E.g., on a machine with a low-DPI monitor, where scale factor higher than 100% is not supposed to be useful, we can set this setting to `1`, and avoid using excessive VRAM for 500%-scale textures. Or, if the scaling factor of the monitor is higher than 100% but much lower than 500%, we can set that exact scaling factor as the maximum scale of the pixmaps, thus reducing VRAM use.

I'm not sure if this 500% increase of texture size really influences performance, but, since @gzotti complained more than once, I propose this patch, which might help.

### DIFF
--- a/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
+++ b/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
@@ -173,13 +173,13 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
 	int scale=static_cast<int>(lineHeight*scv*0.01f);
 	// TODO: change this load-once to interactively editable value of scaling coefficient
 	QPixmap pa(":/graphicGui/btTimeRewind-on.png");
-	QPixmap prevArrow = pa.scaledToHeight(scale * StelButton::getPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
+	QPixmap prevArrow = pa.scaledToHeight(scale * StelButton::getInputPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
 	QPixmap paOff(":/graphicGui/btTimeRewind-off.png");
-	QPixmap prevArrowOff = paOff.scaledToHeight(scale * StelButton::getPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
+	QPixmap prevArrowOff = paOff.scaledToHeight(scale * StelButton::getInputPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
 	QPixmap na(":/graphicGui/btTimeForward-on.png");
-	QPixmap nextArrow = na.scaledToHeight(scale * StelButton::getPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
+	QPixmap nextArrow = na.scaledToHeight(scale * StelButton::getInputPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
 	QPixmap naOff(":/graphicGui/btTimeForward-off.png");
-	QPixmap nextArrowOff = naOff.scaledToHeight(scale * StelButton::getPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
+	QPixmap nextArrowOff = naOff.scaledToHeight(scale * StelButton::getInputPixmapsDevicePixelRatio(), Qt::SmoothTransformation);
 
 	StelActionMgr* actionMgr = StelApp::getInstance().getStelActionManager();
 	QString ocularsGroup = N_("Oculars"); // Possible group name: Oculars on-screen control panel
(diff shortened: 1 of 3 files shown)


gzotti:
I am trying to at least keep (actually already: restore) some way of running on RPi3 (at ~3-5fps it still may have its use), even if some features will no longer be available in full quality. Currently it fails when zooming in while showing the DSO texture loading bar. It seems now it barely loaded again (I saw M42, but it crashed again on further zoom in). Logfile says nothing, though.  

Not loading DSO and just using icons would probably be OK (seems stable). Mars renders, but the Moon has no surface rendering. Equirectangular projection with "spherical" landscapes works. If this PR reduces texture memory used for the actual buttons, even just a few kB, I approve it. Maybe change name to `[gui]/pixmap_scale`. I may try reducing DSO texture sizes (or even numbers) as well. A way to limit texture size on loading may also be good for such limited systems. 

But I have severely not enough time for testing all platforms with the many GLSL updates (although OTOH I welcome them!).

10110111:
The main question is: does this patch change anything usefully? If the change is not noticeable, this would be just a useless complication.

10110111:
Also, it would be useful to have the output of `glxinfo -l` to know the limits of this platform.

gzotti:
I wonder about texture compression. The RPi supports ETC1 compression, but what does it take to activate this?

I just converted all DSO textures to 64x64, and it runs well again. Better blurred than crash, IMO. Maybe we can configure another (config.ini and/or GUI-available?) switch for "limit texture use"?

 
[RPi3_glxinfo-l.txt](https://github.com/Stellarium/stellarium/files/10368162/RPi3_glxinfo-l.txt)
[log_RPi3-master-a195de0aae8186_out-of-memory.txt](https://github.com/Stellarium/stellarium/files/10368163/log_RPi3-master-a195de0aae8186_out-of-memory.txt)
[log_pr2982_maybeSlightlyBetter.txt](https://github.com/Stellarium/stellarium/files/10368164/log_pr2982_maybeSlightlyBetter.txt)

10110111:
> `GL_MAX_TEXTURE_SIZE = 2048`

So this is why the Moon is not rendered: our texture is 4096×2048. May be worth it to resize each texture being loaded if it doesn't fit OpenGL limits.

10110111:
> I just converted all DSO textures to 64x64, and it runs well again.

Where are they stored? What's their initial size?

gzotti:
> > `GL_MAX_TEXTURE_SIZE = 2048`
> 
> So this is why the Moon is not rendered: our texture is 4096×2048. May be worth it to resize each texture being loaded if it doesn't fit OpenGL limits.

Yes, such test and auto-resize should be added in any case! 2048 was the max texture size ever used just for that reason. (And as said, this is why old_style landscapes were required and are still useful.)

> > Video memory: 743MB
> 
> This is a lot. Do DSO really take so much to hit this limit?

I am not sure if this number is correctly reported. The total RAM of this device is 1GB, and raspi-config is used to set a part of this memory for GPU use (64/128/256MB). The documentation is unclear, though, about the right split. Some say the new driver does not use it (set to 64MB), but now I tried 256MB. Maybe this is the total of remaining memory, and the 256MB is even lost...

gzotti:
> > I just converted all DSO textures to 64x64, and it runs well again.
> 
> Where are they stored? What's their initial size?

nebulae/default/*.png
Sizes vary from 256x256 to 2048x2048 IIRC, depending on size in the sky. They could as well become JPG, but I understand this would not reduce texture mem needed. Most are 24bit sRGB PNG, others are 256color sRGB PNGs, a few 256 shade B/W PNG (reported by ImageMagick's identify).
 
> Also, the log doesn't appear to say that the crash is due to lack of memory. How did you come to this conclusion?

The program hangs for a minute, then exits.  dmesg indicates memory issues (The exact wording comes from the VC4 driver. I think there are no gl_error messages available on this platform. )

gzotti:
> The main question is: does this patch change anything usefully? If the change is not noticeable, this would be just a useless complication.

How does it work currently? You load large button textures and show them (usually) smaller? Why not make the actual textures the same size as later used on the screen? Or do you fight the rare occasion when a Stellarium window is moved between different screens of different scalings?

10110111:
> The program hangs for a minute, then exits. dmesg indicates memory issues

This sounds like out of system RAM, rather than VRAM. Namely, heavy swapping (which looks like a hang to the user), and then process termination. Does dmesg say something like "Out of memory: Killed process ..."?

> How does it work currently? You load large button textures and show them (usually) smaller?

I load large textures, and resize them to the configured scale. Then I render them as usual, scaling according to current scale factor as needed.

> Why not make the actual textures the same size as later used on the screen? Or do you fight the rare occasion when a Stellarium window is moved between different screens of different scalings?

Exactly. Stellarium is not always run in fullscreen mode. And when it's windowed, it's not really rare to move the window around. If you want to use one particular scaling factor you want don't care about differences between screens, just put this value in the config. This is the purpose of this PR, after all.

gzotti:
The texture decimation (#3009) seemed to have remedied the basic problem around RPi3. Can you estimate how many MB this patch here would save additionally? The RPi3 has no separate texture memory, just 1GB of system memory for everything. Assume no screen scaling, just FullHD screen (at most) for the RPi3. If this is a few MB only, we can probably close this if you think it just complicates matters. If it's more, we should use it.

10110111:
> Can you estimate how many MB this patch here would save additionally?

About 10 MiB. Maybe less.

gzotti:
Hmm, that's just the size of about 3-5 large DSO textures. Still it is "something", so if sufficiently commented, I still would use it.

gzotti (approved):
I think conserving memory has its merits on limited systems, so yes, Approved!

PR #3002, opened 2023-01-18
Do computations in linear color space when rendering planets

10110111:
Please don't squash if you merge! The commits were carefully separated into logically independent changes!

The main goal of this PR is to make the rocky objects without atmosphere (e.g. the Moon, Mercury, Io etc.) look more correct. In particular, the terminator of the Moon looks much sharper in real life than in Stellarium currently: the brightness comes to zero linearly, while Stellarium renders it superlinearly, which unrealistically smoothens the transition.

Doing this naively reveals a few problems with planets that do have atmospheres: gas giants, Venus etc. These can't be rendered with Oren-Nayar model (much less the special case of it, the Lambert model) due to multiple scattering in the atmosphere that does smoothen the terminator. To work around this I added a hack that emulates the smoothing that used to happen when computations were done in nonlinear sRGB color space. The proper fix would be to find a better BRDF (which may vary between planets).

Also I moved the Lambert model computations from the vertex shader to the fragment shader, because the sharper terminator of the rocky planets reveals the blocky nature of the vertex mesh. 

#### The Moon how it used to look

stellarium-011

#### The Moon how it looks now

stellarium-012

#### A photo of real-life moon, developed from RAW with minimal processing

Screenshot_2023-01-19_01-29-51

#### Saturn + Io how they used to look

stellarium-009

#### Saturn + Io how they look now

stellarium-007

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/data/shaders/planet.frag
+++ b/data/shaders/planet.frag
@@ -175,6 +175,22 @@ lowp float outgasFactor(in mediump vec3 normal, in highp vec3 lightDir, in mediu
     return opac;
 }
 
+vec3 srgbToLinear(vec3 srgb)
+{
+	vec3 s = step(vec3(0.04045), srgb);
+	vec3 d = vec3(1) - s;
+	return s * pow((srgb+0.055)/1.055, vec3(2.4)) +
+		   d * srgb/12.92;
+}
+
+vec3 linearToSRGB(vec3 lin)
+{
+	vec3 s = step(vec3(0.0031308), lin);
+	vec3 d = vec3(1) - s;
+	return s * (1.055*pow(lin, vec3(1./2.4))-0.055) +
+		   d *  12.92*lin;
+}
+
 void main()
 {
     mediump float final_illumination = 1.0;
@@ -201,6 +217,10 @@ void main()
                 mediump float ring_radius = length(P + u * ray);
                 mediump float s = (ring_radius - innerRadius) / (outerRadius - innerRadius);
                 lowp float ringAlpha = texture2D(ringS, vec2(s, 0.5)).w;
+
+				// FIXME: this compensates for too much transparency in the texture (see e.g. Saturn)
+				ringAlpha = pow(ringAlpha, 1./2.2);
+
                 if(ring_radius > innerRadius && ring_radius < outerRadius)
                     final_illumination = 1.0 - ringAlpha;
             }
@@ -343,8 +363,10 @@ void main()
     lum*=shadow;
 #endif
 
+	mediump vec3 ambientLightToUse = srgbToLinear(ambientLight); // FIXME: this should be supplied as linear
+	mediump vec3 diffuseLightToUse = srgbToLinear(diffuseLight); // FIXME: this should be supplied as linear
     //final lighting color
-    mediump vec4 litColor = vec4(lum * final_illumination * diffuseLight + ambientLight, 1.0);
+    mediump vec4 litColor = vec4(lum * final_illumination * diffuseLightToUse + ambientLightToUse, 1.0);
 
     //apply texture-colored rimlight
     //litColor.xyz = clamp( litColor.xyz + vec3(outgas), 0.0, 1.0);
@@ -377,7 +399,9 @@ void main()
 	// We undo the intensity range and gamma transformations here, but leave the
 	// maximum at 1.0 instead of 0.4.
 	texColor.rgb = vec3(0.4) + 0.6 * texColor.rgb;
-    texColor.rgb = pow(texColor.rgb, vec3(2.8/2.2));
+    texColor.rgb = pow(texColor.rgb, vec3(2.8));
+#else
+    texColor.rgb = srgbToLinear(texColor.rgb);
 #endif
 
     mediump vec4 finalColor = texColor;
@@ -400,6 +424,7 @@ void main()
     if(final_illumination < 0.9999)
     {
         lowp vec4 shadowColor = texture2D(earthShadow, vec2(final_illumination, 0.5));
+		shadowColor.rgb = srgbToLinear(shadowColor.rgb);
         finalColor =
 		eclipsePush*(1.0-0.75*shadowColor.a)*
 		mix(finalColor * litColor, shadowColor, clamp(shadowColor.a, 0.0, 0.7)); // clamp alpha to allow some maria detail.
@@ -413,7 +438,7 @@ void main()
     //apply white rimlight
     finalColor.xyz = clamp( finalColor.xyz + vec3(outgas), 0.0, 1.0);
 
-    FRAG_COLOR = finalColor;
+    FRAG_COLOR = vec4(linearToSRGB(finalColor.rgb), finalColor.a);
     //to debug texture issues, uncomment and reload shader
     //FRAG_COLOR = texColor;
 }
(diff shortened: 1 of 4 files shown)


gzotti:
Isn't that too sharp/dark now? Exif date is 2008-05-11-22-29-31 CET


2008-05-11-22-29-31_Mond_Russentonne_Test-0002

gzotti:
This was 2005-05-12 around 22:20 CEDT (20:20UT) with Vienna's 68cm refractor. 

Moon-2005-05-12-9933x4120

10110111:
> Isn't that too sharp/dark now?

Not sure what exactly you mean. The general trend is the same in your photo and in this PR (checked on Mare Serenitatis).

The large contrast in this PR comes (at least partially) from too high-contrast texture. I think the authors of it did more than just applying extra gamma as they say: at least the maria are too dark. This I'm planning to work on later, using the linear LROC data as a source.

gzotti:
I hope I can check your PR on weekend. At first glance the new solution appears to have too large dark craters. But it's a mixed effect with the other change you made recently, where crater shadows are now so strong. These stand out now even stronger.

10110111:
BTW do your photos represent brightness faithfully, or have they been manipulated with some non-linear tone curve? Would be best if you have linear-color raw images.

gzotti:
Processed some 15 years ago... I can provide the CR2 raw images.

10110111:
If we replace the texture with a more correct one that I made from LROC data (lacking the poles), only white-balanced to make it look gray, this would look as in the screenshot below. It already looks more consistent with the JPEG you posted regarding the contrast of the maria, but the JPEG seems to be manipulated in the brightly-lit part, so to compare more fairly it's better to have the CR2.

Note that in fact, Oren-Nayar model is not a good model to describe the BRDF of the lunar surface. NASA uses Hapke model for this, and the albedo map was also created from LROC photos using this model. So we shouldn't expect perfect match with Oren-Nayar. Instead, this PR is just one more step to improve, with further ones to be done.

moon-with-better-texture

Atque:
With that new texture, the low resolution of the craters is obvious. Maybe a 4k or even 8k crater texture would be needed. Could such be generated?

10110111:
> the low resolution of the crater shadows

It's not the low resolution, it's them being aliased. I have a patch locally that enables antialiasing for fragment shader output (basically, supersampling), but this comes at a performance cost. The result looks like this:

stellarium-013

gzotti:
Here are 2 CR2s with different exposure from that 2008-05-11 22:29:31 CET (or CEDT, I cannot rule this out, unfortunately)
[CR2.zip](https://github.com/Stellarium/stellarium/files/10466439/CR2.zip)

gzotti:
Here are CR2s from the big refractor.
[2005_05_12_UniStw.zip](https://github.com/Stellarium/stellarium/files/10467310/2005_05_12_UniStw.zip)

10110111:
So, using a minimally-processed version of your photo, converted to grayscale, and comparing it with Stellarium at this PR + WIP texture + conversion to grayscale + 1.5px Gaussian blur in the GIMP, I get the results below. Switching between them back and forth, I think the photo and this PR version are similar enough, while the photo and before-this-PR version are not as close. The similarity is not ideal, but it's still quite usable IMO. Of course, this is not the result of purely this PR, because the texture is not yet there, but still, this shows that this PR does make sense.

### Photo

2008-05-11-22-29-31_Mond_Russentonne_Test-0002 gray

#### Stellarium after this PR, + WIP texture

stellarium-015gray-blurred

#### Stellarium before this PR, + WIP texture

stellarium-016gray-blurred

10110111:
For reproducibility: the above screenshots were taken with the time of the photo taken as UTC+1 timestamp, position on the Earth at 26°54'48"N, 71°26'45"E (this position was set to reproduce the orientation).

The texture used instead of current `moon_4k.jpg` is this one:
c-wb-4096+

Atque:
Indeed the new texture + PR looks more realistic. Does it look better for crescents and gibbouses?

gzotti:
OK, I did not object, just wanted to allow better comparison. Yes, when you can add polar zones to your texture, it should be OK.

10110111:
> Does it look better for crescents and gibbouses?

It should, but this all is limited by the fact that our model is a sphere, rather than a selenoid, so the shadows in the craters on the limb are visible when they shouldn't be due to being behind the crater walls.

> I did not object, just wanted to allow better comparison

OK, so shall I merge this?

gzotti:
Did you complete your texture?

10110111:
> Did you complete your texture?

No, and it's not a part of this PR. It'll take quite a bit of effort, because the data for the poles need to go from other sources.

gzotti:
Please check whether you can fix some brightness issues which always appear when trying to improve the Moon.

image

Left is this PR, right is V1.2. Atmosphere off. The highlands around Tycho are burned out worse than 1.2 in this PR.

Will the new texture come within days? (At least by, say, late February?) It will likely require yet another round of adaptation. If the answer is yes, I am happy to wait for the fix after the texture. 

Things to check also include mixing with daylight/twilight.

10110111:
> Please check whether you can fix some brightness issues which always appear when trying to improve the Moon.

I can't reproduce them. See the screenshot:

stellarium-010

> Will the new texture come within days? (At least by, say, late February?)

Days most likely no. Late February is possible I think (though I can't guarantee it, of course).

gzotti:
Zoom out a bit so that lunar diameter ~1/4...1/3 of vertical fov.

10110111:
Still doesn't seem to reproduce...

stellarium-010

10110111:
I mean, it does become blown out when FoV is 4.5°, but at this point it doesn't seem to be important, because the details are quite small anyway.

gzotti:
Ah right, try with Lunar scale-up=4.  There are so many settings...

10110111:
OK, I see it now. Let's wait until I make the new texture, and then fix this if it still remains. With the current partially-done one I don't get this problem.

10110111:
Actually, I know what's wrong with the current texture and why it has such a wrong high contrast: I've completely missed this part of the description of the CGI Moon Kit:

> The intensity range (0.16, 0.4) was mapped into the full (0, 255) 8-bit range per channel.

This not only changes maximum brightness, but also shifts the minimum value to 0. After fixing this, the current `master` (without this PR) will look like this:

stellarium-010 1

instead of this:

stellarium-010 2

I'm gonna push this fix, after which I'll update this PR to resolve the conflicts.

10110111:
And for completeness, after fixing the wrong transformation of texture values and applying this PR, we'll get the following:

stellarium-010

10110111:
Now, this fixed version does display more of the highlight blowing-out problem. I think this should be fixed only after this PR is merged, since it's a concern of an unrelated part of the code (the strange scaling of color by 1.3–1.6, which is certain to blow up highlights).

10110111:
OK, I've pushed the fix for blown-out highlights, which appeared to be quite simple. I think nothing should interfere with this PR now.

If you merge it, please _do not_ squash!

10110111:
The result after all the fixes + this PR:

stellarium-011

gzotti (approved):
Compared esp. with the photo from the big refractor. WOW!

PR #3003, opened 2023-01-19
Improve OpenGL text drawing performance.

unpacklo:
### Description
Improve text drawing performance with OpenGL by using a static instance of `QOpenGLPaintDevice`. I still haven't narrowed down exactly what is causing the performance problem but I can see creating a new `QOpenGLPaintDevice` causes every `drawText()` call to create and destroy GL buffer objects.

Yields about a 2x speedup for my typical setup which draws lots of lines for the celestial sphere.

Fixes # (issue)

Before, 10 FPS:
image

After, 20 FPS:
image

### Type of change

### How Has This Been Tested?

Manually ran the change with a locally built stellarium and profiled with Instruments.

**Test Configuration**:
* Operating system: macOS 13.1
* Graphics Card: Radeon Pro 560X 4 GB
Intel UHD Graphics 630 1536 MB

## Checklist:

### DIFF
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -754,7 +754,7 @@ void StelPainter::drawText(float x, float y, const QString& str, float angleDeg,
 	}
 	else
 	{
-		QOpenGLPaintDevice device;
+		static QOpenGLPaintDevice device;
 		device.setSize(QSize(prj->getViewportWidth(), prj->getViewportHeight()));
 		// This doesn't seem to work correctly, so implement the hack below instead.
 		// Maybe check again later, or check on mac with retina..


alex-w (approved):
Thanks!

10110111:
Even though this improves performance, this is still a hack, so it deserves a comment explaining why this object is static and what OS environment needs this.

gzotti:
I tend to agree with @10110111 on this. The Qt docs of this class warns to be careful when interacting with the GL state. Can we guarantee that nothing can break when we leave a static object for re-use? (Or is it better to use an instance member variable in StelPainter and take care of this?)

10110111:
Instance member will be worse because it will partially restore the performance dip, since StelPainter is created and destroyed multiple times per frame.

gzotti:
Maybe then the StelPainter should be kept alive and handed around for a longer time?

gzotti:
Yes, I know. But if creation&destruction, i.e., state switching&restoring, happen so often, of course this is a plausible cause of slowdown. If we can hand around a StelPainter for a longer time, it should save a lot of overhead.

alex-w:
I got report about performance regression in Windows also :(

10110111:
> I got report about performance regression in Windows also

Does this PR improve the situation?

alex-w:
> Does this PR improve the situation?

- 1.2 release with equatorial grid - 17-35 fps
- the current branch with equatorial grid - 17-40 fps

macOS 13.2.1, MBP, M1

10110111:
The code that this PR changed has now been removed, hopefully the performance will be acceptable now. The PR itself is therefore obsolete.

gzotti:
We can close this, thanks to all participants!

PR #3009, opened 2023-01-21
Allow texture decimation

### LINKED ISSUE #550: Stellarium Crashes on Raspberry Pi 3B(+)
### Expected Behaviour
Stellarium will run complex  displays without crashing
### Actual Behaviour
Describe or maybe attach a screenshot?
Find a link to a video showing it. I found that I can get same result running a script
https://drive.google.com/file/d/1T6cfnYuQWCZZQtvNdP5xSiCQmGjgeyKd/view?usp=sharing

### Steps to reproduce
This crashes on raspberry Pi. I first discovered bug when doing rapid remote control changes. I found I could reproduce crash consistently just by running the script double_stars.ssc

It does run some scripts OK, however, I get it to consitently fail on this script

It runs fine on both Mac and Windows - Just Raspberry Pi

### System
* Stellarium version: <Name of downloaded installable file?>
This is shown crashing with https://github.com/Stellarium/stellarium/releases/download/v0.18.2/stellarium-0.18.2.tar.gz

Also tried with latest pull from github - crashes in slightly different place 

This was tried with Raspbian stretch.

* Graphics Card: <Manufacturer (likely Intel, NVidia, AMD?), Model (HD, Geforce, Radeon..., with model
 number), driver version?>
This has been tested on Raspberry Pi 3B and Raspberry Pi 3B+

Open GL tested
http://www.raspberryconnect.com/gamessoftware/item/314-trying_out_opengl_on_raspberry_pi_3

### Logfile 

You will notice in the journal below that this happens as it is crashing

Nov 16 08:46:37 hb-b827eb9089ee kernel: vc4-drm soc:gpu: failed to allocate buffer with size 5595136
Nov 16 08:46:37 hb-b827eb9089ee kernel: vc4-drm soc:gpu: failed to allocate buffer with size 16777216
Nov 16 08:46:37 hb-b827eb9089ee kernel: vc4-drm soc:gpu: failed to allocate buffer with size 16777216

[SystemJournal.txt](https://github.com/Stellarium/stellarium/files/2587199/SystemJournal.txt)

gzotti:
### Description

This is related to #2982. I want to make Stellarium again usable on a Raspberry Pi 3 which has significant problems with limited texture memory. This change allows to limit the texture size on loading the nebula catalog. It should be sufficient to show the "fuzzies" just with less resolution. 

The changes were pretty involved, following the logic of the StelSkyImageTiles until finally loading the textures.
I failed to adapt the formulation for a template, Ruslan provided the solution. 

Fixes #550 (issue)

### Type of change

### How Has This Been Tested?

**Test Configurations**:
- [x] Win11/Geforce M3070Ti.
- [x] Raspberry Pi 3
## Checklist:

### DIFF
--- a/src/core/StelSkyImageTile.cpp
+++ b/src/core/StelSkyImageTile.cpp
@@ -47,13 +47,15 @@ void StelSkyImageTile::initCtor()
 	texFader = Q_NULLPTR;
 	birthJD = -1e10;
 	withAberration = true;
+	decimation = 1;
 }
 
 // Constructor
-StelSkyImageTile::StelSkyImageTile(const QString& url, StelSkyImageTile* parent) : MultiLevelJsonBase(parent)
+StelSkyImageTile::StelSkyImageTile(const QString& url, StelSkyImageTile* parent, int decimateBy) : MultiLevelJsonBase(parent)
 {
 	initCtor();
-	if (parent!=Q_NULLPTR)
+	decimation=decimateBy;
+	if (parent)
 	{
 		luminance = parent->luminance;
 		alphaBlend = parent->alphaBlend;
@@ -62,10 +64,11 @@ StelSkyImageTile::StelSkyImageTile(const QString& url, StelSkyImageTile* parent)
 }
 
 // Constructor from a map used for JSON files with more than 1 level
-StelSkyImageTile::StelSkyImageTile(const QVariantMap& map, StelSkyImageTile* parent) : MultiLevelJsonBase(parent)
+StelSkyImageTile::StelSkyImageTile(const QVariantMap& map, StelSkyImageTile* parent, int decimateBy) : MultiLevelJsonBase(parent)
 {
 	initCtor();
-	if (parent!=Q_NULLPTR)
+	decimation=decimateBy;
+	if (parent)
 	{
 		luminance = parent->luminance;
 		alphaBlend = parent->alphaBlend;
@@ -209,7 +212,7 @@ void StelSkyImageTile::getTilesToDraw(QMultiMap<double, StelSkyImageTile*>& resu
 		{
 			// The tile has an associated texture, but it is not yet loaded: load it now
 			StelTextureMgr& texMgr=StelApp::getInstance().getTextureManager();
-			tex = texMgr.createTextureThread(absoluteImageURI, StelTexture::StelTextureParams(true));
+			tex = texMgr.createTextureThread(absoluteImageURI, StelTexture::StelTextureParams(true, GL_LINEAR, GL_CLAMP_TO_EDGE, false, decimation));
 			if (!tex)
 			{
 				qWarning() << "WARNING : Can't create tile: " << absoluteImageURI;
@@ -237,7 +240,7 @@ void StelSkyImageTile::getTilesToDraw(QMultiMap<double, StelSkyImageTile*>& resu
 #else
 				if (s.type()==QVariant::Map)
 #endif
-					nt = new StelSkyImageTile(s.toMap(), this);
+					nt = new StelSkyImageTile(s.toMap(), this, decimation);
 				else
 				{
 #if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
@@ -245,7 +248,7 @@ void StelSkyImageTile::getTilesToDraw(QMultiMap<double, StelSkyImageTile*>& resu
 #else
 					Q_ASSERT(s.type()==QVariant::String);
 #endif
-					nt = new StelSkyImageTile(s.toString(), this);
+					nt = new StelSkyImageTile(s.toString(), this, decimation);
 				}
 				subTiles.append(nt);
 			}
--- a/src/core/StelSkyImageTile.cpp
+++ b/src/core/StelSkyImageTile.cpp
@@ -469,7 +469,8 @@ void StelSkyImageTile::loadFromQVariantMap(const QVariantMap& map)
 	{
 		const QVariant& polyRaDec = polyList.at(i);
 		QVector<Vec3d> vertices;
-		for (const auto& vRaDec : polyRaDec.toList())
+		const QList<QVariant> polyRaDecList=polyRaDec.toList();
+		for (const auto& vRaDec : polyRaDecList)
 		{
 			const QVariantList vl = vRaDec.toList();
 			Vec3d v;
@@ -484,7 +485,8 @@ void StelSkyImageTile::loadFromQVariantMap(const QVariantMap& map)
 		{
 			const QVariant& polyXY = texCoordList.at(i);
 			QVector<Vec2f> texCoords;
-			for (const auto& vXY : polyXY.toList())
+			const QList<QVariant> polyXYlist=polyXY.toList();
+			for (const auto& vXY : polyXYlist)
 			{
 				const QVariantList vl = vXY.toList();
 				texCoords.append(Vec2f(vl.at(0).toFloat(&ok), vl.at(1).toFloat(&ok)));
--- a/src/core/StelTexture.cpp
+++ b/src/core/StelTexture.cpp
@@ -207,9 +207,9 @@ void StelTexture::waitForLoaded()
 template <typename T, typename...Params, typename...Args>
 void StelTexture::startAsyncLoader(T (*functionPointer)(Params...), Args&&...args)
 {
-    Q_ASSERT(loader==Q_NULLPTR);
-    //own thread pool supported with Qt 5.4+
-    loader = new QFuture<GLData>(QtConcurrent::run(textureMgr->loaderThreadPool, functionPointer, std::forward<Args>(args)...));
+	Q_ASSERT(loader==Q_NULLPTR);
+	//own thread pool supported with Qt 5.4+
+	loader = new QFuture<GLData>(QtConcurrent::run(textureMgr->loaderThreadPool, functionPointer, std::forward<Args>(args)...));
 }
 
 bool StelTexture::load()
--- a/src/core/StelTexture.cpp
+++ b/src/core/StelTexture.cpp
@@ -299,7 +299,7 @@ QByteArray StelTexture::convertToGLFormat(QImage image, GLint& format, GLint& ty
 #endif
 	if(width != image.width() || height != image.height())
 	{
-		if (decimate==1)
+		if ((decimate==1) || (image.width()/decimate>glInfo.maxTextureSize) || (image.height()/decimate>glInfo.maxTextureSize))
 			qWarning().nospace() << "Got a texture with too large dimensions: "
 							 << image.width() << "x" << image.height()
 							 << ", while maximum size is " << glInfo.maxTextureSize
--- a/src/core/StelTexture.cpp
+++ b/src/core/StelTexture.cpp
@@ -207,14 +207,13 @@ void StelTexture::waitForLoaded()
 		loader->waitForFinished();
 }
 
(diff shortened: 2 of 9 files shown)


src/core/StelSkyImageTile.cpp
```diff
@@ -469,7 +469,8 @@ void StelSkyImageTile::loadFromQVariantMap(const QVariantMap& map)
 	{
 		const QVariant& polyRaDec = polyList.at(i);
 		QVector<Vec3d> vertices;
-		for (const auto& vRaDec : polyRaDec.toList())
```
10110111, line 472: What warnings were emitted here?

src/core/StelTexture.cpp
```diff
@@ -204,13 +207,14 @@ void StelTexture::waitForLoaded()
...
 }
 
-template <typename T, typename Param, typename Arg>
-void StelTexture::startAsyncLoader(T (*functionPointer)(Param), const Arg &arg)
-{
-	Q_ASSERT(loader==Q_NULLPTR);
-	//own thread pool only supported with Qt 5.4+
-	loader = new QFuture<GLData>(QtConcurrent::run(textureMgr->loaderThreadPool, functionPointer, arg));
-}
+// GZ Sorry I just cannot get the right template formulation... We need it only 2x, so made explicit.
```
10110111, line 210: Why not simply remove this? This template doesn't seem useful anyway, and even if it were, it's still there in the revision history.

src/core/StelTexture.cpp
```diff
@@ -279,15 +286,21 @@ bool StelTexture::getDimensions(int &awidth, int &aheight)
...
+	height = std::min(image.height()/decimate, glInfo.maxTextureSize);
+
+#ifndef NDEBUG
+	if (decimate>1)
+		qDebug() << "decimated texture width: " << image.width() << "/" << decimate << "->" << width;
+#endif
 	if(width != image.width() || height != image.height())
 	{
-		qWarning().nospace() << "Got a texture with too large dimensions: "
+		if (decimate==1)
```
10110111, line 302: This is not a correct condition. We should still do the print if e.g. `decimate>1`, and the decimated size is too small to fit in the limits.

src/core/StelTextureMgr.cpp
```diff
@@ -38,8 +38,8 @@ StelTextureMgr::StelTextureMgr(QObject *parent)
 	: QObject(parent), glMemoryUsage(0), loaderThreadPool(new QThreadPool(this))
 {
 #ifdef Q_PROCESSOR_X86_64
-	//allow up to 4 textures to be loaded in parallel on 64 bit
-	loaderThreadPool->setMaxThreadCount(std::min(4,QThread::idealThreadCount()));
+	//allow up to 16 textures to be loaded in parallel on 64 bit
```
10110111, line 41: This should be a separate commit, it's unrelated to decimation and may hide if we look for a change that broke something.

gzotti: OK,. Sometimes I find small things that just slip in. I would usually forget about them...

src/core/StelTexture.cpp
gzotti: If anyone of you knows how to fix it, we could fix it. Else sure, if you agree, remove before merge. Maybe it's trivial, just I don't see it.

src/core/StelSkyImageTile.cpp
gzotti: The clangd analysis inside QtCreator often complains in those loop constructs about (IIRC, there are two usual such messages, and I am tired) temporary structures detaching. Better have no warnings. Setting the list const seems to fix that. Weird enough I cannot reproduce this one right now.

src/core/StelTexture.cpp
gzotti: I think the warning should only be emitted if it does not fit for technical reasons. OK, there is a chance that the decimated size still does not fit the limits, but this would e.g. need a 16k texture, decimate=4 and 2k limit.

src/core/StelSkyImageTile.cpp
10110111: I don't find any way this would actually lead to problems, because the temporary bound to the range is life-extended, see [here](https://en.cppreference.com/w/cpp/language/range-for#Temporary_range_expression). Similarly for the same case below.

src/core/StelTexture.cpp
10110111: Well, the fix seems to be simple: just add one more parameter to the function (with template type), or better, make the template variadic to take any number of parameters to pass. But IMO this template function is just an unnecessary complication, so I'd simply remove it.

gzotti: I added the Arg2 argument to the startAsyncLoader template function, but it did not help. Or I forgot something. But where???

@alex-w do you agree that I can just remove the template?

10110111: You forgot to modify the `functionPointer` signature. It should take two parameters:
```cpp
template <typename T, typename Param1, typename Param2, typename Arg1, typename Arg2>
void StelTexture::startAsyncLoader(T (*functionPointer)(Param1,Param2), const Arg1 &arg, const Arg2& arg2)
```
But I'd use a more general approach, if we do leave this function here:
```cpp
template <typename T, typename...Params, typename...Args>
void StelTexture::startAsyncLoader(T (*functionPointer)(Params...), Args&&...args)
{
    Q_ASSERT(loader==Q_NULLPTR);
    //own thread pool only supported with Qt 5.4+
    loader = new QFuture<GLData>(QtConcurrent::run(textureMgr->loaderThreadPool, functionPointer,
                                                   std::forward<Args>(args)...));
}
```

gzotti: d'oh yes I see it now... 
And esp. thank you for the full solution.

src/core/StelTextureMgr.cpp
```diff
@@ -17,12 +17,12 @@
  * Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
  */
 
-#include "StelApp.hpp"
+//#include "StelApp.hpp"
```
10110111, line 20: Why comment out instead of removing?

gzotti: Also these are suggested for removal (or change) by clangd ("not directly used" or so). 
I first commented out to test on other systems (or CI, as it seems some systems react differently, or in case I don't have all plugins active). These will be removed before merge.

src/core/StelTexture.cpp
```diff
@@ -204,12 +207,12 @@ void StelTexture::waitForLoaded()
...
 
-template <typename T, typename Param, typename Arg>
-void StelTexture::startAsyncLoader(T (*functionPointer)(Param), const Arg &arg)
+template <typename T, typename...Params, typename...Args>
+void StelTexture::startAsyncLoader(T (*functionPointer)(Params...), Args&&...args)
 {
-	Q_ASSERT(loader==Q_NULLPTR);
-	//own thread pool only supported with Qt 5.4+
-	loader = new QFuture<GLData>(QtConcurrent::run(textureMgr->loaderThreadPool, functionPointer, arg));
+    Q_ASSERT(loader==Q_NULLPTR);
```
10110111, line 213: Copy-pasting my snippet, you replaced the tabs with spaces.

10110111 (approved):

src/core/StelTexture.hpp
```diff
@@ -20,9 +20,7 @@
 #ifndef STELTEXTURE_HPP
 #define STELTEXTURE_HPP
 
-#include "StelTextureTypes.hpp"
```
10110111, line 23: I wonder why this builds for you and in the CI, because I get the following errors:
```
In file included from /home/ruslan/src/stellarium/src/core/StelTexture.cpp:21:
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:47:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   47 |  StelTextureSP createTexture(const QString& filename, const StelTexture::StelTextureParams& params=StelTexture::StelTextureParams());
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:50:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   50 |  StelTextureSP createTexture(const QImage &image, const StelTexture::StelTextureParams& params=StelTexture::StelTextureParams());
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:58:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   58 |  StelTextureSP createTextureThread(const QString& url, const StelTexture::StelTextureParams& params=StelTexture::StelTextureParams(), bool lazyLoading=true);
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:64:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   64 |  StelTextureSP wrapperForGLTexture(GLuint texId);
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:69:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   69 |  StelTextureSP getDitheringTexture(const int samplerToBindTo);
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:87:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   87 |  StelTextureSP lookupCache(const QString& file);
      |  ^~~~~~~~~~~~~
      |  StelTexture
/home/ruslan/src/stellarium/src/core/StelTextureMgr.hpp:93:2: error: ‘StelTextureSP’ does not name a type; did you mean ‘StelTexture’?
   93 |  StelTextureSP ditheringTexture;
      |  ^~~~~~~~~~~~~
      |  StelTexture
```

10110111: I've fixed it now, but I still wonder why this worked for you.

gzotti: Huh? Interesting. 
What about the PCH setting? I hope this get rebuilt correctly when the relevant headers are touched!

10110111: Oh, interesting. I have PCH disabled, and indeed the CI seems to have them enabled. So looks like you shouldn't remove headers unless you test build without PCH before pushing.

gzotti: But why does the build system not recreate PCHs when the relevant headers change? Sorry for that. I was just assuming there is a `make` rule that ensures this.

10110111: The PCHs are only created when `ENABLE_PCH` is true. The `stelMain_pch.hpp` file is only used in this case. Without PCH each source file has to include the required headers one by one.

gzotti: Yes, sure. But I would assume that the PCH should be rebuilt if any of the included headers changes. And BTW, StelTexture(Types).h* is not included, so where did this come from? 
OK, must remember to test without PCH before removing a header.

10110111: > But I would assume that the PCH should be rebuilt if any of the included headers changes.

Yes, of course they should. But if they are not used, this doesn't affect the build :)

> And BTW, StelTexture(Types).h* is not included, so where did this come from?

`StelCore.hpp` → `StelSkyDrawer.hpp` → `StelTextureTypes.hpp`

PR #3014, opened 2023-01-22
Fix bottom bar position

10110111:
Currently the buttons in the bottom bar are not fully hidden when the bar is retracted down. Instead, top pixel row is visible, which prevents ejection of the button bar if the mouse pointer is moved too fast to the bottom of the screen.

See the screenshot:

Screenshot_2023-01-22_22-56-51

This patch moves it 1 pixel lower to completely hide the buttons.

This is a trivial patch, but I'm not completely sure that the problem is the same on all platforms. I checked it on Ubuntu 20.04. If it's also present on Windows and macOS, feel free to merge. If it's not, but this patch doesn't make it look bad, then I guess we can merge it too.

### DIFF
--- a/src/gui/SkyGui.cpp
+++ b/src/gui/SkyGui.cpp
@@ -340,7 +340,7 @@ void SkyGui::updateBarsPos()
 		updatePath = true;
 	}
 
-	double rangeY = buttonBar->boundingRectNoHelpLabel().height()+0.5-7.-buttonBarPath->getRoundSize();
+	double rangeY = buttonBar->boundingRectNoHelpLabel().height()+0.5-6.-buttonBarPath->getRoundSize();
 	const qreal newButtonBarX = winBar->boundingRectNoHelpLabel().right()+buttonBarPath->getRoundSize();
 	const qreal newButtonBarY = hh-buttonBar->boundingRectNoHelpLabel().height()-buttonBarPath->getRoundSize()+0.5+(1.-animBottomBarTimeLine->currentValue())*rangeY;
 	if (!qFuzzyCompare(buttonBar->pos().x(), newButtonBarX) || !qFuzzyCompare(buttonBar->pos().y(), newButtonBarY))


gzotti:
On Windows the buttons don't stick out. Is that with screen scaling?

10110111:
This doesn't depend on scaling: I can reproduce with 100% and 150% easily.

gzotti:
stellarium-062

alex-w:
@10110111 @gzotti what about font name in your instances of Stellarium?

10110111:
> Do you have any special window manager?

Reproduces with both xfwm4 and kwin. But anyway, a window manager can't affect this, it's an app-internal detail.

10110111:
> what about font name in your instances of Stellarium?

How do I check?

10110111:
Apparently, this only happens with Qt6. I don't reproduce this with Qt 5.15.

gzotti:
Ah, it was Verdana 13pt on Windows. ... OK, also no problem with Noto Sans. 
Here is with Noto Sans 10pt. 
stellarium-064

Font size may play into it, however I am not able to find a setting where the buttons would stick out. What is your font and size?

Check: Configure panel/Tools tab. 

I run Qt6 on both systems.

10110111:
Screenshot_2023-01-23_00-24-03

alex-w:
> Ah, it was Verdana 13pt on Windows. ... OK, also no problem with Noto Sans. Here is with Noto Sans 10pt. stellarium-064
> 
> Font size may play into it, however I am not able to find a setting where the buttons would stick out. What is your font and size?
> 
> Check: Configure panel/Tools tab.
> 
> I run Qt6 on both systems.

I see one pixel shifted text on your screenshot

alex-w:
> Apparently, this only happens with Qt6. I don't reproduce this with Qt 5.15.

Bad news :( Probably coefficient should be related to major version of Qt

alex-w (approved):

gzotti:
Yes, but no button pieces. Probably better make a conditional `#ifdef WIN32 ` or so. I can try on a native Ubuntu next. 
Both my tests were on Qt6. If it has to do with Qt versions, it is rather Qt5 that is affected!

gzotti:
Ah, I see it now on native Ubuntu with some 1.1+ build.

10110111:
I seem to remember seeing this on a Windows machine too, but I don't have access to this machine anymore to re-check.

gzotti:
Also no issue on the Raspberry Pi 3B+ (Raspbian bullseye, Qt5.15.2, Noto Sans 13/13pt). 

Hmm, same Ubuntu with current master, Qt6.2.4, DejaVu Sans, no issue! 

And now it updated, and it seems I have another dead system :-(( Nah! Seriously, I have no time for this!

gzotti:
Aaah: Storing use of DejaVu Sans and restarting, I see the button edges. So the font geometry used at start is used for the buttons, and the issue is visible when DejaVu has been set (away from default Noto Sans)!

And yes, switching to DejaVu Sans on Win11 and storing these settings, I see the button edges on Windows after restarting Stellarium to use DejaVu Sans for computing the button geometry. 

Solution: Use the preselected font Noto Sans, or risk some issues. I would not like to see a trimmed pixel row. Or really test the button construction thoroughly, with a wide variety of fonts and font sizes, before simply shifting the bar down. Maybe the buttons need a recomputation whenever font changes. There was a reason we had this font switching option hidden...

10110111:
OK, looks like this approach is not good. The "Use the preselected font Noto Sans" doesn't sound correct either. I suppose we have to look for mistakes in computation of the geometry of the buttons and the bar.

gzotti:
It comes to the question what goes into `buttonBar->boundingRectNoHelpLabel()`. A proper solution guarantees that the shift just hides the button backgrounds but never can shave off the bottom pixel row of the location name/fov/fps/time etc.

gzotti:
This may be superseded with a more complete solution for more things related to app font size in #3125. If there is no complete solution today, we should move the milestone to 23.2 in the meantime to not block 23.1.

gzotti (commented):
I hope it's superseded now by a more complete solution in #3125

Issue #1482, opened 2021-01-26
transit time incorrect on days changing from/to daylight saving time


CodeBiene:
### Expected Behaviour
calculation of the transit times of a whole year of the sun from a location on earth

### Actual Behaviour
To my honest opinion the two transit times on the days changing from normal to summer time and vice versa are incorrect.

### Steps to reproduce
1. Start Stellarium
   The location is Earth, Graz (Austria), 363m
2. Select the sun
3. press F10
4. select tab "Transits"
5. select start date 1st January 2021 and end date 31st December 2021
6. press "Calculate transits"
7. Save transits

Attached you can find a LibreOffice table with two sheets and two graphs
[transit.zip](https://github.com/Stellarium/stellarium/files/5873346/transit.zip)
.
One containing the results of version 0.20.2 and the other of version 0.20.4
In order to make the calculation error visible I added the colums H and I and the graph.
In case the functions are not part of the table I also attached the macro code.

### System
* Stellarium version: stellarium-0.20.2-win64.exe, stellarium-0.20.4-win64.exe
* Operating system: Windows 10 Pro, 2004, OS build 19041.746
* Graphics Card: NVIDIA GeForce GT 640M
* Screen type (if applicable): 1920x1080

### Logfile
no

Kind regards,
Willy

alex-w:
Duplicate of #554

alex-w:
Please share info about your time zone settings in Stellarium.

CodeBiene:
> 
> 
> Please share info about your time zone settings in Stellarium.

The time zone is "Europe/Vienna", which seems to be the standard setting for Graz Austria

alex-w:
@CodeBiene the issue is still exist for you in latest versions of Stellarium?

CodeBiene:
Hi Alex,
today I installed the latest version stellarium-1.22.3-win64.exe.
As described above I created a table with the transits.
The results you can find in the attached LibreOffice Calc table "RTS.ods".
The graph shows that the big difference of an hour vanished.
But now there are some strange smaller spikes, which don't make sense to me.
Sorry for the delayed reply.

Kind regards,
Willy
[RTS.ods](https://github.com/Stellarium/stellarium/files/9743471/RTS.ods)

CodeBiene:
Now I added the "Rise" and "Set" times to the graph.
What a surprise. Rise and Set times don't show the spikes!


Kind regards,
Willy
[...]

PR #3028, opened 2023-01-28
AstroCalc: Fix RTS

### LINKED ISSUE #2990: moon culmination data problem
Hello, With the astronomical calculations (F10 of the Stellarium interface) I asked to see the culminations of the moon over a given period. I then downloaded the data to excel. The problem is the following: the culmination times and the corresponding information given (height az; observer/moon distance, magnitude) are not always correct. There are in particular duplicates (two culminations for a single day and (then or before in the spreadsheet, everything is shifted. Is this a handling error on my part? a bug in the application? Thank you for your answer !

_Originally posted by @ibaneoudo in https://github.com/Stellarium/stellarium/discussions/2961_

worachate001:
### Description
The problems are:
- Stellarium calculates RTS of the Moon by choosing the nearest instances. Sometimes, rise or transit or set time occurs on previous day or next day.
- There are time spikes in AstroCalc's RTS table which is a remaining issue in #1482 (near bottom of the page).

This PR will fix them by:
1. When select the Moon (or other objects). Info will show '(previous day)' or '(next day)' after the time to inform us that the calculated event doesn't happen on current day.
2. RTS table in AstroCalc will show the events that occur on the same date in the same row. This should make it easier to read/process in other applications.
3. Fixing above helps us eliminate time spikes issue.

Moon-RTS

Fixes #2990 and remaining issue in #1482

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Manjaro 22.0.1
* Graphics Card: -

## Checklist:

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -4912,15 +4912,13 @@ void Planet::setApparentMagnitudeAlgorithm(QString algorithm)
 // We don't compute positions for midnights, but only for two extra positions 1 JD before and after "now", to allow interpolation of positions.
 // Also, the estimate h0 for the Moon in the literature is based on geocentric computation.
 // NOTE: Limitation for efficiency: If this is a planet moon from another planet, we compute RTS for the parent planet instead!
-Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
+Vec4d Planet::getClosestRTSTime(const StelCore *core, const double altitude) const
 {
 	const StelLocation loc=core->getCurrentLocation();
 	if (loc.name.contains("->")) // a spaceship
 		return Vec4d(0., 0., 0., -1000.);
 
-	// Keep time in sync (method from line 592) to fix slow down of time when the moon is selected
 	const double currentJD = core->getJDOfLastJDUpdate();
-	const qint64 millis = core->getMilliSecondsOfLastJDUpdate();
 	const double currentJDE = core->getJDE();
 	double mr, ms, mt, flag=0.;
 
@@ -5051,7 +5049,6 @@ Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
 			ms += ms2;
 		}
 		core1->setJD(currentJD);
-		core1->setMilliSecondsOfLastJDUpdate(millis); // restore millis.
 		core1->update(0); // enforce update
 	}
 	else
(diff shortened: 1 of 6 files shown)


alex-w:
Thank you very much for the fix!

I see sentences “previous day/next day” - why not “yesterday/tomorrow”?

10110111:
> I see sentences “previous day/next day” - why not “yesterday/tomorrow”?

It makes sense if current day is not the realtime "today", and not even a simulated "today".

worachate001:
> Thank you very much for the fix!
> 
> I see sentences “previous day/next day” - why not “yesterday/tomorrow”?

This actually came across my mind at one point during development. But I though the same with @10110111 that it's mostly simulated time. Using yesterday/tomorrow may not make sense.

alex-w:
> This actually came across my mind at one point during development. But I though the same with @10110111 that it's mostly simulated time. Using yesterday/tomorrow may not make sense.

OK, I got it

alex-w (approved):

gzotti:
Many thanks for this. `StelObject::getInfoMap()` may be fixed with this update too. The `transit`, `rise `and `set `entries are QStrings which may have `(previous day)` or `(next day)` appended (I think we usually don't translate the InfoMap to make it easier for programs which are using it). I am not sure how to deal with the `{transit|rise|set}-dhr` entries, those are numerical. Allow hours >24 and <0? Or just keep as-is and let the user of the data find out by parsing the other string?

worachate001:
> Many thanks for this. `StelObject::getInfoMap()` may be fixed with this update too. The `transit`, `rise `and `set `entries are QStrings which may have `(previous day)` or `(next day)` appended (I think we usually don't translate the InfoMap to make it easier for programs which are using it). I am not sure how to deal with the `{transit|rise|set}-dhr` entries, those are numerical. Allow hours >24 and <0? Or just keep as-is and let the user of the data find out by parsing the other string?

I'm thinking about another way that might be better. Create getRTSTimeOfDate() that returns RTS of current date (doing it inside getRTSTime() may be too complicated).

gzotti:
OK. Just that the InfoMap always needs updates in line with the InfoString.

worachate001:
I'm ditching the idea of displaying 'previous day' or 'next day'. It was proposed earlier because that is an easy way with minimal tweak in existing codes. But it becomes more complicated for StelObject::getInfoMap().

New update will show RTS that occur on current date only (RTS will change at midnight, not seem to change in unpredictable way during a day like before). It also makes more sense and in line with the way they show RTS in astronomical almanacs. Please check StelObject::getInfoMap() to see if it's okay with new changes because I'm not familiar with its structure.

alex-w:
Hmm... I see serious regress of performance on the Mac when the Moon is selected

src/core/modules/Planet.cpp
```diff
@@ -4912,15 +4912,13 @@ void Planet::setApparentMagnitudeAlgorithm(QString algorithm)
...
 // Also, the estimate h0 for the Moon in the literature is based on geocentric computation.
 // NOTE: Limitation for efficiency: If this is a planet moon from another planet, we compute RTS for the parent planet instead!
-Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
+Vec4d Planet::getClosestRTSTime(const StelCore *core, const double altitude) const
 {
 	const StelLocation loc=core->getCurrentLocation();
 	if (loc.name.contains("->")) // a spaceship
 		return Vec4d(0., 0., 0., -1000.);
 
-	// Keep time in sync (method from line 592) to fix slow down of time when the moon is selected
```
alex-w, line 4921: Please back this fix

src/core/modules/Planet.cpp
```diff
@@ -5051,7 +5049,6 @@ Vec4d Planet::getRTSTime(const StelCore *core, const double altitude) const
 			ms += ms2;
 		}
 		core1->setJD(currentJD);
-		core1->setMilliSecondsOfLastJDUpdate(millis); // restore millis.
```
alex-w, line 5054: Second part for performance fix

worachate001:
Thanks @alex-w. I have no performance hit so I'm not sure it's getting better or not. The new update should work now for all objects.

alex-w (approved):

PR #3029, opened 2023-01-29
Fix default distance of "observer" to the parent

### LINKED ISSUE #2911: Solar System Observer is located 1AU from the Sun regardless of how it's configured
### Expected Behaviour

In `ssystem_major.ini`, we have the following line for `[solar_system_observer]`:
```
orbit_SemiMajorAxis=30
```

The expected effect is that the distance to the Sun is 30 AU, or at least that the distance changes when this value is changed.

### Actual Behaviour

In practice, the distance to the Sun is stuck at 1 AU no matter what value I configure for the observer's semi-major axis. In fact, all the observers appear to be at 1 AU from their parent.

### Steps to reproduce

1. Run Stellarium
2. Change location to Solar System Observer
3. Locate the Sun
4. Check _Distance_ item in the info panel, notice it's 1 AU
5. Zoom in to see the full Sun, notice that it fills the screen when FoV is ~0.5°, which confirms that it's 1 AU away.
6. Edit `ssystem_major.ini` to increase `orbit_SemiMajorAxis` of the observer 100×
7. Repeat the observation, note that the distance remains the same.

### System
* Stellarium version: v1.1-477-g581cad6fd3

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/10215523/log.txt)

10110111:
The default distance of an "observer" location to its parent is fixed at 1 AU. This means that the properties in `ssystem_major.ini` are ignored. This patch fixes this behavior.

Note that I noticed that the units are specified in a non-intuitive way: the Sun-parented objects have semimajor axis in AU while others are specified in km. Please check that this interpretation is correct, because the User Guide isn't clear.

Fixes #2911

### DIFF
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -760,11 +760,19 @@ bool SolarSystem::loadPlanets(const QString& filePath)
 
 #ifdef USE_GIMBAL_ORBIT
 		// undefine the flag in Orbit.h to disable and use the old, static observer solution (on an infinitely slow KeplerOrbit)
-		// Note that for now we ignore any orbit-related config values from the ini file.
+		// Note that for now we ignore any orbit-related config values except orbit_SemiMajorAxis from the ini file.
 		if (type=="observer")
 		{
+			double unit = 1; // AU
+			double defaultSemiMajorAxis = 1;
+			if (strParent!="Sun")
+			{
+				unit /= AU;  // Planet moons have distances given in km in the .ini file! But all further computation done in AU.
+				defaultSemiMajorAxis *= AU;
+			}
+			semi_major_axis = pd.value(secname+"/orbit_SemiMajorAxis", defaultSemiMajorAxis).toDouble() * unit;
 			// Create a pseudo orbit that allows interaction with keyboard
-			GimbalOrbit *orb = new GimbalOrbit(1, 0., 90.);    // [1 AU over north pole]
+			GimbalOrbit *orb = new GimbalOrbit(semi_major_axis, 0., 90.); // [Over north pole]
 			orbits.push_back(orb);
 
 			orbitPtr = orb;
--- a/data/ssystem_major.ini
+++ b/data/ssystem_major.ini
@@ -1956,7 +1956,7 @@ orbit_Period=100000000000
 ; 1AU to make earth orbit less distorted
 orbit_SemiMajorAxis=149600000
 parent=Earth
-rot_obliquity=90
+rot_obliquity=0
 type=observer
 
 [mars_observer]
@@ -1971,7 +1971,7 @@ orbit_Period = 70000000000
 ; set to 1.5AU: Average light time as seen from earth.
 orbit_SemiMajorAxis = 224400000
 parent = Mars
-rot_obliquity =90
+rot_obliquity=0
 type = observer
 
 [jupiter_observer]
@@ -1986,7 +1986,7 @@ orbit_Period = 70000000000
 ; set to 5.2AU: Average light time as seen from earth.
 orbit_SemiMajorAxis = 777920000
 parent = Jupiter
-rot_obliquity =90
+rot_obliquity=0
 type = observer
 
 [saturn_observer]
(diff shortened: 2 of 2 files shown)


src/core/modules/SolarSystem.cpp
```diff
@@ -763,8 +763,16 @@ bool SolarSystem::loadPlanets(const QString& filePath)
 		// Note that for now we ignore any orbit-related config values from the ini file.
```
gzotti, line 763: // Note that for now we ignore any orbit-related config values from the ini file.

10110111: Is this intended? Why? It makes more harm than good.

gzotti (commented):
Sorry, but I disapprove. It looks like mostly a documentation issue but is actually more. 

The gimbalOrbit is entirely different because you can change distance with Alt+Home/End. 

As written in the comment, while not completely implemented you can recreate the old behaviour by setting

`core/modules/Orbit.hpp:16:#define USE_GIMBAL_ORBIT 1
`
to `0`.

As written in the other comment, this is stuck in development. IMHO StelMovementMgr should be rewritten first, considering other possible "mount frames", and possibly using Quaternions to avoid gimbal lock situations and the "zenith view" upvector problems. If you want we set that value back to zero and use the old observer configuration until this is completely solved. Then users just cannot move around the observed planet but will not complain about odd view angles.

10110111:
> The gimbalOrbit is entirely different because you can change distance with Alt+Home/End.

Well, my proposed change is to use another _default_. The shortcut keys still work, just start from a configurable distance instead of the hard-coded 1 AU.

gzotti:
After lots of discussion about what is currently mostly confusion caused by documentation not updated after introducing the (unfinished) GimbalOrbit (https://github.com/Stellarium/stellarium/commit/6bbff61aab8268243e7f3311cc795662c388255b), it seems the rot_obliquity update plus remembering to set observers always to their planets' north pole (this may need to be automated!?) really advanced the observers. OK, I no longer object, if docs are updated. (SolarSystem.cpp L763; if necessary I rewrite SUG D.2.1. 


Now, we could indeed advance the Observers to make them more useful. 

* Do Location coordinates make sense when used with *observers? Or shall we add Location.getLatitude()  etc which may deliver 90 in case location.planetName.endsWith("_observer")? 

* What to do with Landscape? Inhibit drawing if location.planetName.endsWith("_observer") 

* Also the location name from the last-loaded landscape don't make sense on an Observer location.

10110111:
> Do Location coordinates make sense when used with *observers? Or shall we add Location.getLatitude() etc which may deliver 90 in case location.planetName.endsWith("_observer")?

No, they don't make sense. It's best to automatically force the location to the north pole. Though, detection by name sounds too hacky: it's better to judge by `Planet::getPlanetType()==Planet::isObserver`.

> What to do with Landscape? Inhibit drawing if location.planetName.endsWith("_observer")

Yes, landscape makes no sense for an observer. Again, don't detect by name, detect by type if possible.

> Also the location name from the last-loaded landscape in the status/bottom bar don't make sense on an Observer location.

The location as in city or coordinates indeed aren't useful, but the name of the "planet" should still be there.

10110111:
> if necessary I rewrite SUG D.2.1

It would be great. The docs should reflect the status quo, including the fact that orbital elements except semimajor axis are ignored for observers.

gzotti:
> > Do Location coordinates make sense when used with *observers? Or shall we add Location.getLatitude() etc which may deliver 90 in case location.planetName.endsWith("_observer")?
> 
> No, they don't make sense. It's best to automatically force the location to the north pole. Though, detection by name sounds too hacky: it's better to judge by `Planet::getPlanetType()==Planet::isObserver`.

Ah yes, we also have that :-)

This means, it also should disable most of the Location panel!
 
> > What to do with Landscape? Inhibit drawing if location.planetName.endsWith("_observer")
> 
> Yes, landscape makes no sense for an observer. Again, don't detect by name, detect by type if possible.

sure.
 
> > Also the location name from the last-loaded landscape in the status/bottom bar don't make sense on an Observer location.
> 
> The location as in city or coordinates indeed aren't useful, but the name of the "planet" should still be there.

Yes, certainly! To understand which strange location we are at.

Issue #82, opened 2018-01-22
Orbit lines not hidden properly by the parent planet


alex-w:
### Bug Description
The arcs of orbit lines coinciding with the planet are either shown through the parent planet or completely hidden by it. Please see the attached screen shots.
### Original report
https://bugs.launchpad.net/stellarium/+bug/978576

gzotti:
That's an issue of drawing order. The orbit is drawn as part of the planet/moon draw(), and the orbit is apparently plotted to be lying at the same distance as the moon, without Z buffer test.

PR #3033, opened 2023-01-30
Useful observer orientation

### LINKED ISSUE #1202: Solar System Observer point not above the Sun?
In version 0.20.2 when I "go" to the Solar System Observer location (by pressing CTRL-G), and then center the Sun to look at it and show the orbits of the planets, the view appears to be from off to the side of the line going through the Sun and perpendicular to the ecliptic plane - in other words, the planets' orbits are skewed (see screenshot). This was not the case in previous versions, where the Solar System Observer location appeared to be directly above the Sun and showed the planets orbits' face on. Am I missing something, or has the Solar System Observer location changed? I use it to show my students what the Solar System and planet orbits look like "from above".  
image

gzotti:
### Description
After @10110111 discovered an important detail which improved orientation of the "observer" pseudo-planets (#3029), we can finally make them a bit more useful. Notable changes: 

- [x] when on an observer, freeze location on the North pole
- [x] inhibit landscape drawing
- [x] change location name display in the bottom bar
- [x] don't draw a few coordinate grids (more? please comment)
- [x] Modify Location dialog to inhibit a few settings while in observer mode
- [x] update SUG

Fixes #1202 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: Geforce M3070Ti

## Checklist:


src/core/StelLocation.hpp
```diff
@@ -66,6 +66,13 @@ class StelLocation
 	//! @return azimuth counted from north or south as set in the StelApp preferences, in [0...360].
 	double getAzimuthForLocation(double longTarget, double latTarget) const;
 
+	//! longitude and latitude are private to force use of special getters. These return a location on the north pole if we are located on an "Observer".
+	//! By this we force useful view orientation.
+	float getLatitude()  const { if (planetName.endsWith(" Observer")) return 90.f; else return latitude;}
```
10110111, line 71: [`StelLocation.cpp:228`](https://github.com/Stellarium/stellarium/blob/master/src/core/StelLocation.cpp#L228) shows a precedent of finding a `Planet` object. Same way can be used to check for the actual planet being an observer, rather than relying on its name.

10110111:
> don't draw a few coordinate grids (more? please comment)

What other grids do we have? Azimuthal grid should remain IMO. Equatorial is now equivalent to it (except for the units of azimuth), so can be disabled indeed. What the fixed equator line is I don't know (just a line on the ground?)

gzotti:
> > don't draw a few coordinate grids (more? please comment)
> 
> What other grids do we have? Azimuthal grid should remain IMO. Equatorial is now equivalent to it (except for the units of azimuth), so can be disabled indeed. What the fixed equator line is I don't know (just a line on the ground?)

The fixed equatorial system is the system of hour angle/declination used in classical position astronomy, i.e. does not make sense in a location-less place. Likewise, the horizon/compass marks are useless. Others may be discussed. Just that I need to play a bit more now. It seems the displayed "azimuth" grid is now finally really aligned with the planet's axis.

src/core/modules/Orbit.hpp
```diff
@@ -167,7 +165,7 @@ class GimbalOrbit : public Orbit {
 	void setLatitude(const double lat) { latitude=lat*M_PI_180;}
 	void setDistance(const double dist){ distance=dist;}
 	void addToLongitude(const double dlong){ longitude+=dlong*M_PI_180; }
-	void addToLatitude(const double dlat)  { latitude=qBound(-M_PI_2, latitude+dlat*M_PI_180, M_PI_2);}
+	void addToLatitude(const double dlat)  { latitude=qBound(-89.99*M_PI_180, latitude+dlat*M_PI_180, 89.99*M_PI_180);}
```
10110111, line 168: This is ugly. You'll see this if you try to move to the pole from SSO, looking at the Sun. My attempt came closer to the desired result (basically the same, but with closer value to 90°):

```diff
diff --git a/src/core/modules/Orbit.cpp b/src/core/modules/Orbit.cpp
index b352e53c52..99bdfcebca 100644
--- a/src/core/modules/Orbit.cpp
+++ b/src/core/modules/Orbit.cpp
@@ -293,6 +293,16 @@ GimbalOrbit::GimbalOrbit(double distance, double longitude, double latitude):
 {
     setParentOrientation(0., 0., 0.);
 };
+
+void GimbalOrbit::addToLatitude(const double dlat)
+{
+    // This value lets us avoid jumpiness of view roll when looking at the poles
+    const auto epsilon = 1e-8;
+
+    const auto max = M_PI_2*(1 - epsilon);
+    latitude = qBound(-max, latitude+dlat*M_PI_180, +max);
+}
+
 //! Compute the position (JDE is just a placeholder) and return a "stellarium compliant" position
 void GimbalOrbit::positionAtTimevInVSOP87Coordinates(double JDE, double* v)
 {
diff --git a/src/core/modules/Orbit.hpp b/src/core/modules/Orbit.hpp
index 64e54a1544..71c12a0ad2 100644
--- a/src/core/modules/Orbit.hpp
+++ b/src/core/modules/Orbit.hpp
@@ -167,7 +167,7 @@ public:
     void setLatitude(const double lat) { latitude=lat*M_PI_180;}
     void setDistance(const double dist){ distance=dist;}
     void addToLongitude(const double dlong){ longitude+=dlong*M_PI_180; }
-    void addToLatitude(const double dlat)  { latitude=qBound(-M_PI_2, latitude+dlat*M_PI_180, M_PI_2);}
+    void addToLatitude(double dlat);
     void addToDistance(const double ddist) { distance=qBound(0.01, distance+ddist, 50.);}
 
 private:
```

gzotti: I attempted avoiding the weird flip when going away from 90°, but neither my 0.1° nor your going even closer solves the problem. Not sure what, but something causes the view to rotate by tens of degrees between 85 and ~90°. Go to the maximum latitude now allowed and rotate (Alt-right/left). See also e.g. for Saturn Observer. Maybe the parent orientation for GimbalOrbit needs more thought. Just that previously the odd motion did not even allow thinking about these details!

gzotti:
Now that we can really see Jupiter and Saturn systems more stabilized, I can see (again) that the moons' orbit lines still overdraw the planets when the respective moon is closer than the planet. (#82)

10110111:
Also I get lots of repeated warnings like this:
```
/home/ruslan/src/stellarium/src/core/StelLocation.hpp: In constructor ‘StelLocation::StelLocation()’:
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:140:8: warning: ‘StelLocation::latitude’ will be initialized after [-Wreorder]
  140 |  float latitude;
      |        ^~~~~~~~
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:85:6: warning:   ‘int StelLocation::altitude’ [-Wreorder]
   85 |  int altitude;
      |      ^~~~~~~~
In file included from /home/ruslan/package-builds/stellarium-build-qt6/plugins/Observability/src/Observability-static_autogen/EWIEGA46WW/../../../../../../../src/stellarium/plugins/Observability/src/Observability.hpp:29,
                 from /home/ruslan/package-builds/stellarium-build-qt6/plugins/Observability/src/Observability-static_autogen/EWIEGA46WW/moc_Observability.cpp:10,
                 from /home/ruslan/package-builds/stellarium-build-qt6/plugins/Observability/src/Observability-static_autogen/mocs_compilation.cpp:2:
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:33:2: warning:   when initialized here [-Wreorder]
   33 |  StelLocation() : longitude(0.f), latitude(0.f), altitude(0), population(0), role('X'), isUserLocation(true) {}
      |  ^~~~~~~~~~~~
```

10110111:
> The first (deprecation warnings) is an effect of your changes around light pollution.

I didn't quote the deprecation warnings, only the incomplete type errors (which you have now fixed).

The initialization order warnings are still there, these are introduced by this PR.

src/core/StelLocation.hpp
10110111: OK, this appears to be a serious mistake. After you fixed this I got a performance regression from 38 FPS to 1.2 FPS. I have 186 planets in the list, and the search function appears to be now called several thousand times per frame.

10110111: I think the better fix would be to cache `PlanetP` in the `StelLocation` along with `planetName`.

gzotti: Ouch. I had feared that this was going to cost somewhat, but not that much. Just that the StelLocation has no pointer to its planet. 

Hmm, we can most likely use the Location's role for this! A small o can be used for "observer". I cannot continue today, though.

10110111: > Just that the StelLocation has no pointer to its planet.

Why not introduce it? It would be updated whenever `planetName` changes.

gzotti: No need for another variable if `role` can do it. Hopefully tomorrow...

data/ssystem_major.ini
```diff
@@ -1950,13 +1950,13 @@ color=0., 0., 0.
 halo=false
 hidden=true
 name=Earth Observer
-orbit_Inclination=90
-orbit_MeanLongitude=90
-orbit_Period=100000000000
+; orbit_Inclination=90
```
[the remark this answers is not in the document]
gzotti: The final question about this PR is whether we can already finalize this and remove the define in Orbit.h to go back to the previous Observer solution. 

For this, we should probably think about another solution regarding location. Currently I dislike that the Location panel shows the "nomap.png" planet texture. We could add a decorative "observer" texture. Also, the 90° remains when going to another planet/moon. Is there a way to just deliver 90/0 without changing the other locations' latitude/longitude?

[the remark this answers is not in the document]
10110111: > Currently I dislike that the Location panel shows the "nomap.png" planet texture.

I think at least the map should be removed, since geographic location doesn't make sense for observers. A decorative placeholder can indeed be a nice replacement.

> Also, the 90° remains when going to another planet/moon.

I don't reproduce this. When I switch from Earth/Astana to Solar System Observer and back to Earth, I return to the coordinates of Astana. The _Latitude_ and _Longitude_ spinboxes in the _Location_ dialog keep their values all the time, without changing to 90°. The location name (and its properties like light pollution) is lost though.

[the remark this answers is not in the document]
gzotti: The coordinates of Astana remain? Is this Qt6 on Linux? These are the stupid little details between OSes :-(
I have not tried what happens if I disable and hide the position UI elements as soon as I switch to an "observer" planet. 

I could add planet close-up screenshots for each observer. Or do you have better ideas (and actual images)? Some spaceship cockpit?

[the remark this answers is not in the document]
10110111: Yes it's Qt6 on Linux. Building Qt5 version now, but I don't think it would differ.

As for the artwork, I don't have any good ideas on it. Planet close-up seems a bit misleading, as if we go to this planet. Spaceship cockpit might be OK, though it shouldn't look like a high-detail something that would induce too high expectations.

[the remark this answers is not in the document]
10110111: Indeed, Qt5 version works the same way.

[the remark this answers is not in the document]
gzotti: We actually _are_ going to this planet. How much the closeup goes will be up to the user. Adding orbits will help getting the idea. However, Saturn and Jupiter have those overlapping orbits. May need to paint them away. And yes, I will not paint a cockpit.

gzotti:
The issue about location staying at 90° is not solved yet. Maybe the current solution is even worse as I indeed now manually set the numbers.
When in observer mode, selecting a landscape with location in the View settings seems to miss loading a texture. Only on second attempt the landscape texture is loaded. Not sure what's up here.

data/ssystem_major.ini
```diff
@@ -2002,8 +2002,8 @@ tex_map=obs_saturn.jpg
 ; orbit_Inclination = 90
 ; orbit_MeanLongitude = 90
 ; orbit_Period = 70000000000
-; set to 9.6AU: Average light time as seen from earth.
-orbit_SemiMajorAxis = 1436160000
+; 30 Mio km to see Phoebe's orbit
```
10110111, line 2005: Mio km seems to be a rarely used designation for the standard Gm. At least, I didn't find much use of it online, and most of occurrences were in the context of German language.

data/ssystem_major.ini
```diff
@@ -1954,8 +1954,8 @@ tex_map=obs_earth.jpg
 ; orbit_Inclination=90
 ; orbit_MeanLongitude=90
 ; orbit_Period=100000000000
-; 1AU to make earth orbit less distorted
-orbit_SemiMajorAxis=149600000
+; 800.000 km is about twice Lunar orbit size
```
10110111, line 1957: Oh please, don't separate thousands with comma or period, it's super confusing.

gzotti: OKOK, it just helped me count the zeros...

data/ssystem_major.ini
gzotti: Although few humans understand Gm. I will use M km in these comments.

data/ssystem_major.ini
10110111: ISO 31-0, BIPM, IUPAC, AMA, all recommend using (thin) space for this. C++14 and higher uses apostrophes, and Java uses underscores.

data/ssystem_major.ini
10110111: Well, just over half a century ago few humans understood nm, using mμ instead.

10110111:
Current UI has too large margins in places of the hidden controls. I think the "map" should be stretched into the empty space (preserving aspect ratio).

Screenshot_2023-02-04_23-45-59

src/gui/LocationDialog.cpp
```diff
@@ -291,6 +292,30 @@ void LocationDialog::connectEditSignals()
 	connect(ui->cityNameLineEdit, SIGNAL(textEdited(const QString&)), this, SLOT(reportEdit()));
 }
 
+void LocationDialog::setLocationUIvisible(bool visible)
+{
+	ui->latitudeSpinBox->setEnabled(visible);
```
10110111, line 297: Would be better to just hide the containing widgets like `frame_4`, `frame_2`, and then only the individual controls from `frame_5`. This would simplify the code and avoid the need to revisit it if we add additional elements into these frames.

gzotti: Does the window not collapse if I hide the left-side frame? Then wheeling through the planets would fail.
These changes were now made just to make unfitting controls unavailable without actually changing the layout.

10110111: I think the other panels should expand into the newly freed space. Needs checking how the wheeling will work though.

gzotti:
OK, should be ready for a test. The wheeling through now fails, though.

10110111:
_Return to default location_ button is still useful, so shouldn't be hidden. And then wheeling behavior might be restored.

10110111:
BTW, "Set this location as default" should also be available. Also, for some reason I still get geographical coordinates (90°N 0°E) in the bottom panel after the "planet" name.

gzotti:
> _Return to default location_ button is still useful, so shouldn't be hidden. And then wheeling behavior might be restored.

This was an all-or-nothing option to hide the bottom frame of buttons. OK, hiding the two other buttons individually.

gzotti:
I am afraid the "set location as default" needs more restructuring than I consider useful, so it's even better to hide this option. Is it possible to set an observer via scripting? Those few users who need that may then write a startup.ssc.

10110111:
> Is it possible to set an observer via scripting?

Well it's just another planet, if Stellarium has no special code to forbid setting an observer, it should be handled the same way.

gzotti:
Yes, but it does not work (likely never worked). Allowing observer as default location will be left for another PR.

gzotti:
Yes, but it does not work (likely never worked). Allowing observer as default location will be left for another PR.

10110111:
I still have lots of initialization order warnings about this place:
```
In file included from /home/ruslan/src/stellarium/src/core/StelCore.hpp:26,
                 from /home/ruslan/src/stellarium/src/scripting/StelMainScriptAPI.hpp:30,
                 from /home/ruslan/src/stellarium/src/scripting/StelScriptMgr.hpp:42,
                 from /home/ruslan/src/stellarium/src/gui/ScriptConsole.cpp:24:
/home/ruslan/src/stellarium/src/core/StelLocation.hpp: In constructor ‘StelLocation::StelLocation()’:
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:142:8: warning: ‘StelLocation::latitude’ will be initialized after [-Wreorder]
  142 |  float latitude;
      |        ^~~~~~~~
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:86:6: warning:   ‘int StelLocation::altitude’ [-Wreorder]
   86 |  int altitude;
      |      ^~~~~~~~
In file included from /home/ruslan/src/stellarium/src/core/StelCore.hpp:26,
                 from /home/ruslan/src/stellarium/src/scripting/StelMainScriptAPI.hpp:30,
                 from /home/ruslan/src/stellarium/src/scripting/StelScriptMgr.hpp:42,
                 from /home/ruslan/src/stellarium/src/gui/ScriptConsole.cpp:24:
/home/ruslan/src/stellarium/src/core/StelLocation.hpp:33:2: warning:   when initialized here [-Wreorder]
   33 |  StelLocation() : longitude(0.f), latitude(0.f), altitude(0), population(0), role('X'), isUserLocation(true) {}
      |  ^~~~~~~~~~~~
```

10110111:
The location in the bottom bar still has coordinates after I wheel from Earth (Astana from the list) to Earth Observer. This happens in line 850:
```cpp
		if(loc->name.isEmpty())
			newLocation = planetNameI18n +", "+StelUtils::decDegToDmsStr(loc->getLatitude())+", "+StelUtils::decDegToDmsStr(loc->getLongitude());
```

gzotti:
I still have an issue when selecting a landscape on earth (with "landscape sets location") while on an observer. We are changing planets during the process, obviously. LandscapeMgr::onTargetLocationChanged(const StelLocation &loc) decides it needs to load the default landscape for the new planet, or "zero" if none is defined.

10110111:
So what is the visible problem? I didn't notice anything strange in the scenario of

1. Go to an observer,
2. Check "position from landscape" checkbox,
3. Pick a non-default landscape.

gzotti:
The landscape location loads, but shows the zero landscape.

gzotti:
Following the signals, the setting of location in LandscapeMgr (which successfully has switched to the right landscape) causes a default landscape for the planet to be loaded. Ah, you must enable "Auto select landscapes" in the Config/Tools tab!

alex-w:
> @alex-w can you please also test whether you still find issues?

1. Run Stellarium
2. Open Search tool and find any "x observer" object
3. Ctrl+G (Go to observer)
4. Open Location dialog

Probably this is wrong look

gzotti:
Huh, I never did that... You have always new ways of operation! :-)

gzotti:
Grr, it's simple if Location panel has been created already before searching.  
OK, that was not so hard. But going from Jupiter observer to Saturn shows the same issue as before: default landscape for Saturn does not load. I am afraid I have to add an optional parameter to StelCore::moveObserverTo(., ., ., landscapeID) which is then emitted in an enlarged signal StelCore::targetLoationChanged(StelLocation, QString)

alex-w:
@gzotti please check another "new way of operation" :)

1. Run Stellarium
2. Enable showing nomenclature
3. Open Search tool and find any nomenclature item
4. Ctrl+G (Go to nomenclature item)
5. Open Location dialog

What happenes at step 4 and 5?

alex-w (approved):

gzotti:
> @gzotti please check another "new way of operation" :)
> [...]
> 
> What happenes at step 4 and 5?

It works for me. What did you see?

alex-w:
> It works for me. What did you see?

Good news! I didn't checked it - I just remembered about this behaviour and shared notice about, but in that time I didn't can check it.

gzotti:
@10110111 do you still see anything that should be corrected? 
I think this time I will only rebase, not squash, despite some experimental commits.

10110111:
Functionally I didn't notice any more glitches. Code-wise though, cannot help but object to the committing of commented-out code. It's especially ugly when you doubly comment-out (see the body of `LocationDialog::handleDialogSizeChanged`), or introduce fresh-but-commented-out code (as in `LocationDialog::setLocationUIvisible`).

gzotti:
Ah yes, thanks for reminding me.

gzotti:
Shall I also clean out ssystem_major.ini? Then going back to the old behaviour is more difficult. But maybe it's even time to no longer go back! (remove the define in Orbit.h) @alex-w , @10110111 ?

alex-w:
Clean up the ssystem_major.ini file is OK for me

10110111 (approved):

PR #3036, opened 2023-02-01
Added support 32-wind compass rose

alex-w:
### Description
Currently we have support 8-wind and 16-wind compass rose (used in navigation and meteorology). I've added support 32-wind compass rose as an additional to Navigational Stars plugin for navigational purpose.

### Type of change

### How Has This Been Tested?
1. Run Stellarium
2. Enable cardinal points
3. Enable 8-, 16- and 32-wind compass rose

## Checklist:


10110111:
Is it actually useful? I do realize that these names exist, but IMHO at some point naming ever smaller sectors becomes more cumbersome than simply using numeric azimuth.

gzotti:
1/32 of the full circle is an (outdated?) nautical directional unit ("Strich" in German; sorry, I cannot find an English reference for this! Google translate has "dash".  A similar concept with smaller units is the artilleristic 'mil' (6400 parts to the circle, or 6000 in former Soviet-influenced states). There are nautical commands like "Ruder zwei Strich backbord" (Turn rudder to change course by two of these 1/32 units to the left"). So yes, this is the last partition that may make sense. Maybe even extend this in yet another option to not use these names but show all these directions with the proper unit sign like `16"` for "due south".

alex-w:
> Is it actually useful? I do realize that these names exist, but IMHO at some point naming ever smaller sectors becomes more cumbersome than simply using numeric azimuth.

Yes, but for navigational purposes and partially for marine astronomy. By default these labels are disabled and probably for low count of users this feature will not be extra useful, but this is final part for standard marine/navigational compass points, so, all navigational data is complete now.

alex-w:
> 1/32 of the full circle is an (outdated?) nautical directional unit ("Strich" in German; sorry, I cannot find an English reference for this! Google translate has "dash". A similar concept with smaller units is the artilleristic 'mil' (6400 parts to the circle, or 6000 in former Soviet-influenced states). There are nautical commands like "Ruder zwei Strich backbord" (Turn rudder to change course by two of these 1/32 units to the left"). So yes, this is the last partition that may make sense. Maybe even extend this in yet another option to not use these names but show all these directions with the proper unit sign like `16"` for "due south".

Russian: румб :) These directions still uses by sailors for example.

gzotti:
> Russian: румб :) These directions still uses by sailors for example.

Ah, that brought me to the English term: https://en.wikipedia.org/wiki/Rhumb

src/core/modules/LandscapeMgr.cpp
```diff
@@ -1096,6 +1174,7 @@ void LandscapeMgr::onTargetLocationChanged(const StelLocation &loc)
 				setFlagCardinalPoints(conf->value("viewing/flag_cardinal_points", true).toBool());
 				setFlagOrdinalPoints(conf->value("viewing/flag_ordinal_points", true).toBool());
 				setFlagOrdinal16WRPoints(conf->value("viewing/flag_16wcr_points", false).toBool());
+				setFlagOrdinal32WRPoints(conf->value("viewing/flag_32wcr_points", false).toBool());
```
gzotti, line 1177: what does the wcr actually mean?

src/gui/ViewDialog.cpp
```diff
@@ -373,6 +373,7 @@ void ViewDialog::createDialogContent()
 	connectCheckBox(ui->showCardinalPointsCheckBox,           "actionShow_Cardinal_Points");
 	connectCheckBox(ui->showOrdinal8WRPointsCheckBox,         "actionShow_Intercardinal_Points");
 	connectCheckBox(ui->showOrdinal16WRPointsCheckBox,        "actionShow_Secondary_Intercardinal_Points");
+	connectCheckBox(ui->showOrdinal32WRPointsCheckBox,        "actionShow_Quarter_Intercardinal_Points");
```
gzotti, line 376: Likewise "Tertiary"

src/core/modules/LandscapeMgr.cpp
gzotti: Also, it may make sense to change the whole interface here and use just one value "cardinal_points" with values

0: none
1: cardinal (NESW)
2: intercardinal
3: secondary
4: tertiary
5: Rhumb number (I am not sure about the English terminology. "Strichzahl" in German.)

Likewise, the GUI may then have a Combobox.

With setting "rhumb" the display should also show different partitions. Either just the rhumb ticks, or both, degrees and rhumb tick?

This may happen in this PR or in another one.

gzotti: This also needs (in any case) an update in the SUG.

alex-w: > Also, it may make sense to change the whole interface here and use just one value "cardinal_points" with values
> 
> 0: none 1: cardinal (NESW) 2: intercardinal 3: secondary 4: tertiary 5: Rhumb number (I am not sure about the English terminology. "Strichzahl" in German.)
> 
> Likewise, the GUI may then have a Combobox.
> 
> With setting "rhumb" the display should also show different partitions. Either just the rhumb ticks, or both, degrees and rhumb tick?
> 
> This may happen in this PR or in another one.

Compass marks already exists and combobox is bad idea here IMHO (just a lecturer point of view)

gzotti: Why is a single ComboBox with 6 possible values worse than 4 checkboxes? 

Combobox labeling: Compass Marks

Values:

none          
4 (Tooltip: cardinal) 
8 (Tooltip: cardinal + intercardinal)
16 (Tooltip: up to secondary intercardinal) 
32 (Tooltip: up to tertiary intercardinal partitions found in navigation)
Rhumb (Tooltip: Rhumb number used in navigation) 

Not strictly required now, but please think of it.

alex-w: Currently any lecturer can change the number of compass labels via shortcuts in any time of lecture. To change the values via combobox user/lecturer should use GUI (RCP as an optional UI) - this is not very good behaviour for lecturers.

gzotti: Hm, OK. But as we always say, these could be done with 1-line scripts... I think of typical users who may want to see less GUI elements. 
BTW, the 32 option should also go into the RCP (both files!) before merge of this PR.

alex-w: Good point for RCP!

gzotti: OK. Now I think all that required to be done is a new Figure 4.13 and a line in D.1.25.

gzotti (approved):
OK for me.

PR #3041, opened 2023-02-02
Make color of the Moon closer to reality

10110111:
### Description
If you have ever looked at the Moon through a telescope, you may have noticed its peculiar color, which is definitely not gray nor white. It's closer to beige. This patch makes the color of the simulated Moon like this.

The reference color was taken from a minimally-processed photo I made of the Moon when it was at elevation of about 65.9° — highest achievable now at my current location to minimize the effect of the atmospheric extinction.

#### Old Moon
stellarium-009

#### Updated Moon
stellarium-010


gzotti:
No, this is much too red. Maybe your camera's white balance is off.

10110111:
No it isn't. Lunar albedo is twice as high at 700 nm than at 400 nm and changes almost linearly, so this is an expected result.

gzotti:
Now balance this with mesopic sensitivity. Night scenes look most natural with about 4100 angstrom color temperature where the moon looks IMO pretty much OK as it is. Your moon looks like 15° high or lower.
@Atque you seem also experienced here. Your insight? Others?

Atque:
Sorry, no experience when it comes to this, although I tend to agree with @gzotti; the Moon looks to brown or red in this PR.

10110111:
> Now balance this with mesopic sensitivity.

This doesn't make any sense: the Moon is brighter than any PC monitor is able to show, like daytime cirrus clouds! Have you ever looked at the Moon through a telescope without a lunar filter? You'd have to wait several minutes after removing the telescope from your eye to have it become somewhat night-adapted again. And this is not even at full Moon, at which point the brightness is several times higher!

And without a telescope, if you actually suppose to distinguish the maria, rather than see a bright undetalized white plate, your vision has to be in the daylight regime (at least the fovea), the rest may still be able to see dim objects, although glare from the Moon might interfere). In particular, with the full Moon the pupils must be contracted, otherwise you indeed will see just a bright white plate.

> Are there unprocessed photos of the Moon taken from space to compare with?

Well, not unprocessed, but I guess we should be interested in linear data with sensible relative spectra. There is LROC normalized albedo for multiple wavelengths in the visible spectrum in a quite detailed map spanning latitudes -70°..70°, which can be found [here](http://pds.lroc.asu.edu/data/LRO-L-LROC-5-RDR-V1.0/LROLRC_2001/DATA/MDR/WAC_HAPKE/). If we take these data, interpolate, multiply by extraterrestrial solar irradiance spectrum, and integrate against XYZ color matching functions, converting then to sRGB, we'll get this for e.g. E350N0450 sector (scaled down to reduce size):

E350N0450

gzotti:
I am less interested in  "linear data with sensible relative spectra" but rather how it appears to humans. A moon high in the night sky is an annoyingly bright white thing. On the horizon it may appear more yellow/reddish, but extinction should take care for this already. 

You can play many evil tricks with photos. Moon was 35° high at this exposure.

An ordinary automatic exposure
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WBauto

Explicitly setting white balance to "daylight"
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WBday

Carefully setting color temperature to the conventional wisdom setting for nighttime photography, 4100K
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WB4100

Stellarium master:
image


5200K?
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WB5200

Daylight at 5600K
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WB5600

Or rather 6500K? (Too yellow for my taste)
2008-05-11-22-29-31_Mond_Russentonne_Test-0002_Alt35_WB6500

If your aim is to improve the screen filling telescopic eye-blinding experience, then improve just that: telescopic view. I.e., make the color tinge appear dependent on zoom level or rather zoom+magnification. When the moon fills e.g. 1/8 of the vertical fov, start transition from "auto" (whatever is set at current master) to "daylight" (5200 setting? Something less brown than yours).

10110111:
> rather how it appears to humans

Well, I go into the street, look up at the sky and see a "piece of cheese" quarter Moon. Then open Stellarium and get a "wrinkled aluminium foil". No it doesn't look like in reality.

> I.e., make the color tinge appear dependent on zoom level or rather zoom+magnification.

This is a job for tone mapper actually. But we already have something simplistic like this: increase of exposure on zooming out. This should automatically whiten the Moon because of overexposure, just as it happens with human vision. Maybe just needs some manipulation to work better.

Atque:
I think I’ve said this earlier at some point, but I would prefer if the Moon has some kind of glare effect which decreased when zooming in and increased when zooming out. Also, no halo when the FOV is large enough. Rather, an overexposed crescent or whatever the form the lunar disc is at.

worachate001:
We may have to check appearance of the Moon during eclipse. The new update looks weird in this half penumbral lunar eclipse.

stellarium-023

And this is NASA's visualization of the Moon https://svs.gsfc.nasa.gov/5048

10110111:
> We may have to check appearance of the Moon during eclipse. The new update looks weird in this half penumbral lunar eclipse.

Doesn't it look similarly weird without this PR?

worachate001:
> > We may have to check appearance of the Moon during eclipse. The new update looks weird in this half penumbral lunar eclipse.
> 
> Doesn't it look similarly weird without this PR?

Yes, it looks better in version 1.2.

10110111:
> Yes, it looks better in version 1.2.

This PR is not the only change since 1.2.

gzotti:
Yes, the current outer edge of the penumbra looks really weird now. In reality the edge is invisible. Traditionally we showed a tiny little brightness falloff that slowly increased towards umbra. I had conceded the umbra edge may need improvement in form of a somewhat steeper falloff of the penumbra's inner edge. This may then require another rebalancing of the necessary apparent brightness surge. What amount and how many cycles of re-optimisations do we need for just the Moon?  IMHO a better re-balancing of stellar/MW/ZL visibility in light of the new atmosphere model and LP settings, and being able to create variants of the latter with CalcMySky (its documentation!), are far more important issues.

alex-w:
I agree with the patch and disagree in same time. In this evening I observed the white Moon, but I remember that color of the Moon is not constant and it depends by atmosphere properties. What about adding config option for color temperature (atmosphere settings?) and use it for color balance of the Moon?

10110111:
> What about adding config option for color temperature (atmosphere settings?) and use it for color balance of the Moon?

Yes, of course it should be done, but using it only for the Moon makes all this rendering even more inconsistent.

The correct solution for color balancing is to do it in the tone mapper that would apply to the whole scene and do all the post-processing like scotopic/mesopic transformation, white balancing, sensor saturation etc. Stellarium simply doesn't have this now. What it has is bluing of the atmosphere on the fly (via ToneReproducer), irreversibly blued Milky Way texture, and some ad hoc coefficients here and there.

> (atmosphere settings?)

No, it's not related to atmosphere in any way. It depends on light pollution first of all (at night in particular), and generally on irradiance of the observer by all the sky.

alex-w:
> > What about adding config option for color temperature (atmosphere settings?) and use it for color balance of the Moon?
> 
> Yes, of course it should be done, but using it only for the Moon makes all this rendering even more inconsistent.

But this patch touches the Moon only.

> The correct solution for color balancing is to do it in the tone mapper that would apply to the whole scene and do all the post-processing like scotopic/mesopic transformation, white balancing, sensor saturation etc. Stellarium simply doesn't have this now. What it has is bluing of the atmosphere on the fly (via ToneReproducer), irreversibly blued Milky Way texture, and some ad hoc coefficients here and there.

Well, this is big task. Could it be sliced into few subtasks?

> 
> > (atmosphere settings?)
> 
> No, it's not related to atmosphere in any way. It depends on light pollution first of all (at night in particular), and generally on irradiance of the observer by all the sky.

OK, another option in View/Sky section.

10110111:
> But this patch touches the Moon only.

Yes, because it fixes the initially wrong colors that the Moon texture has. It does no attempt to implement any tone mapping.

> Well, this is big task. Could it be sliced into few subtasks?

It's really a big task, especially given Stellarium's targeting at the smallest devices with GLES2 and GL2- support. Looks like at some point we'll have to split the graphics engine into two parts, so that not only atmosphere, but also planet rendering and all other elements are done differently on obsolete/small devices and on modern ones.

Issue #1793, opened 2021-07-11
altitude line: partitioning should take scale into account


axd1967:
**Is your feature request related to a problem? Please describe.**

When zooming in, the altitude line subdivisions become useless because there are only markers at 30 and 60 degrees.

**Describe the solution you'd like**
Divide the visible vertical line in a few visible partitions, similar to the  subdivision algorithm used for the azimuth grid.

![image](https://user-images.githubusercontent.com/3529789/125200607-229d1f80-e26c-11eb-83bd-37a65c199443.png)

**Describe alternatives you've considered**

## workaround
The azimuth grid gives reasonable partitionings, but adds noise to the image.

**Additional context**

alex-w:
Probably other partitioned lines should be checked also.

PR #3057, opened 2023-02-12
Grids: accurate ecliptic calendar marks

### LINKED ISSUE #2175: Calendar on Ecliptic
Stellarium 0.21.2 - It would be useful to have a calendar with division into days on the Ecliptic, to show the positions of the sun during the months

gzotti:
### Description

This is an alternative to #2216 with accurate Solar positions labelled above the ecliptic.

Fixes #2175 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: (Geforce)

## Checklist:


Atque:
I see the dates don't move when changing timezone or even location. Is that intended?

gzotti:
No, I am just adding such refinements. A problem is DST change. I had thought core->getUTCOffset(jd) would deliver that. No, it doesn't. But I will add signal connections to recreate partitions when timezone changes. Location does not change ticks.

alex-w:
Maybe using Rome numbers for monthes to avoid misunderstanding the dates (e.g. 20/II for February 20)?

Atque:
> Maybe using roman numbers for months to avoid misunderstanding the dates (e.g. 20/II for February 20)?

Or perhaps Jan, Feb, Mar etc, but only at the 1st each month?

10110111:
Why does it show dates in format that alternates between `DD.MM.` and `DD.`? This looks confusing, and the alignment between these alternations is broken. Also the trailing dots don't make sense to me.

stellarium-011

10110111:
> have no month attached (avoiding some clutter)

Since the user needs to understand the meaning of a figure from the context, it's not helpful to have half with the month, and another without. It's better to just reduce number of ticks when the clutter would be unavoidable.

> The trailing dots may be a reason for Localisation of dates. In my part of the world it is common to write dots to indicate ordinals e.g. in dates.

Well, I have `LC_NUMERIC=ru_RU.UTF-8` and `LC_TIME=ru_RU.UTF-8`. In `ru` locales it's not common (not even heard of unless one has studied German) to end the numbers with periods to indicate an ordinal. To me it just looks like a C++ way of spelling a `double` literal with integral value. To non-programmers in similar locales to me it would just look weird (especially after a Roman numeral).

gzotti:
OK. I did not want to go as far as localizing or even user-customizing those labels as this breaks some of the fov thinning logic. Shall I use Roman numerals for the first of each month, and "Arab" numerals for the other days? This will obviously make users unhappy when they zoom in to see just a couple of days which will then have lost their clear indication of "10.V." etc. There are always compromises to take.

gzotti:
> I have now used this line for some time, and it's incomprehensible at high zoom levels when it doesn't have any labels at all, except "Ecliptic of Date". Maybe some refinement could be made, e.g. adding time points in `HH:MM` format in such cases.

Adding intermediate values to the lines is another issue. (#1793)

gzotti:
While I somewhat disagree about showing the month on also the 5-ticks, I can live with them.

alex-w (approved):

src/core/modules/GridLinesMgr.cpp
```diff
@@ -1129,8 +1195,60 @@ void SkyLine::draw(StelCore *core) const
...
+				Vec3d end10=part10;
+
+				const Mat4d& rotDay = Mat4d::rotation(partZAxis, lng);
+				start.transfo4d(rotDay);
+				end.transfo4d(rotDay);
+				end10.transfo4d(rotDay);
+
+				sPainter.drawGreatCircleArc(start, end, Q_NULLPTR, Q_NULLPTR, Q_NULLPTR);
+
+				if ((label.length()>0) && (
```
10110111, line 1232: There's way too many parentheses here. At least the single logical expressions should be simpler, i.e. instead of `(x)` or `((x))` it should be just `x`. Like so:
```cpp
if (label.length()>0 && (
					currentFoV<60. // all labels
					|| (currentFoV<=180. && !label.contains("5"))) // 1.MM/10.MM/20.MM
					|| label.startsWith("1.") // in any case
					)
```
Also, the indentation is misleading: the final opening parenthesis on the first line is closed on the third line after `contains("5")`, which is not obvious at a glance from the indentation structure.

Moreover, the compiler (g++-9) hints that the `&&` and the last `||` are on the same level, which is even less obvious at a glance, where the two `||` seem to be on the same level. Maybe the whole complicated expression should be split into several `bool` variables that would then be more simply combined in the `if` condition.

src/core/modules/GridLinesMgr.cpp
```diff
@@ -651,6 +658,60 @@ void SkyLine::init()
...
+		lng+=psiA;
+		double aberration=-20.4898/3600.*M_PI_180/r; // Meeus AA (25.10)
+		double deltaEps, deltaPsi;
+		getNutationAngles(jde, &deltaPsi, &deltaEps);
+
+		QString label;
+		if ((year==1582) && (date[1]==10))
+		{
+			// Gregorian calendar reform. Duplicate ticks are harmless, but we must tweak date labels.
+			if (QList<int>({1, 4, 15, 20, 25}).contains(date[2]))
```
10110111, line 705: When a constructor accepts `std::initializer_list`, the outer parentheses can be omitted at the call site (since C++14), resulting in a neater expression: `QList<int>{1, 4, 15, 20, 25}`.

src/core/modules/GridLinesMgr.cpp
```diff
@@ -651,6 +658,60 @@ void SkyLine::init()
...
+	Q_ASSERT(earth);
+	StelCore *core=StelApp::getInstance().getCore();
+	eclipticOnDatePartitions.clear();
+
+	if (year==std::numeric_limits<int>::min())
+	{
+		int m, d;
+		StelUtils::getDateFromJulianDay(core->getJD(), &year, &m, &d);
+	}
+	for (int day=1; day <= (StelUtils::isLeapYear(year) ? 366 : 365); day++)
```
10110111, line 683: It would be cleaner to compute the upper bound once.

src/core/modules/GridLinesMgr.cpp
gzotti: Nice!

src/core/modules/GridLinesMgr.cpp
gzotti: No, this was indeed a logic error (without visible consequences). Too much of reformulating the rules... Thanks.

10110111 (approved):

PR #3061, opened 2023-02-15
Fix screenshot sizes

### LINKED ISSUE #2926: It is abnormal to generate large screenshots
### Expected Behaviour
Generate screenshots correctly and quickly.

### Actual Behaviour
When generating 16K (15360 * 8640, 72 dpi) pictures or 8K (7680 * 4320, 72 dpi) pictures, the program will not respond, and the generated pictures will be black.

### Steps to reproduce
Adjust the screenshot size to 16K (15360 * 8640, 72 dpi) or 8K (7680 * 4320, 72 dpi) in the settings, and then exit the settings. Press Ctrl+S to screenshot.

### System
* Stellarium version: stellarium-1.1.1-qt6-win64.exe
* Operating system: Microsoft Windows 11 Pro for Workstations, version 22H2 (OS Build 22621.963)
* Graphics Card: NVIDIA GeForce RTX 2080 Super, Game Ready driver 522.25
* Screen type (if applicable): 1920*1080

gzotti:
### Description

This is an attempt to fix a scaling issue where scaling was never expected: screenshots.

- A regular screenshot of the full-size screen shall have as many pixels as the screen, or less if screen scaling is set in the OS. In this case the image will have less pixels.
- A screenshot of a window smaller than screen size is expected to have just as many pixels as the window. 
- A screenshot with custom size has just that: a user-configured pixel count. Whatever screen scaling is set in the OS should not matter. 
- There are technical limits to screenshot sizes. Setting hardware maximum and then multiply by a scale factor is an obvious bug. 

Please explain why fiddling with a scaling factor here was important in the first place. 

For a first test, I have trivially set the pixelScale for screenshots to 1. Of course, remove code before merge. 

Fixes #2926 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: (Geforce)

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1631,7 +1631,7 @@ void StelMainView::doScreenshot(void)
 	// HiDPI screens interfere, and the viewing angle has to be maintained.
 	// First, image size:
 	glWidget->makeCurrent();
-	float pixelRatio = static_cast<float>(QOpenGLContext::currentContext()->screen()->devicePixelRatio());
+	float pixelRatio = 1.f; // static_cast<float>(QOpenGLContext::currentContext()->screen()->devicePixelRatio());
 	int imgWidth =static_cast<int>(stelScene->width());
 	int imgHeight=static_cast<int>(stelScene->height());
 	bool nightModeWasEnabled=nightModeEffect->isEnabled();
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1634,11 +1634,12 @@ void StelMainView::doScreenshot(void)
 	const bool nightModeWasEnabled=nightModeEffect->isEnabled();
 	nightModeEffect->setEnabled(false);
 	const float pixelRatio = static_cast<float>(QOpenGLContext::currentContext()->screen()->devicePixelRatio());
+	qDebug() << "pixelRatio:" << QString::number(pixelRatio, 'f', 2);
 	int imgWidth =static_cast<int>(stelScene->width());
 	int imgHeight=static_cast<int>(stelScene->height());
 	// Screen scaling makes things worse again.
-	int finalImgWidth  = int(static_cast<float>(imgWidth)  * pixelRatio);
-	int finalImgHeight = int(static_cast<float>(imgHeight) * pixelRatio);
+	//int scaledImgWidth  = int(static_cast<float>(imgWidth)  * pixelRatio);
+	//int scaledImgHeight = int(static_cast<float>(imgHeight) * pixelRatio);
 
 	if (flagUseCustomScreenshotSize)
 	{
@@ -1676,8 +1677,8 @@ void StelMainView::doScreenshot(void)
 			int maximumFramebufferSize = qMin(texSize,qMin(rbSize,qMin(viewportSize[0],viewportSize[1])));
 			qCDebug(mainview)<<"Maximum framebuffer size:"<<maximumFramebufferSize;
 
-			finalImgWidth =qMin(maximumFramebufferSize, customScreenshotWidth);
-			finalImgHeight=qMin(maximumFramebufferSize, customScreenshotHeight);
+			imgWidth =qMin(maximumFramebufferSize, customScreenshotWidth);
+			imgHeight=qMin(maximumFramebufferSize, customScreenshotHeight);
 		}
 		else
 		{
@@ -1694,19 +1695,22 @@ void StelMainView::doScreenshot(void)
 	fbFormat.setInternalTextureFormat(isGLES ? GL_RGBA : GL_RGB); // try to avoid transparent background!
 	if(const auto multisamplingLevel = configuration->value("video/multisampling", 0).toInt())
         fbFormat.setSamples(multisamplingLevel);
-	QOpenGLFramebufferObject * fbObj = new QOpenGLFramebufferObject(finalImgWidth, finalImgHeight, fbFormat);
+	//QOpenGLFramebufferObject * fbObj = new QOpenGLFramebufferObject(scaledImgWidth, scaledImgHeight, fbFormat);
+	QOpenGLFramebufferObject * fbObj = new QOpenGLFramebufferObject(imgWidth, imgHeight, fbFormat);
 	fbObj->bind();
 	// Now the painter has to be convinced to paint to the potentially larger image frame.
-	QOpenGLPaintDevice fbObjPaintDev(finalImgWidth, finalImgHeight);
+	//QOpenGLPaintDevice fbObjPaintDev(scaledImgWidth, scaledImgHeight);
+	QOpenGLPaintDevice fbObjPaintDev(imgWidth, imgHeight);
 
 	// It seems the projector has its own knowledge about image size. We must adjust fov and image size, but reset afterwards.
 	StelCore *core=StelApp::getInstance().getCore();
-	StelProjector::StelProjectorParams pParams=core->getCurrentStelProjectorParams();
+	const StelProjector::StelProjectorParams pParams=core->getCurrentStelProjectorParams();
 	StelProjector::StelProjectorParams sParams=pParams;
 	qCDebug(mainview) << "Screenshot Viewport: x" << pParams.viewportXywh[0] << "/y" << pParams.viewportXywh[1] << "/w" << pParams.viewportXywh[2] << "/h" << pParams.viewportXywh[3];
 	qCDebug(mainview) << "Screenshot: setting dimensions to" << imgWidth << "x" << imgHeight;
 	sParams.viewportXywh[2]=imgWidth;
 	sParams.viewportXywh[3]=imgHeight;
+	sParams.devicePixelsPerPixel=1.; // Forget about scaling. We know what we want!
 
 	// Configure a helper value to allow some modules to tweak their output sizes. Currently used by StarMgr, maybe solve font issues?
 #if (QT_VERSION>=QT_VERSION_CHECK(5,12,0))
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -818,7 +818,7 @@ QSurfaceFormat StelMainView::getDesiredGLFormat(QSettings* configuration)
 #else
 	const bool vsdef = true;
 #endif
-	if (configuration->value("video/vsync", vsdef).toBool())
+	if (configuration && configuration->value("video/vsync", vsdef).toBool())
 		fmt.setSwapInterval(1);
 	else
 		fmt.setSwapInterval(0);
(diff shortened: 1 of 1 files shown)


10110111:
Apparently, it's there for a reason, otherwise I get this:

stellarium-007

10110111:
With the latest commit it looks like this:

stellarium-009

gzotti:
Is this a regular screenshot (unexpected) or a custom-sized one (what I have commented as still buggy)?
The key question is what messes this up. IMHO we both know what a user expects, but something in the elaboration of projection parameters breaks.

10110111:
In both cases it was custom size (1152×864 was chosen in the GUI). The default size works fine. Didn't notice the comment (and still don't find it).

gzotti:
Second line in my commit message. You must click on the dots in the GH page to see. 
I think we need to configure the sParams variable properly, also with custom dimensions.

gzotti:
@alex-w , @10110111 is it OK to disable scaling for screenshots entirely? I don't see the point for scaling when we have a preset window size or a configuration for custom screenshot size. And also a scaled full-screen window just has a limited native size (smaller than the screen resolution). 
Obviously I will clean up the current messy state. Else I need your comments/input/alternative solution.

gzotti:
@alex-w , @10110111 is it OK to disable scaling for screenshots entirely? I don't see the point for scaling when we have a preset window size or a configuration for custom screenshot size. And also a scaled full-screen window just has a limited native size (smaller than the screen resolution). 
Obviously I will clean up the current messy state. Else I need your comments/input/alternative solution.

gzotti:
@alex-w , @10110111 is it OK to disable scaling for screenshots entirely? I don't see the point for scaling when we have a preset window size or a configuration for custom screenshot size. And also a scaled full-screen window just has a limited native size (smaller than the screen resolution). 
Obviously I will clean up the current messy state. Else I need your comments/input/alternative solution.

10110111:
The custom size could be at 100% scale, but the window-based size should still use the scaling of the window. This would ensure consistency between what we'd get when taking screenshots via <kbd>PrtScr</kbd> and via <kbd>Ctrl</kbd>+<kbd>S</kbd>.

gzotti:
But when I have set a window size of 1920x1080 to always have FullHD screenshots in non-fullscreen operation, this would result in oversized images. This comes unexpected IMHO. If scaling depends on actual monitor, this may also lead to weird behaviour on multi-monitor setups with mixed scaling (it then depends on which monitor the window happens to be).  Likewise, when I run 125% scaling in full screen on a 3440x1440 monitor, the GL window is actually only 2752x1152. Why should a screenshot then become larger in resolution? Yes, a system screenshot makes it 3440x1440. I would leave *this* to the users to make system screenshots in the physical resolution of the monitor.

10110111:
> But when I have set a window size of 1920x1080 to always have FullHD screenshots in non-fullscreen operation, this would result in oversized images.

This is not the expected behavior. A window with size 1920x1080 has OpenGL viewport with a smaller size. This is the size for the screenshot.

> If scaling depends on actual monitor, this may also lead to weird behaviour on multi-monitor setups with mixed scaling (it then depends on which monitor the window happens to be).

If the current window has e.g. button size of 40 px, the same should be in the screenshot. We just assume that the monitor that contains this window will be the monitor on which the screenshot will be viewed.

> Likewise, when I run 125% scaling in full screen on a 3440x1440 monitor, the GL window is actually only 2752x1152.

This is the point of view of QtWidgets. From OpenGL perspective (e.g. `glGetViewport`) we should see the physical resolution. This resolution is what should be the size of the window-sized screenshot.

> Yes, a system screenshot makes it 3440x1440. I would leave this to the users to make system screenshots in the physical resolution of the monitor.

This goes against the principle of least surprise.

gzotti:
> > But when I have set a window size of 1920x1080 to always have FullHD screenshots in non-fullscreen operation, this would result in oversized images.
> 
> This is not the expected behavior. A decorated window with size 1920x1080 has OpenGL viewport with a smaller size. This is the size for the screenshot.

At least not on Windows. In my config.ini I have 
```
[video]
screen_h                                  = 1080
screen_w                                  = 1920

```
and screenshots while in windowed mode at 125% scaling now (with this branch) are back to 1920x1080. The given dimensions are the viewport area. The image has no window decorations. If I use Greenshot to create an image of the upscaled window with decoration, it is sized 2402x1398. 
Do you have a window of 1920x1080 *including* decoration on Linux when set with these configs?

> > If scaling depends on actual monitor, this may also lead to weird behaviour on multi-monitor setups with mixed scaling (it then depends on which monitor the window happens to be).
> 
> If the current window has e.g. button size of 40 px, the same should be in the screenshot. We just assume that the monitor that contains this window will be the monitor on which the screenshot will be viewed.

Usually a screenshot will be printed on paper or placed e.g. in a slide show to be displayed elsewhere. (Other screen, other potential scaling, ...) The relative size of buttons in the upscaled window or in the correct-sized image are now the same. 

 
> > Likewise, when I run 125% scaling in full screen on a 3440x1440 monitor, the GL window is actually only 2752x1152.
> 
> This is the point of view of QtWidgets. From OpenGL perspective (e.g. `glGetViewport`) we should see the physical resolution. This resolution is what should be the size of the window-sized screenshot.

I have to check that sometime next weekend. 

> > Yes, a system screenshot makes it 3440x1440. I would leave this to the users to make system screenshots in the physical resolution of the monitor.
> 
> This goes against the principle of least surprise.

Currently I am very surprised about a mismatch between configured window size and screenshot size when I have some scaling set.

10110111:
> Do you have a window of 1920x1080 including decoration on Linux when set with these configs?

Stellarium fails to apply this size, just maximizes the window whenever the size of contents + decorations exceeds the screen size. But, if I choose something smaller to fit in the screen, like 1920×1000, I get total decorated size of 1920×1042 (and `xwininfo` reports 1920×1000). So decoration doesn't count in this setting (which I didn't expect). This was with 100% scaling. And with 150% scaling the physical size is upscaled from this setting. I.e. with the size 800×400 in the settings I get 1208×642 total size of the decorated window on the screen (`xwininfo` reports 1200×600).

> Usually a screenshot will be printed on paper or placed e.g. in a slide show to be displayed elsewhere.

Well it's one possible use case of course, but even then using 100% scale for screenshots might be even less desirable unless you want to emphasize pixelization of the GUI. Just look into the SUG and notice how pixelized the screenshots look (figures 3.1, 3.2, chapter backgrounds etc.). To get nicer illustrations in the books the creators should want higher scale factors, allowing for higher DPI on printing and leaving the GUI elements of readable size.

> Currently I am very surprised about a mismatch between configured window size and screenshot size when I have some scaling set.

Me too. This is due to the very unintuitive behavior of the whole pixel-ratio kludge. I'd really like to be able to simply get the correct scale factor, a single number, from Qt and then have Qt behave as if it was 100%. Then it'd be possible to do all the required scaling in a custom, well thought out, way. But alas, Qt either reports the correct scaling factor and additionally distorts all the sizes in an inconsistent way, or doesn't support scaling at all.

gzotti:
> > Do you have a window of 1920x1080 including decoration on Linux when set with these configs?
> 
> Stellarium fails to apply this size, just maximizes the window whenever the size of contents + decorations exceeds the screen size. But, if I choose something smaller to fit in the screen, like 1920×1000, I get total decorated size of 1920×1042 (and `xwininfo` reports 1920×1000). So decoration doesn't count in this setting (which I didn't expect). This was with 100% scaling. And with 150% scaling the physical size is upscaled from this setting. I.e. with the size 800×400 in the settings I get 1208×642 total size of the decorated window on the screen (`xwininfo` reports 1200×600).

OK, so at least this is consistent between Win and Linux. 
 
> > Usually a screenshot will be printed on paper or placed e.g. in a slide show to be displayed elsewhere.
> 
> Well it's one possible use case of course, but even then using 100% scale for screenshots might be even less desirable unless you want to emphasize pixelization of the GUI. Just look into the SUG and notice how pixelized the screenshots look (figures 3.1, 3.2, 

These are ancient (2008, when screens were often 1280x768 or so, and no custom resolution was available.)

> chapter backgrounds etc.). To get nicer illustrations in the books the creators should want higher scale factors, allowing for higher DPI on printing and leaving the GUI elements of readable size.

We have introduced custom size in 0.18.2 and user-controlled dpi (metadatum only) in 0.22.0. The chapter heads are from 0.14 or so. So yes, we could care to remake custom-sized hi-res screenshots for the Guide. 

> > Currently I am very surprised about a mismatch between configured window size and screenshot size when I have some scaling set.
> 
> Me too. This is due to the very unintuitive behavior of the whole pixel-ratio kludge. I'd really like to be able to simply get the correct scale factor, a single number, from Qt and then have Qt behave as if it was 100%. Then it'd be possible to do all the required scaling in a custom, well thought out, way. But alas, Qt either reports the correct scaling factor and additionally distorts all the sizes in an inconsistent way, or doesn't support scaling at all.

The scaling is an ugly mess, also as it was introduced within the platforms, as it appears to differ by-platform. However, Qt has improved documentation a lot. See https://doc.qt.io/qt-5/highdpi.html and https://doc.qt.io/qt-6/highdpi.html. I never saw the point for 21" 4k screens that then needed upscaling to be usable.  I prefer good colorspace coverage over excessive pixel count.

So, we could decide to even wrap one of the environment variables around the program and fully activate and follow one option. However, I have no overdense screens and always run at 100%, so did not care until now when I saw this oversize fbo issue which really is a bug. 

For me as user, I see a window size is configured in pixels, is apparently upscaled by OS setting, but a Ctrl-S snapshot of this window from within our program can be expected to have the size given in the configuration file, regardless on which monitor the window was displayed (see https://github.com/Stellarium/stellarium/wiki/Common-Problems-for-the-current-version#mixing-screens-or-projectors-with-high-and-normal-resolutions). Capturing the monitor with system tools may provide something else. The scaling apparently should influence GUI elements like buttons and font size only. Screenshots with visible buttons are IMO aimed at documentation/tutorials and are rarely hi-res, and a very high-res (8k or better) screenshot is aimed for large poster prints, i.e. usually goes without any GUI. We should then maybe even consider adding a switch or percentage setting to upscale GUI in screenshots or not (but how to re-arrange GUI panels from a horizontal screen onto a vertical poster?), i.e. make huge fonts on a poster print, or leave them in original size to have more and smaller labels. But this is a next round, which would require a deeper HiDPI redesign of deciding which screen element has to be scaled in which context.

alex-w:
Screenshots from macOS 13.2.1

stellarium-005
stellarium-006
stellarium-007

gzotti:
I cannot see clearly, is that OK then? Looks good to me. 
The question is, if you have configured your window to be 1200x800 in config.ini, run Stellarium in windowed mode on a 4k screen at scaling 150% so that it appears with 1800x1200 screen pixels, and make a regular screenshot, is the screenshot to be expected by the user to be 1200x800, or 1800x1200?

If you then have a setup with one screen at 100%, the other at 150%, prepare some presentation, and you move an 800x600 Stellarium window from one screen to the other and make the next default-size screenshot, would you expect the screenshot dimensions to become larger when the screenshot was made when Stellarium was displayed on the scaled screen? I would be surprised and would report it as bug.

alex-w:
> I cannot see clearly, is that OK then? Looks good to me. 

It's OK here

> The question is, if you have configured your window to be 1200x800 in config.ini, run Stellarium in windowed mode on a 4k screen at scaling 150% so that it appears with 1800x1200 screen pixels, and make a regular screenshot, is the screenshot to be expected by the user to be 1200x800, or 1800x1200?
> 
> If you then have a setup with one screen at 100%, the other at 150%, prepare some presentation, and you move an 800x600 Stellarium window from one screen to the other and make the next default-size screenshot, would you expect the screenshot dimensions to become larger when the screenshot was made when Stellarium was displayed on the scaled screen? I would be surprised and would report it as bug.

config.ini: 900x600

standard resolution (1440x900), windowed mode, MBP:
stellarium-002

low resolution (1024x640), windowed mode, MBP:
stellarium-005

hi resolution (1680x1050), windowed mode, MBP:
stellarium-006

standard resolution, windowed mode, iPad as second monitor:
stellarium-001

alex-w (approved):

10110111:
Current state of this branch results in 1280×720 screenshot with 100%-scaled GUI, when default size is specified (non-custom) and the window is 1920×1080 fullscreen with scale factor 150%. This is unacceptable, whatever decision for custom-sized screenshots is taken.

gzotti:
Maybe #3071 is the better solution.

Issue #2899, opened 2022-12-08
Pixelating bottom panel icons on screenshots


alex-w:
### Expected Behaviour
Smooth bottom panel icons on the screenshots

### Actual Behaviour
On HiDPI devices screenshots has pixelized bottom panel icons when it taken by Stellarium itself. See bottom panel on screenshots below.

Screenshot from Stellarium (72 DPI):
![stellarium-003](https://user-images.githubusercontent.com/88731/206483566-81eb5983-eeec-45c6-82c7-a84f421c332b.png)

Screenshot from Stellarium (144 DPI):
![stellarium-004](https://user-images.githubusercontent.com/88731/206485215-2379a8f5-2496-4179-9b74-01dbaa8d37ba.png)

Screenshot from Stellarium (300 DPI):
![stellarium-005](https://user-images.githubusercontent.com/88731/206486489-56f2190a-c7b8-47da-b9ad-986f2e4108c5.png)

Screenshot from macOS:
![Снимок 08 12 2022 в 22 04](https://user-images.githubusercontent.com/88731/206483892-7ab9c44f-54eb-4e5f-b0ed-47da42ee26e5.jpeg)

### System
* Stellarium version: master
* Operating system: macOS 13.0.1 on MacBook Pro
* Screen type (if applicable): 2880x1800

10110111:
Also reproduces on GNU/Linux/X11 with Qt scaling factor 1.5.

PR #3071, opened 2023-02-25
Fix screenshot size

### LINKED ISSUE #2926: It is abnormal to generate large screenshots

10110111:
### Description

This is an alternative solution to #3061.

The custom-sized screenshots now have the exact dimensions specified in the GUI, while the default-sized screenshots just follow the current physical size of the window.

Scaling factor is retained, so if the window is currently on a high-DPI monitor, GUI is upscaled to the same factor as shown on screen. The idea for the behavior of custom-sized screenshots is that the results should be as if the user had resized the window to the target size, took default-sized screenshot, and then resized the window back. Obviously, such actions wouldn't alter scaling factor.

I think this is the most intuitively sensible behavior for normal users.

~~One problem that remains unsolved is that the upscaled panel icons are pixelized (which doesn't happen with QWidget GUIs). This one is tracked in #2899. I sort of know what's going on, but Qt behaves unintuitively with `QOpenGLPaintDevice::setDevicePixelRatio()`, so I don't attempt try to fix it in this PR.~~

Fixes #2926

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1631,9 +1631,9 @@ void StelMainView::doScreenshot(void)
 	// HiDPI screens interfere, and the viewing angle has to be maintained.
 	// First, image size:
 	glWidget->makeCurrent();
-	float pixelRatio = static_cast<float>(QOpenGLContext::currentContext()->screen()->devicePixelRatio());
-	int imgWidth =static_cast<int>(stelScene->width());
-	int imgHeight=static_cast<int>(stelScene->height());
+	const auto pixelRatio = QOpenGLContext::currentContext()->screen()->devicePixelRatio();
+	int physImgWidth  = std::lround(stelScene->width() * pixelRatio);
+	int physImgHeight = std::lround(stelScene->height() * pixelRatio);
 	bool nightModeWasEnabled=nightModeEffect->isEnabled();
 	nightModeEffect->setEnabled(false);
 	if (flagUseCustomScreenshotSize)
@@ -1672,8 +1672,8 @@ void StelMainView::doScreenshot(void)
 			int maximumFramebufferSize = qMin(texSize,qMin(rbSize,qMin(viewportSize[0],viewportSize[1])));
 			qCDebug(mainview)<<"Maximum framebuffer size:"<<maximumFramebufferSize;
 
-			imgWidth =qMin(maximumFramebufferSize, customScreenshotWidth);
-			imgHeight=qMin(maximumFramebufferSize, customScreenshotHeight);
+			physImgWidth =qMin(maximumFramebufferSize, customScreenshotWidth);
+			physImgHeight=qMin(maximumFramebufferSize, customScreenshotHeight);
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


10110111:
Special notice about `config.ini:[video]/screen_{w,h}`: these values are buggy, they shouldn't be used directly for default-sized screenshots. Instead the code that reads and writes them must be fixed to operate on physical window dimensions.

gzotti:
I think custom sizes now need their own scaling setting. When I want 1024x768 screenshots for some manual, the buttons look oversized when I prepare this on an upscaled 4k screen. When I then continue writing on another system with a FullHD screen (scaling=1.0), result looks different. 

It's not me, I don't scale.  But I can see the demand.

10110111:
> I think custom sizes now need their own scaling setting.

Yes, this would be a good solution. But it's not the focus of this PR, which is about image dimensions and the principle of least surprise.

alex-w (approved):

gzotti:
And we disagree on what surprises us the least. For me a window with given size in config.ini should have same-sized screenshots. You started caring about scaling them, so I ask for a completion of these concerns in the context of "Fix screenshot size". I think it may be enough to add a property "customScreenshotScaling" (double; with GUI doubleSpinner [1...5 in 0.25 increments?] in Config/Tools), and -- maybe (untested) reset the y value of the gui and rootItem in L1725ff. But I have not dug deep enough to see how the buttons and dialogs are scaled.

10110111:
> For me a window with given size in config.ini should have same-sized screenshots.

Fine, but this setting should also result in this size applied to the window on the screen, which it doesn't. This is a separate bug. Screenshot functionality reflects the reality (just like it would be if done by the OS/windowing system) rather than what's written in the config file. Ideally, these should coincide, but due to the bug in settings I/O they don't.

> I think it may be enough to add a property "customScreenshotScaling"

GUI-wise, it might be enough, but I'm afraid it would take quite a bit of effort to pass it through to yield consistent output. This is why I think it should be done in a separate PR. The current PR just brings the custom-sized screenshots to the actual dimensions requested.

gzotti:
When starting Stellarium and then changing the scaling, the screenshot size of a 1920x1080 windowed mode scaled in an 125% screen is now 1920x1072. And when switching between Fullscreen and windowed mode shows ugly pixelation in one mode. Funny, when starting scaled and switching to unscaled, the fullscreen screenshot resolution is too large. 

However, these are minor quibbles. The most important result in this effort to kill #2926 is that screenshots now come in the expected size, at least when changing screen scaling *before* launching the program.

10110111:
> and then changing the scaling

IIRC, dynamic change of scaling (other than by moving the window to another monitor) is not well-supported by Qt: in particular, the GUI becomes blurry instead of correctly rescaling everything. So I think we can ignore this for now.

gzotti (approved):
Good enough for now. Some scaling issues remain, though.

PR #3075, opened 2023-02-26
Render ShowMySky atmosphere calculating view direction per fragment

10110111:
### Description

This PR makes atmosphere look better in the parts of the screen where view direction changes rapidly.

One problem that we have here is that switching between projections takes about 3 seconds to recompile all the ShowMySky shaders. After the particular projection has been switched to, usually the shaders are cached by the GPU driver (at least by NVIDIA and Mesa), so this has a limited impact. But I'm not sure if the initial delay is OK to have.

An alternative solution is to render the view directions to a separate texture and use it as an additional input for the renderer. This gets rid of the switching delay, but reduces performance for all other times (by about 9% of frame rate on my machine).

#### Old

Screenshot_2023-02-27_00-01-40

#### New

Screenshot_2023-02-27_00-02-12

### Type of change

### How Has This Been Tested?
**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/modules/AtmosphereShowMySky.cpp
+++ b/src/core/modules/AtmosphereShowMySky.cpp
@@ -96,7 +96,7 @@ struct SkySettings : ShowMySky::Settings, QObject
 	bool useEclipseShader_ = false;
 
 	// These variables are not used by AtmosphereRenderer, but it's convenient to keep them here
-	QMatrix4x4 projectionMatrix_;
+	QMatrix4x4 projectionMatrixInverse_;
 
 private:
 	static constexpr const char* zeroOrderScatteringPropName() { return "LandscapeMgr.flagAtmosphereZeroOrderScattering"; }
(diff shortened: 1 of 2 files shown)


gzotti:
Looks great! 
I think 3s for shader recompile is acceptable (switching is not done often IMO). 9% performance loss would be bad though.

10110111:
Did you try it out? I'd like to know whether it's acceptable after you actually have experienced it.

gzotti:
Not tried yet.

gzotti:
Works great on the Geforce. Switching delay totally acceptable IMO. Testing others...

alex-w:
Similar visual issue I see for Preetham model too - is it possible resolve that too?

alex-w (approved):

10110111:
> Similar visual issue I see for Preetham model too - is it possible resolve that too?

No, Preetham is in principle CPU-based. Any attempt to move it to the GPU is going to slow it down. Given that it's useful only on slow machines, I wouldn't try to do this, just I didn't try to support GLES2 with new-way old-style landscapes.

gzotti:
Works well as-is on my Intel. Anything to try out? Qt5/6? Angle and Raspi have no SMS, so I skip those.

10110111:
> Works well as-is on my Intel

Was it on Windows? AFAIR, Intel@Windows doesn't even cache the shaders. On Mesa and `nvidia` at least the delay happens only once per projection, because the compilation&linking results are cached.

gzotti:
Yes, WIn10. Core i5-4570. (Intel HD4600). Performance is low (about 6-7fps with SMS, about 17fps with Preetham). Stellarium 1.2 has 6-7fps with SMS. Switching causes a delay, but is acceptable.

alex-w (approved):

PR #3080, opened 2023-03-04
Add a workaround for poor NPOT mipmap implementations

10110111:
### Description

We use the deepest mipmap level as the average of texture data in order to compute brightness of the scene in AtmosphereShowMySky. Some OpenGL implementations, however, appear to handle non-power-of-two textures poorly when computing mipmap levels. The result is that, when a texture dimension is close to a power of two but not exactly it, the whole mip map displays ugly aliasing, and the deepest 1×1 level is very far from the average of the texels of level 0.

The effect can be seen when resizing the window on e.g. Mesa-based drivers on Linux (and, IIRC, Intel on Windows). As you resize the window, the daytime atmosphere will change its brightness erratically, which shouldn't really happen.

To work around such implementations, we resize the base level texture to the closest smaller power-of-two texture, and compute the mip map from it instead of the initial texture. The resize introduces a small inaccuracy, but the result is still much higher quality than if we used glGenerateMipmap() on the initial NPOT texture.

On good implementations such as NVIDIA GPU drivers, the workaround is disabled.

The functionality of such averaging is implemented in a class `TextureAverageComputer` that's only available on desktop OpenGL. On OpenGL ES 2 it's `#ifdef`ed out.

### Type of change

### How Has This Been Tested?
**Test Configuration**:

## Checklist:

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -126,6 +126,8 @@ SET(stellarium_lib_SRCS
      core/StelOpenGLArray.cpp
      core/StelHips.hpp
      core/StelHips.cpp
+     core/TextureAverageComputer.hpp
+     core/TextureAverageComputer.cpp
 
      ${spout_SRCS}
 
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -33,6 +33,7 @@
 #include "StelOpenGL.hpp"
 #include "StelOpenGLArray.hpp"
 #include "StelProjector.hpp"
+#include "TextureAverageComputer.hpp"
 
 #include <QDebug>
 #include <QDir>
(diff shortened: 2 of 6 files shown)


src/CMakeLists.txt
```diff
@@ -126,6 +126,8 @@ SET(stellarium_lib_SRCS
      core/StelOpenGLArray.cpp
      core/StelHips.hpp
      core/StelHips.cpp
+	 core/TextureAverageComputer.hpp
```
gzotti, line 129: please fix indentation

gzotti (approved):
Tested on my Intel Core i5-4570. I never saw the brightness issue during windows resize before, but on this slow system I usually have the old sky model. Yes, I confirm this works here!

alex-w (approved):

PR #3087, opened 2023-03-07
Losslessly recompress PNGs to save space

afontenot:
### Description
Many of the PNGs included for textures or other reasons in Stellarium were poorly optimized.
I ran all of them through a PNG compressor to reduce their size. The resulting images are not
changed in any way - the operation is fully lossless - but the installed size of Stellarium is reduced
by more than 32 MB.

This makes it easier to download and use Stellarium on metered connections that are common
worldwide.

I used [ect](https://github.com/fhanau/Efficient-Compression-Tool) with options

    ect --strict -9

### Type of change

### How Has This Been Tested?
As no changes to code are involved here, testing and code review don't appear to be relevant.
If there's something I've missed I'm happy to look at this further.


gzotti:
Good idea, thank you. However, as the buttons are created from SVGs, this should be added to the processing chain, and recompressing PNG screenshots for the Guide should be recommended in the Guide's README (The optimal way would be a fast check to attempt recompression of updated or new PNGs). I don't know ect, does this build&run on all our platforms?

alex-w:
Good idea, thanks!

I agreed with Georg, some documentation/instructions are required.

alex-w (approved):

afontenot:
One option for now is I could do a version of this PR skipping the changes to automatically generated button PNGs. There's still a ton of space to be saved on long term assets like space imagery. 

ECT compiles for Win / Mac / Linux and compiles on ARM as well I believe. You could also potentially download static builds for Win / Mac if that's appropriate for your build pipeline. One issue is that you wouldn't want to run this tool on the whole build every time, because it's quite slow (especially in the maximum compression mode that I used - I left it to run overnight). So you'd want to set it just to run on the button images, maybe with `ect -3` (the default) instead of `-9`.

If you wanted to require PRs that make changes to images to always pre-optimize them you could use a simple test like

    ect -9 --strict $imagefile | grep -q '^Saved 0B out of'

as the grep would fail whenever ect manages to reduce the size of an image.

10110111:
> as the buttons are created from SVGs, this should be added to the processing chain

This is not done automatically. One just opens the SVG in Inkscape, selects a square as described there and exports. There's no automatic processing chain.

gzotti:
All this doc about ect should go into e.g. a new section in MAINTAINER_BUSINESS. Occasionally, recently added textures can then be recompressed. (E.g. during preparation of each release?)

alex-w:
> E.g. during preparation of each release?

For newly added images - of course!

alex-w:
@afontenot please resolve conflict

afontenot:
I've rebased the PR and also added some explanatory text to MAINTAINER_BUSINESS. This could be revised or expanded as necessary, perhaps in future commits.

alex-w (approved):
Thanks!

gzotti (approved):
Good description, thank you!

Issue #1638, opened 2021-05-02
right click in toolbar to open plugin config dialog


axd1967:
**Is your feature request related to a problem? Please describe.**
To open some config dialogs of plugins, I need to go via F2 > Plugins > [plugin] > `Configure`.

**Describe the solution you'd like**
Right-click on the plugin's button in the Toolbar opens that plugin's config dialog.

PR #3093, opened 2023-03-10
Added GUI to manually entering FOV

### LINKED ISSUE #3013: finer grained zooming - manual plate solving
Sometimes I want to compare Stellarium output to real images -  a kind of "manual plate solving".
For this it is important to be able to select a precise zoom factor + (added:) rotation angle.

image

image
(Source: https://www.facebook.com/groups/sterrenkundeforumVVS/posts/1958468424356632/)

This shows a small offset between computed and actual position (in this case probably about 1 arcmin).
Such an offset can be expected, nevertheless it gives an indication of precision.

The problem is that I cannot accurately select a scale so that image and viewport coincide sufficiently, because there is no way to select fractional zoom levels. As a result, there is a significant scale difference on the right side of the images.

A text box that allows to enter a more precise zoom/scale level; or, some way to expand the viewport with the mouse (e.g. ctrl-drag).

Approximation.

**Additional context**

alex-w:
### Description
In this pull request I've proposes the optional GUI (available through hotkey F9) to manually entering the field of view (simlar to Date/Time window). Of course this tool changes FOV only - no rotation!

Parially fix #3013 (issue)

### Screenshot
stellarium-003

### Type of change

## Checklist:


10110111:
"Use decimal degrees" option should be honored.

10110111:
Also, I think the UX would be best if the FOV field in the bottom panel became a button that would open this dialog (and have a tooltip to hint at the <kbd>F9</kbd> shortcut).

alex-w:
> "Use decimal degrees" option should be honored.

OK, this is good point.

alex-w:
ok, it’s done.

@10110111 @gzotti please review the feature.

@axd1967 please check it too

alex-w:
> Why show both D°M'S" and D.D° when it's not done this way in any other angle entry field (e.g. in _Location_ dialog)?

Why not? :)

Of course I can add code to change the UI to follow "Use decimal degrees option on-the-fly, but… 1) the FOV value on the bottom bar are configurable and I don’t want to add the all possible combinations; 2) Date/Time window has in addition to the date and time JD/MJD - here is similar feature; 3) someone can use this tool as convertor between DMS and DD values.

Or you expect to see one field of input for FOV value (like as in Location dialog)?

10110111:
> Or you expect to see one field of input for FOV value (like as in Location dialog)?

Yes, this is definitely what I expected. Otherwise why do we even have this checkbox in the _Tools_ dialog?

I don't quite like the current two-formats form, but OK, let's see what others think.

gzotti:
Do we really need a full new dialog class, css addition and precious hotkey for such rarely used fine-tweak functionality? A simple AngleSpinBox in ViewSettings/Sky/Projection should be enough IMHO. Please avoid code bloat! Most users (all-1?) will continue to use regular zoom and maybe the ten configurable FOV shortcuts. 

If you think this additional dialog is necessary, I have a few remarks:

- It seems too wide and could be made more compact.
- It should then have fixed (minimal) size.
- when I spin the arcminutes with mouse wheel, the zero seems to block advancing if rotated slowly, but I can "crash through the wall" if I rotate the mouse wheel fast enough. The arrow buttons seem to never go through the zero.

Atque:
> 3) someone can use this tool as convertor between DMS and DD values.

I think this is a useful feature, so I vote for showing both in the same window.

10110111:
> I think this is a useful feature, so I vote for showing both in the same window.

Purporting a completely unrelated dialog for such a converter is really bad design. Not only because the purposes are mixed so that using the converter has a side effect of changing the scene, but also:

 * a subsequent redesign can break down the secondary purpose;
 * discoverability of such a tool is poor.

Besides, when done in a single-format way, if you enter an angle in the current format and then toggle the setting in _Tools_ dialog, the conversion would happen automatically, just as it is now in _Location_ dialog.

> Do we really need a full new dialog class, css addition and precious hotkey for such rarely used fine-tweak functionality?

+1 about the hotkey, and then my suggestion to use a panel button for this would help avoid spending a key.

As for the separate class, I think it's completely OK to use a new class since it improves modularity while letting one connect spinboxes with each other, leaving this invisible to other code.

But the CSS additions, albeit hacky (as most CSS-based styles in Qt), is basically the same design as in `DateTimeDialog`, as you might notice by its look (and by the same problem of horizontal bloating).

> Most users (all-1?) will continue to use regular zoom and maybe the ten configurable FOV shortcuts

All users will continue using the regular zoom, but some of them will also be happy to be able to set the exact desired zoom level on the fly.

axd1967:
looking good.

1. the UI should be opened by right-clicking on the status bar FOV field, similar to the behaviour of right-clicking in the bottom toolbar. this is to stay in line with the philosophy of opening related dialogs via the toolbar. both UI elements are intimately related. see also #1638 as well as https://github.com/Stellarium/stellarium/pull/3093#issuecomment-1463638229. this might require extra coding/separate issue, as this logic is not yet present for the status bar
1. the mouse wheel action is a bit confusing. an improvement is to ensure that the behaviour is the same whether inside or outside the F9 dialog. an improvement is a checkbox in the F2 (of F9) dialog to inverse the zoom behaviour of the main window so that zooming requires the same mouse wheel actions. 
1. I'm not sure F9 is a good way to spend a precious shortcut on this less used feature; I would use a more distant key sequence, such as e.g. Alt-F9
2. could not replicate auto-sliding window behaviour (I use a laptop)

10110111:
> the UI should be opened by right-clicking on the status bar FOV field, similar to the behaviour of right-clicking in the bottom toolbar

Right-clicking in the bottom toolbar triggers a side function ("configure"), while left-clicking toggles the plugin. Doing it the same way with FOV doesn't make sense, since the primary purpose of this button would be to configure FOV (we can't "toggle the field of view"). And if the button highlights as a button and left-clicking doesn't do anything, it's poor UX, and it won't make the user expect that right-clicking might work.

> could not replicate auto-sliding window behaviour

It's the panel that slides, not a window. If you have the bottom panel always expanded (this is toggled by a button in the corner), then you'll not have this problem.

gzotti:
It would be an added primary dialog of the same rank as the time dialog, so a high-valued F9 keystroke (so far unused) and even button in the left bar would be called for. However, I really see no need for the whole dialog with all required plumbing and boilerplating when the same rarely used action can be added easily in an obvious location of an existing dialog. I would rather allow plugins with large GUIs take left-bar buttons (E.g. Scenery3D, Oculars), but this demand has been remedied by plugin display buttons with right-click config action in the bottom bar. (We should probably add right-click actions to the leftmost bottom bar buttons to show the respective tab in the view dialog!)
Adding buttons to the purely text-driven output ("status line") in the sliding bottom bar requires thorough redesign of the menu system. Currently it is possible to configure which fields are shown in the text output.

10110111:
Well, one option is to just add a spinbox near the "Custom FoV limit" in View dialog, naming it something like "Current FoV". It just must not interfere with possible very small values of `navigation/min_fov`, i.e. not force its digit limits on normal wheel-driven control.

gzotti:
> Well, one option is to just add a spinbox near the "Custom FoV limit" in View dialog, naming it something like "Current FoV". 

Exactly here. 

>It just must not interfere with possible very small values of `navigation/min_fov`, i.e. not force its digit limits on normal wheel-driven control. Maybe even set the number of fractional digits in the spinbox to fit this minimum.

Ah yes. I did not check behaviour of the proposed dialog in this respect. But TBH, is a fov of <0.01" useful in practice?

10110111:
> But TBH, is a fov of <0.01" useful in practice?

Maybe not, but it would be unfortunate to not have a way to zoom deeper merely because of a misbehaving GUI control.

alex-w:
Please check the changes

10110111:
Seems to work, but has one annoyance: pressing <kbd>Enter</kbd> opens _Skylight Details_ dialog (I have `Skylight/enable_gui` set to `true`).

src/core/StelMovementMgr.hpp
```diff
@@ -100,6 +100,10 @@ class StelMovementMgr : public StelModule
 		   READ getUserMaxFov
 		   WRITE setUserMaxFov
 		   NOTIFY userMaxFovChanged)
+	Q_PROPERTY(double currentFov
+		   READ getCurrentFov
+		   WRITE setFov
+		   NOTIFY currentFovChagned)
```
gzotti, line 106: typo: changed

data/gui/normalStyle.css
```diff
@@ -879,6 +879,8 @@ QLineEdit[collision="true"] {
 	padding-top: 20px;
 }
 
+
```
gzotti, line 882: Please remove two new empty lines. (i.e., normalStyle has no changes now)

src/core/StelMovementMgr.hpp
alex-w: Thanks!

data/gui/normalStyle.css
alex-w: Ok

alex-w:
> Seems to work, but has one annoyance: pressing Enter after typing the value opens _Skylight Details_ dialog (I have `Skylight/enable_gui` set to `true`).

Hmm… I don’t understand why it happens

gzotti:
> > Seems to work, but has one annoyance: pressing Enter after typing the value opens _Skylight Details_ dialog (I have `Skylight/enable_gui` set to `true`).
> 
> Hmm… I don’t understand why it happens

This may be the age-old issue with UI elements having focus for the Enter key. (See also #1971.) I almost always only operate the GUI with the mouse, so I almost never press enter or tab...

alex-w:
> This may be the age-old issue with UI elements having focus for the Enter key. (See also #1971.) I almost always only operate the GUI with the mouse, so I almost never press enter or tab...

No, this is unrelated issue. 

P.S. I found and fixed the issue.

10110111:
Your fix will break the intention of #1971. This is more a workaround for some other problem than a fix.
I had a similar problem when I was adding ShowMySky support into the Atmosphere dialog, and I didn't have to make anything unfocusable to fix it. Need to recall what I did instead...

alex-w:
> Your fix will break the intention of #1971. This is more a workaround for some other problem than a fix. I had a similar problem when I was adding ShowMySky support into the Atmosphere dialog, and I didn't have to make anything unfocusable to fix it. Need to recall what I did instead...

`focusPolicy::NoFocus` for `pushButtonAtmosphereDetails`

10110111:
Not really, this part came in long before me, in 916d26b6b53be3ee87df39c716c37d8381ea5988.

10110111:
OK, recalled: [this](https://github.com/Stellarium/stellarium/blob/master/src/gui/AtmosphereDialog.cpp#L123) was my solution.

alex-w:
Stop! You are proposed add eventFilter to ViewDialog?

10110111:
> You are proposed add eventFilter to ViewDialog?

Yes, I proposed to add a filter to ViewDialog and reject <kbd>Enter</kbd> and <kbd>Return</kbd> keys. They are never used for accepting this dialog anyway (unlike <kbd>Esc</kbd>, which rejects it).

10110111:
> Probably I doing something wrong

You might have forgotten to install the event filter: simply overriding the method is not enough.

Anyway, I've pushed the version that works for me.

alex-w:
> You might have forgotten to install the event filter: simply overriding the method is not enough.

Yes, this was correct point :-/

> Anyway, I've pushed the version that works for me.

You have forgotten remove the focus property from "atmosphere" button.

@gzotti @10110111 This feature may be added into 23.1 release if it OK for you.

gzotti (approved):
Works for me, thanks!

PR #3110, opened 2023-03-17
Changes for proper motion

### LINKED ISSUE #3108: Including conversion of spherical plane into star's propermotion calculations
### Discussed in https://github.com/Stellarium/stellarium/discussions/3101

<sup>Originally posted by **vreijs** March 14, 2023</sup>
Can someone tell me in which module the proper motion of a star is being included in de declination/right ascension?
I am not that used to github, so searching in the source-code has not been successful for me.
Thanks for the help.

Why I am asking is: Is the RA's proper motion from the star database changed with a factor 
1/cos(declination)? https://www.fxsolver.com/browse/formulas/Magnitude+of+proper+motion+%28%CE%BC%CE%B1%2A%29

I hope to find that in the source code.

Thanks for your help.

All the best,

Victor
</div>

pekartt:
### Description

Proposal for proper motion fix
Fixes #3108 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: <MacOS, 12.5.1>

## Checklist:

### DIFF
--- a/src/core/modules/Star.hpp
+++ b/src/core/modules/Star.hpp
@@ -24,6 +24,7 @@
 
 #include "ZoneData.hpp"
 #include "StelObjectType.hpp"
+#include "StarMgr.hpp"
 #include <QString>
 #include <QtEndian>
 
@@ -110,9 +111,29 @@ public:
 	void getJ2000Pos(const ZoneData *z,float movementFactor, Vec3f& pos) const
 	{
 		pos = z->axis0;
-		pos*=(static_cast<float>(getX0())+movementFactor*getDx0());
-		pos+=(static_cast<float>(getX1())+movementFactor*getDx1())*z->axis1;
+		pos*=static_cast<float>(getX0());
+		pos+=(static_cast<float>(getX1()))*z->axis1;
 		pos+=z->center;
+		PMData pm=StarMgr::getProperMotion(getHip());
+		if((pm.first!=NAN) && (pm.second!=NAN))
+		{
+			float dra=pm.first*movementFactor*10.0;
+			float ddec=pm.second*movementFactor*10.0;
+			float r=pos.norm();
+			float sd=pos[2]/r;
+			float dec=std::asin(sd);
+			float phi=pos.longitude();
+			float cd=std::cos(dec);
+			float cp=std::cos(phi);
+			float sp=std::sin(phi);
+			float a1=ddec*sd;
+			float a2=r*a1;
+			float a3=r*dra;
+			float dx=-a2*cp-a3*sp;
+			float dy=a3*cp-a2*sp;
+			float dz=r*ddec*cd;
+			pos+=Vec3f(dx,dy,dz);
+		}
 	}
 	inline int getBVIndex() const {return d.uint8[12];}
 	inline int getMag() const {return d.uint8[13];}
--- a/src/core/modules/StarWrapper.hpp
+++ b/src/core/modules/StarWrapper.hpp
@@ -79,7 +79,7 @@ protected:
 	{
 		static const double d2000 = 2451545.0;
 		Vec3f v;
-		s->getJ2000Pos(z, (M_PI/180.)*(0.0001/3600.) * ((core->getJDE()-d2000)/365.25) / a->star_position_scale, v);
+		s->getJ2000Pos(z, (M_PI/180.)*(0.0001/3600.) * ((core->getJDE()-d2000)/365.25),v);
 
 		// Aberration: Explanatory Supplement 2013, (7.38). We must get the observer planet speed vector in Equatorial J2000 coordinates.
 		if (core->getUseAberration())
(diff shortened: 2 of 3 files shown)


gzotti:
Thanks, but: 
Try to see the frame rate drop. With many stars: Absolute Scale: 2.0, Relative scale: 0.25. With a particular setup I have:
Master: 20fps
This: 16fps

If it was that easy...

The whole file hip_pm.dat is a temporary kludge. It should be possible to restore proper motion elements from the star data by reversing the triangle vector math. Then this file can go away again. 
The solution is most likely to be found in fixing something in util/ParseHip.C and creating a new star catalog.

Atque:
> Thanks, but: Try to see the frame rate drop. With many stars: Absolute Scale: 2.0, Relative scale: 0.25. With a particular setup I have: Master: 20fps This: 16fps
> 
> If it was that easy...
> 
> The whole file hip_pm.dat is a temporary kludge. It should be possible to restore proper motion elements from the star data by reversing the triangle vector math. Then this file can go away again. The solution is most likely to be found in fixing something in util/ParseHip.C and creating a new star catalog.

Is a frame drop of 4 FPS really anything to worry about? At a first glance, the proper motion mess seems to be solved by this PR.

10110111:
> Is a frame drop of 4 FPS really anything to worry about?

Of course it is when it's 20% of a frame rate that was already far from ideal frame rate for an animation!

Atque:
> Of course it is when it's 20% of a frame rate that was already far from ideal frame rate for an animation!

Fair enough, but the framerate is only noticeably lower when using extreme settings, such as 0.25 scale for stars. We all know Stellarium already has framerate problems when showing too many grids or DSO's.

gzotti:
This slowdown applies to all users, also those for which the current flaw around proper motion is almost irrelevant. (I mean those living in the present times looking into the present-time sky.) We have discussed this for days now (#3101, #3108; also check links or related projects linked therein).

Yes, I also want proper motion done properly for simulation in prehistory. I did most of these fine corrections and new models for the planets, and now we can simulate e.g. Solar eclipses to the second. Yes, a handful of stars are a few arcminutes off in the stone age. The problem has existed since 2006 or so. It is well documented. Show me the essential use case or research case where this currently matters, with bleeding-edge urgency. The urgent case can still be researched tonight by setting a Marker to the correct coordinates of &alpha; Cen or &alpha; Boo as found in CdC. 

I am very sure that the problem _should and can be solved elsewhere_ while _keeping efficiency_. Just do it at the right spot, not with kludges added to add-on kludges. 

First, the extra text file delivering proper motion data has been added as openly declared temporary kludge because the star catalog seems to be hard to decipher. Just to fill somebody's request for PM data. However, PM data already exists in the packed star data. To be consistent, graphics and number output should come from the same data. 

Someone, analyze the data format and geometry model of the star catalog. It's just vector math. If PM has been packed in, it can likely be unpacked. And if you feel like it, while you are "doing a new star catalog", use Gaia instead of NOMAD, or at least add the umpteen "missing stars" reported in various issues. In case the star catalog format really has a principal technical flaw that cannot be fixed, think about a new format that can then also fix the binary-flying-apart problem. Even if that takes a year of your spare time.

Atque:
It's not just a handful of stars. Compare this PR with the master at 10,000 BCE, and look toward north (Polaris in center). Many stars show quite different positions, some constellations look very strange in the master compared to this PR (Cassiopeia being probably the most noticeable, but also the Big Dipper looks a bit different). I'm not even sure there really is a difference. I use absolute scale = 2 and relative scale = 0.25, and see a difference of at most tenths of a frame per second. Is that _really_ a problem?

Frankly, I don't understand how planetary positions can be more important in the stone age than constellation shapes. Shouldn't users be just as interested in knowing where Arcturus was placed in the Old Kingdom of Egypt, as they would be of the position of Jupiter in the same era? If positional accuracy is needed, I don't see how Stellarium could be called 1.0 if PM issues are not solved.

pekartt:
I'm new to Stellarium development and GitHub. I think the calculations can be done when loading the Hipparchus stars and the proper motion data. This would improve frame rate.

gzotti:
I may repeat myself, but: 

Have you tried to understand the vector math around the original star catalog? PM is encoded with the programs existing in the util subdirectory. 

From the Star1 and Star2 classes, 

`J2000pos = z.center + z.axis0*(x0+movementFactor(t)*dx0) + z.axis1*(x1+movementFactor(t)*dx1)`.

It appears to me that a 2D proper motion arrow given "on the sphere" as d&alpha;/d&delta; in the original catalog can be modelled with this simple and fast vector approach with position plus correction in 2 axes that does not require any sin/cos. Maybe something like the cosine correction was overlooked in the encoding of `dx0 `and `dx1`, though. This can most likely be corrected in the catalog creation program in the util subdirectory. There may be other causes for problems, though. Currently I cannot go deeper into the ZoneArray::initTriangle() where star_position_scale is derived which is then used for the movementFactor(t).

Have you tried to understand the vector math around the original star catalog? PM is encoded with the programs existing in the util subdirectory. I have not studied this code yet. With some luck, the star_position_scale there is compatible to that used in ZoneArray. 

If there are no principal fundamental mathematical flaws in the star catalog designed in 2006, the **final solution** to both the PM problem and the PM display **should work without the extra PM file** but uses data from the catalog. As first step, a PM extraction code should be added to the Star classes to extract PM data for the purpose of comparative display in the InfoString. If some cos() was missed when deriving `dx0` and `dx1`, this can be added to the tools in util and the catalog recreated. It is better to run a single acos() if (not when) displaying Infostring for one star than running 2 million cos() per frame.  When we properly understand the catalog format, we can debug it. This need not be finished next week, but it can probably be solved over the summer.

PR #3111, opened 2023-03-18
Disable "dangerous" scripting calls unless --unsafe is passed

randomstuff:
### Description

Disable "dangerous" scripting API usage by default. It can be enable with --unsafe.

This mitigates CVE-2023-28371 from both file association (opening a .ssc file from a GUI) and commandline usage.

This could break some commandline usage. The --unsafe flag must be added.

### Type of change

### How Has This Been Tested?

Tested by triggering some unsafe scripting API with and without the flag.

**Test Configuration**:
* Operating system: Debian testing
* Graphics Card: not relevant

## Checklist:

WIP

- [X] I have made corresponding changes to the documentation (header file)
- [X] I have updated the respective chapter in the Stellarium User Guide

### DIFF
--- a/src/scripting/StelMainScriptAPI.cpp
+++ b/src/scripting/StelMainScriptAPI.cpp
@@ -342,8 +342,10 @@ QStringList StelMainScriptAPI::getAllTimezoneNames()
 
 void StelMainScriptAPI::screenshot(const QString& prefix, bool invert, const QString& dir, const bool overwrite, const QString &format)
 {
+	const bool allowUnsafe = qApp->property("allow_unsafe").toBool();
+
 	QString realDir("");
-	if ((!dir.isEmpty()) && (!StelApp::getInstance().getScriptMgr().getFlagAllowExternalScreenshotDir()))
+	if (!allowUnsafe && (!dir.isEmpty()) && (!StelApp::getInstance().getScriptMgr().getFlagAllowExternalScreenshotDir()))
 	{
 		qWarning() << "SCRIPT CONFIGURATION ISSUE: the script wants to store a screenshot" << prefix << "." << format << "to an external directory " << dir;
 		qWarning() << "  To enable this, check the settings in the script console";
@@ -354,7 +356,7 @@ void StelMainScriptAPI::screenshot(const QString& prefix, bool invert, const QSt
 
 
 	const bool oldInvertSetting = StelMainView::getInstance().getFlagInvertScreenShotColors();
-	const QString oldFormat=StelMainView::getInstance().getScreenshotFormat();
+	QString oldFormat=StelMainView::getInstance().getScreenshotFormat();
 	if ((format.length()>0) && (format.length()<=4))
 		StelMainView::getInstance().setScreenshotFormat(format);
 	// Check requested against set image format.
--- a/cmake/stellarium.pod.cmake
+++ b/cmake/stellarium.pod.cmake
@@ -94,6 +94,10 @@ latitudes should be prefixed with C<->.
 
 Print a list of landscape names and exit.
 
+=item B<--unsafe>
+
+Allow unsafe script execution.
+
 =item B<--landscape> I<name>
 
 Start Stellarium using landscape I<name>.
--- a/guide/app_config_ini.tex
+++ b/guide/app_config_ini.tex
@@ -736,10 +736,8 @@ password   & string & Password for proxy. E.g. \emph{xxxxx}\\\bottomrule
 \begin{tabularx}{\textwidth}{l|l|l|X}\toprule
 \emph{ID}                  & \emph{Type} & \emph{Default} & \emph{Description}\\\midrule
 startup\_script                            & string & startup.ssc & name of script executed on program start\\
-flag\_script\_allow\_absolute\_path        & bool   & false       & set true to allow running scripts from absolute pathnames.
-                                                                    This may pose a security risk if you run arbitrary scripts.\\
-flag\_allow\_screenshots\_dir              & bool   & false       & Allow scripts to set a target directory.\\
-flag\_allow\_write\_absolute\_path         & bool   & false       & set true to let scripts store output files to absolute pathnames.
+flag\_allow\_screenshot\_dir               & bool   & false       & Allow scripts to set a target directory.\\
+flag\_script\_allow\_write\_absolute\_path & bool   & false       & set true to let scripts store files to absolute pathnames.
                                                                     This may pose a security risk if you run scripts from other authors
                                                                     without checking what they are doing.\\\bottomrule 
 %% These are unused as on 0.15pre.
--- a/guide/ch_advanced_use.tex
+++ b/guide/ch_advanced_use.tex
@@ -489,6 +489,7 @@ For example, using the option \texttt{-c\ my\_config.ini} would resolve to the f
 -\/-sky-date         & date       & The initial date in \texttt{yyyymmdd} format. \\
 -\/-sky-time         & time       & The initial time in \texttt{hh:mm:ss} format. \\\midrule
 -\/-startup-script   & script name & The name of a script to run after the program has started. [\file{startup.ssc}] \\\midrule
+-\/-unsafe           & path        & Allow unsafe script execution \\\midrule
 -\/-fov              & angle (degrees) & The initial vertical field of view in degrees. \\\midrule
 -\/-scale-gui        & scale factor & Scaling the GUI according to scale factor\\
 -\/-gui-css style    & name       & Use \file{name.css} to define GUI style\\\midrule
--- a/src/CLIProcessor.cpp
+++ b/src/CLIProcessor.cpp
@@ -130,6 +130,7 @@ void CLIProcessor::parseCLIArgsPreConfig(const QStringList& argList)
 		          << "--restore-defaults      : Delete existing config.ini and use defaults\n"
 		          << "--multires-image        : With filename / URL argument, specify a\n"
 			  << "                          multi-resolution image to load\n"
+			  << "--unsafe                : Allow unsafe script execution\n"
 #ifdef Q_OS_WIN
 			  << "--angle-mode (or -a)    : Use ANGLE as OpenGL ES2 rendering engine (autodetect driver)\n"
 			  << "--angle-d3d9 (or -9)    : Force use Direct3D 9 for ANGLE OpenGL ES2 rendering engine\n"
@@ -176,7 +177,7 @@ void CLIProcessor::parseCLIArgsPostConfig(const QStringList& argList, QSettings*
 {	
 	// Over-ride config file options with command line options
 	// We should catch exceptions from argsGetOptionWithArg...
-	int fullScreen, altitude;
+	int fullScreen, altitude, allowUnsafe;
 	float fov;
 	QString landscapeId, homePlanet, longitude, latitude, skyDate, skyTime;
 	QString projectionType, screenshotDir, multiresImage, startupScript, customCSS;
@@ -195,6 +196,7 @@ void CLIProcessor::parseCLIArgsPostConfig(const QStringList& argList, QSettings*
 		latitude = argsGetOptionWithArg(argList, "", "--latitude", "").toString();
 		skyDate = argsGetOptionWithArg(argList, "", "--sky-date", "").toString();
 		skyTime = argsGetOptionWithArg(argList, "", "--sky-time", "").toString();
+		allowUnsafe = argsGetOption(argList, "", "--unsafe");
 		fov = argsGetOptionWithArg(argList, "", "--fov", -1.f).toFloat();
 		projectionType = argsGetOptionWithArg(argList, "", "--projection-type", "").toString();
 		screenshotDir = argsGetOptionWithArg(argList, "", "--screenshot-dir", "").toString();
@@ -213,6 +215,8 @@ void CLIProcessor::parseCLIArgsPostConfig(const QStringList& argList, QSettings*
 		qCritical() << "ERROR while checking command line options: " << e.what();
 		exit(0);
 	}
+
+	qApp->setProperty("allow_unsafe", allowUnsafe);
 
 	// Will be -1 if option is not found, in which case we don't change anything.
 	if (fullScreen==1)
--- a/src/scripting/StelScriptOutput.cpp
+++ b/src/scripting/StelScriptOutput.cpp
@@ -60,7 +60,10 @@ void StelScriptOutput::saveOutputAs(const QString &name)
 	const QDir dir=outputInfo.dir(); // will hold complete dirname
 	const QFileInfo newFileNameInfo(name);
 
-	const bool okToSaveToAbsolutePath=StelApp::getInstance().getSettings()->value("scripts/flag_allow_write_absolute_path", false).toBool();
+	const bool allowUnsafe = qApp->property("allow_unsafe").toBool();
+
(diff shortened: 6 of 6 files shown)


data/splash.png
gzotti, line 1: why bother with the splash screen? Color is for releases, b/w for betas.

data/stellarium.ico
gzotti, line 1: why is this deleted?

src/scripting/StelMainScriptAPI.cpp
```diff
@@ -342,6 +342,11 @@ QStringList StelMainScriptAPI::getAllTimezoneNames()
 
 void StelMainScriptAPI::screenshot(const QString& prefix, bool invert, const QString& dir, const bool overwrite, const QString &format)
 {
+	if (!qApp->property("allow_unsafe").toBool()) {
```
gzotti, line 345: Do I understand that this inhibits any screenshots requested by scripts unless the program is started with --unsafe? This means no screenshots (and complaints to us) for the average user, and permanently including --unsafe in the STEL_OPTS envvar (or Windows start link) for advanced users.

data/stellarium.ico
randomstuff: I really don't know how this file and spalsh.png were modified. Not intentional. I'm removing these changes.

src/scripting/StelMainScriptAPI.cpp
randomstuff: > This means no screenshots (and complaints to us) for the average user

This does not block screenshot using for example Control+S. This part still works.

This only blocks screenshots through the scripting API. I guess the average user does not use this (?).

> Note that if dir is empty (as is common), storing goes to the preconfigured screenshot directory. IMO when dir=="" there should not be any danger.

This this should be improved by allowing safe calls to screenshot().

10110111: > I guess the average user does not use this (?).

Periodic screenshots is a way to make a time lapse, and scripting seems like an ideal means to make such screenshots. Does "average user" not want time lapses?

randomstuff:
I have added a commit which allows some forms of core.sreenshot() without the --unsafe flag.

src/scripting/StelMainScriptAPI.cpp
randomstuff: > IMO when dir=="" there should not be any danger.

Event in this case, it would be possible to writ to other directories through path traversal, for example using prefix=".././../foo".

gzotti: OK. This means we must add a test for "simple filename without devious traversal" for prefix. 
... done in 9ecd967f0a103b34f217c97c095ff1c6dfae8fb8

Other than that, running scripts which just store into the default screenshot directory should not pose a problem.

I can consider this as start of the method:

```
	if (dir.length()>0 && !qApp->property("allow_unsafe").toBool()) {
		qWarning() << "SCRIPTING CONFIGURATION ISSUE: You are trying to save a screenshot to a separate directory.";
				qWarning() << "  To enable this, pass the --unsafe flag";
				return;
	}
```

I still don't know what a QImage stored to some arbitrary path can break. Yes, you can overwrite an existing other file named like a supported image type, if you are entitled to such writing operation in that directory. Scenario example?

src/scripting/StelMainScriptAPI.cpp
```diff
@@ -342,8 +342,29 @@ QStringList StelMainScriptAPI::getAllTimezoneNames()
...
+		&& !prefix.isEmpty()
+		&& !format.isEmpty()
+		// Protection against path traversal:
+		&& !prefix.contains("/")  && !format.contains("/")
+		// Protection against path traversal (Windows):
+		&& !format.contains("\\") && !prefix.contains("\\")
+		// Protect against using alternate data streams (Windows)?
+		&& !prefix.contains(":") && !format.contains(":")
+		// Make sure we don't overwrite config.ini and such:
+		&& format != "ini");
```
gzotti, line 359: only image formats supported by the current build can be stored. .ini is not a supported image type.

src/scripting/StelMainScriptAPI.cpp
```diff
@@ -342,8 +342,29 @@ QStringList StelMainScriptAPI::getAllTimezoneNames()
 
 void StelMainScriptAPI::screenshot(const QString& prefix, bool invert, const QString& dir, const bool overwrite, const QString &format)
 {
+	const bool allowUnsafe = qApp->property("allow_unsafe").toBool();
+	const bool allowed = allowUnsafe || (
+		// Only allow writing in the default directory:
```
gzotti, line 347: I have just added path stripping to the `StelMainView::saveScreenShot()`. Only a "naked" filename remains. 

It should be enough to ensure that prefix is not empty (but how risky is a ".tif" file? Oh well, yes, it's hidden on Linux. Just a nameless but visible TIF on Windows.)  and dir is empty.

src/scripting/StelScriptMgr.cpp
```diff
@@ -794,10 +794,8 @@ bool StelScriptMgr::runPreprocessedScript(const QString &preprocessedScript, con
 bool StelScriptMgr::runScript(const QString& fileName, const QString& includePath)
 {
 	QString preprocessedScript;
-	if (prepareScript(preprocessedScript,fileName,includePath))
-		return runPreprocessedScript(preprocessedScript,fileName);
-	else
-		return false;
+	prepareScript(preprocessedScript,fileName,includePath);
```
gzotti, line 797: prepareScript() returns false if preprocessing fails. In this case, the preprocessedScript should not be run.

src/StelMainView.cpp
```diff
@@ -1618,6 +1618,8 @@ void StelMainView::setScreenshotDpi(int dpi)
 void StelMainView::saveScreenShot(const QString& filePrefix, const QString& saveDir, const bool overwrite)
 {
 	screenShotPrefix = QFileInfo(filePrefix).fileName(); // Strip away any path elements (Security issue!)
+	if (screenShotPrefix.isEmpty())
+		screenShotPrefix = "screenshot";
```
gzotti, line 1622: The default name is "stellarium-".

src/scripting/StelMainScriptAPI.cpp
```diff
@@ -342,8 +342,14 @@ QStringList StelMainScriptAPI::getAllTimezoneNames()
 
 void StelMainScriptAPI::screenshot(const QString& prefix, bool invert, const QString& dir, const bool overwrite, const QString &format)
 {
+	const bool allowUnsafe = qApp->property("allow_unsafe").toBool();
+	if (!allowUnsafe && dir.isEmpty()) {
+		qWarning() << "SCRIPTING CONFIGURATION ISSUE: You are trying to save a screenshot in some dangerous path";
```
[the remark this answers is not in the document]
randomstuff: Yes, the text was not really intended to be the final version. Do you intend to say the wording should be changed (maybe "suspect" ?) or that this check is not super useful? The main danger this way is for an attacker to achieve to overwrite some of your important images (which would not be easily achievable because the attacker would have to guess the filename or existing files) or fill the disk (which is absolutely not mitigated).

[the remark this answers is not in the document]
gzotti: Overwriting existing files is indeed intended when setting overwrite=true. The script this mechanism for external directories and overwriting existing files was created for is the skybox script which can be called repeatedly to overwrite skybox images which are then auto-loaded in a running (serious) game (simulation). If you really feel this is a catastrophic risk for the user's data, we can add yet another configuration option even visible in the GUI to "allow overwriting existing images when taking screenshots". This will need a longish tooltip explaining that this applies only to scripts that actually request overwriting. Manual screenshots never overwrite. This GUI checkbox should therefore best go into the scripting GUI. 

"Fill the disk" is the only relevant danger I can now identify that may affect usability of a system.  This applies to every program that can create files automatically. How does e.g. a movie renderer like Blender or movie cutting programs that can dissect a movie into single frames handle decreasing disk space? Do they output images until disk is full and get stuck with the next frame they can no longer produce? Or do they have a setting "emergency stop when disk has only ...MB free"? We could add such setting as well, and check free space before writing a screenshot.

10110111:
> Or do they have a setting "emergency stop when disk has only ...MB free"?

Sounds like overengineering for casual programs. So far I've only seen such behavior in installers. All others just write and let the errors happen (and then most report them gracefully).

src/scripting/StelMainScriptAPI.cpp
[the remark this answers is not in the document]
10110111: > Or do they have a setting "emergency stop when disk has only ...MB free"?

Sounds like overengineering for casual programs. So far I've only seen such behavior in installers. All others just write and let the errors happen (and then most report them gracefully).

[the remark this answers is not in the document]
randomstuff: > I would not call just any external and dedicated target directory "dangerous".

Actually, the idea of this change was to prevent the script from writing outside of dedicated directories.

> Overwriting existing files is indeed intended when setting overwrite=true. The script this mechanism for external directories and overwriting existing files was created for is the skybox script which can be called repeatedly to overwrite skybox images which are then auto-loaded in a running (serious) game (simulation). If you really feel this is a catastrophic risk for the user's data, we can add yet another configuration option even visible in the GUI to "allow overwriting existing images when taking screenshots".

I am afraid my previous comment was not clear at all :) No, I do **not** believe overwriting a file is problematic. What I am not super confortable with is to be able to overwrite images to arbitrary locations (unrelated directories) because it might be possible to overwrite existing user files but this is quite far-fetched.

randomstuff:
Alternative proposal regarding how the "dir" parameter of `core.screenshot()` is handled.

src/scripting/StelMainScriptAPI.cpp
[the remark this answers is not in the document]
gzotti: far-fetched indeed. Is this not a fully constructed case? you must guess what filename exists (or maybe check for existing PNG in that directory, select a random file and overwrite it? Can we just assume users check scripts when they have downloaded them from external creators? 
I can add another tickbox in the script console. However, everybody will activate it and forget about the issue.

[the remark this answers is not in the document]
gzotti: ...done in 5d15479fdb73ad975ce23b71539d9b90531d3736

gzotti:
I have now added the suggestion to avoid empty filename prefix (with proper default name prefix). Also, storing screenshots is rejected if less than 50MB are free. It is still possible to fill the disk with output text files which usually are measured in kB. Do we have to fully protect the user from him/herself?

I would like to close this today or tomorrow, in time for 23.1.

10110111:
> Also, storing screenshots is rejected if less than 50MB are free.

Why so restrictive? A typical screenshot has size of 1-2 MiB! Don't do this in any form, it's not a normal behavior for casual software. Have you ever seen GIMP or Inkscape discriminating you on the basis of storage space available?

gzotti:
In times of multi-TB disks IMHO 50MB free space can be regarded "critically full" in any case. A TIFF screenshot may be large, tens of MB. The limit is arbitrary, sure, but It may be good not to be the last program to fill the disk. I did not test the situation, though, so don't know what happens if we really fill up storage.

10110111:
> The limit is arbitrary, sure, but It may be good not to be the last program to fill the disk.

I'd rather have the program just do its job, not impose arbitrary limits on me. It's none of its business to know how much space is left. Each time I found a program that thought it's smarter than its user, it happened to get in my way rather than help.

Besides, at least ext3 and ext4 on Linux by default reserve some percentage of the storage to `root`, so while a normal user might see no space left (and get an error trying to write into a file), the system will still have some wiggle room.

gzotti:
I reverted the emergency leftover. Seriously, I don't care. A disk with <1GB is full for me.

src/scripting/StelMainScriptAPI.cpp
randomstuff: > fill the disk (which is absolutely not mitigated).

For reference, I mentionned this for exhaustivity, not because I felt it was necessary to have some mitigation for this.

randomstuff:
Rebased the PR. For reference, this PR aims to:

1. remove `flag_script_allow_absolute_path` in order to fix file association (which is mostly broken unless you set flag_script_allow_absolute_path = true) under the assumption that the other mitigations are enough.
2. allow a transient `--unsafe` CLI argument which can be used instead of the (persistent) configurations. The idea is that it is safer to use a transient option than a persistent configuration which he user will never switch back to the safer mode.

data/splash.png
gzotti, line 1: deletion is back again (?)

data/stellarium.ico
gzotti, line 1: (please undelete as well)

gzotti:
I am at least not convinced about the CLI option. This only applies to those Linux users (or maybe Macs) who commonly start GUI programs by typing their name into a console window. The typical Windows user clicks on the desktop icon, and does not open a CMD shell (yuck!) to type the full path to the program. If given instructions, user would modify the desktop icon's associated command and quickly forget about it. This is equivalent to setting the flag right in the program.  The important thing to take home around this is that the user is made aware of a potential chance of an issue when running malicious scripts from random locations, and to concede more liberties to scripts when required. A similar "dangerous" option is the CORS access in the RemoteControl plugin. If needed, you can. At your own risk.

If we can take away the absolute path option again, fine for me. But I could live with a situation where only SSC files in one of the two usual paths would be allowed. And for me, SSC is associated with an editor...

gzotti:
@10110111 , @alex-w any opinion on this?

10110111:
> If we can take away the absolute path option again, fine for me.

No, this would throw the baby out with bathwater. Choosing location from a script is a legitimate need (and I actually used this when making my time lapses). If you're going to restrict it, the option to turn it on should still be available.

randomstuff:
Another argument against my proposal is that it's currently possible to achieve a similar behavior by passing an alternate ini file : `stellarium -c ./stellarium-unsafe.ini ...`.

gzotti:
Yes. That's also a CLI option, and would likely start with other settings, but could also work to disable further tests. Again, in this case the user must actively do that, so we can assume "advanced level". 

@10110111 not sure if you are discussing the same option. Writing to other directories is also my aim. 
Double-clicking an SSC icon after first install of the program now should emit a warning and not run the script until user has given permission by editing config.ini. Removing the `flag_script_allow_absolute_path` option would probably allow running SSC files sitting anywhere in the system by following some configured file association, which is what @randomstuff has warned about some days ago. 
Is it reasonably safe and acceptable now to take away this test again?

gzotti:
> This [modifying the startup link on Windows] is equivalent to setting the flag right in the program.

It is even worse, as the GUI setting remains easier visible in the GUI than the once-edited startup link. 

So, in total, seriously, I would not like to add this CLI option when the in-program GUI options do the same with better visibility. By default, Stellarium installs "safe", and if needed, users can set the flag in the GUI. It is highly likely that they will run the same script again, so those who need it will need the setting permanently. But they will have read the warning and implications once, and are effectively responsible for their own systems. 

@randomstuff  @10110111 @alex-w agreed?

10110111:
> So, in total, seriously, I would not like to add this CLI option when the in-program GUI options do the same with better visibility.

I agree.

gzotti:
That means, can we close this? Or what remains? I thank @randomstuff for pointing out issues. But @alex-w surely wants to pack release 23.1 as soon as we have finished this discussion.

randomstuff:
Yes, we can close/reject this PR :+1:

PR #3125, opened 2023-03-25
Fix: bottom bar fontsize adaptations

gzotti:
This should fix a few issues which have become more apparent around the now default-visible font selection and font size switches. A preliminary fix was attempted in #3014.

GUI help text did not follow font and font size changes. 
A non-default font and font size may lead to a visible (1-pixel) peek of the sliding buttons. The bottom bar must be properly reconstructed whenever font/size change.

A few other little quirks as I find them.

### Description

Fixes #3014 (issue)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Geforce 

## Checklist:

### DIFF
--- a/src/gui/StelGuiItems.cpp
+++ b/src/gui/StelGuiItems.cpp
@@ -384,6 +384,7 @@ LeftStelBar::~LeftStelBar()
 
 void LeftStelBar::addButton(StelButton* button)
 {
+	prepareGeometryChange();
 	double posY = 0;
 	if (QGraphicsItem::childItems().size()!=0)
 	{
@@ -392,7 +393,7 @@ void LeftStelBar::addButton(StelButton* button)
 	}
 	button->setParentItem(this);
 	button->setFocusOnSky(false);
-	button->prepareGeometryChange(); // could possibly be removed when qt 4.6 become stable
+	//button->prepareGeometryChange(); // could possibly be removed when qt 4.6 become stable
 	button->setPos(0., qRound(posY+10.5));
 
 	connect(button, SIGNAL(hoverChanged(bool)), this, SLOT(buttonHoverChanged(bool)));
@@ -557,6 +558,7 @@ BottomStelBar::BottomStelBar(QGraphicsItem* parent,
 //! connect from StelApp to resize fonts on the fly.
 void BottomStelBar::setFontSizeFromApp(int size)
 {
+	prepareGeometryChange();
 	QFont font=QGuiApplication::font();
 	// Font size was developed based on base font size 13, i.e. 12
 	font.setPixelSize(size-1);
@@ -628,6 +630,7 @@ BottomStelBar::~BottomStelBar()
 
 void BottomStelBar::addButton(StelButton* button, const QString& groupName, const QString& beforeActionName)
 {
+	prepareGeometryChange();
 	QList<StelButton*>& g = buttonGroups[groupName].elems;
 	bool done = false;
 	for (int i=0; i<g.size(); ++i)
@@ -1264,7 +1267,7 @@ QRectF CornerButtons::boundingRect() const
 	if (QGraphicsItem::childItems().size()==0)
 		return QRectF();
 	const QRectF& r = childrenBoundingRect();
-	return QRectF(0, 0, r.width()-1, r.height()-1);
+	return QRectF(0, 0, r.width(), r.height());
 }
 
 void CornerButtons::setOpacity(double opacity)
--- a/src/gui/StelGuiItems.cpp
+++ b/src/gui/StelGuiItems.cpp
@@ -135,7 +135,7 @@ void StelButton::initCtor(const QPixmap& apixOn,
 	StelGui* gui = dynamic_cast<StelGui*>(StelApp::getInstance().getGui());
 	connect(gui, SIGNAL(flagUseButtonsBackgroundChanged(bool)), this, SLOT(updateIcon()));
 
-	if (action!=Q_NULLPTR)
+	if (action!=nullptr)
 	{
 		if (action->isCheckable())
 		{
@@ -148,7 +148,7 @@ void StelButton::initCtor(const QPixmap& apixOn,
 			QObject::connect(this, SIGNAL(triggered()), action, SLOT(trigger()));
 		}
 	}
-	if (secondAction!=Q_NULLPTR)
+	if (secondAction!=nullptr)
 	{
 			QObject::connect(this, SIGNAL(triggeredRight()), secondAction, SLOT(trigger()));
 	}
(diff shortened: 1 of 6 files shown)


10110111:
At the current stage the 1px sticking of the bottom panel buttons still remains. Nothing seems to have changed.

gzotti:
Of course not. Draft/working on it. But systematically. I try to understand the various components and how they work together. When I feel confident to have solved it, I will switch to "Ready for review".

gzotti:
I had to change quite a bit in the bottom bar to allow even larger fonts. Also, the rescaling has to be called twice to catch all boundingbox updates. Now you can try mousewheeling the font setting in the config dialog and should hopefully always see text&buttons well arranged.

What I don't know is why bounding boxes returned previously were 1 pixel narrower than the actual tested&combined geometries. I also changed this. 

I left a few dev parts commented out. These I will clean up before an eventual merge.

src/gui/StelGuiItems.cpp
```diff
@@ -1030,29 +1030,36 @@ void BottomStelBar::updateText(bool updatePos, bool updateTopocentric)
...
 		QRectF rectCh = getButtonsBoundingRect();
-		location->setPos(0, 0);		
-		int dtp = static_cast<int>(rectCh.right()-datetime->boundingRect().width())-5;
-		if ((dtp%2) == 1) dtp--; // make even pixel
-		datetime->setPos(dtp,0);
+		location->setPos(0, 0);
+		QFontMetrics locationMetrics(fov->font());
+		int locationWidth=locationMetrics.boundingRect(location->text()).width() + locationMetrics.boundingRect("MMM").width();
+
+		int dateTimePos = static_cast<int>(rectCh.right()-datetime->boundingRect().width())-5;
```
10110111, line 1065: Since you're replacing the numbers by semantically-loaded expressions, why is this -5 still here as a literal?

gzotti: I actually just *did not change* this buffer width of 5 pixels. Whoever introduced "5" likely did so for aesthetic reasons. It could probably be replaced by something like 'StelBarsPath::roundSize-1'. 
I just saw that CornerButtons::boundingRect() also returns a boundingRect that appears to be 1 pixel too small in my view. I don't understand why boundingRect sizes are reduced by 1. Insights welcome.

10110111: > I actually just _did not change_

Yes, this is what makes me wonder, because you did change some magic numbers above but didn't change this one.

gzotti: Because I did not see an aesthetic issue about the change in X coordinates. You see the purpose of this, "keep a small distance from the right edge". This still works. Keeping the constant may be faster than retrieving the 'StelBarsPath::roundSize-1' which is actually still just a named constant without in-program setting option.

10110111:
The FOV field is now too close to the FPS one, which makes it harder to parse. It used to be like
```
FOV 80.3°       29 FPS      date
```
But now it's more like
```
FOV 80.3° 30 FPS  date
```

The `80.3° 30` is hard to decypher compared to the way it used to be. Here're the screenshots: [old](https://user-images.githubusercontent.com/6376882/229346105-35748bf3-94cc-4442-88db-e950ef9aaa3a.png), [new](https://user-images.githubusercontent.com/6376882/229346116-6bc58935-037b-4952-88bb-24d35831537d.png).

gzotti:
Ah yes, this used to be like that before, just with too narrow field width. The jumping is indeed bad.

alex-w:
At least 2 digits after comma need for decimal FOV. 

Important note: the format of FOV may be changed via `config.ini` file (_gui/flag_fov_dms_) - I hope you're checked it too (includes changes in date & time - _gui/flag_time_jd_)

10110111:
The jumps are still present, though they are a bit smaller.

gzotti:
May depend on your font settings. Can you please experiment: StelGuiItems.cpp, LL 1070-72. 
The "MMM" are placeholders. Maybe we need more. However, at some point it is too much to be pretty with only few buttons.

10110111:
OK, my experiments say that "3" is wider than "1", so when I have 34.2 FPS real-life value, it appears wider than the 12.3 placeholder, so this dominates the `max()`, while 34 FPS is narrower. This results in the jumps. I suggest that `W` or `M` be used instead of the digits, as these seem to be the widest characters.

gzotti:
> OK, my experiments say that "3" is wider than "1", so when I have 34.2 FPS real-life value, it appears wider than the 12.3 placeholder, so this dominates the `max()`, while 34 FPS is narrower. This results in the jumps. I suggest that `W` or `M` be used instead of the digits, as these seem to be the widest characters in all fonts (recall the "m dash").

Ah yes, the 12.3 in the placeholder 'loses' against your 34.2.  Will switch to 'M's (and yes, those are meant as m-dash...)

gzotti:
OK, this seems to work. 

For a few weeks now I have issues in Full-screen (W11, Geforce). First, I see no tooltips. Second, when I am trying font sizes, every few seconds I experience a brief flicker where the image turns black, then there is one frame with different fontsize, and the next frame has those font sizes which I had set. The flicker is not triggered by the font size change. In windowed mode, I have tooltips and no flicker. This is not related to this PR, 23.1 has the same problem. I don't use scaling. Do you have any idea?

10110111:
Does it happen with legacy atmosphere?

gzotti:
both. We had the no-tooltips problem some years ago when I had an M580. I think this was a Geforce driver vs. Qt bug that went away in M960 times.

10110111:
Well, this symptom is similar to something we had on Intel/Windows, and it was worked around by the `--single-buffer` option.

gzotti:
Indeed! That fixes both issues. :-)

alex-w (approved):

10110111 (approved):

gzotti:
@alex-w Do you know why the boundingrects for some GUI items are set on purpose to be 1 pixel smaller than their childrenRects? See e.g. `BottomStelBar::boundingRect() const`. I have fixed that in some places, but not in all (like `CornerButtons::boundingRect() const`)

alex-w:
> @alex-w Do you know why the boundingrects for some GUI items are set on purpose to be 1 pixel smaller than their childrenRects? See e.g. `BottomStelBar::boundingRect() const`. I have fixed that in some places, but not in all (like `CornerButtons::boundingRect() const`)

Sorry, no. But maybe @xalioth can explains why…

10110111:
Why have you closed this?

alex-w:
> Why have you closed this?

Damn!!! This is my mistake!

gzotti:
Must have been an error. I hope @xalioth can answer the question: Why do the StelGUI elements have boundingrects which are 1 pixel smaller than their contents? This must have been something important concluded in 2006...2008.

Issue #3067, opened 2023-02-21
Daytime Moon is too visible


Atque:
In reality, the daytime sky is so bright that it renders the Moon difficult, almost impossible, to see. At gibbous phase, it might be visible if high enough in the sky, but when the gibbous Moon has just risen, and the Sun is decently high up, the Moon will be incredibly difficult to spot. In Stellarium, however, the Moon is about as visible at all times. Furthermore, the terminator ought to be smoother in daylight, see the photo below:

![0_Close-Up-Of-Moon-In-Clear-Blue-Sky](https://user-images.githubusercontent.com/37028368/220376288-98bec08f-49fb-47f4-a8af-673f30b98df7.jpg)

gzotti:
I agree on your screenshot it appears too bright. I had to fight this burnout several times in the last years, but recent changes re-introduced the issue. It's a real pain! We have an unfinished discussion in #3041. 
The additive nature of our rendering will always show the dark edge of the moon also in daylight sky. I see it as observing hint. 

I have no trouble seeing a gibbous moon in daylight, maybe it's a question of air quality? 
Seeing your image, maybe the moon should be allowed to become as ruddy as @10110111 wants it *for daytime skies*, but have a more yellowish tint when it is dark.

Atque:
> I have no trouble seeing a gibbous moon in daylight, maybe it's a question of air quality?

High in the sky, yes, but not low in the sky. I lived by the sea some years ago, and I had troubles seeing a gibbous Moon about 5 degrees above the horizon when the Sun was high. It looked clear.


> Seeing your image, maybe the moon should be allowed to become as ruddy as @10110111 wants it _for daytime skies_, but have a more yellowish tint when it is dark.

That might be a good compromise. But mostly, make the Moon less visible (to the very least not burnt out).

gzotti:
Ah, sure, in 5° and lower there will be considerable absorption. So whatever rebalancing will have to be made now also must take Lunar magnitude and altitude-dependent absorption into account.

10110111:
There are multiple issues at play here:

[...]

PR #3156, opened 2023-04-02
Use a linear color space when rendering

10110111:
Historically, Stellarium did all the rendering in nonlinear sRGB color space. This means that all the operations like brightening and blending were done not quite correctly, sometimes leading to serious overshooting of colors, like in bug #3067.

This PR is an attempt to switch the rendering to linear-sRGB color space, i.e. the color space where the primaries and the white point are set to sRGB, but the nonlinear transfer function AKA gamma has not been applied. To achieve this, the following steps are done:

1. Have a separate high-color-depth FBO for linear-color scene rendering
2. When the linear-color scene is ready, blit this FBO to the target framebuffer, applying the sRGB transfer function and dithering
3. To avoid the need for explicit conversion of texture data from sRGB to linear-sRGB inside the shaders, the internal format of the textures is set to `GL_SRGB8` or `GL_SRGB8_ALPHA8`, which makes the hardware perform the conversion when sampling (and additionally, do minification filtering in linear colors).

There's a **problem** though: in my tests on Intel UHD Graphics 620 with Mesa 21.2.6 in OpenGL ES 2 mode the `GL_SRGB8*` formats appeared to be not color renderable, which precludes the use of `glGenerateMipmap()` for them. This in particular makes anisotropic filtering impossible even if we generate the mipmaps manually. The particular objects affected are the Moon, Jupiter, bottom of the old-style landscape (dunno why the sides are not): they become black in the current state of this branch.

The question is whether it's actually a problem in real-life OpenGL ES use cases. Does this problem happen on Windows/ANGLE? What about the small Linux systems where OpenGL ES is the only 3D API? I don't have any hardware to test this, so please try on yours.

### DIFF
--- a/data/shaders/planet.frag
+++ b/data/shaders/planet.frag
@@ -422,6 +422,7 @@ void main()
 		// FIXME: this should be calculated properly in linear space as
 		// extinction of sunlight, and with subsequent tone mapping.
 		// Current implementation is a legacy from older times.
+		// shadowColor is passed as a GL_RGBA texture, so the sample is sRGB-encoded
 		lowp vec4 color = vec4(linearToSRGB(finalColor.rgb), finalColor.a);
 		lowp float alpha = clamp(shadowColor.a, 0.0, 0.7); // clamp alpha to allow some maria detail
         finalColor = eclipsePush * (1.0-0.75*shadowColor.a) * mix(color, shadowColor, alpha);
--- a/data/shaders/planet.frag
+++ b/data/shaders/planet.frag
@@ -174,7 +174,7 @@ void main()
 	{
 		// We are drawing the Sun
 		vec4 texColor = texture2D(tex, texc);
-		texColor.rgb = srgbToLinear(texColor.rgb * sunInfo.rgb);
+		texColor.rgb = texColor.rgb * srgbToLinear(sunInfo.rgb);
 		// Reference: chapter 14.7 "Limb Darkening" in "Allen’s Astrophysical Quantities",
 		//            A.N.Cox (ed.), 4th edition, New York: Springer-Verlag, 2002.
 		//            DOI 10.1007/978-1-4612-1186-0
@@ -189,7 +189,7 @@ void main()
 		float cosTheta2 = cosTheta*cosTheta;
 		vec3 limbDarkeningCoef = a0 + a1*cosTheta + a2*cosTheta2;
 		vec3 color = texColor.rgb * limbDarkeningCoef;
-		FRAG_COLOR = vec4(linearToSRGB(color), texColor.a);
+		FRAG_COLOR = vec4(color, texColor.a);
 		return;
 	}
 #endif
@@ -398,8 +398,6 @@ void main()
         texColor = texture2D(tex, texc);
     }
 
-    texColor.rgb = srgbToLinear(texColor.rgb);
-
     mediump vec4 finalColor = texColor;
 	// apply (currently only Martian) pole caps. texc.t=0 at south pole, 1 at north pole. 
 	if (texc.t>poleLat.x-0.01+0.001*sin(texc.s*18.*M_PI)) {	// North pole near t=1
@@ -434,7 +432,5 @@ void main()
     //apply white rimlight
     finalColor.xyz = clamp( finalColor.xyz + vec3(outgas), 0.0, 1.0);
 
-    FRAG_COLOR = vec4(linearToSRGB(finalColor.rgb), finalColor.a);
-    //to debug texture issues, uncomment and reload shader
-    //FRAG_COLOR = texColor;
+    FRAG_COLOR = finalColor;
 }
--- a/data/shaders/xyYToRGB.glsl
+++ b/data/shaders/xyYToRGB.glsl
@@ -45,10 +45,10 @@ vec3 xyYToRGB(highp float x, highp float y, highp float Y)
 		if (flagUseTmGamma)
 		{
 			Y = pow(abs(Y*pi*1e-4), alphaWaOverAlphaDa*oneOverGamma)* term2TimesOneOverMaxdLpOneOverGamma;
-			return vec3(0.787077, 0.9898434, 1.9256125) * Y * brightnessScale;
+			return srgbToLinear(vec3(0.787077, 0.9898434, 1.9256125) * Y * brightnessScale);
 		}else{
 			Y = pow(abs(Y*pi*1e-4), alphaWaOverAlphaDa) * term2TimesOneOverMaxdL;
-			return pow(vec3(0.787077, 0.9898434, 1.9256125) * abs(Y * brightnessScale), vec3(oneOverGamma));
+			return vec3(0.787077, 0.9898434, 1.9256125) * abs(Y * brightnessScale);
 		}
 	}
 	else
@@ -93,9 +93,9 @@ vec3 xyYToRGB(highp float x, highp float y, highp float Y)
                                vec3(-0.3446944, +0.0415560, +1.0154096));
 		}
 	  	if (flagUseTmGamma){
-			return XYZ2RGB * XYZ * brightnessScale;
+			return srgbToLinear(abs(XYZ2RGB * XYZ * brightnessScale));
 		}else{
-			return pow(abs(XYZ2RGB * XYZ * brightnessScale), vec3(oneOverGamma));
+			return XYZ2RGB * XYZ * brightnessScale;
 		}
 	}
 }
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -873,10 +873,7 @@ void StelMainView::init()
 	glInfo.vendor = QString(reinterpret_cast<const char*>(glInfo.functions->glGetString(GL_VENDOR)));
 	glInfo.renderer = QString(reinterpret_cast<const char*>(glInfo.functions->glGetString(GL_RENDERER)));
 	const auto format = glInfo.mainContext->format();
-	glInfo.supportsLuminanceTextures = format.profile() == QSurfaceFormat::CompatibilityProfile ||
-									   format.majorVersion() < 3;
 	glInfo.isGLES = format.renderableType()==QSurfaceFormat::OpenGLES;
-	qDebug().nospace() << "Luminance textures are " << (glInfo.supportsLuminanceTextures ? "" : "not ") << "supported";
 	glInfo.isCoreProfile = format.profile() == QSurfaceFormat::CoreProfile;
 
 	auto& gl = *QOpenGLContext::currentContext()->functions();
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -913,6 +913,46 @@ void StelMainView::init()
 	gl.glGetIntegerv(GL_MAX_TEXTURE_SIZE, &glInfo.maxTextureSize);
 	qDebug() << "Maximum 2D texture size:" << glInfo.maxTextureSize;
 
+	// sRGB support in GLES is a mess
+	if(glInfo.isGLES)
+	{
+		if(format.majorVersion() >= 3 || glInfo.mainContext->hasExtension("GL_EXT_texture_sRGB"))
+		{
+			glInfo.srgbTextureFormatRGB = GL_RGB;
+			glInfo.srgbTextureFormatRGBA = GL_RGBA;
+			glInfo.srgbTextureInternalFormatRGB = GL_SRGB8;
+			glInfo.srgbTextureInternalFormatRGBA = GL_SRGB8_ALPHA8;
+			glInfo.supportsGenerateMipmapSRGB = false;
+		}
+		else if(glInfo.mainContext->hasExtension("GL_EXT_sRGB"))
+		{
+			glInfo.srgbTextureFormatRGB = GL_SRGB;
+			glInfo.srgbTextureFormatRGBA = GL_SRGB_ALPHA_EXT;
+			glInfo.srgbTextureInternalFormatRGB = GL_SRGB;
+			glInfo.srgbTextureInternalFormatRGBA = GL_SRGB_ALPHA_EXT;
+			glInfo.supportsGenerateMipmapSRGB = false;
+		}
+		else
+		{
+			qCritical() << "No support for sRGB textures found, expect rendering problems!";
+			// Let the user still have gamma-incorrect rendering if possible (though
+			// configurations without sRGB support would likely have other problems).
+			glInfo.srgbTextureFormatRGB = GL_RGB;
+			glInfo.srgbTextureFormatRGBA = GL_RGBA;
+			glInfo.srgbTextureInternalFormatRGB = GL_RGB;
+			glInfo.srgbTextureInternalFormatRGBA = GL_RGBA;
+			glInfo.supportsGenerateMipmapSRGB = true;
+		}
+	}
+	else
+	{
+		glInfo.srgbTextureFormatRGB = GL_RGB;
+		glInfo.srgbTextureFormatRGBA = GL_RGBA;
+		glInfo.srgbTextureInternalFormatRGB = GL_SRGB8;
+		glInfo.srgbTextureInternalFormatRGBA = GL_SRGB8_ALPHA8;
+		glInfo.supportsGenerateMipmapSRGB = true;
+	}
+
 	gui = new StelGui();
 
 	// Should be check of requirements disabled? -- NO! This is intentional here, and does no harm.
--- a/src/StelMainView.hpp
+++ b/src/StelMainView.hpp
@@ -77,6 +77,11 @@ public:
 		PFNGLMINSAMPLESHADINGPROC glMinSampleShading = nullptr;
 		GLint maxAnisotropy = 0;
 		GLint maxTextureSize = 2048;
+		GLenum srgbTextureFormatRGB = GL_RGB;
+		GLenum srgbTextureFormatRGBA = GL_RGBA;
+		GLenum srgbTextureInternalFormatRGB = GL_SRGB8;
+		GLenum srgbTextureInternalFormatRGBA = GL_SRGB8_ALPHA8;
+		bool supportsGenerateMipmapSRGB = true;
 		bool isCoreProfile = false;
 		bool isGLES = false;
 	};
--- a/src/StelMainView.hpp
+++ b/src/StelMainView.hpp
@@ -77,7 +77,6 @@ public:
 		PFNGLMINSAMPLESHADINGPROC glMinSampleShading = nullptr;
 		GLint maxAnisotropy = 0;
 		GLint maxTextureSize = 2048;
-		bool supportsLuminanceTextures = false;
 		bool isCoreProfile = false;
 		bool isGLES = false;
 	};
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -19,6 +19,8 @@
 
 #include "StelApp.hpp"
 
+#include "StelSRGB.hpp"
+#include "Dithering.hpp"
 #include "StelCore.hpp"
 #include "StelMainView.hpp"
 #include "StelSplashScreen.hpp"
@@ -79,7 +81,10 @@
 #include <QNetworkDiskCache>
 #include <QNetworkProxy>
 #include <QNetworkReply>
+#include <QOpenGLBuffer>
 #include <QOpenGLContext>
+#include <QOpenGLShaderProgram>
+#include <QOpenGLVertexArrayObject>
 #include <QOpenGLFramebufferObject>
 #include <QString>
 #include <QStringList>
@@ -101,6 +106,10 @@
 #ifdef ENABLE_SPOUT
 #include <QMessageBox>
 #include "SpoutSender.hpp"
+#endif
+
+#ifndef GL_RGBA16
+# define GL_RGBA16 0x805B
 #endif
 
 #ifdef USE_STATIC_PLUGIN_HELLOSTELMODULE
@@ -665,6 +674,9 @@ void StelApp::init(QSettings* conf)
 
 	// Animation
 	animationScale = confSettings->value("gui/pointer_animation_speed", 1.).toDouble();
+
+	ditherPatternTex = StelApp::getInstance().getTextureManager().getDitheringTexture(0);
+	setupLinearToSRGBBlitter();
 	
 #ifdef ENABLE_SPOUT
 	//qDebug() << "Property spout is" << qApp->property("spout").toString();
@@ -811,6 +823,62 @@ void StelApp::applyRenderBuffer(GLuint drawFbo)
 	viewportEffect->paintViewportBuffer(renderBuffer);
 }
 
+void StelApp::setupLinearToSRGBBlitter()
+{
+	linearToSRGB_VBO.reset(new QOpenGLBuffer(QOpenGLBuffer::VertexBuffer));
+	linearToSRGB_VBO->create();
+	linearToSRGB_VBO->bind();
+	const GLfloat vertices[]=
+	{
+		// full screen quad
+		-1, -1,
+		 1, -1,
+		-1,  1,
+		 1,  1,
+	};
+	GL(gl->glBufferData(GL_ARRAY_BUFFER, sizeof vertices, vertices, GL_STATIC_DRAW));
+
+	linearToSRGB_VAO.reset(new QOpenGLVertexArrayObject);
+	linearToSRGB_VAO->create();
+	linearToSRGB_VAO->bind();
+	gl->glVertexAttribPointer(0, 2, GL_FLOAT, false, 0, 0);
+	linearToSRGB_VBO->release();
+	gl->glEnableVertexAttribArray(0);
+	linearToSRGB_VAO->release();
+
+	linearToSRGB_Program.reset(new QOpenGLShaderProgram);
+	linearToSRGB_Program->addShaderFromSourceCode(QOpenGLShader::Vertex,
+		StelOpenGL::globalShaderPrefix(StelOpenGL::VERTEX_SHADER) + R"(
+ATTRIBUTE vec3 vertex;
+VARYING vec2 texcoord;
+void main()
+{
+	gl_Position = vec4(vertex, 1.);
+	texcoord = 0.5*vertex.xy+0.5;
+}
+)");
+	linearToSRGB_Program->addShaderFromSourceCode(QOpenGLShader::Fragment,
+		StelOpenGL::globalShaderPrefix(StelOpenGL::FRAGMENT_SHADER) + R"(
+VARYING vec2 texcoord;
+uniform sampler2D tex;
+
+)" + makeDitheringShader() + makeSRGBUtilsShader() + R"(
+
+void main()
+{
+	vec4 rgba = texture2D(tex, texcoord);
+	vec3 srgb = linearToSRGB(clamp(rgba.rgb, vec3(0), vec3(1)));
+	FRAG_COLOR = vec4(dither(srgb), rgba.a);
+}
+)");
+	StelPainter::linkProg(linearToSRGB_Program.get(), "Linear RGB to sRGB conversion program");
+	linearToSRGB_Program->bind();
+	linearToSRGBUniformLocations.tex           = linearToSRGB_Program->uniformLocation("tex");
+	linearToSRGBUniformLocations.ditherPattern = linearToSRGB_Program->uniformLocation("ditherPattern");
+	linearToSRGBUniformLocations.rgbMaxValue   = linearToSRGB_Program->uniformLocation("rgbMaxValue");
+	linearToSRGB_Program->release();
+}
+
 //! Main drawing function called at each frame
 void StelApp::draw()
 {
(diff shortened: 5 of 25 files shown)


Atque:
Hm, stars and DSO's look a bit weird (too obvious compared to the master-branch), and the cardinals are no longer red.

stellarium-176

Built with Qt 6.4.0, Windows 10 64-bit, RX 5700 XT.

10110111:
This is not the final state, some items haven't been fully transformed. I just stumbled upon the problem with GLES2, which may be a blocker for the current approach.

10110111:
OK, it appears that only `GL_SRGB8` format is not color renderable in the problematic GLES2 mode. `GL_SRGB8_ALPHA8` gets mipmaps as expected, which explains why only the bottom of the old-style landscape was black.

Moreover, manual specification of isotropically-filtered mipmap levels appears to be sufficient to let even anisotropic filtering work. Looks like the GL implementation does the necessary actions to automagically generate the additional anisotropically-filtered levels from what we specify manually (or maybe this just happens on the fly during texture sampling?).

10110111:
If we ignore the text gamma issue discussed in #3213, I think this PR is ready. Please check that there are no modules left with gamma-incorrect colors, and that this version works on both desktop OpenGL and GLES.

alex-w:
macOS 13.3.1 / M1

23.1:
stellarium-001
stellarium-002
stellarium-003

this branch:
stellarium-004
stellarium-005
stellarium-006

10110111:
Looks OK, doesn't it? BTW, does ShowMySky not work on M1, or you don't enable it for another reason?

Atque:
I noticed that using ShowMySky, the display gamma option is unresponsive. It works for the Preetham model though.

bild

10110111:
> I noticed that using ShowMySky, the display gamma option is unresponsive.

This is intended. ShowMySky uses a real sRGB EOTF, not a gamma power function. This option exists only to make the legacy atmosphere look as it used to look.

alex-w:
> Looks OK, doesn't it? 

Yes, looks OK

> BTW, does ShowMySky not work on M1, or you don't enable it for another reason?

Just use standard options for atmosphere.

> I have to note though, that you shouldn't expect the colors to be identical. The whole reason for this PR is to make them more sensible when we render physically-based scenery. (Currently not much of it is physically-based.)

I understand it, but I share screenshots for you to check expected visuals :)

Current branch with ShowMyShy:
stellarium-001
stellarium-002
stellarium-003

alex-w (approved):

Atque:
Lunar eclipses look much greyer now than in the master:

Master:
stellarium-199

PR:
stellarium-198

Some kind of tweak might be needed for `textures/earth-shadow.png`.

10110111:
@gzotti what about your comments/review?

gzotti:
Sorry, I hope later next week! This is something that needs to be tested thoroughly, not only if it runs on ANGLE and Raspberries. We also have many tweaks that may have been added rather ad-hoc (visual fiddling), these probably need to be re-configured. The eclipsed Moon is just one example. What about apparent Zodiacal light brightness etc.? I need to run RemoteSync'ed instances in parallel to check color/brightness differences. 

In any case this (~the PR text and motivation) needs to be added somewhere to the source documentation.

data/shaders/planet.frag
```diff
@@ -434,7 +433,5 @@ void main()
     //apply white rimlight
     finalColor.xyz = clamp( finalColor.xyz + vec3(outgas), 0.0, 1.0);
 
-    FRAG_COLOR = vec4(linearToSRGB(finalColor.rgb), finalColor.a);
-    //to debug texture issues, uncomment and reload shader
```
gzotti, line 438: for the sake of less experienced GLSL devs, please leave such comments/alternative code paths visible.

10110111: It would mean maintaining dead code. In particular, the following commented out line would be out of date if left as is.

gzotti:
RPi3: fail! :-(
[log.txt](https://github.com/Stellarium/stellarium/files/11516643/log.txt)
I see a naked (Preetham) sky, no planets, no ground, no DSO (just the symbols).

gzotti:
RPi4: master still fails to link when building ShowmySky (I think it did so before):

>[ 75%] Linking CXX shared library libShowMySky-Qt6.so
/usr/bin/ld: ../libcommon.a(util.cpp.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `_ZZNSt8__detail18__to_chars_10_implIjEEvPcjT_E8__digits' which may bind externally can not be used when making a shared object; recompile with -fPIC
../libcommon.a(util.cpp.o): in function `openglErrorString[abi:cxx11](unsigned int)':
util.cpp:(.text+0x23c): dangerous relocation: unsupported relocation
/usr/bin/ld: ../libcommon.a(util.cpp.o): relocation R_AARCH64_ADR_PREL_PG_HI21 against symbol `_ZZNSt8__detail18__to_chars_10_implIjEEvPcjT_E8__digits' which may bind externally can not be used when making a shared object; recompile with -fPIC
util.cpp:(.text+0x474): dangerous relocation: unsupported relocation
collect2: error: ld returned 1 exit status
make[2]: *** [_deps/showmysky-qt6-build/ShowMySky/CMakeFiles/ShowMySky.dir/build.make:153: _deps/showmysky-qt6-build/ShowMySky/libShowMySky-Qt6.so.15.0.0] Error 1
make[1]: *** [CMakeFiles/Makefile2:2459: _deps/showmysky-qt6-build/ShowMySky/CMakeFiles/ShowMySky.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

Building without SMS works, but appears much darker:

Master:
stellarium-002
[log.txt](https://github.com/Stellarium/stellarium/files/11518368/log.txt)


This PR:

stellarium-001

 
[log_PR3156.txt](https://github.com/Stellarium/stellarium/files/11518126/log_PR3156.txt)
There are also a few ICC warnings from libpng, not sure if master also has the same.

10110111:
> RPi4: master still fails to link when building ShowmySky

Dunno what changed, but try configuring cmake with `-DCMAKE_CXX_FLAGS=-fPIC`.

gzotti:
Yes, building with -fPIC worked. Now I face the next issue. 
stellarium-003
Not sure if the error message is meaningful This is also after make install. But it's off-topic to this PR,and RPi4 is OpenGL2.1 only in any case.  -- Ah, solved. Forgot LD_LIBRARY_PATH. Now the OpenGL3.3 error shows, OK.

alex-w:
> Dunno what changed, but try configuring cmake with `-DCMAKE_CXX_FLAGS=-fPIC`.

Hmm… maybe this need add for all linux/arm in cmake?..

10110111:
> Hmm… maybe this need add for all linux/arm in cmake?..

Normally, Qt already enables `-fPIC` for both libraries and executables (and fails to build if this option is omitted). Why should we do anything special for ARM64?

If you find an explanation that tells why manual messing with compilation options is the proper solution in this case, please let me know. Otherwise I consider it a kludge.

gzotti:
There must be a reason for -fPIC not activated on RPi4. Is it activated on arm64/macOS? Then it is RPi-specific and goes into the FAQ. If not, it looks ARM-specific and may be activated on all ARM(64?) builds.

alex-w:
> > Hmm… maybe this need add for all linux/arm in cmake?..
> 
> Normally, Qt already enables `-fPIC` for both libraries and executables (and fails to build if this option is omitted). Why should we do anything special for ARM64?

However, Qt already enabled it, but not Stellarium - we do not changes flags for compilers for specific architectures.

> If you find an explanation that tells why manual messing with compilation options is the proper solution in this case, please let me know. Otherwise I consider it a kludge.

In manual for GCC and on stackoverflow I see `-fPIC` is important for arm64 architecture.

10110111:
> There must be a reason for -fPIC not activated on RPi4.

It would be useful to see the build output with `--verbose` option, so that the build tool prints compiler command lines for each file. CMake should enable `-fPIC` for shared libraries by default, so it would be strange if it doesn't on some platforms.

> QOpenGLFramebufferObject: Framebuffer incomplete attachment.

What if you replace the format in `format.setInternalTextureFormat(GL_RGBA16)` inside `StelApp::draw()` with `GL_RGBA12`, `GL_RGBA16F`, `GL_RGBA32F`? Does any of these work on RPi4?

10110111:
> However, Qt already enabled it, but not Stellarium - we do not changes flags for compilers for specific architectures.

Stellarium does use Qt, so it must already have corresponding options on.

alex-w:
> There must be a reason for -fPIC not activated on RPi4. Is it activated on arm64/macOS? 

No, it's not activated on macOS/arm64 by default and everything works as expected with and without `-fPIC`. 

> Then it is RPi-specific and goes into the FAQ. If not, it looks ARM-specific and may be activated on all ARM(64?) builds.

`-fPIC` is important for some architectures (and arm64 in the list)

gzotti:
> It would be useful to see the build output with `--verbose` option, so that the build tool prints compiler command lines for each file. CMake should enable `-fPIC` for shared libraries by default, so it would be strange if it doesn't on some platforms.

Which program shall run --verbose? gcc? Do I just add this as another -DCMAKE_CXX_FLAGS="-fPIC --verbose" or so?

I will try building Qt5/ANGLE tonight.

What do we do with the outdated Mesa from Qt5 which only has OpenGL3.0? Shall we pack some more recent version, with the necessity to set a few  environments variables? ( I have found some solution, but hope it still works.)

10110111:
> Which program shall run --verbose? gcc? Do I just add this as another -DCMAKE_CXX_FLAGS="-fPIC --verbose" or so?

If you use `cmake --build` for building, then add `--verbose` to this CMake command line. If you use `make`, then pass `VERBOSE=1` option to it. Not sure about `ninja`, it seems to accept `--verbose` option.

> What do we do with the outdated Mesa from Qt5 which only has OpenGL3.0?

Mesa is a project independent from Qt, so for all versions of Qt we should pack the latest version (after having checked that it works). Whatever environment variables are needed, should be applied without requiring any user interaction (but don't break system Mesa on Linux).

gzotti:
This is Win11 with Qt5.12.12:
[log_pr1356_Qt5-Geforce3070Ti.txt](https://github.com/Stellarium/stellarium/files/11522907/log_pr1356_Qt5-Geforce3070Ti.txt)
[log_pr3156_Qt5-angle_blackScreen.txt](https://github.com/Stellarium/stellarium/files/11522908/log_pr3156_Qt5-angle_blackScreen.txt)
[log_pr3156_Qt5-mesa_default_.txt](https://github.com/Stellarium/stellarium/files/11522909/log_pr3156_Qt5-mesa_default_.txt)

So far we have only packed whatever comes with Qt5, but it's really outdated. We could just take one working version and put it into the git (this has been criticised in the past), or s.o. finds out how to download/unpack current releases via CPM(?). Of course, this only goes into Windows/Qt5-packages.  It needs setting envvars on Windows when calling --angle-mode or --mesa-mode. 

It appears Angle would acccess the more modern Mesa23 with GLES3.1 when the library is in the path. However, it is very slow. 
[log_pr3156_Qt5-angle_whyMesa23P_errors.txt](https://github.com/Stellarium/stellarium/files/11522927/log_pr3156_Qt5-angle_whyMesa23P_errors.txt) No idea how to fix that.

10110111:
> It appears Angle would acccess the more modern Mesa23 with GLES3.1 when the library is in the path

This is of no use: the only real reason to use ANGLE is to get access to HW acceleration of Direct3D. If it just provides software rendering, we'd better use Mesa with desktop OpenGL directly.

> This is Win11 with Qt5.12.12

Please check if it works on any of your platforms with format replacements as described in [this comment](https://github.com/Stellarium/stellarium/pull/3156#issuecomment-1555746802).

gzotti:
Just have troubles with QtCreator. I can open any other source file, but opening StelApp.cpp crashes. Even after a reboot! What kind of nonsense is that?

gzotti:
Here are the logfiles with Qt's Angle mode.
[log_pr3156_Qt5-angle_GL_RGBA12.txt](https://github.com/Stellarium/stellarium/files/11523151/log_pr3156_Qt5-angle_GL_RGBA12.txt)
[log_pr3156_Qt5-angle_GL_RGBA32F.txt](https://github.com/Stellarium/stellarium/files/11523152/log_pr3156_Qt5-angle_GL_RGBA32F.txt)
[log_pr3156_Qt5-angle_RGBA16F.txt](https://github.com/Stellarium/stellarium/files/11523153/log_pr3156_Qt5-angle_RGBA16F.txt)

10110111:
> Just have troubles with QtCreator. I can open any other source file, but opening StelApp.cpp crashes

Try disabling CLang code model plugin.

gzotti:
Did not help. Disabled all clang... plugins. No help. Disabled even more plugins. Updated Qt stuff. No help either. 
Then: git checkout master. Load StelApp.cpp successfully. git checkout pr/3156. QtCreator asks to reload, and... boom! 
This is super weird. 
Now I edited the file with another editor and just compiled without opening. That works (see logs above).
Loading StelApp.cpp in master works. When I now quit QtCreator, switch branches and store the new file with the other editor, then start QtCreator, it crashes on loading the "last project".

gzotti:
Now, back to content. Running 23.1 (right) and this (left) in RemoteSync, Qt6.5/OpenGL. 

First observations:
- Sun glare smaller. Better in this context.
- ashen moonlight is less apparent. Admittedly, the model is not based on an astrophysics paper (I could not find one) but based on my own photos. Could be tweaked to become a bit brighter again.
- blue sky has a different hue. Cannot say what is better.
- Milky Way (and likely, ZL) is again less visible. We really have to define a simpler to tweak mixing function. Visibility depends on sky brightness, and this is very different between old sky model and SMS, for reasons not fully identified.
- gridlines mix differently. I would say, even better. (They can almost fade into the horizon)
- fps indicator seems to show double framerate when Stellarium is idle (intent 15fps). Does that count some step twice per final frame now?

Screenshot (62)
Screenshot (63)
Screenshot (69)
Screenshot (70)

10110111:
I think I've fixed the black screen problems now.

The framebuffer problems (which lead to the dark image) seem more serious. Please show the log from RPi4 with stellarium run with `-d` option, so that I could check what extensions are supported.

Also, is there any way you could check with a newer build of ANGLE, so that it supports `GL_EXT_color_buffer_half_float`, and try with `GL_RGBA16F` FBO format?

gzotti:
I can only test RPi4 next weekend. 
On ANGLE, I am afraid we must use QT5.6's ANGLE. Later versions had problems. OTOH, maybe with your modernization going on, this may have remedied those problems (cannot remember what it was!) Maybe later next week (~Thu?).

10110111:
BTW, do we actually intend to support Windows systems that are not capable of Direct3D 11? If not, we could require OpenGL ES 3 (which ANGLE implements on top of D3D11).

gzotti:
We should keep supporting Windows 7 systems which need ANGLE (OpenGL ES2.0 in the way supported by Qt5.6, or, at your discretion, update to some more recent version of ANGLE if you find out what kind of different initialization (?) this requires) or MESA, i.e. those supported by Qt5. I have no connection/relation to OpenGLES3 and don't know where this is used and what it provides. I hope it means 100.0% compatibility with OpenGL ES2.0.  

Qt6-based systems are 64bit systems with Win10 and later. What is the minimum graphics spec for that? And what kind of 64bit system actually needs MESA? Non-Core-i* Intel CPUs? Older AMD CPUs? Or modern systems with specialized setups? (Any benefits for remote access?)

10110111:
I've now checked on Wine using Qt5's ANGLE that's packaged with Stellarium 23.1 (adding my changes as CI-generated patches), and it doesn't support anything that's required to get an sRGB or 16 bpc or any other wide-bit FBO. It provides GLES2. Then I tried a weekly build of ANGLE from [this generator](https://github.com/mmozeiko/build-angle). This version uses D3D11 to provide GLES3, and with it things worked well. Someone still needs to check that this works well on real MS Windows, rather than Wine.

> We should keep supporting Windows 7 systems which need ANGLE

Our installer barks if I have Windows 7 set in Wine config. Not sure if it was for Qt5 or Qt6 version though.

Note that whether a user needs ANGLE is defined not by the OS, but rather by the quality of their GPU OpenGL drivers. Windows 7 supports Direct3D 11, but it's just API-level support. The GPU needs to be capable enough for D3D11 to be useful. What desktop GPUs do you still want to support? Already in 2008 GeForce 8 series supported OpenGL 3.3, so maybe we could simply require a D3D11/OpenGL3.3-capable GPU (with an exception for the Raspberries, but these don't run Windows)?

> update to some more recent version of ANGLE if you find out what kind of different initialization (?) this requires

In my checks on Wine I simply replaced the libraries (`libGLESv2.dll` and `libEGL.dll`) with the newer ones, and this worked without any additional changes. I got OpenGL ES 3, which worked well with this PR.

> Qt6-based systems are 64bit systems with Win10 and later. What is the minimum graphics spec for that? And what kind of 64bit system actually needs MESA? Non-Core-i* Intel CPUs? Older AMD CPUs? Or modern systems with specialized setups? (Any benefits for remote access?)

No idea. Every Windows system I had access to (there were few) worked well enough with OpenGL 3.3. This included laptop Intel GPUs and NVIDIA discrete ones.

gzotti:
I can confirm that master works with this new ANGLE on W11, Qt5.12.12, Laptop Geforce 3070Ti. Can we adjust our CI/Makefile/whatever-it-takes? We should also move MESA on Windows to Mesa22/23.

I have no more Win7 to test all this, though, and have no systems with too weak GPUs. There are some observers who are using old PCs in their setups. And while indeed my first laptop (2008) had a Geforce 9800 with OpenGL3.3 (and had all graphics extensions we ever needed on XP and later Ubuntu), I made the error to buy an Atom-"powered" travel netbook in 2010 which only had OpenGL1.4. Qt5-based version never worked, so I sold it around 2013. (Maybe it ran Mesa, but what can one do at 1.3 fps (?) on a weak Atom anyways? Clearly a no-fun system...)

(I was writing "Win7" and "Win10" because maybe Win10 has some higher hardware-related specs. Of course the GPU+drivers matter. And I have zero insight into graphics capabilities of these various micro PCs which are now en vogue as telescope computers.)

I have a Core-i7-860 (?) (1st-gen) laptop which had a switchable (pre-optimus) Geforce 340. Due to missing drivers I lost the Geforce part, and it runs ANGLE on the Intel GPU. That should provide some insight, hopefully also next week. 

If anyone has a non-Core-i (Pentium 6? Celeron?) without appropriate GPU, it will be something to test. I am however unsure whether we should rush that now into 23.2 with the risk of thousands of failing systems, or rather test this over the summer for 23.3.  It can go into weekly betas in early July. 

[log.txt](https://github.com/Stellarium/stellarium/files/11715426/log.txt)

10110111:
Even though this isn't going into the soon-to-happen release, I need to know how this PR behaves on Raspberries or any other low-end systems (and see logs from stellarium run with `-d` option) — to know how to develop this further.

I suspect that we may have to split rendering modes (and visual results) between modernish (GL3.3-capable) systems and these poor ones.

Also, I think we need to drop support for any desktop GPUs that don't support OpenGL 3.3. It's been lots of time since GL3.3 support appeared, and Stellarium has already had a precedent of dropping support for some old GPUs (e.g. Intel Atom's GMA950, which used to work with Stellarium 0.12.4 on Ubuntu 14.04). Extra-small machines as Raspberries we might continue to support, but at least desktops should be required not to be ancient.

gzotti:
I was away from my device zoo for a good week, and have also other tasks and not enough time for my own PRs....
If we drop non-3.3 systems, we can as well drop Qt5. However, this is a very bad idea IMHO, as many people run low-spec or old PCs for their observatories. The GMA950/3150 was dropped eons ago, with introduction of Qt5 in V0.13 in mid-2014. Many people however still run first-gen Intel-only Core-i (or not even Core-i!) which are blacklisted by Qt w.r.t. OpenGL and use ANGLE or even Mesa. Running non-SMS is less dramatic than not running Stellarium. At earliest drop-out date I would propose October 2025 when Win10 has EOL and a few more functionally critical bugs have been solved. We can however really update the packed versions of Mesa and ANGLE. I have two 1st-gen Core-i systems with build environment (ANGLE), access to one 2008 tri-core AMD Phenom X3 with Geforce 210 (?; runs OpenGL sufficiently; installers only) but nothing still inferior. However, I will again be off my device zoo mid-next week.

Other testers more than welcome!!

alex-w:
32-bit applications can be produce with Qt5 only, so, we cannot drop support Qt5 at the moment IMHO, but we may planned EOL for 32-bit Stellarium packages

10110111:
> 32-bit applications can be produce with Qt5 only

Depends on your will :) I successfully use Stellarium+Qt6 on a Linux-based OS with 32-bit userspace.

> we may planned EOL for 32-bit Stellarium packages

Should this maybe be created as a separate issue, so that we could plan there what exactly and when would be dropped?

alex-w:
> > 32-bit applications can be produce with Qt5 only
> 
> Depends on your will :) I successfully use Stellarium+Qt6 on a Linux-based OS with 32-bit userspace.

I don't want self-build Qt6 from source code in cross-platform environment :)

gzotti:
Please stay reasonable. Building Qt6-32bit just because it's possible is not on our agenda. Again: There are users of older PCs with basic graphics capabilities from the early years of Qt5. Not everybody can afford to just buy a new €600 PC (and €400 PCs are usually really no fun). For those, and all telescope users with equipment for which only older 32-bit ASCOM drivers exist we need 32-bit, Win7-compatible, no-OpenGL3.3 requirement delivered by Qt5. The 0.12 (Qt4) or also the "Stellarium classic" packages (WinXP/32bit) were dropped when download count went into the low thousands. But it is not a problem on Windows only. I know some old-MacOS user who can sadly no longer upgrade. And a few SBCs also are EOL and "final" with Ubuntu 18LTS, but still somewhat useful.

alex-w:
> I know some old-MacOS user who can sadly no longer upgrade. And a few SBCs also are EOL and "final" with Ubuntu 18LTS, but still somewhat useful.

I have many old macs - m68k, ppc and x86 - but I need reinstall macOS onto my old iMac (here was produces packages for MacOS X since 0.10.6) and in theory I can produce a "classic" 32-bit packages for macOS :)

10110111:
> all telescope users with equipment for which only older 32-bit ASCOM drivers exist we need 32-bit, Win7-compatible

OK, no big deal. Though the real reason I see for Qt5 here is Windows 7, not 32-bitness.

> no-OpenGL3.3 requirement
> Not everybody can afford to just buy a new €600 PC (and €400 PCs are usually really no fun)

Speaking of desktop PCs, one can buy a PCIE OpenGL4-level video card for less than 100$, especially if looking for a used one. And then even a Pentium4-based PC will work with modern Stellarium (provided it has enough RAM). I actually have an experience that a simple PCIE card turned such a PC from junk to an enjoyable machine.

> when download count went into the low thousands

Well, we cannot rely on this currently until all Telescope plugin issues with Qt6 are fixed, and the default download link on the main site is replaced with Qt6 version.

> Running non-SMS is less dramatic than not running Stellarium

The problem here is that these super-old GPUs hold me back from doing more graphic improvements than just ShowMySky for Stellarium. I'd really not like having to split the graphic engine to legacy+modern.

gzotti:
The reason for 32-bit are legacy ASCOM drivers for legacy hardware. This implies Qt5 for all practical purposes, and must such users may not use SMS for efficiency.
The reason for Win7/non-OpenGL3.3 are old computers (desktop+laptops) for many observers and users who don't want or cannot afford upgrades for various reasons. I know, a desktop PC can be revived with cheap GPU and SSD. An old laptop can usually also be revived with an SSD, but is usually limited to its original GPU (or even worse if, like on my otherwise OK-ish little 2010 Core i7-(1stGen) VAIO, the switchable Geforce lost driver support one version upgrade after the Win10 upgrade). I can run 23.1/Qt5/ANGLE on this.  The system is faster than my 2015 OpenGL4.2 1GHz AMD netbook of the same size which can run Qt6 with SMS (but non-SMS is way faster here).

Seriously, we should try to identify and fix existing bugs this year before advancing in other areas which finally need to shed off older systems. We have two issues, the location fixes and ObsLists, which should reach completion in 23.2. I hope to have the location fixes done this weekend and try my Raspberries with -d for you while also preparing for a symposium, and hope Alexander can work on ObsList to complete what he sees required. When there is a version without severe bugs/crashes under regular use conditions, we can raise the hardware requirements.

gzotti:
> Even though this isn't going into the soon-to-happen release, I need to know how this PR behaves on Raspberries or any other low-end systems (and see logs from stellarium run with `-d` option) — to know how to develop this further.
> 

Here is the log of RPi4/UbuntuMate. Tomorrow I will send an RPi3 log. 
[log.txt](https://github.com/Stellarium/stellarium/files/11777243/log.txt)

10110111:
> Here is the log of RPi4/UbuntuMate

Did the image look OK? Was the performance usable?

gzotti:
This was master (I think), sent just for the -d output. apt dist-upgrading my RPi3 now...

10110111:
Please send the log with this PR, and observations of the rendering. I need to know how it works in the current state of the PR.

gzotti:
Here is a log and even screenshot from RPi3. It works, but there are some errors (invalid GL values, and some malloc error at shutdown). There is also now a seam where the texture wraps near the South point. 
[log.txt](https://github.com/Stellarium/stellarium/files/11778664/log.txt)
stellarium-001

Uh... switched to Guereins landscape (old_style): out of memory. Do you create some extra buffers for this linear CS?
[log:Pi3-crashWithOldstyle.txt](https://github.com/Stellarium/stellarium/files/11778679/log.Pi3-crashWithOldstyle.txt)

10110111:
> There is also now a seam where the texture wraps near the South point

This should also be present in `master`.

> Do you create some extra buffers for this linear CS?

Not sure what CS is, but yes I certainly create a framebuffer that keeps linear-color image. But no, this crappy GPU doesn't succeed in creating it, so effectively no, there's no extra memory use. Maybe the same error would happen on `master`.

The ugliest part of this RPi3 is this:

> No support for sRGB textures found, expect rendering problems!

Can we maybe require RPi4 at least and drop support for RPi3? Though, I still need to make sure that RPi4 actually works, for which I need a log from you.

gzotti:
Linear CS = Linear ColorSpace

Here is the log of RPi4/UbuntuMate of pr/3156.  It also shows errors.
[log.txt](https://github.com/Stellarium/stellarium/files/11779146/log.txt)

It seems the Milky Way is missing. Ah no, it is super dim. Exaggerating 10x and zooming in makes it appear. This needs remix. DSO textures seem OK.

With this linear CS, it appears the extra gamma term in the tonemapper may become unneeded. Now it's too colorful.
stellarium-002
stellarium-001

gzotti:
Do you have a Qt5-based build to test on an Core-i5-1stGen with ANGLE? I have no build environment on this one (yet).

10110111:
> It seems the Milky Way is missing. Ah no, it is super dim.

We get `QOpenGLFramebufferObject: Unsupported framebuffer format.` here, so actually the render is gamma-incorrect. This is also the reason for your super-colorful second screenshot.

But, the list of extensions contains `GL_ARB_color_buffer_float`. What if you replace the `GL_RGBA16` [here](https://github.com/Stellarium/stellarium/pull/3156/files#diff-222f2b6a02a9a285b2043af84d0cc28600137c9a8df02f8d479cba88966c072cR910) with `GL_FLOAT32F` or `GL_FLOAT16F`? Does anything change?

> Do you have a Qt5-based build to test on an Core-i5-1stGen with ANGLE? I have no build environment on this one (yet).

I can push a temporary `[publish]` commit so that you could get a patch (don't forget to replace the shaders from the source when using it).

10110111:
You can find Qt5-based patches in [this build](https://ci.appveyor.com/project/10110111/stellarium/builds/47328817) artifacts after it finishes, if I didn't mess it up.

gzotti:
Thanks. Here are logs from 23.1 and with this patch. The new version looks again darker/more colorful, and planets are black by shader error. Then, zooming around, nebua texture creation started, and it crashed without further notice.
[log_23.1_Corei5-650_Win10_ANGLE.txt](https://github.com/Stellarium/stellarium/files/11780903/log_23.1_Corei5-650_Win10_ANGLE.txt)
[log_23.1+_Corei5-650_Win10_ANGLE_PlanetShaderError_crashLoadingDSOtextures.txt](https://github.com/Stellarium/stellarium/files/11780904/log_23.1%2B_Corei5-650_Win10_ANGLE_PlanetShaderError_crashLoadingDSOtextures.txt)

This is a 2010 PC (Lenovo ThinkCentre) which is otherwise fine (upgraded with SSD), and could perfectly well serve as office or observatory PC. Millions of those should exist. Of course, unsuited for Win11 by MS caprice. Will run Linux in 2025... No Intel OpenGL drivers installed, standard Stellarium fails with OpenGL1.1. 

I have no Intel non-Core-i to test cheaper PCs of similar age.

If updating ANGLE and MESA to a later version helps for older PCs, we can do that after release of 23.2.

10110111:
> ```
> QOpenGLShader::compile(Fragment): ERROR: 0:170: 'srgbToLinear' : function already has a body 
> ERROR: 0:178: 'linearToSRGB' : function already has a body
> ```

You seem to have ignored my

> (don't forget to replace the shaders and textures from the source when using it)

The patch only contains the EXE. All the resources are still from 23.1 in your installation.

10110111:
> OpenGL ES 2.0 (ANGLE 2.1.99f075dade7c)

This is not going to work in any case. You need the modern ANGLE version I linked above.

gzotti:
I don't have sources and dev tools on this PC. You mean I should use shaders and textures from sources of your branch? Which exactly? I really have not enough time for much of further testing until after next week, sorry. 

Just installed the new ANGLE. Still reports OpenGL ES 2.0. 
[log.txt](https://github.com/Stellarium/stellarium/files/11781812/log.txt)

If this means upgrading ANGLE does not help with this still widespread PC class, we have a problem with this.

10110111:
> Just installed the new ANGLE. Still reports OpenGL ES 2.0.

Oh well. OK, so I'll have to somehow split the engine to provide anything new or improving what's already there. This in particular means copy-pasting some files and doing diverging changes to them afterwards.

10110111:
What if we stop supporting dithering on the machines where OpenGL Core Profile is not available? I.e. on Raspberries etc., in ANGLE mode, in `--opengl-compat` mode, and the likes. These systems would get the banded image like Stellarium used to have in 2018, but this will help me reduce the amount of code duplication for the further development of the high-graphics mode.

PR #3166, opened 2023-04-09
Vectorize markers of diffuse and dark nebulas

10110111:
### Description
Currently the nebular markers are made of a hierarchy of similar rectangles differing only by width of their lines. This is ugly technically, and it doesn't look nice when the nebula has large size on the screen.

This PR replaces the texture-based *diffuse* (and now also *dark*) nebula markers with a line-based scalable one. I tried to make it look mostly the same as the old version, just less pixelized, and without the fancy gradient (this is for simplicity and performance; tell me if you dislike it).

#### Old
stellarium-009
#### New
stellarium-010

### Type of change

### How Has This Been Tested?
**Test Configuration**:

## Checklist:


alex-w (approved):

gzotti:
Yes, very welcome!  I did not test performance, though. Any change you can see?
I still remember changing from generic DSO blob to type icons, and later doing a "preliminary upscale, until s.o. can vectorize it...".

src/core/modules/Nebula.cpp
```diff
@@ -811,13 +808,68 @@ void Nebula::drawHints(StelPainter& sPainter, float maxMagHints, StelCore *core)
...
-				Nebula::texDiffuseNebulaLarge->bind();
-			else
-				Nebula::texDiffuseNebula->bind();
-			break;
+		{
+			// Take into account device pixel density and global scale ratio, as we are drawing 2D stuff.
+			const auto pixelRatio = sPainter.getProjector()->getDevicePixelsPerPixel();
+			const auto scale = pixelRatio * StelApp::getInstance().getGlobalScalingRatio();
+			finalSize *= scale;
+			// FIXME: this size reduction is just for compatibility with the
```
gzotti, line 816: I guess the 0.89 ensures that the box line is as large as the blurred line that had to be technically inside the icon's pixels.  We can of course get rid of it. The symbol should approximately represent the size given in the data. (But keep a minimum size for small objects or large fov.)

10110111:
> I did not test performance, though. Any change you can see?

No difference on my machine.

10110111:
So, this is how it looks now after removing the hack that reduced marker size:

stellarium-009

src/core/modules/Nebula.cpp
```diff
@@ -811,13 +808,64 @@ void Nebula::drawHints(StelPainter& sPainter, float maxMagHints, StelCore *core)
...
+			const float leftInnerX = leftOuterX + roundRadius;
+			const float bottomOuterY = y - finalSize;
+			const float bottomInnerY = bottomOuterY + roundRadius;
+			const float rightOuterX = x + finalSize;
+			const float rightInnerX = rightOuterX - roundRadius;
+			const float topOuterY = y + finalSize;
+			const float topInnerY = topOuterY - roundRadius;
+			using std::sin;
+			using std::cos;
+			for(int n = 0; n < numPointsInArc; ++n)
```
gzotti, line 833: Not sure: I see four loops which all use the same angle and sin/cos values. Would it not be advisable (esp. for lesser systems) to create those values only once, cache them in arrays and re-use in the 4 loops?

10110111: Maybe, but do you have any evidence that this is a critical place to add more code for this caching? It'll obviously make the code less straightforward to read.

10110111: OK, using a function from StelUtils seems to be an acceptable compromise between efficiency and readability. So I've pushed this.

gzotti: But don't you now compute values around a full circle and use 1/4 of it? ComputeCosSinRhoZone computes the necessary parts of a semicircle, this could avoid some wasted cycles. Admittedly, the StelUtils functions themselves are efficiently avoiding the sin/cos by using the addition theorem, so it may not matter much. 

I think the call would be 
StelUtils::ComputeCosSinRhoZone(M_PI/(2.f*(numPointsInArc-1)), (numPointsInArc-1), M_PIf/2.f);

but I could be 1 off. 

I can test versions on a RPi3 with many nebulae and see if that matters anywhere. I should test on an RPi3 in any case for fps differences.

10110111:
I've also changed the marker of the *dark* nebulas, the result is as follows.

### Old

stellarium-009

### New

stellarium-010

alex-w:
Good!

gzotti:
Tested on RPi3 for the same sky part: no speed penalty. It even may be faster in the hundredths (seen typically 4.88 for master vs 4.91 for this, but both values were fluctuating by &pm;0.2 ... no, cannot really say which is faster).

gzotti:
Do you change all icons? Globulars are also multi-resolution, and galaxies may want that as well. Open clusters may need a very different method for the multi-dots. Maybe make separate private drawDiffuseNebula(pos, size, ...) etc. methods for them (also to combine "cluster with nebulosity" - this could even become two-colored), codefactor.io already complains about "very complex method", and it can only get longer.

10110111:
> Do you change all icons?

No, I only changed two types of nebulas: the rounded-rectish ones. I intend to work on others separately, since they might require more work (especially the dotted ones, where simple lines might not look nice enough).

gzotti (approved):
It looks good and works well, thanks. If you want to change others later, I can approve it now. 
It would be wise to - at least in the next round - refactor a bit and make separate drawing methods to avoid too complex code.

10110111:
> It would be wise to - at least in the next round - refactor a bit and make separate drawing methods to avoid too complex code.

Yes, I did think of this, but decided that I'll first make all the meat and see what common parts can be combined and extracted, and then decide how exactly to refactor.

PR #3172, opened 2023-04-11
Nebula markers vectorization

10110111:
This is a continuation of #3166, now vectorizing markers of open clusters: simple and with nebulosity.

The number of dots now depends on the size of the marker. The markers of open clusters with nebulosity might look not the best way when they are small. Do you think the points should be moved closer to the center? Or is the way it looks now OK?

#### Old
stellarium-009
#### New
stellarium-010


gzotti:
Nice! Yes, IMHO the dots in CL+Neb should be a bit contracted. And now that this would technically be possible, can you please try mixing "green" nebula color box with "yellow" cluster color dot circle?

10110111:
> can you please try mixing "green" nebula color box with "yellow" cluster color dot circle?

I can, but won't it look as if we have two nebulas of different kinds at the same spot? Or do you want to visually separate the nebulosity as a separate nebula?

10110111:
This is how the color-separated version looks. Maybe the circle doesn't have to be contracted in this case.

stellarium-009

gzotti:
I like the bicolor version. However, I'd still make the circle smaller to keep some buffer around the dots.

alex-w:
The bicolor version of C+N markers is acceptable, but IMHO the circle should be a bit small - does not touche the rectangle.

Atque:
Shouldn’t the circles also be vectorized?

Atque:
> > Shouldn’t the circles also be vectorized?
> 
> Why do you think they haven't been? Or do you want to also vectorize the fluffy points that make up the circles?

Yes, I would expect that vectorized markers be fully vectorized.

10110111:
> Yes, I would expect that vectorized markers be fully vectorized.

I don't find any use in this, only potential performance ramifications.

gzotti:
There is of course a change in overall appearance. The old icons had soft edges which became more blurry by upscaling (mitigated by a few icon sizes), while the new line-based symbols are hard-edged, but the dots are still "fluffy". Maybe the dots should become hard-edged as well? I understand drawing blurred lines needs more GPU code.

10110111:
> However, I'd still make the circle smaller to keep some buffer around the dots

New version:

10110111:
> Maybe the dots should become hard-edged as well?

Not sure about this. Some users have multisampling disabled, and such circles will look ugly when aliased.

alex-w (approved):

src/core/modules/Nebula.cpp
```diff
@@ -967,11 +980,13 @@ void Nebula::drawHints(StelPainter& sPainter, float maxMagHints, StelCore *core)
...
 			return;
 		case NebCn:
-			if (finalSize > 35.f)
-				Nebula::texOpenClusterWithNebulosityLarge->bind();
-			else
-				Nebula::texOpenClusterWithNebulosity->bind();
-			break;
+		{
+			renderMarkerRoundedRect(sPainter, XY[0], XY[1], finalSize, col);
+			if(objectInDisplayedType())
```
gzotti, line 985: I think this test is not required. The symbol should be drawn completely or not at all. 

With that bicolor icon, we can and should remove `NebulaMgr::clusterWithNebulosityColor` etc., as it should be combined from openClustersColor and NebulaColor.

src/core/modules/Nebula.cpp
```diff
@@ -887,6 +883,28 @@ void Nebula::renderMarkerRoundedRect(StelPainter& sPainter, const float x, const
 	sPainter.enableClientStates(false);
 }
 
+void Nebula::renderMarkerPointedCircle(StelPainter& sPainter, const float x, const float y,
+									   float size, const Vec3f color, const bool insideRect) const
```
gzotti, line 887: I see the intent of this "insideRect", but where is it used? Maybe a line like 
```
if (insideRect)
    size*=0.95;
```
is still missing? (The scaling value may be something else than 0.95...)

10110111: You're right, I forgot to put an if before line 899.

src/core/modules/Nebula.cpp
10110111: This test is taken from line 913. If it is to be removed here, the same should be done there.

gzotti: Uh, line 913 is funny. It should not set black, but I think it could simply return at this point. @alex-w ?

alex-w: @gzotti within `drawHints()` ?

10110111: @alex-w [here](https://github.com/Stellarium/stellarium/pull/3172/files/0e9d3c620bb3542d76763e88696df64f609c836e#diff-e081b39a4a7e50860a2c73cfa76ccb2d5ba2df7fb09c7b716a96ab096f11b33fL913)

alex-w: Yes, it should be `return;` here

alex-w: > I think this test is not required. The symbol should be drawn completely or not at all.
> 
> With that bicolor icon, we can and should remove `NebulaMgr::clusterWithNebulosityColor` etc., as it should be combined from openClustersColor and NebulaColor.

Hmm... in this case the UI for DSO color and config data should be updated too.

gzotti: Yes, of course. Removing one color will even look better in the current UI.

10110111:
All the comments seem to have been addressed now.

gzotti (approved):
I did not run the latest changes, but they seem complete now (including manual! :-), thank you!

PR #3179, opened 2023-04-14
Vectorize marker of globular cluster and the default circular marker

10110111:
This continues #3172 with two kinds of circular markers.

#### Old
stellarium-011

#### New
stellarium-012


gzotti (approved):
Sure!

10110111:
Now that these easy ones are in, how do you think the galaxy marker should look? Currently it's fluffy, and I think it's meant to somehow reflect the fluffiness of galaxies in real life. Is it something that we must try to preserve when vectorizing, or will conversion to a simple ellipse be OK?

gzotti:
These can become just plain ellipses, just as the nebula boxes lost their blurred lines. Note that I have once added a drawing angle to them. The ellipse drawing could include this angle directly.

Issue #3173, opened 2023-04-11
Clicking on map to change location gets increasingly slower from click to click


gzotti:
### Expected Behaviour

Click on the map in LocationPanel sets new location within a short reaction time with every click

### Actual Behaviour

Clicking 3-6 times to set various random locations takes longer and longer to set new location

### System
* Stellarium version: 23.1
* Operating system: Win11
* Graphics Card: Geforce (likely irrelevant)

### Logfile 
Nothing suspicious

10110111:
Does it happen with Qt 5?

gzotti:
Unfortunately, yes. At least 5.15.2.

alex-w:
I can't reproduce the issue in macOS

P.S. When I/O is happened then choosing the location is a bit slowly of course

gzotti:
But it should not become slower and slower. Why does the 5th click already take several seconds (~twice the time of the 4th click) to compute coordinates from the map (or do something else in the background)? There is no landscape to load, what sort of heavy IO would be triggered?

alex-w:
> But it should not become slower and slower.

Not in the macOS

alex-w:
Stop! This is really slow when atmosphere is disabled!

gzotti:
I have atmosphere on. But something unexpected must be going on here. 
Is there a way (debug command) to let Qt write out triggered signal connections ?

alex-w:
Georg, please try follow:
* Run Stellarium
* Disable atmosphere
* Open Location dialog
* Disable showing stars (S)
* Enable showing stars (S)
* Random click on the map

What happening on your side?

alex-w:
Hmm... Setting time zones in `setLocationFromMap` may be removed, because it set in `setFieldsFromLocation` and the last one is called within `setLocationFromMap`

gzotti:
Maybe. But what happens here that slows it down? I had to set this to return to default location/landscape. 
Why are there in StelCore::moveObserverTo() two emits which send locations?

alex-w:
I can't reproduce the issue in Stellarium 23.1/Win10 O_o

alex-w:
OK, I've remarked line 199, builded the planetarium and run it.

* Open Location dialog
* Random clicks on the map without slowings
* Select the Venus and Ctrl+G
* Of course the map is not changed, but when I click on the map I see in the log:
```
"Can't create StelObserver on planet SpaceShip because it is unknown. Use Earth as default."
[...]

PR #3180, opened 2023-04-14
StelCore: reduce confusion around location signals

gzotti:
### Description

StelCore has two signals after a location change. 

- locationChanged(loc)
- targetLocationChanged(loc, landscapeID)

Those were likely introduced with different intentions in mind, but now are both emitted. Their documentation was lame. 
Several other components are connected to the signals or emit them by themselves, with possible side effects.
See #3173 and see difficulties around the "best eclipse sites". 

IMHO this area of location settings needs to be checked thouroughly and slimmed down or at least made more intuitive. 

These points are to be considered

- [x] Removed default value in signal, force default second arg.
- [x] Emit only one of the two in StelCore::moveObserverTo() EDIT: Reverted. They have both their uses.
- [x] Clicking on the map sets a "zero" landscape with color taken from the clicked point in the map. This is limited to terrestrial locations. For other planets, a "planet" landscape is always used.
- [ ] if default planet landscape cannot be found, a gray zero landscape seems useful

Fixes: #3173 and other possible signal loops or unexpected results or non-results.

I invite @alex-w and @10110111 to collaborate on that.

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:


alex-w:
locationChanged() uses in many places - SolarSystem, AstroCalc, ArchaeoLines, Meteor Showers, Navigational Stars, Observability, Satellites, etc.

gzotti:
Do I understand that targetLocationChanged is "next location to be set, but with a transition animation". This means, LandscapeMgr should receive targetLocation to prepare a new landscape, but without changing location. And the actual location switch should not trigger other things. Currently the RemoteSync landscape switch seems to fail.

alex-w:
> Do I understand that targetLocationChanged is "next location to be set, but with a transition animation". This means, LandscapeMgr should receive targetLocation to prepare a new landscape, but without changing location. And the actual location switch should not trigger other things. Currently the RemoteSync landscape switch seems to fail.

Broke everything except one feature is not a good solution :) In Location dialog `targetLocationChanged` can be replaced by `locationChanged` without problem.

gzotti:
We are breaking things in circles without knowing. Somebody fixes something, breaking others in the way. That needs a bit more study, and documentation of developers intents and observations. Not "not a good solution", but "not a good solution because...". Also, if a comment was "NOTE: Use ...", the next "fix" should not break what is written in the note and remove it. 

Probably the SpaceShipObserver::update(double deltaTime) should NOT set a new landscape when it has reached its target, when that has been done already because LandscapeMgr has received the targetLocationChanged signal.

gzotti:
I now try 10s transitions for the solar eclipses (1-2s would be good for a soft move, this is just for demonstration). They are really bad, at around 1fps. Are there just too many computations (RTS, local circumstances, always requiring lots of extra computations...), or is there still some signal loop? 
In the first case, using immediate view is the only useful way...

gzotti:
D'oh! Found while testing #3187... I had a huge ssystem_minor.ini. With default file, it is just as I wanted it!  *-/

Probably when auto-moving between locations, updating SolarSystem or at least MinorPlanets and Comets should be inhibited. It's usually just for a few seconds.

alex-w (commented):
Great! I forget about `Qt::UniqueConnection` for signal/slot

gzotti:
Seems we need a fallback to "zero" landscape when going e.g. to Venus (as unattractive as that location may be) or other sites with no default planet landscape.

gzotti:
In the Config/Tools tab we have "Auto-enabling for the environment" and "auto select landscapes". These may be better in the Location dialog or in the View/Landscape tab. Also, one sets landscape, the other both landscape and atmosphere. Seems somewhat redundant.

alex-w:
> In the Config/Tools tab we have "Auto-enabling for the environment" and "auto select landscapes". These may be better in the Location dialog or in the View/Landscape tab. Also, one sets landscape, the other both landscape and atmosphere. Seems somewhat redundant.

Probably splitting onto 2 options - to follow atmosphere and landscape settings from the planet - and moving both into Location dialog seems acceptable to avoid misunderstandings

gzotti:
The auto-colored zero horizon works in the location dialog. Clicking e.g. into the desert now shows a sandy color sampled from the map. However, selecting best place from the solar eclipse list now shows the default green zero landscape. I could add a general coordinate->color lookup to StelLocationMgr. Before I continue, @alex-w , @10110111, @worachate001, do you agree on this feature? Should then any location change switch to zero+mapColor? (only for "flagLandscapeAutoSelection"). Probably even when storing zero as default landscape, it could then be re-colored on startup (or color stored in config.ini?). To continue on that, this could be extended to a LandscapeMgr.PolygonalLandscapeColor property that is applied to all polygonal landscapes. Maybe in the next round, however.

10110111:
Just tried going to the best point of a solar eclipse, and both atmosphere and ground were disabled. I enabled them, chose another eclipse, but again they get disabled. The same happens when I click on the map.

If I enable atmosphere and ground and save settings, then they don't get disabled on location switch.

Generally, the choice of color from the map sounds good. Though, if we appear in the ocean, the landscape could better be "Ocean", rather than "Zero Horizon".

gzotti:
Have you enabled both "Auto-enable environment" and esp. "Auto select landscapes"? I must check (and probably fix) behaviour with these settings disabled. These switches should also move into the LocationDialog. 

The problem about using ocean is that the color of the ocean in our map is not singular, and the dark-blue may be used in small interior lakes as well. See the bright-blue Caribbean sea. If we had another 1-bit map or 4-bit land characteristics map, we could develop/define "typical" landscape panos with as close-to-zero as possible skylines. These could be used for flat snow fields and open grassland. But problems are of course mountaineous landscapes or forested areas...  this is beyond the current PR. 

Another coordinate-based lookup could define time zones, but this may have to change occasionally. 

Another way to extend the location map would be a zooming option. I think you have added the larger map image. This is never fully utilized. Or we extend with the latest Qt feature, https://doc.qt.io/qt-6/qml-location5-maps.html. This could start as optional plugin on Qt6.5+. The map feature must remain available also for offline use of course.

10110111:
> Have you enabled both "Auto-enable environment" and esp. "Auto select landscapes"?

- [x] Auto-enabling for the environment
- [ ] Auto-select landscapes

alex-w:
> Have you enabled both "Auto-enable environment" and esp. "Auto select landscapes"? I must check (and probably fix) behaviour with these settings disabled. These switches should also move into the LocationDialog.

+1 for moving these settings into LocationDialog + I think it should be "Auto select landscape" and "Auto enable atmosphere" (replacement for "Auto-enable environment")

> The problem about using ocean is that the color of the ocean in our map is not singular, and the dark-blue may be used in small interior lakes as well. See the bright-blue Caribbean sea. If we had another 1-bit map or 4-bit land characteristics map, we could develop/define "typical" landscape panos with as close-to-zero as possible skylines. These could be used for flat snow fields and open grassland. But problems are of course mountaineous landscapes or forested areas... this is beyond the current PR.

What about adding new type of landscapes for open space environments (ocean, grassland, sea, etc.) - textures + shader for colorification of textures - @10110111 is it possible?

> Another coordinate-based lookup could define time zones, but this may have to change occasionally.

[ZoneDetect ](https://github.com/BertoldVdb/ZoneDetect) library + [Time Zone Builder](https://github.com/evansiroky/timezone-boundary-builder) or using some web API, like [GeoNames](http://www.geonames.org/export/web-services.html#timezone)

> Another way to extend the location map would be a zooming option. I think you have added the larger map image. This is never fully utilized. Or we extend with the latest Qt feature, https://doc.qt.io/qt-6/qml-location5-maps.html. This could start as optional plugin on Qt6.5+. The map feature must remain available also for offline use of course.

We cannot use QML within QWidget and we cannot use QWidget within QML scene, so, you need full replacement of LocationDialog for zooming the maps, but this feature is not necessary for planetarium, because this is not a GIS. In my humble opinion the map with fixed size in the widget is a better solution here. If you want to add some orienteers for users you can re-use data from the locations list to mark capital cities on the Earth or big features for non-terrestrial locations (from nomenclature data).

10110111:
> What about adding new type of landscapes for open space environments (ocean, grassland, sea, etc.) - textures + shader for colorification of textures - @10110111 is it possible?

Well, this is of course possible, but I'm not sure you want to colorify a landscape that you've already specially made for each kind of land. It's better to just have then a map of kinds of land: ocean, grass, sand, ice, and use them without additional coloration.

gzotti:
> > What about adding new type of landscapes for open space environments (ocean, grassland, sea, etc.) - textures + shader for colorification of textures - @10110111 is it possible?
> 
> Well, this is of course possible, but I'm not sure you want to colorify a landscape that you've already specially made for each kind of land. It's better to just have then a map of kinds of land: ocean, grass, sand, ice, and use them without additional coloration.

This is what I mean with "4-bit land characteristics map". If anybody creates that, we could load a "typical subtropical grassland", "typical snowfield", etc. (These would be pre-made fixed "spherical" landscapes.) But this is IMHO a next milestone. And it does not solve the correct skyline, so probably the ZeroColor is "good enough". It still has to be completed, though. Whoever needs a better landscape: Photography, GoogleEarth, Horizon, HeyWhat'sThat, all has been described in the SUG.

Yes, QML cannot be integrated in QWidgets, but still mixed. However, this was just an idea reading Qt6.5 release notes. With the quite large earth map, a scaling/zoom extension of our MapWidget (and maybe immediate display of stored locations) seems simpler to do.

gzotti:
You may give it a try. Of course, some messages will have to be cleaned up.

alex-w:
I've added a small cosmetic fix for GUI (localization related issue), but everything is OK for me. Of course clean up debug/warning messages is needed.

10110111:
```
/home/ruslan/src/stellarium/src/gui/MapWidget.cpp: In member function ‘virtual void MapWidget::mousePressEvent(QMouseEvent*)’:
/home/ruslan/src/stellarium/src/gui/MapWidget.cpp:63:34: error: ‘StelApp’ has not been declared
   63 |  StelLocationMgr* locationMgr = &StelApp::getInstance().getLocationMgr();
      |                                  ^~~~~~~
```

10110111:
Switching locations now forces atmosphere and ground to be enabled, even though I explicitly disabled them. This didn't use to happen, but happens both in `master` and in this PR.

Atque:
> Switching locations now forces atmosphere and ground to be enabled, even though I explicitly disabled them. This didn't use to happen, but happens both in `master` and in this PR.

Try unchecking "Auto-enabling for the environment" in Configurations -> Tools.

10110111:
> Try unchecking "Auto-enabling for the environment" in Configurations -> Tools.

OK, this worked, but it doesn't make any sense, since I was on the Earth, and I didn't change the planet.

alex-w:
> ```
> /home/ruslan/src/stellarium/src/gui/MapWidget.cpp: In member function ‘virtual void MapWidget::mousePressEvent(QMouseEvent*)’:
> /home/ruslan/src/stellarium/src/gui/MapWidget.cpp:63:34: error: ‘StelApp’ has not been declared
>    63 |  StelLocationMgr* locationMgr = &StelApp::getInstance().getLocationMgr();
>       |                                  ^~~~~~~
> ```

It should be fixed now

alex-w:
> > Try unchecking "Auto-enabling for the environment" in Configurations -> Tools.
> 
> OK, this worked, but it doesn't make any sense, since I was on the Earth, and I didn't change the planet.

Please see the screenshot - I cannot reproduce the issue
stellarium-043

10110111:
> I cannot reproduce the issue

Did you check with the checkbox unchecked? I can confirm that when it's unchecked, things work as expected. My last comment was about it not making sense to enable atmosphere and ground when I don't switch planets—that's implied by the tooltip of the Auto-enablement checkbox.

src/gui/AstroCalcDialog.cpp
```diff
@@ -2738,7 +2735,7 @@ LunarEclipseIteration::LunarEclipseIteration(double &JD, double &positionAngle,
 void AstroCalcDialog::selectCurrentLunarEclipseDate(const QModelIndex& modelIndex)
 {
 	double JDMid = modelIndex.sibling(modelIndex.row(), LunarEclipseDate).data(Qt::UserRole).toDouble();
-	if (objectMgr->findAndSelectI18n("Moon") || objectMgr->findAndSelect("Moon"))
```
10110111, line 2741: Why is the search by localized name even needed, if it's statically defined here to be the Moon?

10110111:
When clicking the map, I get the following message in the log:
```
QImage::scaled: Image is a null image
QImage::pixelColor: coordinate (0,0) out of range
```
And the Zero Horizon auto-selected landscape is black.

10110111:
I also get this warning at the loading time:
```
qt.core.qobject.connect: QObject::connect: No such signal StelCore::StelCore::locationChanged(StelLocation)
qt.core.qobject.connect: QObject::connect:  (sender name:   'StelCore')
```

Full log: [log.txt](https://github.com/Stellarium/stellarium/files/11773669/log.txt).

src/core/StelLocationMgr.cpp
```diff
@@ -491,6 +491,10 @@ StelLocationMgr::StelLocationMgr()
 #endif
 	// Init to Paris France because it's the center of the world.
 	lastResortLocation = locationForString(conf->value("init_location/last_location", "Paris, Western Europe").toString());
+
+	planetName="Earth";
+	planetSurfaceMap.load(StelFileMgr::findFile("data/gui/miscWorldMap.jpg", StelFileMgr::File));
```
10110111, line 496: `MapWidget` initializes its pixmap as `QPixmap(":/graphicGui/miscWorldMap.jpg")`. I suppose you'll want to use the same path here.

Atque:
> And the Zero Horizon auto-selected landscape is black. This is on Qt 6.

Huh? It works on Windows 10 64-bit, Qt 6.4.2.

src/gui/AstroCalcDialog.cpp
10110111: That's OK, the replacement doesn't invalidate my question.

src/core/StelLocationMgr.cpp
gzotti: Ah, did you make install before that test? Thanks, this would have been stupid!

10110111:
> Huh? It works on Windows 10 64-bit, Qt 6.4.2.

Maybe `StelFileMgr::findFile` somehow finds the file in your source tree. At least on Linux the installation prefix doesn't contain `miscWorldMap.jpg` at all.

src/core/StelLocationMgr.cpp
10110111: Yes, I always make install.

gzotti:
Yes, I usually run my test builds from the source dir (probably @Atque does the same), so these files are found.

src/core/StelLocationMgr.cpp
```diff
@@ -1569,18 +1569,19 @@ void StelLocationMgr::changePlanetMapForLocation(StelLocation loc)
...
-
-	if (!planetSurfaceMap.load(StelFileMgr::findFile(mapName, StelFileMgr::File)))
+	if (planetName=="Earth")
+		planetSurfaceMap=QImage(":/graphicGui/miscWorldMap.jpg");
+	else
 	{
-		// texture not found. Use a gray pixel.
-		planetSurfaceMap=QImage(16,16,QImage::Format_RGB32);
-		planetSurfaceMap.fill(QColor(64, 64, 64));
+		QString mapName="textures/" + planetName + ".png";
```
10110111, line 1577: `LocationDialog` searches for the map to pass to `MapWidget` as follows:
```
StelFileMgr::findFile("textures/"+p->getTextMapName())
```
Here `p = ssm->searchByEnglishName(loc.planetName)`. Maybe this should be done here the same way.

10110111:
> I also get this warning at the loading time

This warning originates from `StelLocationMgr::StelLocationMgr()`. Why do we even use string-based connect here? The warning is gone if I switch to the pointer-based connect.

gzotti:
The connect(sender, SIGNAL(...), receiver, SLOT(...)) syntax is used almost everywhere simply by tradition.

10110111:
> almost everywhere simply by tradition

This should change just as `Q_OVERRIDE` tradition was changed.

src/gui/AstroCalcDialog.cpp
gzotti: I don't find this code in active use any longer. Of course this is unnecessary.

src/core/StelLocationMgr.cpp
gzotti: Done.

alex-w:
> Did you check with the checkbox unchecked?

Yes

src/gui/AstroCalcDialog.cpp
10110111: Oh, just noticed that it was later replaced by `objectMgr->findAndSelectI18n(name) || objectMgr->findAndSelect(name)`, which makes a bit more sense as a more general code.

10110111:
```
/home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp: In member function ‘void StelLocationMgr::changePlanetMapForLocation(StelLocation)’:
/home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp:1578:45: error: expected primary-expression before ‘)’ token
 1578 |   SolarSystem *ssm=GETSTELMODULE(SolarSystem);
      |                                             ^
/home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp:1578:20: error: ‘GETSTELMODULE’ was not declared in this scope
 1578 |   SolarSystem *ssm=GETSTELMODULE(SolarSystem);
      |                    ^~~~~~~~~~~~~
```

alex-w:
> /home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp: In member function ‘void StelLocationMgr::changePlanetMapForLocation(StelLocation)’:
> /home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp:1578:45: error: expected primary-expression before ‘)’ token
>  1578 |   SolarSystem *ssm=GETSTELMODULE(SolarSystem);
>       |                                             ^
> /home/ruslan/src/stellarium/src/core/StelLocationMgr.cpp:1578:20: error: ‘GETSTELMODULE’ was not declared in this scope
>  1578 |   SolarSystem *ssm=GETSTELMODULE(SolarSystem);
>       |                    ^~~~~~~~~~~~~

Missing include, let me check

gzotti:
Sorry, I seem to have missed a few includes because of PCH.

10110111 (approved):

PR #3186, opened 2023-04-15
Satellites: Allow Satellite Number to be zero

normanr:
### Description
6a7afb8f9e43236b746b22fcd19adf19ce80e07f added stripping of leading zeros, but a temporary/manually created Orbital Element may have a Satellite Number of zero, and that should be allowed.

I considered replacing the regular expression with `QString::toInt()` and `QString::number()`, but that's a larger change and doesn't allow non-numeric values. Adding the "not at a word boundary" assertion at the end is much simpler.

### Type of change

### How Has This Been Tested?
I've added a unit test to test that `00000` and `    0` both parse and return `0`.

**Test Configuration**:
* Operating system: Debian Linux
* Graphics Card: Intel(R) UHD Graphics 620

## Checklist:

### DIFF
--- a/plugins/Satellites/src/Satellites.cpp
+++ b/plugins/Satellites/src/Satellites.cpp
@@ -2685,7 +2685,7 @@ QString Satellites::getSatIdFromLine2(const QString& line)
 	if (!id.isEmpty())
 	{
 		// Strip any leading zeros as they should be unique ints as strings.
-		static const QRegularExpression re("^[0]*");
+		static const QRegularExpression re("^[0]*\\B");
 		id.remove(re);
 	}
 	return id;
--- a/plugins/Satellites/src/test/testSatellites.cpp
+++ b/plugins/Satellites/src/test/testSatellites.cpp
@@ -46,4 +46,14 @@ void TestSatellites::testNoSatDuplication()
     QVERIFY(dutA == dutB);
 }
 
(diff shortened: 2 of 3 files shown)


gzotti:
The NORAD number in my understanding is not for the user to play. I can neither approve nor disapprove this. Is Zero allowed with other programs that process TLE?

normanr:
> The NORAD number in my understanding is not for the user to play. I can neither approve nor disapprove this. Is Zero allowed with other programs that process TLE?

I got TLE data from flightclub.io for the (currently) _future_ Starship Orbital Flight Test, so there's no official NORAD number yet. I hand edited the data to give it a non-zero number, but I think it would be nice if Stellarium handled it without the work-around.

Comments in https://celestrak.com/columns/v04n03/ seems to indicate that leading spaces and zeros should be treated the same, but then gives an example of non-leading zeros being replaced with spaces. Stellarium's parsing worked fine before 6a7afb8f9e43236b746b22fcd19adf19ce80e07f. I suspect that replacing all spaces with zeros would also be a valid way to parse the field.

In the long term NORAD is switching to Orbit Mean-Elements Message which is XML based, and will probably have less ambiguous representation of data.

alex-w (approved):
The use case is very specific, but the fix is acceptable

Issue #3183, opened 2023-04-15
AstroCalc: Avoid the frequent core updates when computing eclipses


gzotti:
### Actual Behaviour
Currently, when searching for eclipses, I see frequent calls of 
```
core->setJD(jd);
core->update(0);
```

This triggers a computation of the whole SolarSystem, but later in these methods all we need are positions of Sun, Earth and Moon. 

It would certainly speed up things if we would call the respective position functions just for the three. 

### System
* Stellarium version: 23.1
* Operating system: all

PR #3187, opened 2023-04-15
Render some wide lines as sets of triangles

### LINKED ISSUE #2836: Bad performance for boundaries
### Expected Behaviour
Good performance to rendering boundaries of constellations

### Actual Behaviour
Bad performance to rendering boundaries of constellations - after enabling displaying boundaries FPS dramatic decreases in 4 times

### Steps to reproduce
1. Run Stellarium
2. Enable boundaries (`B`)

### System
* Stellarium version: master
* Operating system: macOS 13.0 M1

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/9974143/log.txt)

10110111:
This is an experimental partial conversion of some wide-line `LineStrip` primitives to rendering via manual tesselation. May fix #2836.

Please check whether
1. It renders correctly, including when the lines are wide;
2. Performance improves.

### DIFF
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -2400,6 +2400,128 @@ void StelPainter::enableClientStates(bool vertex, bool texture, bool color, bool
 	normalArray.enabled = normal;
 }
 
+void StelPainter::drawFixedColorWideLinesAsQuads(const ArrayDesc& vertexArray, int count, int offset,
+													 const Mat4f& projMat, const DrawingMode mode)
+{
+	if(count < 2) return;
+
+	GLint viewport[4] = {};
+	glGetIntegerv(GL_VIEWPORT, viewport);
+	const auto viewportSize = Vec2f(viewport[2], viewport[3]);
+
+	std::vector<Vec4f> newVertices;
+	if(mode == LineStrip)
+		newVertices.reserve((count-1)*6);
+	else
+		newVertices.reserve(count*6);
+	newVertices.resize((count-1)*6);
+	assert(vertexArray.type == GL_FLOAT);
+	const auto lineVertData = static_cast<const float*>(vertexArray.pointer) + vertexArray.size*offset;
+	const int step = mode==Lines ? 2 : 1;
+	for(int n = 0; n < count-1; n += step)
+	{
+		Vec4f in0;
+		Vec4f in1;
+		switch(vertexArray.size)
+		{
+		case 4:
+			in0 = Vec4f(lineVertData+4*(n+0));
+			in1 = Vec4f(lineVertData+4*(n+1));
+			break;
+		case 3:
+			in0 = Vec4f(lineVertData[3*(n+0)], lineVertData[3*(n+0)+1], lineVertData[3*(n+0)+2], 1);
+			in1 = Vec4f(lineVertData[3*(n+1)], lineVertData[3*(n+1)+1], lineVertData[3*(n+1)+2], 1);
+			break;
+		case 2:
+			in0 = Vec4f(lineVertData[2*(n+0)], lineVertData[2*(n+0)+1], 0, 1);
+			in1 = Vec4f(lineVertData[2*(n+1)], lineVertData[2*(n+1)+1], 0, 1);
+			break;
+		case 1:
+			in0 = Vec4f(lineVertData[n+0], 0,0,1);
+			in1 = Vec4f(lineVertData[n+1], 0,0,1);
+			break;
+		default:
+			in0 = Vec4f(0.f);
+			in1 = Vec4f(0.f);
+			assert(!"Bad number of elements in vertex array");
+		}
+
+		const Vec4f clip0 = projMat * Vec4f(in0);
+		const Vec4f clip1 = projMat * Vec4f(in1);
+
+		const Vec3f ndc0 = Vec3f(clip0.v) / clip0[3];
+		const Vec3f ndc1 = Vec3f(clip1.v) / clip1[3];
+
+		const Vec2f lineDir2d = normalize((Vec2f(ndc1.v) - Vec2f(ndc0.v)) * viewportSize);
+		const Vec2f perpendicularDir2d = Vec2f(-lineDir2d[1], lineDir2d[0]);
+
+		const Vec2f offset2d = (Vec2f(glState.lineWidth) / viewportSize) * perpendicularDir2d;
+
+		// 2D screen coordinates of the new pairs (a,b) of vertices for each input vertex (0,1)
+		const Vec2f v0a_xy = Vec2f(clip0.v) + offset2d*clip0[3];
+		const Vec2f v0b_xy = Vec2f(clip0.v) - offset2d*clip0[3];
+		const Vec2f v1a_xy = Vec2f(clip1.v) + offset2d*clip1[3];
+		const Vec2f v1b_xy = Vec2f(clip1.v) - offset2d*clip1[3];
+
+		// Final 4D coordinates of the new vertices
+		const auto v0a = Vec4f(v0a_xy[0], v0a_xy[1], clip0[2], clip0[3]);
+		const auto v0b = Vec4f(v0b_xy[0], v0b_xy[1], clip0[2], clip0[3]);
+		const auto v1a = Vec4f(v1a_xy[0], v1a_xy[1], clip1[2], clip1[3]);
+		const auto v1b = Vec4f(v1b_xy[0], v1b_xy[1], clip1[2], clip1[3]);
+
+		newVertices[6*n+0] = v0a;
+		newVertices[6*n+1] = v0b;
+		newVertices[6*n+2] = v1a;
+
+		newVertices[6*n+3] = v1a;
+		newVertices[6*n+4] = v0b;
+		newVertices[6*n+5] = v1b;
+	}
+	if(mode == LineLoop)
+	{
+		// Connect the ends
+		const auto lastN = newVertices.size()-1;
+		assert(lastN >= 2);
+		newVertices.push_back(newVertices[lastN-2]);
+		newVertices.push_back(newVertices[lastN]);
+		newVertices.push_back(newVertices[0]);
+
+		newVertices.push_back(newVertices[0]);
+		newVertices.push_back(newVertices[lastN]);
+		newVertices.push_back(newVertices[1]);
+	}
+	vao->bind();
+	verticesVBO->bind();
+
+	verticesVBO->allocate(newVertices.data(), newVertices.size() * sizeof newVertices[0]);
+
+	const auto& pr = basicShaderProgram;
+	pr->bind();
+	pr->setAttributeBuffer(basicShaderVars.vertex, vertexArray.type, 0, 4);
+	pr->enableAttributeArray(basicShaderVars.vertex);
+	pr->setUniformValue(basicShaderVars.projectionMatrix, QMatrix4x4{});
+	pr->setUniformValue(basicShaderVars.color, currentColor.toQVector());
+
+#ifdef GL_MULTISAMPLE
+	const bool multisampleWasOn = multisamplingEnabled && glIsEnabled(GL_MULTISAMPLE);
+	if(multisamplingEnabled && glState.lineSmooth)
+		glEnable(GL_MULTISAMPLE);
+#endif
+
+	// TODO: convert this to a triangle strip, but mind the overlaps and gaps of non-parallel lines' corners
+	glDrawArrays(GL_TRIANGLES, 0, newVertices.size());
+
+	verticesVBO->release();
+	vao->release();
+
+	if (pr) pr->release();
+
+#ifdef GL_MULTISAMPLE
+	if(multisamplingEnabled && !multisampleWasOn && glState.lineSmooth)
+		glDisable(GL_MULTISAMPLE);
+#endif
+}
+
 void StelPainter::drawFromArray(DrawingMode mode, int count, int offset, bool doProj, const unsigned short* indices)
 {
 	if(!count) return;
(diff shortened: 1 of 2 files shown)


alex-w:
Tested in macOS/arm64 and yes, this patch improves performance and solves #2836

src/core/StelPainter.cpp
```diff
@@ -2400,6 +2400,103 @@ void StelPainter::enableClientStates(bool vertex, bool texture, bool color, bool
 	normalArray.enabled = normal;
 }
 
+void StelPainter::drawFixedColorWideLineStripAsQuads(const ArrayDesc& vertexArray, int count, int offset, const Mat4f& projMat)
+{
+	GLint viewport[4] = {};
+	glGetIntegerv(GL_VIEWPORT, viewport);
+	const auto viewportSize = Vec2f(viewport[2], viewport[3]);
```
gzotti, line 2407: for me, when reading code, I find the "auto" type specifier cumbersome, especially for actually simple data types. It always requires a second mental translation from the right side. It may be helpful in some situations to reduce complex casting syntax.

10110111: Generally, it serves several purposes, one of which is DRY (and this one in particular lets avoid annoying situations when you forget to update one of the sides when changing types). Another is just that it doesn't really matter: the name is descriptive, and the type is "whatever is on the RHS, it's not important for reading".

10110111: In cases when one might lose track of types or get hard to understand compiler errors I indeed use explicit types, like [around line 2440](https://github.com/Stellarium/stellarium/pull/3187/files#diff-824dbcd0573315173d549be8b09852aca7c49fbc51fb70efc1031fd583466465R2440).

gzotti: "viewportSize" may be reasonably descriptive, but "v0axy" (below) certainly is not. In QtCreator, type mismatch is nicely highlighted by clangd.

10110111: Regarding `v0axy` I agree.
As for Clang model in QtCreator, it's a piece of crap that can't really parse lots of C and C++ code in my experience, and additionally, uses loads of RAM (or at least used to do so).

alex-w: Clang model in Qt Creator 10 is unusable

10110111:
@gzotti you noted in #2836 that you saw performance problems on Windows. Does this PR change anything?

gzotti:
Just developing my own stuff around finding out why the transitions are so damn slow. (#3183). Will do in 1-2 hrs.

src/core/StelPainter.cpp
gzotti: I see a few exceptional cases where it fails in the Stars area, but in general works pretty well IMHO. It may use lots of RAM, but I have more than enough for that. It finds type mismatches, unused headers and many more useful things. Meanwhile it even gives git info about who last edited the line. As always, YMMV, use what you like best. On Raspi I use emacs...

10110111:
Aren't asteroids created at random times? This might introduce some uncertainties.
Besides, running two apps simultaneously on Windows gives the one with the active window higher priority, which once again can skew the benchmarks.

gzotti:
> Aren't asteroids created at random times? This might introduce some uncertainties. 

No, maybe you mean meteors/shooting stars? I mean the minor planets. 

> Besides, running two apps simultaneously on Windows gives the one with the active window higher priority, which once again can skew the benchmarks.

But not by much. Both show a few tenths of fps up and down, but maybe its something like 37.2 vs 36.8. &pm;0.5. I have plenty of cores free, overall CPU load at 15%. (And yes, we should identify ways to distribute load better!)

The major effect of #2836 was on Apple. If this solves it, it's OK for me.

gzotti:
Seems to be the same as before.
Is this now also used in the gridlines? I see a drop from 37 to 32fps when activating both equatorial and horizon grid. 
The Borders are far fewer lines, probably making detection of differences more difficult.

alex-w:
stellarium-002
stellarium-003

10110111:
@alex-w Is this lines disabled vs lines enabled? What about the case with lines enabled before this commit?

10110111:
> Is this now also used in the gridlines?

Should be I think.

> I see a drop from 37 to 32fps when activating both equatorial and horizon grid.

Is it much worse than before this commit? Do you run Release version of Stellarium?

> The Borders are far fewer lines, probably making detection of differences more difficult.

You can enable all of them for greater effect. The borders had a large impact in #2836.

alex-w:
> @alex-w Is this lines disabled vs lines enabled? 

Yes

> What about the case with lines enabled before this commit?

OK, the screenshots from the current master:
stellarium-004
stellarium-005

Same master, but with `--opengl-compat` option:
stellarium-006
stellarium-007

alex-w (approved):
Great!

10110111:
Your screenshots have strangely inconsistent frame rates, maybe because of the time taken to switch to screenshot mode. Could you type the values that you see on the screen in real time instead of the screenshots?

gzotti:
> > Is this now also used in the gridlines?
> 
> Should be I think.
> 
> > I see a drop from 37 to 32fps when activating both equatorial and horizon grid.
> 
> Is it much worse than before this commit? Do you run a Release build?

23.1: 37fps
This: 32fps. 
All in Release build. I activated Equatorial and Horizontal grids. Framerate intent was higher, so it's maxed out. 

> 
> > The Borders are far fewer lines, probably making detection of differences more difficult.
> 
> You can enable all of them for greater effect. The borders had a large impact in #2836.

Yes, I have read and seen that. However strange that sounds if there apparently were no issues around gridlines on Macs.

10110111:
> All in Release build. I activated Equatorial and Horizontal grids.

Strange behavior (and not really acceptable). Is the CPU used 100%? What about the GPU?

alex-w:
> Your screenshots have strangely inconsistent frame rates, maybe because of the time taken to switch to screenshot mode. Could you type the values that you see on the screen in real time instead of the screenshots?

master, FOV 180 degrees, OpenGL compatibility mode - same FPS (17.9) for both cases - without and with borders.

master, FOV 180 degrees, OpenGL core profile mode - 17.9 and 2.4 FPS for without and with borders respectively.

Max FPS for master is over 30.

current branch, FOV 180 degrees, OpenGL core profile mode - same FPS (17.9) for both cases - without and with borders.

P.S. v23.1, FOV 180 degrees, OpenGL core profile mode - 17.9 and 10.9 FPS for without and with grids (equatorial + horizontal) respectively.

gzotti:
> > All in Release build. I activated Equatorial and Horizontal grids.
> 
> Strange behavior (and not really acceptable). Is the CPU used 100%? What about the GPU?

The resource monitor reports several core are "parked".

Screenshot (34)

10110111:
> P.P.P.S. Max FPS for master and this branch is over 30.

For the future, any benchmarks should be done without frame rate restrictions.

> P.S. v23.1, FOV 180 degrees, OpenGL core profile mode - 17.9 and 10.9 FPS for without and with grids (equatorial + horizontal) respectively.

Does this mean that master has something that drastically degrades performance compared to 23.1?

10110111:
> The resource monitor reports several core are "parked".

Could you check one process at a time? Such simultaneous run really confuses all the measurement, especially system load.

alex-w:
v23.1, FOV 180, FPS 35.6 (without borders) / FPS 2.5 (with borders) - min_fps=100
this branch, FOV 180, FPS 35.3 (without borders) / FPS 27.5 (with borders) - min_fps=100

> Does this mean that master has something that drastically degrades performance compared to 23.1?

No, both has similar performance

gzotti:
> > The resource monitor reports several core are "parked".
> 
> Could you check one process at a time? Such simultaneous run really confuses all the measurement, especially system load.

To ensure identical settings, I first synchronize with RemoteSync, then switch off RemoteSync. On this PC, with the workload displayed, the available resources should allow that, if not 4 parallel sessions. 
Sorry, I must stop here and do mostly sth else in the next days.

10110111:
> On this PC, with the workload displayed, the available resources should allow that, if not 4 parallel sessions.

The problem is that when you run two processes simultaneously, they share RAM bandwidth and GPU processing power (and require context switches and cache invalidation on the GPU). This doesn't let me know what's actually being used for a single Stellarium process.

> To ensure identical settings

For this it's sufficient to have a pre-configured `config.ini` that has everything required for the benchmark enabled and all extraneous (and randomized) disabled. Then just launch two versions one after another, testing them separately.

10110111:
@gzotti When you have time, please test the following scenario with the new debug commit:

1. Set frame rate to virtually unlimited
1. Set large FOV
1. Disable everything: stars, planets, atmosphere, DSO markers, DSO, cardinal directions, etc. — so that the sceen is black
1. Enable azimuthal grid
1. Note the frame rate
1. Note CPU & GPU load
1. Disable azimuthal grid, re-enable it (this should toggle the tessellation, you can see it in the log)
1. Note the frame rate
1. Note CPU & GPU load

I think this will be the fairest comparison of the two modes.

gzotti:
I had to switch several grids to go below 60fps. Both ways fluctuate 36-40 fps, with 9-10% CPU and 24% GPU load. I may have to reboot to activate >60fps which I will not do right now,  I seriously must do sth else now.

gzotti:
With VSync disabled and only horizontal grid 
enabled: 85-93fps GPU fluctuating 28-30% CPU 10-11%
disabled: 88...95fps GPU fluctuating 28..34% CPU 9-12%

Still almost inconclusive. Probably a draw on the latest and greatest system. I can test on Intel only later next week.

alex-w:
Intel UHD Graphics 630; Windows 10

A black screen, FOV 235 - average FPS 49; azimuthal grid - avr. FPS 31; equatorial grid - avr. FPS 31; azimuthal + equatorial grids - avr. FPS 25

alex-w:
This is full crap - this branch with Qt 6.5 on M1 (OpenGL core profile mode) is slow as it was before now :(

10110111:
Have you checked in the log that tessellation is enabled?

alex-w:
> Have you checked in the log that tessellation is enabled? Or simply checkout the previous commit, [278d653](https://github.com/Stellarium/stellarium/commit/278d653b1e4b7a275de57445b853095eb23fbca7).

No tessellation in the log :zany_face:

10110111:
No tessellation — no optimization. You're in the wrong mode :P . Press <kbd>Z</kbd> to enable azimuthal grid and tessellation. Each enable toggles. Or, once again, just go to the commit before the debug one.

10110111:
> Still almost inconclusive. Probably a draw on the latest and greatest system.

I measured a tiny difference on GeForce GTX 750Ti, which is not really latest nor greatest. On my Intel UHD Graphics 620 the difference is also imperceptible. So if you don't mind, I can simply merge this PR.

gzotti:
Maybe I should still test on Intel i5-4xxx and RPi3/4. Sorry, next weekend.

gzotti:
Intel Core i5-4570 (HD4600), Qt6.5: disabled: 15.2&pm;0.3 -- enabled: 15.0&pm;0.3. Again, hard to see the difference. 
Intel Core i5-4570 (HD4600), Qt5.15.2: disabled: 17.5&pm;0.3 -- enabled: 17.2&pm;0.4. Wide lines show about 0.5fps penalty in both paths.
Intel Core i5-4570 (HD4600), Qt5.15.2, ANGLE: disabled: 31.4&pm;0.3 -- enabled: 31.5&pm;0.3. ***Still no wide lines on ANGLE!***

(The fps difference between OpenGL and ANGLE may come from a different FoV.)

RPi3 (smaller screen, not comparable): disabled: 9.37&pm;0.1 -- enabled: 9.43&pm;0.2. A tiny improvement!
RPi4/UbuntuMate22.04/OGL2.1/GLSL1.2/Mesa22.2.5 (same screen as RPi3): disabled: 32.4&pm;0.1 -- enabled: 32.4&pm;0.2. No change.

10110111:
> disabled: 31.4±0.3 -- enabled: 31.5±0.3. _**Still no wide lines on ANGLE!**_

What does this mean? Nothing rendered when thickness is >1?

gzotti:
Yes, lines stay thin. This was always the case on ANGLE (~GLES2.0). I just hoped this may enable wide lines. `StelPainter::setLineWidth(float width)` does nothing when not on a Core profile. Is that outdated?

10110111:
> `StelPainter::setLineWidth(float width)` does nothing when not on a Core profile. Is that outdated?

It doesn't do nothing, it sets `glState.lineWidth` that is used when rendering wide lines. But yes, you pointed me to the reason of this problem. It's because `wideLineMode` checks whether Core profile is active, so in GLES mode we don't set line width. I'll try to fix this.

Does your testing show that this PR is OK other than this?

gzotti:
Yes, I think other than that this would be OK. If this enables wide lines in ANGLE, it is also visible progress here!

10110111:
OK, it should work in GLES2 now. Caveat as before: wide lines with color varying per vertex are still rendered with geometry shader on Core profile, and as thin lines in GLES2. This use case is not so frequent though, so I think it should be OK.

10110111:
@alex-w Please check that the last commit hasn't broken anything on macOS.

gzotti:
> OK, it should work in GLES2 now. Caveat as before: wide lines with color varying per vertex are still rendered with geometry shader on Core profile, and as thin lines in GLES2. This use case is not so frequent though, so I think it should be OK.

Confirm this works! I think we currently (never had) lines with separate vertex colors. Ah, not sure about the technical details of meteors. They should not be affected, though. OK, from my side it's fine now, thank you!

alex-w:
> @alex-w Please check that the last commit hasn't broken anything on macOS. The logic is a bit convoluted now.

30FPS with boundaries and 27FPS without boundaries - so, this is acceptable

PR #3202, opened 2023-04-25
Refactoring TZ support

alex-w:
### Description
The fixator for missing and wrong time zones is refactored in this pull request - the old code are splitted into code and data now.

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -1214,15 +1214,12 @@ void StelLocationMgr::loadTimeZoneFixes()
 			if (line.startsWith("//") || line.startsWith("#") || line.isEmpty())
 				continue;
 
-			if (!line.isEmpty())
+			QRegularExpressionMatch dataMatch = dataRx.match(line);
+			if (dataMatch.hasMatch())
 			{
-				QRegularExpressionMatch dataMatch=dataRx.match(line);
-				if (dataMatch.hasMatch())
-				{
-					// The first entry is the DB name, the second is as we display it in the program.
-					locationDBToIANAtranslations.insert(dataMatch.captured(1).trimmed().toUtf8(), dataMatch.captured(2).trimmed().toUtf8());
-					readOk++;
-				}
+				// The first entry is the DB name, the second is as we display it in the program.
+				locationDBToIANAtranslations.insert(dataMatch.captured(1).trimmed().toUtf8(), dataMatch.captured(2).trimmed().toUtf8());
+				readOk++;
 			}
 		}
 		qDebug() << "Loaded" << readOk << "fixes for time zones";
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -411,7 +411,7 @@ StelLocationMgr::StelLocationMgr()
 	QSettings* conf = StelApp::getInstance().getSettings();
 
 	// N.B. Further missing TZ names will be printed out in the log.txt. Resolve these by adding into data/timezone.tab file.
-	loadTimeZones();
+	loadTimeZoneFixes();
 
 	loadCountries();
 	loadRegions();
@@ -1182,17 +1182,17 @@ void StelLocationMgr::loadCountries()
 	countryNameToCodeMap.insert("Taiwan (Provice of China)", "tw");
 }
 
-void StelLocationMgr::loadTimeZones()
+void StelLocationMgr::loadTimeZoneFixes()
 {
-	QString tzFilePath = StelFileMgr::findFile(tzfFileName, StelFileMgr::File);
+	QString tzFilePath = StelFileMgr::findFile(tzfFileName);
 	if (tzFilePath.isEmpty())
 	{
 		tzFilePath = StelFileMgr::findFile(tzfFileName, StelFileMgr::New);
-		// Create a default TZF (time zone fixes) file
+		// Create a default TZF (timezone fixes) file
 		QFile tzSrc(":/data/timezone.tab");
 		if (!tzSrc.copy(tzFilePath))
 		{
-			qWarning() << "Cannot copy time zones file to " + QDir::toNativeSeparators(tzFilePath);
+			qWarning() << "Cannot copy timezone fixes to " + QDir::toNativeSeparators(tzFilePath);
 			return;
 		}
 	}
@@ -1202,8 +1202,8 @@ void StelLocationMgr::loadTimeZones()
 	{
 		locationDBToIANAtranslations.clear();
 		QString line;
-		int readOk=0;
-		locationDBToIANAtranslations.insert("", "UTC");
+		locationDBToIANAtranslations.insert("", "UTC"); // a first fix
+		int readOk = 1;
 		while(!tzFile.atEnd())
 		{
 			line = QString::fromUtf8(tzFile.readLine());
@@ -1219,8 +1219,11 @@ void StelLocationMgr::loadTimeZones()
 				#endif
 
 				// The first entry is the DB name, the second is as we display it in the program.
-				locationDBToIANAtranslations.insert(list.at(0).trimmed().toLocal8Bit(), list.at(1).trimmed().toLocal8Bit());
-				readOk++;
+				if (list.count()==2) // to avoid crashes
+				{
+					locationDBToIANAtranslations.insert(list.at(0).trimmed().toUtf8(), list.at(1).trimmed().toUtf8());
+					readOk++;
+				}
 			}
 		}
 		qDebug() << "Loaded" << readOk << "fixes for time zones";
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -1195,6 +1195,9 @@ void StelLocationMgr::loadTimeZoneFixes()
 			qWarning() << "Cannot copy timezone fixes to " + QDir::toNativeSeparators(tzFilePath);
 			return;
 		}
+		QFile dest(tzFilePath);
+		// fix file permissions
+		dest.setPermissions(dest.permissions() | QFile::WriteOwner);
 	}
 	QFile tzFile(tzFilePath);
 
@@ -1204,6 +1207,7 @@ void StelLocationMgr::loadTimeZoneFixes()
 		QString line;
 		locationDBToIANAtranslations.insert("", "UTC"); // a first fix
 		int readOk = 1;
+		static const QRegularExpression dataRx("^([\\w\\/]+)\\t([\\w\\/\\+\\-:]+)\\s*(.*)$");
 		while(!tzFile.atEnd())
 		{
 			line = QString::fromUtf8(tzFile.readLine());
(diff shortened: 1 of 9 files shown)


src/core/StelLocationMgr.cpp
```diff
@@ -1246,6 +1182,41 @@ void StelLocationMgr::loadCountries()
 	countryNameToCodeMap.insert("Taiwan (Provice of China)", "tw");
 }
 
+void StelLocationMgr::loadTimeZones()
+{
+	QFile tzFile(":/data/timezone.tab");
```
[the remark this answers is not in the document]
alex-w: > In principle we had planned to do that. But now the file remains still hidden from the user, right? 

Yes

> It should be copied to the user data directory and maybe even updated semiregularly (from stellarium.org; either monthly on start or by manual "update timezones" button?). [See comment in deleted line 475]

Well, this is possible to implement

alex-w:
Ok, it’s ready

alex-w:
The time zone support in Qt is not very good (esp. in Windows) and if we want to get the better support then we need use 3rd-party library and IANA data.

gzotti:
Actually Qt uses OS resources, but TZ support on Windows is limited. I have seen interesting results in Linux (like TZ rule details applied in previous years) which were different in Windows. However, when we have a TZ entry for a city, and city changes the actual timezone at some point, this may also not help. It may be more important to find some TZ lookup per geographic coordinates, i.e. clicking on the map. Or replacing (or just extending when online) the map with Qt6.5 by the new map QML interface (with zoom etc.). However, as I am struggling with #3180, this may also need some more extensive redesign.

gzotti:
From a user perspective: what should the user see when pressing "Update time zones"? IMHO this should happen behind the screen. I have not looked into the relevant code, but what about, at startup:

```
If (unknownTimezonesDetected())
    updateTimezones()    // check file age, download new file if possible. 
If (unknownTimezonesDetected()) // still unknown?
    complainAsBefore()   // qDebug() messages.
```

The local TZ file might include an optional timestamp in line 1 to avoid online lookup at every restart. (Don't lookup again on the same date.) It should also contain a warning in the header that discourages users from editing:

`# This file is automatically retrieved on demand. Do not edit!
`

src/gui/LocationDialog.cpp
```diff
@@ -981,3 +991,125 @@ void LocationDialog::filterSitesByRegion()
...
+			qWarning() << "Updating timezone fixes are complete... The new data will be apply on next startup.";
+			updateState = LocationDialog::CompleteUpdates;
+		}
+		else
+		{
+			qWarning() << "Downloaded timezone fixes are not completed, restore the data!";
+			updateState = LocationDialog::OtherError;
+		}
+	}
+	catch (std::runtime_error &e)
```
10110111, line 1070: What is the expected source of this exception?

src/gui/LocationDialog.cpp
```diff
@@ -981,3 +991,125 @@ void LocationDialog::filterSitesByRegion()
...
+
+	int currentValue = 0;
+	int endValue = 0;
+
+	if (bytesTotal > -1 && bytesReceived <= bytesTotal)
+	{
+		//Round to the greatest possible derived unit
+		while (bytesTotal > 1024)
+		{
+			bytesReceived = static_cast<qint64>(std::floor(static_cast<double>(bytesReceived) / 1024.));
```
10110111, line 1095: Why so many explicit casts? You already divide a `qint64` by a `double`, the integer would be implicitly promoted to `double`. Then you floor it, which makes the conversion to `double` senseless, since the same result, without any need for rounding and back-conversion, could be obtained from an integer division:
```
bytesReceived /= 1024;
```

alex-w:
> Actually Qt uses OS resources, but TZ support on Windows is limited. I have seen interesting results in Linux (like TZ rule details applied in previous years) which were different in Windows. However, when we have a TZ entry for a city, and city changes the actual timezone at some point, this may also not help. 

IANA TZDB has all changes in timezones. I don't like the latest changes in TZ data (removing TZ data from main file, renaming locations, using fictional TZ for locations by political reasons instead of factical data), but this is a comprehensive timezones database.

> It may be more important to find some TZ lookup per geographic coordinates, i.e. clicking on the map. Or replacing (or just extending when online) the map with Qt6.5 by the new map QML interface (with zoom etc.). However, as I am struggling with #3180, this may also need some more extensive redesign.

Mixing Widgets and QML is a bad idea - we've already had this mix and it was removed due problems.

gzotti:
Where and when did we have QML? It's just new, and mixing should work. Likely not in an embedded dialog but separate window, though. If there was a better Map widget with nice built-in features (e.g. geodata/tz lookup), of course it would be the clearer way to go.

alex-w:
> From a user perspective: what should the user see when pressing "Update time zones"? 

Progress bar for downloading the data.

> IMHO this should happen behind the screen. I have not looked into the relevant code, but what about, at startup:
> ```
> If (unknownTimezonesDetected())
>     updateTimezones()    // check file age, download new file if possible. 
> If (unknownTimezonesDetected()) // still unknown?
>     complainAsBefore()   // qDebug() messages.
> ```

I can implement it, but this is bad idea:
 1. Moving the code from `LocationDialog` class to `StelLocationMgr` class will adding extra dependencies for all plugins, where locations are using;
 2. The time of loading the Stellarium will be increasing due downloading TZF file every startup (esp. for low speed connections or offline systems);
 
Of course we can add the code for updating the data by some frequency (every month or something else) like in pluigns, but this is will not resolve point 1 from the list above.

> The local TZ file might include an optional timestamp in line 1 to avoid online lookup at every restart. (Don't lookup again on the same date.) 

How? With standard Qt's classes for network management this is impossible, because Qt doesn't support partially loading the data. We may use HEAD request and GET request later if filesize is changed, but this is not give us guarantee for following a real changing the data.

>It should also contain a warning in the header that discourages users from editing:
> 
> `# This file is automatically retrieved on demand. Do not edit! `

OK

gzotti:
> > From a user perspective: what should the user see when pressing "Update time zones"?
> 
> Progress bar for downloading the data.

I mean, when and how should John Simple (who never reads the logfile) ever want to press this button? Just because it's "There is a button, let's see what happens!"? Pressing this button appears useful if there is a complaint about missing timezone in the logfile. But this condition can trigger the update automatically.
 
> > IMHO this should happen behind the screen. I have not looked into the relevant code, but what about, at startup:

AMENDED:

```
If (unknownTimezonesDetected() && lastUpdateLongEnoughAgo())
     updateTimezones()    // download new file if possible, save date of last retrieval. 
 If (unknownTimezonesDetected() && !lastUpdateLongEnoughAgo()) // still unknown?
     complainAsBefore()   // qDebug() messages.
 ```
> 
> I can implement it, but this is bad idea:
> 
>     1. Moving the code from `LocationDialog` class to `StelLocationMgr` class will adding extra dependencies for all plugins, where locations are using;

Why? The LocationMgr is the central authority for location-specific stuff. 
 
>     2. The time of loading the Stellarium will be increasing due downloading TZF file every startup (esp. for low speed connections or offline systems);

Only if missing timezones are detected shall there be any attempt at downloading. And of course, not at every start!
  
> Of course we can add the code for updating the data by some frequency (every month or something else) like in pluigns, but this is will not resolve point 1 from the list above.

Not even regular frequency. Update attempt shall be done on demand. (Usually after update of tzdata OS package...)

> > The local TZ file might include an optional timestamp in line 1 to avoid online lookup at every restart. (Don't lookup again on the same date.)
> 
> How? With standard Qt's classes for network management this is impossible, because Qt doesn't support partially loading the data. We may use HEAD request and GET request later if filesize is changed, but this is not give us guarantee for following a real changing the data.

Then use some other mechanism, like a datestamp in config.ini. "Don't attempt to download on the same day" (or +1 week? This gives us a few days to fix such missing timezone). The pseudo-function lastUpdateLongEnoughAgo() would test this.

alex-w:
> Where and when did we have QML? It's just new, and mixing should work. Likely not in an embedded dialog but separate window, though. If there was a better Map widget with nice built-in features (e.g. geodata/tz lookup), of course it would be the clearer way to go.

See https://github.com/Stellarium/stellarium/commit/db64b13bf203c2e4882d737c7e1ca4e25a6c8ddd

gzotti:
Ah, interesting, 2014... This was QML1. Whatever this was used for, it looked like some testing environment.

alex-w:
> I mean, when and how should John Simple (who never reads the logfile) ever want to press this button? Just because it's "There is a button, let's see what happens!"? Pressing this button appears useful if there is a complaint about missing timezone in the logfile. But this condition can trigger the update automatically.

Obviously that John Simple also never updating OS and, of course, time zones. I don't see a big problem if users will be press the button.

>  > I can implement it, but this is bad idea:
> > ```
> > 1. Moving the code from `LocationDialog` class to `StelLocationMgr` class will adding extra dependencies for all plugins, where locations are using;
> > ```
> 
> Why? The LocationMgr is the central authority for location-specific stuff.

Because this is C++. Adding `#include <QNetworkReply>` into `StelLocationMgr.hpp` file will add requirement `Qt{5|6}::Network` as dependencies for all plugins where `StelLocationMgr.hpp` is included.

> 
> > ```
> > 2. The time of loading the Stellarium will be increasing due downloading TZF file every startup (esp. for low speed connections or offline systems);
> > ```
> 
> Only if missing timezones are detected shall there be any attempt at downloading. And of course, not at every start!

Seriously? Do you think John Simple above will report about missing time zone? :) Of course, no. So, we don't have data about new missing time zone and we don't add the data into file and result is obvious - attempt to downloading the fixes at every start.
 
> > Of course we can add the code for updating the data by some frequency (every month or something else) like in pluigns, but this is will not resolve point 1 from the list above.
> 
> Not even regular frequency. Update attempt shall be done on demand. (Usually after update of tzdata OS package...)

So, you expected that user will see the log and press the button. But John Simple...

gzotti:
> > I mean, when and how should John Simple (who never reads the logfile) ever want to press this button? Just because it's "There is a button, let's see what happens!"? Pressing this button appears useful if there is a complaint about missing timezone in the logfile. But this condition can trigger the update automatically.
> 
> Obviously that John Simple also never updating OS and, of course, time zones. I don't see a big problem if users will be press the button.

TZ updates are done by the operating system. Typically this is automated today. If there are no OS updates, timezones don't change, and there will be no need for updates. 

> > > I can implement it, but this is bad idea:
> > > ```
> > > 1. Moving the code from `LocationDialog` class to `StelLocationMgr` class will adding extra dependencies for all plugins, where locations are using;
> > > ```
> > 
> > 
> > Why? The LocationMgr is the central authority for location-specific stuff.
> 
> Because this is C++. Adding `#include <QNetworkReply>` into `StelLocationMgr.hpp` file will add requirement `Qt{5|6}::Network` as dependencies for all plugins where `StelLocationMgr.hpp` is included.

Ah, header inclusion. You can probably declare "class QNetworkReply" in the header file (if at all needed), and only include the header in the cpp file. 

 
> > > ```
> > > 2. The time of loading the Stellarium will be increasing due downloading TZF file every startup (esp. for low speed connections or offline systems);
> > > ```
> > 
> > 
> > Only if missing timezones are detected shall there be any attempt at downloading. And of course, not at every start!
> 
> Seriously? Do you think John Simple above will report about missing time zone? :) Of course, no. So, we don't have data about new missing time zone and we don't add the data into file and result is obvious - attempt to downloading the fixes at every start.

Not John Simple, but those few who have reported will hopefully continue to do so. And no, with a timestamp "last-attempted-to-update" stored, there can be a few days of updating silence.

I have somehow forgotten what actually happens if tz is not found. Invalid time for those locations? Other strange behaviour? 
 
> > > Of course we can add the code for updating the data by some frequency (every month or something else) like in pluigns, but this is will not resolve point 1 from the list above.
> > 
> > 
> > Not even regular frequency. Update attempt shall be done on demand. (Usually after update of tzdata OS package...)
> 
> So, you expected that user will see the log and press the button. But John Simple...

The current situation is IMHO just not good. The test for unknown timezones can simply trigger an update action (... or just the visibility of this button! But one must open the location panel to see that in the first place.), but only if the latest update attempt was already some time ago, *preventing update attempts at every start*. Usually there will then be no unknown timezones, therefore no network activity. If needed, another config entry could even inhibit this lookup attempt, e.g. on offline installations or when network activity must be inhibited. Those offline sites will also not get tzdata updates from their OS services, so, no problems.

gzotti:
I did not intend you to kill this at 80%. We had planned such auto-update in the future when we see how the missing timezones issue develops. Just that I think this update should run without user interaction.

Issue #3206, opened 2023-04-26
Importing observe list doesn't replace default observe list


ilkant:
### Expected Behaviour
When you import observe list from computer disk and allow it to overwrite default list, it replaces it.

### Actual Behaviour
Imported observe list doesn't replace default list. So you cant see at all imported observe list.

I think that there is other similair behaviors. So may be vice to draw finite automata and design what should be done each case. Default list can be saved default name json and then replaces. So this does not forget default observations.

Imported list might appear to observe window top to drop-down menu. Now it doesn't appear.

### Steps to reproduce
1. Create new observe list
2. Export it to pisces.json (or what you want)
3. Quit Stellarium
4. Start Stellarium
5. Import pisces.json observe list to Stellarium
6. Allow pisces.json to overwrite default list
7. BUG Stellarium does not overwrite default list. So you cant see pisces.json observations at all.

### System
* Stellarium version: <Stellarium 23.1+ 23.1.113-38dbf6a [master]>
* Operating system: <Kubuntu 23.04>
* Graphics Card: <MSI GeForce GT 1030/PCIe/SSE2 2GH LF OC, OpenGL 3.3, driver 3.3.0 NVIDIA 515.105.01>
* Screen type LG 34GK950F 3440x1440

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/11328131/log.txt)

gzotti:
Seems exported list is malformatted.

gzotti:
The point was not anything around default list, but a newly-imported list was not available in the list selection.

ilkant:
Well done! Now you can save and download lists. Another comment is that it would be good if this announcement could be localized for different languages.
![image](https://github.com/Stellarium/stellarium/assets/18149564/43692f43-3d8e-49a8-b188-d87176ac6b98)

PR #3207, opened 2023-04-26
Fix observing list import/export 

gzotti:
Fix a few quirks and a real bug in ObsList import and export.

### Description

List export exports a single list. 
On import, the name of the list is checked. 

List date is date of creation, not of last edit. Maybe we should change that?

The oh-so-important OLUD is actually not tested. If a list is created, an OLUD is given. Exporting to file, editing the "live" list, storing and importing the previously exported would (I think) just overwrite the existing list with same OLUD.  Is this not what should have been avoided? What is the scenario?

Fixes 
- [x] #3206 (issue)
- [x] after loading a list, show it immediately and fill its name in the name combo
- [x] On export, let users export single lists (ending: `.sol`) or the whole list of lists `.ol`
- [x] Change timestamp from list creation date to date of last edit.
- [x] On import, make sure to test existing lists not by name and last date, but by OLUD (or all 3?)
- [ ] (other observations)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/guide/ch_tour.tex
+++ b/guide/ch_tour.tex
@@ -298,9 +298,7 @@ also section~\ref{sec:gui:configuration:tools} for more screenshot options.
 \section{Observing Lists (Bookmarks)}
 \label{sec:tour:bookmarks}
 
-You can store your favourite objects or views in observing lists. This
-feature extends the earlier bookmarks feature.
-
+You can store your favourite objects or views in observing lists. 
 Press \key{Alt+B} or \guibutton[0.75]{2.5}{bt_bookmarks.png} to call
 up the dialog.  A bookmarks file from previous versions will be
 imported and converted to the new format. You can however create an
@@ -319,14 +317,17 @@ a particular location. Likewise, the displayed landscape may be nice
 to restore. Additionally, the field of view may be relevant. To store
 these data and later retrieve them, use the respective checkboxes.
 
-You can also export the current list with the according
-button, or import such an exported list on another system. Lists are identified with an 
+You can also export the current list or all lists with the according
+button, or import such an exported list on another system. 
+To export the current list, keep the default \texttt{.sol} file type in the filename dialog.
+To export all your lists, use the \texttt{.ol} file type in the filename dialog.
+Lists are identified with an 
 Universally Unique Identifier (UUID) which we call OLUD (Observing List Unique Identifier). 
 On importing a list, existing lists with the same OLUD may be replaced after a warning. 
 During editing, storing a list with an existing name is prevented. 
 The list import may however import a list with an existing name as long as the OLUD is different. 
-To transfer all your lists to another system, copy, rename and import 
-(or just copy, to replace the existing) the full file \file{observingLists.json} 
+To transfer all your lists to another system, export them as \texttt{.ol} file, or
+copy, rename and import (or just copy, to replace the existing) the full file \file{observingLists.json} 
 from the \file{data/} subdirectory of your user directory (see section~\ref{sec:Directories}).
 
 
--- a/src/gui/ObsListDialog.cpp
+++ b/src/gui/ObsListDialog.cpp
@@ -926,8 +926,9 @@ void ObsListDialog::exportListButtonPressed()
 		{KEY_VERSION, FILE_VERSION}};
 
 	QVariantMap currentListMap={{selectedOlud, observingLists.value(selectedOlud).toMap()}};
-	QVariantMap oneListMap={{KEY_OBSERVING_LISTS, currentListMap}};
-	exportJsonMap.insert(KEY_OBSERVING_LISTS, oneListMap);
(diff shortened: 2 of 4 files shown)


alex-w:
I didn't check the code yet, but I have important notification for import/export feature. In the original idea the OL is contains many lists of objects with some description and settings, of course, the OLUD is unique identifier for each list. The import/export feature should import/export one list per file and by the fact OL should support 2 formats of files - OL (all lists) and SOL (single observing list - the format for exchange the data). By the fact currently the format of SOL file is wrong. :(

See my observing lists file and exported default list from v23.1:
[observingList.json.txt](https://github.com/Stellarium/stellarium/files/11332114/observingList.json.txt)
[observingList_default.json.txt](https://github.com/Stellarium/stellarium/files/11332116/observingList_default.json.txt)

gzotti:
The SOL/OL distinction was never discussed before January I think. Can you please edit the PR description with checkable list entries what should be done. And then please help to implement these things, esp. OL/SOL stuff. Not starting on June 15, but this weekend, finish by ~May 20 to allow real beta testing before release. I have never seriously used the lists in the new format, so some issues evaded my attention. 

I just see the danger written above that the whole OLUD idea is useless if it allows edit/export/import without giving a new OLUD. Probably a new OLUD should be given whenever content changes (basically, create a checksum from list contents: entries, name, date), so that each import can immediately see whether this list exists already of if it was changed. We could then as well use the list name as key (instead of OLUD), plus a checksum entry for that purpose of identification. (Format 3.0...)

ilkant:
I think that two timestaps to observe lists are good. Creation timestamp and last edited timestamp. Like Linux files.

alex-w:
> The SOL/OL distinction was never discussed before January I think. Can you please edit the PR description with checkable list entries what should be done. 

I should find all my record for this feature before editing checklist...

> And then please help to implement these things, esp. OL/SOL stuff. Not starting on June 15, but this weekend, finish by ~May 20 to allow real beta testing before release.

Good point

> I have never seriously used the lists in the new format, so some issues evaded my attention.

Same problem

> I just see the danger written above that the whole OLUD idea is useless if it allows edit/export/import without giving a new OLUD. Probably a new OLUD should be given whenever content changes (basically, create a checksum from list contents: entries, name, date), so that each import can immediately see whether this list exists already of if it was changed. We could then as well use the list name as key (instead of OLUD), plus a checksum entry for that purpose of identification. (Format 3.0...)

No. The OLUD should be unique within one collection of observing lists, but we've never guaranteed unique OLUD for whole world span. I hope in some time later on our website will be added new category for user contributed data - favorites/observing lists with good description and interesting data.

gzotti:
Is the OLUD ever checked? Or will importing a renamed list with same OLUD overwrite an existing list? The latter case must be avoided, else the whole idea is broken! So, at least when importing, the check must be extended to not just name, but also OLUD.

ilkant:
I have a vision: Observe list could be a set of objects. And then there can be a script which creates a star trek using the observation list. And the user can share and receive lists of observation points from others.

I have also thought that in those previous situations it is not necessary that all the information about the target has been stored in the list. However, if the data is in Stellarium data. But if that list is a list of the user's own observations, then there should be places for all the information obtained from the observation.

gzotti:
These lists retrieve data from the program. Some people seem to have interest in using these lists externally. 

Storing one's own observation notes is not intended at this stage. Years ago, an observation logger was allegedly planned, but without an actual usage scenario apparently went nowhere. I have read about a 3rd-party library/format/spec, but others may be more in actual "observing" than me.

alex-w:
> Is the OLUD ever checked? Or will importing a renamed list with same OLUD overwrite an existing list? 

Well, as far as I remember the process of importing list into collection should add new list - not just adding some elements into existing one. Of course at this step OLUD is checked and the list will not import if collection have data with same OLUD.

> The latter case must be avoided, else the whole idea is broken! So, at least when importing, the check must be extended to not just name, but also OLUD.

I fear in some chunk of refactoring (or probably someone requested it) the behaviour was changed to allow the ability of merging the lists.

gzotti:
@alex-w, Storing (exporting) the whole ObsList as .ol is identical to copying the observingList.json file, right? 
What about the file type options? Should I allow import of "*.json, *.ol, *.sol"  (do we need to support *.json for legacy, or just skip that?)  and, on export, discern "export all" and "export current list" just by the selected extension, one of "*.ol" or "*.sol"?

gzotti:
You had expressed earlier that you want OL and SOL. It was not terribly hard to do so, so now it's done. A .sol is just a one-list .ol, and .ol has the same format as the observing list flavor of what was by now  .json. 

I can export and import lists. Hopefully it is done.

alex-w (approved):
Thanks!

PR #3230, opened 2023-05-17
Discovery Circumstances for Solar System Objects

### LINKED ISSUE #3224: Discovery Date
Would it be possible to include the discovery date and possibly who discovered a selected celestial object in the information panel?

—
List added by @alex-w:
* [x] Deep-Sky Objects: M
* [ ] Deep-Sky Objects: C
* [ ] Deep-Sky Objects: NGC/IC
* [x] Solar system objects: planets 
* [x] Solar system objects: moons
* [x] Solar system objects: minor objects
* [x] Solar system objects: comets

alex-w:
### Description
This pull request added support the discovery circumstances for solar system objects (except comets, but the date of discovery included into designation of the comets).

Features:
* [x] Added discovery circumstances for major planets and their moons into `ssystem_major.ini` file
* [x] Added discovery circumstances for minor planets and comets into `ssystem_minor.ini` file
* [x] Solar System Editor plugin adds discovery circumstances for numbered minor planets now

Extra features:
* [x] Solar System Editor plugin adds IAU designations for periodical comets with old-style designations now
* [x] Search Tool support search minor planets and comets by IAU designations also

Partially fixes #3224 (issue)

### Type of change

### How Has This Been Tested?
* Run Stellarium
* Find Ceres and check the info
* Run Solar System Editor (SSE) plugin and update the data for Ceres
* Check Ceres info again
* Find Quaoar and check the info
* Run SSE plugin and update the data for Quaoar
* Check Quaoar info again
* Find 2P/Encke and check the info
* Run SSE plugin and update the data for comet Encke
* Check comet Encke info again
* Try find Quaoar and comet Encke by extra designations

## Checklist:


alex-w:
OK, it's ready to review

data/ssystem_minor.ini
```diff
@@ -947,7 +1046,8 @@ coord_func                     = comet_orbit
 dust_brightnessfactor          = 1.5
 dust_lengthfactor              = 0.4
 dust_widthfactor               = 1.5
-name                           = 2I/Borisov [C/2019 Q4 (Borisov)]
+name                           = 2I/Borisov
+provisional_designation        = C/2019 Q4
```
gzotti, line 1050: The C/2019 Q4 is not a provisional code. It is the current IAU style of a unique designation of a comet. 
See your list periodic_comet_codes.fab: What we know as 1P/Halley has indeed the discovery-based code "C/1682 Q1", although it has never been observed under this code (ok, maybe with 4m-class instruments :-) According to https://en.wikipedia.org/wiki/Halley%27s_Comet, this designation should actually change to indicate which perihel run is meant.

When I started in the field, there were provisional designations like  "1982i (Halley)". These year-small-letter codes were clearly "year/sequence of (re)discovery", and when identity with a known comet was confirmed, the known name was added. E.g., when the first comet of 1910 was discovered, it was labelled 1910a (Not sure if the YYYYx labelling was even used in the early 20th ct. or this was the label used in the early 1990s). After perihel (or sometimes later?), the final name was given as "1910 I (Johannesburg)". (Generally. YYYY of perihel and Roman sequence numbers, making it clear which apparition was discussed for a periodic comet) And now it is labelled "C/1910 A1 Great January Comet". 

Halley can be identified with e.g. 1P/1909 R1, 1910 II, 1909c and 1P/1982 U1, 1986 III, 1982i (see WP). The first labels are the currently favoured style of discovery year/half-month. The second was the traditional numerical order of perihel dates. Third the provisional discovery code. 

Maybe rename this entry to "date_code". This would not necessarily mean it was labelled like this just after discovery (this would be 1982i for Halley). provisional_designation for me would read as "label used when it was first discovered and news were circulated." This could be an optional entry when labelling style has been changed. I know, this is a bit messy. If we want to show all flavours (which I would like to have!), Infotext name could look like

1P/Halley (P/1982 U1 - 1982i - 1986 III)

For this, we'd need additional optional entries "perihel_code=1986 III" and "discovery_code=1982i".

So, to sum it up, please extend this list to 

date_code ... current date-based code like "P/1982 U1" or (for Steins above) "1969 VC"
perihel_code ... used for comets during most of the 19/20th centuries, e.g. "1986 III" 
discovery_code ... provisional labelling made immediately after discovery, e.g. "1982i".

alex-w: I use `iau_designation` field for minor planets and comets now. I think this is enough for current task :)

alex-w: I'll use data from Kronk's _Cometography_ to add discovered/recovered dates and _perihel_code_/_discovery_code_ for comets in our `ssystem_minor.ini` file. Probably on next iteration we should update `ssystem_1000comets.ini` file by data from _Cometography_.

src/core/modules/SolarSystem.cpp
```diff
@@ -1621,22 +1633,52 @@ void SolarSystem::fillEphemerisDates()
 	}
 }
 
+// a wrapper for getting an IAU designations of minor bodies
+QString SolarSystem::getIAUDesignation(PlanetP minorBody) const
+{
+	QString iau;
+	if (minorBody->getPlanetType()==Planet::isComet)
```
gzotti, line 1640: This would invite implementing a virtual method 
`Planet::getIAUDesignation() const {return englishName;}`

which is overridden in MinorPlanet and Comet classes. Then this method is trivially:

```
QString SolarSystem::getIAUDesignation(PlanetP p) const
{ 
    return p->getIAUDesignation();
}
```

gzotti: It could also simplify the methods below.

src/core/modules/MinorPlanet.hpp
```diff
@@ -90,17 +88,20 @@ class MinorPlanet : public Planet
 	//! @param slope Slope parameter G. This is usually [0..1], sometimes slightly outside. Allowed here [-1..2].
 	void setAbsoluteMagnitudeAndSlope(const float magnitude, const float slope);
 
-	//! renders the subscript in a minor planet provisional designation with HTML.
-	//! \returns an empty string if the source string is not a provisional
-	//! designation.
-	static QString renderProvisionalDesignationinHtml(QString plainText);
+	//! renders the subscript in a minor planet IAU provisional designation with HTML.
+	static QString renderIAUDesignationinHtml(QString plainText);
```
gzotti, line 92: please use string references (QString &plainText) [everywhere]

gzotti:
Thanks for including the extra comet labels! This is great for reviewing 20th century reports. However, a few observations:

- planet InfoText now appears like "Jupiter (Jupiter)". I think, if bracketed phrase is equal to first name, there is no need to show it.
- After a comet, there is an empty bracket "P/Halley (1982i=1986III)()". And comets miss the serial number "1P" This should probably appear as "1P/Halley (1982 U1=1982i=1986III)" ... Ah, my error. The data in the 1000comets file include the extras, but no 1P, in the name entry, and not yet in the added fields. However, the empty bracket should not be displayed if the extra names don't exist.

src/core/modules/MinorPlanet.hpp
alex-w: Why?

10110111: I suppose Georg meant using `const` references, rather than what he wrote. A small advantage of this is avoidance of the atomic check-and-increment of the refcount inside the Qt's implicit sharing mechanism (I doubt this would make any measurable difference in this case, but it makes sense to use const reference by default in such situations).

gzotti: Indeed, `(const QString &arg)` is even clearer. See e.g. https://www.ibm.com/docs/en/zos/2.4.0?topic=calls-pass-by-reference-c-only
> Pass-by-references is more efficient than pass-by-value, because it does not copy the arguments.

This may not mean a lot for simple types (int, double, ...) but seems recommended for QStrings and other complex types. I try to update these things as I find them.

data/ssystem_minor.ini
```diff
@@ -448,6 +502,7 @@ absolute_magnitude             = 22.08
 albedo                         = 0.07
 coord_func                     = comet_orbit
 name                           = 1I/2017 U1 (`Oumuamua)
+iau_designation                = A/2017 U1
```
gzotti, line 505: This could be used as discovery_code. The official IAU designation is 1I/... Probably

name=`Oumuamua
iau_designation=1I/2017 U1
discovery_code=A/2017 U1

data/ssystem_minor.ini
```diff
@@ -947,7 +1048,8 @@ coord_func                     = comet_orbit
 dust_brightnessfactor          = 1.5
 dust_lengthfactor              = 0.4
 dust_widthfactor               = 1.5
-name                           = 2I/Borisov [C/2019 Q4 (Borisov)]
+name                           = 2I/Borisov
```
gzotti, line 1051: Also here:

name=Borisov
iau_designation=2I/2019 Q4
discovery_code=C/2019 Q4

data/ssystem_minor.ini
```diff
@@ -1799,7 +1992,10 @@ dust_brightnessfactor          = 1.5
 dust_lengthfactor              = 0.4
 dust_widthfactor               = 1.5
 model                          = 1682q1halley_MLfix.obj
-name                           = P/Halley (1982i=1986III)
+name                           = Comet Halley
+date_code                      = 1P/1982 U1
```
gzotti, line 1996: This should now be iau_designation

data/ssystem_minor.ini
```diff
@@ -1822,6 +2018,7 @@ dust_lengthfactor              = 0.4
 dust_widthfactor               = 1.5
 model                          = 1682q1halley_MLfix.obj
 name                           = 1P/Halley (2061)
+iau_designation                = P/1682 Q1
```
gzotti, line 2021: I am unsure about using the 1682 label to the future apparition. Do recurring comets keep the label of earlier apparition? I know, WP is not the most reliable source, but they provide a list of new designations for 1P/Halley in all known apparitions.

src/core/modules/SolarSystem.cpp
```diff
@@ -1059,12 +1066,27 @@ bool SolarSystem::loadPlanets(const QString& filePath)
...
-			{
+			if (magnitude > -99.f)
 					mp->setAbsoluteMagnitudeAndSlope(magnitude, slope);
-			}
+
+			QString iauDesignation = pd.value(secname+"/iau_designation", "").toString();
+			if (!iauDesignation.isEmpty())
+				mp->setIAUDesignation(iauDesignation);
+			// order of codes: date_code [P/1982 U1] - perihelion_code [1986 III] - discovery_code [1982i]
+			QStringList codes = { pd.value(secname+"/date_code", "").toString(),
```
gzotti, line 1078: In light of the recent updates, it seems date_code is now redundant (?). iau_designation looks like P/1982 U1, same as date_code. 

The only (minor) issue now is the collision of "P/1982 U1" and "name=1P/Halley". I am unsure whether this would better be written "1P/1982 U1 Halley". However, "1P/Halley (P/1982 U1 -  1982i - 1986III)" is perfectly understandable.

data/ssystem_minor.ini
alex-w: `discovery_code` is used for comets only

src/core/modules/SolarSystem.cpp
alex-w: `date_code` will be empty almost always, but in some cases the supporting this parameter will be important for periodical comets with standard designations XXP.

data/ssystem_minor.ini
alex-w: Well, according MPC the comet _1P/Halley_ has IAU designation _P/1682 Q1_, so, the designation _1P/1982 U1_ should be added as `date_code`

data/ssystem_minor.ini
alex-w: We really need development the rule for adding the extra parameters for periodic comets, such, `*_designation`, ` *_code` and `discover*` because we should follow historical data in the one hand and resolve messy issues in the other hand. Plus maybe we should all some extra parameters for comets, like `last_seen` for unperiodical comets.

I hope I'll find the monography _Физические характеристики комет_ and all appticaltions for it and all volumes of _Cometography_ to adding missing data and fix mistakes. Anyone can help me find these books?

data/ssystem_minor.ini
alex-w: The `discovery_code`, `date_code` and `perihelion_code` supporting by comets only. Probably I should add support of them for all minor bodies (at least for `A/` and `I/` objects).

This object hasn't proper name, so, `name` should follow IAU (provisional or standard) designation.

data/ssystem_minor.ini
gzotti: Ah, OK. So my fix for all Halley entries in 1000_comets.ini is incomplete/wrong, I should replace iau_designation by date_code. But then this entry still needs iau_designation=1P/1682 Q1.

data/ssystem_minor.ini
gzotti: last_seen can only be useful for periodical comets which were lost and now should carry D/  labels ;-) If really the first "modern" discovery code (YYYY H1) is used for periodical comets, we need to specify which apparition we are describing. For now, we do this as part of the name (e.g. "1P/Halley (2061)". But who then specified the neo-historical Halley date codes?  Of course, the apparition is coded in the date of perihel, but humans rarely decode JD directly.

src/core/modules/SolarSystem.cpp
gzotti: That means, the complete infoString label according to MPC standards and our aim to protect existing labels should be "1P/1682 U1 Halley (P/1982 U1 - 1982i - 1986III)"? And the screen label "1P/1682 U1 Halley"?
 
Ouch, this reads so wrong! It would make life easier to read the label as "Halley in its apparition that started in 1982". 

But I agree, we must show the standardized name first.

alex-w: No, `1P/Halley (P/1982 U1 - 1982i - 1986III)`

alex-w: Hmm… Kronk use `1P/1682 Q1` style, MPC use `P/1682 Q1` style

alex-w: https://pds-smallbodies.astro.umd.edu/data_sb/resources/designation_formats.shtml

gzotti:
The valid range (orbit_good) is only a rough subjective estimate where positions are "probably good enough to show the object's location". There is no formula for this (or is there?). It depends on gravitational interaction with the major planets and (for comets) other perturbations, and especially on users' expectations for accuracy. For high-accuracy demands, you can also say "elements 1 week old are too old". For NEAs during flyby, JPL/Horizon can deliver hourly elements.   For 1-time comets it may generally be enough to have some time window around perihel (with an epoch also close to perihel) while it is bright and interesting enough to be observed with small instruments. (mag<12??)

alex-w:
> The valid range (orbit_good) is only a rough subjective estimate where positions are "probably good enough to show the object's location". There is no formula for this (or is there?). It depends on gravitational interaction with the major planets and (for comets) other perturbations, and especially on users' expectations for accuracy. For high-accuracy demands, you can also say "elements 1 week old are too old". For NEAs during flyby, JPL/Horizon can deliver hourly elements. For 1-time comets it may generally be enough to have some time window around perihel (with an epoch also close to perihel) while it is bright and interesting enough to be observed with small instruments. (mag<12??)

I know it, but when infobox just had notification "orbital elements are outdated" without any dates then this is hard to understanding how much orbital elements are outdated (without reading the config data). Probably displaying the epoch of orbital data for minor bodies may be helpful for users.

gzotti:
Maybe more general: "Elements for epoch YYYY-MM-DD probably outdated. Consider updating data!" 
Then we can leave it to the user to decide whether 600-day old elements are still good enough. (I hope users will not just try to update Stellarium, but the data file!)

Speaking of updating: When installing a new version of Stellarium, ssystem_minor.ini will be kept, including old and possibly outdated data. Maybe add an update option at first program start (do we see if a new version has been installed? Yes, if config.ini:main/version mismatches!) to "Replace minor bodies data with default file? [Y/N/Never]". If yes, copy file from program location. If "Never", add another flag "main/restore_sso_defaults=false" to config.ini to suppress this dialog...

Another problem with the new addition is really the screen space. When displaying everything, even on my large screen I have strings overlapping the left menu bar. We should try to not lightheartedly produce additional lines. 

This is also true for the name! IMO there is enough space on even a 4:3 screen for the name and all designations in the top line.

alex-w:
Damn! You're speaking about installing new version of Stellarium as upgrade an existing one. See issue #3240

alex-w:
OK, it's ready to review. Some data is missing, but I haven’t the full books at the moment.

alex-w:
OK, the pull request ready to review.

P.S. The question to updating the default list of comets has been moved into separate issue - see #3261

gzotti (approved):
The data are enormous! I have not seen an issue so far, and cannot test for a week. Probably OK for beta testing.

PR #3254, opened 2023-06-01
Fix shapes of ocular frames

### LINKED ISSUE #3249: Incorrect ocular rectangle for a photo lens
### Expected Behaviour

When choosing focal length of the lens and the sensor corresponding to my camera, and choosing perspective projection (which is what most non-fisheye lenses try to simulate), the red rectangle that's turned on by the _Image sensor frame_ button should be roughly the same as the rectangle of the photo I'd take with my camera, up to distortions of the lens.

### Actual Behaviour

* A change of projection doesn't change the shape of the rectangle, this is already suspicious.
* Aspect ratio of the red rectangle differs a lot from aspect ratio of my photos.
* Angle per pixel figures in the bottom-right part of the rectangle are different for X and Y direction, which doesn't make sense with my usual isotropic camera sensor.

### Steps to reproduce

* Make a photo with an APS-C camera like Canon 80D and e.g. a 10mm lens.
* Try to position the red frame in Stellarium so that the stars in the photo are situated similarly to the stars in Stellarium.
* Notice the discrepancy that can't be fixed by panning around.
* Note that the camera has square pixels (which you can check by taking a picture of a square) and a resolution of 6000×4000 implying aspect ratio of 1.5, but the rectangle in Stellarium has aspect ratio ~1.3.

### System
* Stellarium version: 23.1.126-17cbd6b

10110111:
### Description

Now sensor frame, crop overlay, crop overlay grid, OAG frame its inner and outer circles, and the focuser circles — all are computed in terms of view direction, and correctly projected according to the current projection. This results in non-rectangular/non-circular shapes in projections other than perspective, which is to be expected, and is a good thing.

The problem remains with positioning of the text, and here I'd like to hear some opinions on how it should be fixed: should I try to rotate it along the tangent to the curve that represents the frame (see the last screenshot), or maybe better to just move it closer to the curve?

Fixes #3249 

#### Perspective projection, old

stellarium-010

#### Perspective projection, new

stellarium-011

#### Stereographic projection, old

stellarium-008

#### Stereographic projection, new

stellarium-009

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/plugins/Oculars/src/CCD.cpp
+++ b/plugins/Oculars/src/CCD.cpp
@@ -250,6 +250,22 @@ double CCD::getActualFOVy(const Telescope *telescope, const Lens *lens) const
 	return fovY;
 }
 
+double CCD::getCentralAngularResolutionX(const Telescope *telescope, const Lens *lens) const
+{
+	const auto lensMultipler = lens ? lens->getMultipler() : 1.0;
+	const auto pixelSize = chipWidth() * binningX() / resolutionX();
+	const auto focalLen = telescope->focalLength() * lensMultipler;
+	return RADIAN_TO_DEGREES * 2 * std::atan(pixelSize / (2 * focalLen));
+}
+
+double CCD::getCentralAngularResolutionY(const Telescope *telescope, const Lens *lens) const
+{
+	const auto lensMultipler = lens ? lens->getMultipler() : 1.0;
+	const auto pixelSize = chipHeight() * binningY() / resolutionY();
+	const auto focalLength = telescope->focalLength() * lensMultipler;
+	return RADIAN_TO_DEGREES * 2 * std::atan(pixelSize / (2 * focalLength));
+}
+
 double CCD::getFocuserFOV(const Telescope *telescope, const Lens *lens, double focuserSize) const
 {
 	// note: focuser size in inches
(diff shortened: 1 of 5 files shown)


gzotti:
Nice work!

The text placement should be visually guided by the frame corners. The easiest in the stereographic example may be to use a boundingrect for the pincushion shape and show text along its frame. Not sure how a tangential but straight orientation would look.  Of course, following the arc of the frame would be extremely elegant. But I am not sure how weird a frame rectangle may look in cylindrical or Hammer-Aitoff projections, so this may become complicated.

10110111:
> did you comparing new frame and photo from the report?

Yes, and at least the frame now has correct shape. The scale is not ideal, but this may be due to unspecified way lens distortions are corrected by the camera, so that we don't know the exact effective focal length. At least theoretically, the frame rendering now works correctly, I checked this using Cursor Coordinates plugin to check the fields of view.

10110111:
OK, I've now fixed the positioning of the text around the sensor frame. Note that "Max exposure" label is now moved to top-right position, because top row has two lines, with both being occupied on the left side, but only one on the right side. This way the overall look becomes a bit more symmetric.

Screenshots:

stellarium-010

stellarium-011

alex-w (approved):
Thanks!

gzotti:
If the telescope (photo lens) has equatorial mount set, the labels are now off. 
stellarium-083

10110111:
Should be fixed now.

PR #3260, opened 2023-06-03
Fix orientation of sensor rectangle in equatorial mode

10110111:
Now we get rid of the concept of polar angle inside Oculars::paintCCDBounds(), instead the correct orientation is automatically defined by the projector corresponding to the mount of the current telescope.

This should fix the wrong orientation of the rectangle in equatorial mode. Please check that it works in all relevant modes (some of which I might be unaware of).

### DIFF
--- a/plugins/Oculars/src/Oculars.cpp
+++ b/plugins/Oculars/src/Oculars.cpp
@@ -1529,12 +1529,12 @@ bool Oculars::isBinocularDefined()
 	return binocularFound;
 }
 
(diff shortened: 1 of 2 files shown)


alex-w:
The orientation is fixed, but the text in corners has density issue now on HiDPI devices

stellarium-002

gzotti (approved):
Looks good for me, thanks. My approval depends on HiDPI fixes confirmed by @alex-w, of course.

10110111:
> My approval depends on HiDPI fixes

Well, this is a separate problem orthogonal to this PR.

PR #3262, opened 2023-06-03
Fix positioning of sensor labels with high-DPI scaling on

10110111:
It includes a kludge that was necessary on my system (Ubuntu 20.04 Qt 6.4.1). Please check that it works on your systems correctly not only with integral scaling, but also with fractional scaling like 150%.


alex-w (approved):
Thanks for quick fix!

PR #2891, opened 2022-12-04
Update to INDI 2.0.0

ckuethe:
### Description
Update to INDI 1.9.9 (current version) when building from source which handles remote server shut down a lot better.

TelescopeClientINDI itself doesn't require any change to link against 
INDI 1.9.9 either using `libindi-dev 1.9.9` from the INDI PPA, or compiling it as part of the stellarium build.

Fixes #2864 

### Screenshots (if appropriate):
Here's the log output of me connecting to my stellarmate, and then shutting down INDI remotely... with Stellarium not crashing!

```
...
Loaded plugin "TelescopeControl"
[TelescopeControl] Only embedded telescope servers are available.
[TelescopeControl] Adding device model: "Meade AutoStar compatible" "Any telescope or telescope mount compatible with Meade's AutoStar controller." "TelescopeServerLx200" 500000
[TelescopeControl] Adding device model: "Meade LX200 (compatible)" "Any telescope or telescope mount compatible with Meade LX200." "TelescopeServerLx200" 500000
[TelescopeControl] Adding device model: "Meade ETX70 (#494 Autostar, #506 CCS)" "Meade's ETX70 with the #494 Autostar controller and the #506 Connector Cable Set." "TelescopeServerLx200" 1500000
[TelescopeControl] Adding device model: "Losmandy G-11" "Losmandy's G-11 telescope mount." "TelescopeServerLx200" 500000
[TelescopeControl] Adding device model: "Wildcard Innovations Argo Navis (Meade mode)" "Wildcard Innovations' Argo Navis DTC in Meade LX200 emulation mode." "TelescopeServerLx200" 500000
[TelescopeControl] Adding device model: "Celestron NexStar (compatible)" "Any telescope or telescope mount compatible with Celestron NexStar." "TelescopeServerNexStar" 500000
[TelescopeControl] Adding device model: "Sky-Watcher SynScan (version 3 or later)" "Any Sky-Watcher mount that uses version 3 or later of the SynScan hand controller." "TelescopeServerNexStar" 500000
[TelescopeControl] Adding device model: "Sky-Watcher SynScan AZ GOTO" "The Sky-Watcher SynScan AZ GOTO mount used in a number of telescope models." "TelescopeServerNexStar" 500000
connectionType: TelescopeControl::ConnectionINDI  initString: "stellarmate:INDI:J2000:stellarmate:7624:EQMod Mount"
[...]

PR #3269, opened 2023-06-13
Port code to INDI 2.0.2 and update CPM

paolostivanin:
This PR is based on the work done by some people at https://github.com/Stellarium/stellarium/pull/2891. In particular:
* patch created by @panicgh was applied as is, since it worked fine
* the code created by @ckuethe was updated to work with v2.0 since it was not compiling

What was tested:
* compilation using the OS' INDI 2.0.2 works fine
* compilation using the CPM's INDI 2.0.2 works fine
* the TelescopeControl plugin could be loaded and connected to a INDIserver instance using Simulators.

What was **not** tested:
* connection to real hardware
* real hardware working properly


gzotti:
Looks interesting, thank you! Anybody with real hardware willing to test?

10110111:
> the TelescopeControl plugin could be loaded and connected to a INDIserver instance using Simulators

IIRC there were some problems only present in Qt6 but not Qt5. Did you check this?

alex-w:
macOS:
```
[  4%] Building CXX object plugins/TelescopeControl/src/INDI/CMakeFiles/TelescopeControl_INDI.dir/TelescopeControl_INDI_autogen/mocs_compilation.cpp.o
In file included from /Users/aw/stellarium/builds/macosx/plugins/TelescopeControl/src/INDI/TelescopeControl_INDI_autogen/mocs_compilation.cpp:2:
In file included from /Users/aw/stellarium/builds/macosx/plugins/TelescopeControl/src/INDI/TelescopeControl_INDI_autogen/EWIEGA46WW/moc_INDIConnection.cpp:9:
/Users/aw/stellarium/builds/macosx/plugins/TelescopeControl/src/INDI/TelescopeControl_INDI_autogen/EWIEGA46WW/../../../../../../../../plugins/TelescopeControl/src/INDI/INDIConnection.hpp:23:10: fatal error: 'libindi/baseclient.h' file not found
#include "libindi/baseclient.h"
         ^~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [plugins/TelescopeControl/src/INDI/CMakeFiles/TelescopeControl_INDI.dir/TelescopeControl_INDI_autogen/mocs_compilation.cpp.o] Error 1
make[1]: *** [plugins/TelescopeControl/src/INDI/CMakeFiles/TelescopeControl_INDI.dir/all] Error 2
make: *** [all] Error 2
Stellarium.app: No such file or directory

```

paolostivanin:
@alex-w thanks for checking macos! I'm gonna look into it

paolostivanin:
That's weird though, I can't reproduce that issue locally :thinking:

alex-w:
> That's weird though, I can't reproduce that issue locally 🤔

Probably you have locally installed INDI

paolostivanin:
yep I do have INDI installed, but I'm running cmake with `-DPREFER_SYSTEM_INDILIB=OFF` and cmake is executing the CPM action while generating stuff.
I'll try on a vm

panicgh:
Thanks for following this up. I am super busy at the moment and got stuck in the CMake porting part.

I wonder if it is really necessary to build INDI through `CPMAddPackage` and then repeating parts of INDI's CMakeFile (including version, so-version, dependencies between some files etc.) and only building selected parts of INDI. Wouldn't it be simpler to either build full INDI as CMake external project or not build it at all and simply require a system-provided INDI? It appears to me that the current approach makes it unnecessarily hard to port to new INDI versions.

paolostivanin:
I agree that bundling INDI is, IMHO, not the best solution.
I also think that we should treat INDI like any other dep: is it installed? Yes, then build the plugin. No? Then don't build the plugin.

gzotti:
If the existence of compatible versions is guaranteed by simple dependencies, and we don't change anything in the INDI code, this would be OK for me.

10110111:
Wouldn't this go against the practice used for `QXlsx` and `CalcMySky`, which were also once optional but are now CPM-required?

gzotti:
If INDI is available as standard package from default repositories just like Qt stuff, and people may have installed them already for other programs, it would be preferred over CPM and private compilations. But I had understood our first adoption of INDI had some non-standard edits in the INDI 0.* sources, so required a private build. The other packages for QXlsx and CMS are also still optional, but there are no apt packages from Debian, right?

10110111:
> there are no apt packages from Debian, right?

Right.

BTW I just noticed that in Ubuntu, even in 23.04 and the development version Stellarium still is 0.22.2. Do you happen to know why it's so old?

gzotti:
No, I don't know how the official Ubuntu maintainers upgrade their packages. We provide an Ubuntu repo with the latest version, see Guide 2.3.3. If INDI does the same, it should be fine for us. But CPM may be the solution if INDI is not available everywhere.

paolostivanin:
Hello :smile: do we have a final decision on what to do with INDI? Treat it as a dep or bundle it?

alex-w:
INDI client within Telescope Control plugin expected to all platforms, so, e.g. Windows user can connect direct to INDI server onto Linux. The current branch has few `#include` directives to building the INDI-client, but when INDI is not installed locally, then headers are missing (INDI2 has other locations for these files) - the code of Telescope Control plugin should be updates to reflect the changes paths in INDI.

paolostivanin:
I guess we better keep CPM then. Otherwise Windows support could be an issue :thinking: 
I'll update the PR

gzotti:
Any news? It seems other projects are expecting Stellarium finally moving to INDI 2.* 'really soon'.

gzotti:
Any news here?

hmartinez82:
I'm working on getting Stellarium running on Windows on ARM. I already got it working without CalcMySky and without the Telescope plugin (due to INDI). But I just packaged CalcMySky and I'm going to package INDI, but it will have to be the latest INDI (2.0.4 as of this writing) so support for INDI 2.x should happen soon™️ .

alex-w:
Any progress?

hmartinez82:
Well since I got Stellarium building for Windows on Arm with the current INDI I did now pursue this any further. It was not my main objective.

 I don't know this library nor how it's used by Stellarium so someone who knows this stuff should do this upgrade

panicgh:
As already mentioned four months ago: if we didn't try to build INDI from Stellarium but treat it like any other dependency, the issue would have been long solved. Some distros, e.g. NixOS, apply the patch for new INDI support and just use the system-provided INDI.

gzotti:
If we can depend on optionally system-installed stock INDI2 (build yourself where required), this would be the best solution IMO. When users don't need telescope control, there is no need for INDI either. That would make it a "recommended" package for apt-get or similar systems.  We still need to know what "our own copy" has changed, and a maintainer who occasionally cares for INDI updates as years go by.

gzotti:
- the pull request has conflicts
- From @panicgh 's wording ("we") I had expected final changes, build and usage instructions to get that working on all platforms, etc., for people not familiar with the internals of INDI. This must come from someone with a goto telescope which then must be sure to work. 
- seems too late for tomorrow's 23.4, but let's target it for 24.1

gzotti:
Anybody using INDI and interested to move forward?

gzotti:
Is anybody out there using Stellarium with INDI interested in going forward?
Is anybody out there on Windows using Stellarium with some INDI server on the local network? 
Is there a header package for Windows available from INDI?
Should we ask in some INDI forum for multiplatform guidance?

carlosal1015:
Arch Linux users would like using INDI, for Windows cases I do not know.

gzotti:
Yes, there are people who want to USE it. Are there users who want to use this part of the program and are also capable of MAKING it work? Or, given that it somehow works, still improve what's wrong with the current situation.  If INDI is unsupported in Windows (which I don't know) we can also exclude it on this platform.

alex-w:
Tried testing on INDI mount simulator - it can be connected/disconnected, but not targeting

alex-w:
OK, so, 4 subtasks are open at the moment: 
- restoring unit testing 
- building INDIclient in FreeBSD
- restoring speed chooser for directional movements in INDIclient
- movement of telescope circles onto sky.

alex-w:
Well, the PR is ready to review and merge almost - I don’t understand why marker of telescope is not showed a targeting to the selected object or position. Is it problem for INDI only or for any telescope?

alex-w:
> Well, the PR is ready to review and merge almost - I don’t understand why marker of telescope is not showed a targeting to the selected object or position. Is it problem for INDI only or for any telescope?

Hmm… I can check the problem in master with INDI 1.8.5 (why I didn’t think about this possibility before?)

alex-w:
OK, the problem in current port of INDI

PR #3283, opened 2023-06-23
Fixed prism rotation

alex-w:
### Description
This pull request fixed issue for prism rotation in Oculars plugin.

Fixes #3281 (issue)

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #3284, opened 2023-06-23
Specify tab size explicitly in coding style document

10110111:
Due to a recent confusion in #3283, I found the need to explicitly specify tab size in the coding style. The statement that using tabs lets developers choose their favorite tab size is wrong whenever we find the need to align lines of code. So this patch alters the document.

Alternatively, while the alignment in the current code base is messed up (due to me in particular, since I never expected 8-char tabs to be used here), we might choose to set e.g. 4 chars per tab if everyone agrees.

### DIFF
--- a/doc/codingConventions.doxygen
+++ b/doc/codingConventions.doxygen
@@ -59,8 +59,7 @@ Settings for <a href="http://doc.qt.io/qtcreator/index.html">QtCreator</a> IDE y
 #define MIN_WIDTH 3
 static const QString VERSION = "0.10.1";
  @endcode
- - Indentation should be done with tabs, not spaces. This allows developers
- to use their favorite indent size without changing the code. 
+ - Indentation should be done with tabs, not spaces. Tab size is 8 characters.
  - When wrapping lines from long function calls, where the wrapped line does not start at the same level of indentation as the start of the function call, tab up to the start of the function call, and then use spaces to the opening parenthesis.
  @verbatim
  [--tab-][--tab-][--tab-]someFunction(param, ...


gzotti:
There is a code style file for QtCreator which was mandatory since 2008 or earlier. I think it sets some tab width which should of course match this style file which is completely unknown to me. However, I have frequently observed indentation mismatches between AW and myself, but assumed different interpretations between Win and Linux, and either ignored that or fixed by replacing with spaces. I would prefer not to use tabs, just spaces, just because of that perennial tab with nonsense.

10110111:
> There is a code style file for QtCreator which was mandatory since 2008 or earlier. I think it sets some tab width which should of course match

It sets 8 characters, which this PR now mentions in the human-readable coding style document. XML file is not what a web search for "stellarium coding style" yields.

> this style file which is completely unknown to me.

This file is published at https://stellarium.org/doc/0.22/codingStyle.html (not sure if it's the latest published version, this is what Google suggested).

> I would prefer not to use tabs, just spaces, just because of that perennial tab with nonsense.

Me too, but this would mean reformatting the whole code base—and change the code style.

gzotti:
Maybe there is a mismatch between this file (see git log for updating frequency...) and what has been used in practice in QtCreator's file (which is linked in BUILDING.md and which has been considered mandatory for all (2) team members since 2012). It is always good to weed out inconsistencies, however both of us have used the IDE file and may not have updated that documentation. There is even a chance that tab with is set to 8 after all, and there is still an inconsistency Win/Linux that I have experienced. The important thing is to use this style file in QtCreator consistently. In case we change anything, there is still no need to reformat everything ASAP.

10110111:
So, this is what we have in the XML file:
```
   <value type="bool" key="SpacesForTabs">false</value>
   <value type="int" key="TabSize">8</value>
```
And this PR just makes the Doxygen document say the same. Shall I merge it?

alex-w (approved):

alex-w:
I don't remember why we use 8 chars for tab size - probably for better visual style within QtCreator, but for me 4 chars will be OK too if it will OK for all others.

gzotti (approved):
OK for me. And please let's keep indentation at 8 charwidth per tab.

10110111:
Funny, I only now noticed a peculiarity in the coding style document:

> When wrapping lines from long function calls, where the wrapped line does not start at the same level of indentation as the start of the function call, tab up to the start of the function call, and then use spaces to the opening parenthesis.
> ```
>  [--tab-][--tab-][--tab-]someFunction(param, ...
>  [--tab-][--tab-][--tab-][  spaces   ]moreparams, ...);
> ```
> This method will handle different tab widths gracefully.

So this is how it was intended to let the code be tab-size-independent. But from my experience, this is not used by any of the developers, which is what led to the original confusion. Instead, QtCreator was assumed to know better, and it only uses spaces to adjust indentation for less than tab size, not for the whole duration of non-whitespace code in a line.

So should we remove this paragraph?

gzotti:
Ah, that sounds reasonable. Just that devs have to be (made) aware of this. Is there any way to convince the QtCreator XML style file to adhere to this rule?

10110111:
> Is there any way to convince the QtCreator XML style file to adhere to this rule?

Most likely no. This rule sounds like something atypical. Qt Creator *Options*→*Text Editor*→*Behavior*→*Align continuation lines* only contains:

* Not At All,
* With Spaces,
* With Regular Indent.

So you'll have to manually replace tabs with spaces in part of the line. Sounds easy, right? ;)

gzotti:
Is "With spaces" useful? Or does it start the spaces in column 1?

10110111:
Hmm, you made me doubt my assumptions, so I did an experiment. It appears that indentation with "Tabs Only" and continuation "With Spaces" appears to follow exactly this:

Screenshot_2023-07-08_15-32-17

So, well, if we do go this route, then this PR will have to be reverted.

10110111:
This setting corresponds to the following change:
```diff
--- stellarium-ide.xml  2023-06-22 09:23:03.000000000 +0400
+++ stellarium-ide-edited.xml   2023-07-08 15:51:32.261119205 +0400
@@ -27,7 +27,7 @@
    <value type="int" key="IndentSize">8</value>
    <value type="bool" key="IndentStatementsRelativeToSwitchLabels">true</value>
    <value type="bool" key="IndentSwitchLabels">true</value>
-   <value type="int" key="PaddingMode">2</value>
+   <value type="int" key="PaddingMode">1</value>
    <value type="bool" key="SpacesForTabs">false</value>
    <value type="int" key="TabSize">8</value>
   </valuemap>
```

10110111:
I didn't find a reliably working plugin for Vim that would implement this rule though. Tried SmartTabs and stabs, they don't always produce the correct tab-spaces pattern.

alex-w:
Maybe the time for configuring and using the tool for [Beautifying Source Code](https://doc.qt.io/qtcreator/creator-beautifier.html) is come, like clang-format?

10110111:
I'd not trust it: it's likely to break some things, e.g. raw string literals, and maybe some other aligned parts of code (e.g. when we align sides around the `=` operator).

Issue #3006, opened 2023-01-19
Telescope control via RS232 USB broken


Baudouin0:
### Expected Behaviour
Connects to telescope
### Actual Behaviour
Does not connect. Shows as COM4 on GUI connect to skywatcher syncscan controller. Does not connect.
Works fine if you roll back to last version 1,1 and 0,22.2

### System
* Stellarium version: 1.2 latest
* Operating system: Windows 64 bit
* Graphics Card:  NVidia
* Screen type (if applicable): HD

gzotti:
Duplicate of #2934?

gzotti:
Did it work with a Qt6-based 1.1? I wonder whether we have yet another new error between 1.1 and 1.2. Reportedly a Qt5-based build of 1.2 works.

TonyBhargav:
hi @gzotti  , I would like to work on this issue as a part of academic course work ,It would be a great opportunity for me to learn about the project architecture and design as well. Could you please assign it to me

gzotti:
Thanks. Please have a look into all these reports since September around telescope plugin, Qt6 and failed connections. It may be just a small thing, but I think you need the actual hardware to test and fix this. 

Start with cloning our repo, creating a branch, build with Qt5.12 and Qt6 (6.4 or later), and then you can go after fixing the issues.

TonyBhargav:
Sure, i will look into it .
Thank you

Baudouin0:
Cheers for looking into this. You can test just with the controller EQ6 handset only without the mount, that how I regression tested.



Charles Baudouin
Chief Technical Architect

Email: ***@***.******@***.***>
Mobile / WhatsApp: +44 7973 146 671
Telephone: +44 1489 884 430
Skype / Teams: ***@***.***


DOMO TACTICAL COMMUNICATIONS (DTC) LTD TRADING AS DTC
Registered office Fusion 2, 1100 Parkway, Whiteley, Hampshire, PO15 7AB, United Kingdom
Registered in England and Wales, company number 1456922
[cid:dtclogo203x88_8fe7e5cd-210b-49fc-b334-f34189be7d63.png]
[...]

PR #3301, opened 2023-07-04
Ensure Unicode strings are converted when compiling  with Unicode enabled

### LINKED ISSUE #2934: Telescop control : cannot open serial device
### Expected Behaviour:
I have a NexStar compatible telescop controler which can conect at Stelarium throught Bluetooth on my celularphone an works prety well. I  expect to do the same on Stelarium 1.1 or 1.2  on Windows 10  ( the controler was working ok with somme old version of stelarium on a Lenovo YOGA Laptop Windows 10)

### Actual Behaviour
I get in the log file: "ERROR creating TelescopeClientDirectNexStar: cannot open serial device "COM5"
TelescopeClient::create(): Unable to create a telescope client." 

I can tell you that if I connect on COM 5 with Putty and send some comands to the controler i receive good answers ( so the BT connection via COM 5 is working).

### System
* Stellarium version: <stellarium-1.2-qt6-win64?>
* Operating system: <Windows 10 Version 22H2>
*Laptop ASUS ExpertBook 

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/10300576/log.txt)

A-j-K:
### Description

Qt6 appears to have introduced Unicode compilations. Type casting is insufficient and a full byte conversion is required.

Fixes #2934 

N/A

### Type of change

### How Has This Been Tested?

This was tested using my SkyWatcher SynScan Handcontroller AZ-GTiX Mount.

Note, this fix was at the serial port level and therefore I believe will address any mount that uses a serial port and not just my own SkyWatcher mount.

**Test Configuration**:
* Operating system: Windows 11 Qt6 6.3.4

## Checklist:

- [n/a] I have made corresponding changes to the documentation (header file)
- [ n/a] I have updated the respective chapter in the Stellarium User Guide
- [ n/a] I have added tests that prove my fix is effective or that my feature works
- [ n/a] New and existing unit tests pass locally with my changes
- [ n/a] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/plugins/TelescopeControl/src/common/SerialPort.cpp
+++ b/plugins/TelescopeControl/src/common/SerialPort.cpp
@@ -32,6 +32,10 @@ Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
 
 #include <cstring> // memset
 
+#ifdef UNICODE
+#include <codecvt> // for std::wstring_convert
+#endif
+
 SerialPort::SerialPort(Server &server, const char *serial_device)
 	: Connection(server, INVALID_SOCKET)
 	#ifndef Q_OS_WIN
@@ -40,7 +44,13 @@ SerialPort::SerialPort(Server &server, const char *serial_device)
 {
 #ifdef Q_OS_WIN
 #if (QT_VERSION>=QT_VERSION_CHECK(6,0,0))
-	handle = CreateFile(LPCWSTR(serial_device), GENERIC_READ|GENERIC_WRITE, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+#ifdef UNICODE
+	std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> wconverter;
+	std::wstring wdevice = wconverter.from_bytes(std::string(serial_device));
+	handle = CreateFile(wdevice.c_str(), GENERIC_READ | GENERIC_WRITE, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+#else
+	handle = CreateFile(LPCWSTR(serial_device), GENERIC_READ | GENERIC_WRITE, 0, nullptr, OPEN_EXISTING, 0, nullptr);
+#endif // UNICODE
 #else
 	handle = CreateFile(serial_device, GENERIC_READ|GENERIC_WRITE, 0, nullptr, OPEN_EXISTING, 0, nullptr);
 #endif
@@ -75,7 +85,12 @@ SerialPort::SerialPort(Server &server, const char *serial_device)
 				memset(&dcb, 0, sizeof(dcb));
 				dcb.DCBlength = sizeof(dcb);
 #if (QT_VERSION>=QT_VERSION_CHECK(6,0,0))
+#ifdef UNICODE
+				std::wstring wbaud = wconverter.from_bytes(std::string("9600,n,8,1"));
+				if (!BuildCommDCB(wbaud.c_str(), &dcb))
+#else
 				if (!BuildCommDCB(LPCWSTR("9600,n,8,1"), &dcb))
+#endif // UNICODE
 #else
 				if (!BuildCommDCB("9600,n,8,1", &dcb))
 #endif
--- a/plugins/TelescopeControl/src/common/SerialPort.cpp
+++ b/plugins/TelescopeControl/src/common/SerialPort.cpp
@@ -30,11 +30,9 @@ Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
 #include <unistd.h>
 #endif
 
-#include <cstring> // memset
+#include <QString>
 
-#ifdef UNICODE
-#include <codecvt> // for std::wstring_convert
-#endif
+#include <cstring> // memset
 
 SerialPort::SerialPort(Server &server, const char *serial_device)
 	: Connection(server, INVALID_SOCKET)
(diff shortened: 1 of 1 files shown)


10110111:
Wouldn't it be better to always use Unicode-enabled `CreateFileW`, regardless of Qt version and `UNICODE` macro, instead of all this useless conditional compilation?

10110111:
Also, `QString::fromUtf8("...").constData()` already provides a sequence of UTF-16 code units, so no need to create an `std::string` and convert via codecvt.

10110111:
> I currently don't have a system that can build a Qt5 version

You can use the AppVeyor build that's configured in `.appveyor.yml` in the source tree root. Prepending `[publish]` in commit messages produces binary artifacts. You should choose the "patch" one (which only installs `stellarium.exe`), otherwise if you download the full installer several times, you'll run over download limit.

> Quite happy to rework it that way

Yes, please do. I really dislike how complicated the code is becoming. Even before this change the code looked needlessly complicated. So please make it simple, there's not a single reason to use 8-bit legacy code pages with Qt where `QString` already stores data in UTF-16.

gzotti:
You can use the Qt MaintenanceTool in your Qt directory to add Qt5.15.2 (make sure to add QtScript and the other required additional libraries). Then in QtCreator, select the Kit for Qt5.15.2 and in the  cmake configuration, make sure to set ENABLE_QT6 to false/0. You may need to set Qt5 libs instead of Qt6 into PATH as well. 

+1 for simpler code. Note that most of the telescope plugin code roots in Qt4 times.  I did the Qt6 ifdefs and was happy I could compile it, but had no hardware to test. If there are functions that work on both, Qt5 and Qt6, it's preferrable. Methods introduced after Qt5.12 would still be problematic, though, as the Qt5 releases are AFAIK still made with 5.12. 

@alex-w is Qt5.9 still used somewhere, or could we raise at least to 5.12?

alex-w:
> Qt5.9 still used somewhere, or could we raise at least to 5.12?

No, we can update requirements up to Qt 5.12

alex-w:
Probably this PR fixed #3006 already

alex-w (approved):
Can't test on hardware, but looks ok

gzotti (approved):
If it did not work before, and works with this on actual hardware, sure it's OK for me!

gzotti:
It is possible that this fix has already solved a handful of them. OPs would need to test their gear now with a new build. It's a pity 23.2 just came out, so this will likely take months...

gzotti:
I also have just fixed another error around a message connection that was renamed in Qt5.15 and would never have worked on Qt6. Whatever reported issue this probably fixes I cannot say as we rarely receive useful logfiles (where this issue would have been visible). This is one of the big problems with the TC plugin: we would need ~30 devices to test on 3 OS and 2 CPU families, and 30 user-developers to debug and keep the program working with their own gear, not withdrawing to obsolete versions on the slightest hint of problem. So, any help is welcome, and esp. working fixes after deep reverse-engineering insights!

gzotti:
Qt5 builds are required on older systems:
- 32bit versions. This also includes users of older instrumentation with 32bit-only ASCOM drivers. 
- Windows 7 systems
- Windows systems with weak graphic subsystems, e.g. Intel Core-i of 1st/2nd generation are blacklisted by Qt and need to run ANGLE mode, which is no longer available with Qt6. And there are even weaker systems (Pentium6, Celeron, Atom, ...), things that users with higher demands may just mask out when shopping for new systems. 
 
It seems many observers use such oldish, otherwise obsolete systems to drive their telescopes, or they are in regions of this planet where PCs don't rain from the sky, and even old, used PCs are a major investment. A 2008 PC with Win10 (works in most cases!) and ANGLE is still better than no PC. Qt5 based releases for Windows should IMHO be provided as long as Win10 is widely used (i.e., at least V25.4), and as long as the code differences are miniscule. On Linux, some LTS distros may still have no Qt6. And who knows what will happen when probably hundreds of millions systems will (at least officially) have to move from Win10 to Linux in late 2025?

After release of 1.0 reports about issues with the TC plugin came in. About every report that included version number indicated Qt6, which is why we recommend using Qt5-based builds (but not obsolete 0.21 builds!). This most likely indicates I messed up something while porting to Qt6, but was not able to identify what. People complained in some astro forums instead of reporting with us, and often declared to have reverted to older versions.

10110111:
> Pentium6

What is this? Do you mean Pentium Pro, with the P6 microarchitecture? This won't even run Windows 7. Pentium 4? This one is usable with a discrete GPU.

> On Linux, some LTS distros may still have no Qt6

They are irrelevant, their official Stellarium packages won't be upgraded anyway until OS release upgrade.

> have to move from Win10 to Linux

No one will have, just as many haven't moved from Windows 7 to anything, because at least this may bring problems with driver support.

gzotti:
> > Pentium6
> 
> What is this? Do you mean Pentium Pro, with the P6 microarchitecture? This won't even run Windows 7. Pentium 4? This one is usable with a discrete GPU.

Whatever, maybe my memory fails with all these CPU names. A year ago I was asked for help with an older sub-Core-i laptop. I installed Win10, and Stellarium 0.22.1 worked in ANGLE mode. The thing serves the daily needs of an older gentleman.  Then maybe it was Pentium-IV. Already then I associated "Pentium" with the 1990s. Now they seem to also call Atoms "Pentium". And I admit, I will never learn their "Lakes".

Now, fit a discrete GPU into such an old laptop!

> > On Linux, some LTS distros may still have no Qt6
> 
> They are irrelevant, their official Stellarium packages won't be upgraded anyway until OS release upgrade.

We provide Ubuntu packages for 18.04LTS. Does it have Qt6? 

> > have to move from Win10 to Linux
> 
> No one will have, just as many haven't moved from Windows 7 to anything, because at least this may bring problems with driver support.

Requote: **"(at least officially)"**

alex-w:
> We provide Ubuntu packages for 18.04LTS. Does it have Qt6?

Sorry, but no, we do not provide packages for Ubuntu 18.04 LTS

gzotti:
I was referring to https://launchpad.net/~stellarium/+archive/ubuntu/stellarium-releases
image
OK, maybe 23.1 was the last then? 

I have two Odroid SBCs which are stuck with 18.04LTS. On these I like to have the latest features available, but can only build with Qt5. I hope I could still point out: being able to build with Qt5 will be needed for a few more years.

Issue #2916, opened 2022-12-14
Wrong phase angle of waning moon


SMi1984DE:
### Expected Behaviour

The lunar phase angle should be 72 degrees + 180 degrees (waning moon) = 252 degrees, in this example, not 72 degrees, because the phase angle should represent the different phases of the moon (and it's movement arround the earth/a full circumnavigation) from 0 degrees (=New Moon) over 90 degrees (first quarter), 180 degrees (=Full Moon), 270 degrees (last quarter) (back to) to 360 degrees/0 degrees (New Moon).

Otherwise the phase angle will never reach a value above 180 degrees. So it's only nessecary to add a value of 180 degrees after Full Moon, if the "phase angle" is beeing calculated with help of the "moon phase" value "P" in percentage.

(for example: The moon has 50% illumination at two times. At first quarter (waxing moon) = 50% and at last quarter (waning moon) = 50% as well. The calculation of the phase angle (f) with help of the common formula below only considers the phase from waxing moon.

`f = cos-1 [ 2P / 100 - 1]`

see also
http://castor2.ca/16_Calc/01_Moon/03_Phase/index.html
and
https://www.projectglobalawakening.com/moon-phases/

### Actual Behaviour
phase angle is +72 degrees

![phase_angle](https://user-images.githubusercontent.com/120596291/207688507-b304cdde-2d62-4b91-bbb2-f884dda34892.jpg)

### Steps to reproduce
n/a

### System
* Stellarium version: 1.1.1-macOS (Stellarium.app Version 0.22.4-8cc1565
* Operating system: macOS Big Sur (Version 11.7.1)
* Graphics Card: AMD Radeon R9 M290X 2GB
* Screen type (if applicable): iMac Retina 5k, 27")

### Logfile 
n/a

SMi1984DE:
btw. Stellarium is a great planetarium software. I like it very much. It's easy to use and has a lot of functions and informations. I think, you spend a lot of time for developing. ;) Thank you... 👍  👍 👍 👍 👍 👍

gzotti:
The phase angle is the angle Sun-object-observer and is 0...180°. Your idea implies that "phase angle" for the Moon must be defined differently than for planets. Can you give a reference? Or do you just mean elongation? 
And yes, we spend a lot of time, maybe too much. And then some.

SMi1984DE:
[...]

PR #3306, opened 2023-07-05
Fix computation of signed phase angle to determine apparent magnitude

10110111:
### Description

Current computation of apparent magnitude uses two similar quantities: phase angle and something that approximates it, adding a sign to indicate whether the it's a waxing or waning phase.

As I understand, the only reason of determining the sign is to make magnitude depend on what half of the Moon is illuminated, due to asymmetry in the albedo of the visible side of the Moon. So the quantity should actually be equal to the phase angle, up to a sign.

The unsigned phase angle is computed correctly, taking into account positions of the observer and the Moon. But the way the signed quantity is currently computed has some problems:

1. It has some daily ripples compared to the phase angle that shouldn't be there (possibly the shift of the observer from planet center isn't taken into account);
2. It switches sign not at the point of 0 or 180° phase, but a few dozens of minutes away from it. 

This PR implements the correct computation of whether the phase angle is increasing or decreasing (this is done in `Planet::isWaning()`). This computation uses heliocentric position and velocity of the observer (the velocity is computed in `StelCore::getObserverHeliocentricEclipticVelocity()`).

~I spent some time time trying to reverse-engineer orientation of the axes of the AltAz frame, so the first commit here documents my findings. Please check that it's actually correct.~

Also, it appeared that the computation to determine the phase angle tendency is quite nontrivial (the simple elongation-based trick mentioned in #2916 won't yield precise results), so the third commit here exports a flag to the scripting API that indicates whether the planet is in the waning phase, i.e. whether its phase angle is increasing.

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/StelCore.hpp
+++ b/src/core/StelCore.hpp
@@ -305,6 +305,9 @@ public:
 
 	//! Return the observer heliocentric ecliptic position (GZ: presumably J2000)
 	Vec3d getObserverHeliocentricEclipticPos() const;
+	//! Return the observer heliocentric ecliptic velocity. This includes orbital and diurnal motion;
+	// diurnal motion is omitted if planetocentric coordinates are in use.
+	Vec3d getObserverHeliocentricEclipticVelocity() const;
 
 	//! Get the information on the current location
 	const StelLocation& getCurrentLocation() const;
--- a/src/core/StelCore.cpp
+++ b/src/core/StelCore.cpp
@@ -1072,6 +1072,23 @@ Vec3d StelCore::getObserverHeliocentricEclipticPos() const
 	return Vec3d(matAltAzToHeliocentricEclipticJ2000[12], matAltAzToHeliocentricEclipticJ2000[13], matAltAzToHeliocentricEclipticJ2000[14]);
 }
 
+Vec3d StelCore::getObserverHeliocentricEclipticVelocity() const
+{
+	const auto& planet = *position->getHomePlanet();
+	const Vec3d planetVelocity = planet.getHeliocentricEclipticVelocity();
+	if (!flagUseTopocentricCoordinates)
+		return planetVelocity;
+
+	const auto off = position->getTopographicOffsetFromCenter();
+	const auto rotRadius = off.v[0];
+	const auto rotPeriod = planet.getSiderealDay();
+	const auto linearSpeed = 2 * M_PI * rotRadius / rotPeriod; // AU/day
+	const auto toJ2000 = matAltAzToHeliocentricEclipticJ2000.upper3x3();
+	const auto eastwardVelocity = Vec3d(0,linearSpeed,0);
+	const auto velocity = toJ2000 * eastwardVelocity;
+	return planetVelocity + velocity;
+}
+
 // Set the location to use by default at startup
 void StelCore::setDefaultLocationID(const QString& id)
 {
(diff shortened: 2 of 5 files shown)


src/core/StelCore.hpp
```diff
@@ -305,6 +305,9 @@ class StelCore : public QObject
 
 	//! Return the observer heliocentric ecliptic position (GZ: presumably J2000)
 	Vec3d getObserverHeliocentricEclipticPos() const;
+	//! Return the observer heliocentric ecliptic velocity. This includes orbital and diurnal motion;
+	// diurnal motion is omitted if topocentric coordinates are in use.
```
gzotti, line 309: omitted if planetocentric coordinates are in use. (?)

10110111: Oops, right.

src/core/StelCore.hpp
```diff
@@ -74,7 +74,7 @@ class StelCore : public QObject
 	enum FrameType
 	{
 		FrameUninitialized,			//!< Reference frame is not set (FMajerech: Added to avoid condition on uninitialized value in StelSkyLayerMgr::draw())
-		FrameAltAz,				//!< Altazimuthal reference frame centered on observer.
+		FrameAltAz,				//!< Altazimuthal reference frame centered on observer: +x=south, +y=east, +z=zenith.
```
gzotti, line 77: Correct. See also external use in http://stellarium.org/doc/head/remoteControlApi.html#rcMainServiceView

src/core/StelCore.hpp
```diff
@@ -305,6 +305,9 @@ class StelCore : public QObject
 
 	//! Return the observer heliocentric ecliptic position (GZ: presumably J2000)
 	Vec3d getObserverHeliocentricEclipticPos() const;
+	//! Return the observer heliocentric ecliptic velocity. This includes orbital and diurnal motion;
```
gzotti, line 308: If you include diurnal changes in velocity, this should be added to the aberration correction of the observer in SolarSystem::computePositions().

10110111: I suppose then `observerPlanet->getHeliocentricEclipticPos()` should also be replaced with `core->getObserverHeliocentricEclipticPos()`, right? Because currently that computation takes "observer" to be in the center of the planet.

gzotti: Yes, I also think that's all that has to be changed. When I did aberration correction, I said we won't include the diurnal aberration, but when you are doing this computation now, it's even one more step. Not sure about the magnitude, though, it may introduce an observable milliarcsecond wobble.  Of course, switching off topographic correction will also disable that.

10110111: Do you think it should go right into this PR? This seems quite unrelated.

alex-w (approved):
Thanks!

10110111:
@gzotti What is your verdict on this PR?

gzotti:
Just having a look into its differences, but answering other duties at the same time. 

Why is the Moon about 0.1-0.2mag less bright? It may not matter a lot in reality, but still it's a notable difference between my model and your change in computing one of the required angles.

10110111:
The change in magnitude reflects the change in the phase angle that's fed into the algorithm that computes the magnitude. A totally expected result.

10110111:
The difference between the old angle and the new one is that the old angle was computed between projected positions of the observer and the Moon, while the new one is between the actual 3D positions.

As we take into account different *z* components of the positions, we should expect the angle between the objects to increase, which will reduce the visible lit part.

src/core/StelCore.hpp
gzotti: It will be an obvious next change then. Probably we can even make a switch aberration: none/annual/diurnal (for didactic purposes).

gzotti (approved):
OK, the original paper was in terms of phase angle, not dLong, so it should be fine. 

This change will introduce localized heliocentric velocity, so despite previous reservation, diurnal aberration should come soon.

src/core/StelCore.hpp
10110111: Why do you only say that it should be applied in `SolarSystem`? Shouldn't it equally affect everything?

PR #3327, opened 2023-07-14
Add Skywatcher SynScan HC and Pro App alignment stars

A-j-K:
### Description

No issue, just a small addition. I saw the Meade telescope alignment stars added so I thought I'd add the Skywatcher ones.

### Type of change

- [ n/a] Bug fix (non-breaking change which fixes an issue)
- [n/a ] Breaking change (fix or feature that would cause existing functionality to change)
- [ n/a] This change requires a documentation update

### How Has This Been Tested?

Tested manually. This update doesn't lend itself well to automation.

Tested on Windows 11 MSVC 2019 QT 6.5.1

## Checklist:

- [x ] I have made corresponding changes to the documentation (header file)
- [ n/a] I have updated the respective chapter in the Stellarium User Guide
- [n/a] I have added tests that prove my fix is effective or that my feature works
- [n/a] New and existing unit tests pass locally with my changes
- [n/a] Any dependent changes have been merged and published in downstream modules


plugins/NavStars/src/gui/NavStarsWindow.cpp
```diff
@@ -251,6 +251,7 @@ void NavStarsWindow::populateNavigationalStarsSets()
 	nsSets->addItem("Celestron", "Celestron");
 	// TRANSLATORS: Part of full phrase: Apollo space program set of navigational stars
 	nsSets->addItem(q_("Apollo space program"), "Apollo");
+	nsSets->addItem("Skywatcher", "Skywatcher");
```
gzotti, line 254: I think the Apollo list should go between the Nautical lists and the telescope lists.

plugins/NavStars/src/NavStars.cpp
```diff
@@ -660,6 +664,26 @@ void NavStars::populateNavigationalStarsSet(void)
 				     << 102098 << 107315 << 113368;
 			break;
 		}
+		case Skywatcher:
+		{
+			// Skywatcher SynScan Hand Held Controller and SynScan Pro mobile and desktop App
+			// Sources: https://www.iceinspace.com.au/63-501-0-0-1-0.html
+			//          https://www.iceinspace.com.au/download.php?10d72300ac5d3762f0529eb11977fb68
+			starNumbers << 13847 <<  33579 <<  95947 <<  67301 <<  17702 <<  21421 << 105199
```
gzotti, line 672: My gut feeling is that `starnumbers = {13847, 33579, ...}` would be more efficient than this subsequent list buildup. It's minuscule, but maybe you change all lists accordingly while you're at it.

gzotti:
Thanks, yes. This is the place to add the various lists if somebody has them.

plugins/NavStars/src/NavStars.cpp
```diff
@@ -450,6 +450,10 @@ QString NavStars::getCurrentNavigationalStarsSetDescription() const
 			txt = q_("These 37 stars were used by the Apollo space program to navigate to the Moon from 1969-1972, Apollo 11 through Apollo 17.");
 			break;
 		}
+		case Skywatcher:
+		{
+			txt = q_("This list contains all the alignment stars for the Skywatch SynScan hand controller and SynScan Pro App.");
+		}
```
gzotti, line 456: maybe add a `break` also here, even if it's the last case.

gzotti: Yes, not my code. Indeed the braces could go away, would also save a few lines. Can you also clean this up while you are here?

10110111: > some compilers complain if a switch statement is missing a default

They only complain if the switch parameter is an enum, and there are enum values that are not represented in the cases. By adding a default to an enum that must consider all defined enum values you lose a useful diagnostic.

gzotti (approved):
Looks OK to me, thanks!

alex-w (approved):

PR #3343, opened 2023-07-25
Always use texture-based text rendering in StelPainter

10110111:
This will let us control gamma correction by simply changing texture format when we convert our rendering pipeline to linear colors. We can't do this when using QOpenGLPaintDevice.

Text rendering quality doesn't change. From my observations via apitrace, QOpenGLPaintDevice does basically the same, even for rotated text: just renders textured quads with the texture containing text.

A side effect is a small performance improvement when the scene has a lot of text.

The `-t` mode now only disables antialiasing in `StelPainter::drawText`, and also does something for the GUI, which I didn't change.

Also, this PR will make #3003 obsolete and, possibly, fix the cause of it.

### Type of change

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/StelPainter.cpp
+++ b/src/core/StelPainter.cpp
@@ -674,6 +674,7 @@ StringTexture* StelPainter::getTextTexture(const QString& str, int pixelSize) co
 		return cachedTex;
 	QFont tmpFont = currentFont;
 	tmpFont.setPixelSize(currentFont.pixelSize()*static_cast<int>(static_cast<float>(prj->getDevicePixelsPerPixel())*StelApp::getInstance().getGlobalScalingRatio()));
+	tmpFont.setStyleStrategy(QFont::NoSubpixelAntialias); // The text may rotate, which would break subpixel AA
 	QRect strRect = QFontMetrics(tmpFont).boundingRect(str);
 	int w = strRect.width()+1+static_cast<int>(0.02f*strRect.width());
 	int h = strRect.height();
@@ -681,12 +682,20 @@ StringTexture* StelPainter::getTextTexture(const QString& str, int pixelSize) co
 	QPixmap strImage = QPixmap(StelUtils::getBiggerPowerOfTwo(w), StelUtils::getBiggerPowerOfTwo(h));
 	strImage.fill(Qt::transparent);
 	QPainter painter(&strImage);
-	tmpFont.setStyleStrategy(QFont::NoAntialias);
+	if (qApp->property("text_texture").toBool()) // CLI option -t given?
+	{
+		// This is essential on devices like Raspberry Pi (2016-03).
+		tmpFont.setStyleStrategy(QFont::NoAntialias);
+	}
+	else
+	{
+		painter.setRenderHints(QPainter::TextAntialiasing);
+	}
 	painter.setFont(tmpFont);
-	//painter.setRenderHints(QPainter::TextAntialiasing);
 	painter.setPen(Qt::white);
 	painter.drawText(-strRect.x(), -strRect.y(), str);
 	StringTexture* newTex = new StringTexture(new QOpenGLTexture(strImage.toImage()), QSize(w, h), QPoint(strRect.x(), -(strRect.y()+h)));
+	newTex->texture->setMinMagFilters(QOpenGLTexture::Linear, QOpenGLTexture::Linear);
 	texCache.insert(hash, newTex, 3*w*h);
 	// simply returning newTex is dangerous as the object is owned by the cache now. (Coverity Scan barks.)
 	return texCache.object(hash);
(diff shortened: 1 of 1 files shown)


alex-w:
The `-t` mode enable very strange look for text in bottom bar

10110111:
> The `-t` mode enable very strange look for text in bottom bar

Not my fault, it's similarly ugly in `master`.

alex-w:
As far as I remember `-t` mode was added for low-cost devices such RPi or ANGLE-“powered” (low-graphics?) systems. Bad news, that I can’t saw when we broke GUI in `-t` mode.

10110111:
It was for RPi, and the aliasing was intended. As for the wrong placement of the panel (not sure if you see it too), I suspect it might have been broken by #3125.

alex-w (approved):

alex-w:
But in this case we can remove support -t mode

10110111:
So, should we do it now?

alex-w:
I think yes. @gzotti?

gzotti:
I hope I can test with RPi3&4 later tonight.

gzotti:
What is ugly or broken with -t in master?

gzotti:
Ok, looks like the older composition/layout of that status line. It still works, but just looks less perfect than the reworked solution. I may have not cared, given the lack of actual necessity of the -t option.

gzotti:
Please compare: right: 23.2. left: this branch, regular mode. At least on Windows, there is a significant loss of quality. 
Screenshot (82)

gzotti:
RPi3, this PR
stellarium-rp3_pr3343
RPi3, master
stellarium-rp3_master

gzotti:
RP4, master
stellarium-rp4_master

RP4, master with -t
stellarium-rp4_master-t

RP4, this PR
stellarium-rp4_pr3343

RP4, this PR with -t
stellarium-rp4_pr3343-t

This really gives a nice speed gain!

10110111:
> At least on Windows, there is a significant loss of quality.

What if you run with `-platform windows:fontengine=freetype` command line option?

10110111:
So, except for Windows, everything looks OK, right?

gzotti:
> > At least on Windows, there is a significant loss of quality.
> 
> What if you run with `-platform windows:fontengine=freetype` command line option?

Yes, that works. Fonts are a bit narrower now. Right: 23.2, left: this PR
Screenshot (83)

OK, to accept this, the platform option must go into all startup links, like `-platform windows:altgr,fontengine=freetype`, into the cmake/*.iss.* files.

I can also confirm a great speed improvement here (~40->65 fps!). 

WAIT: I did a few more runs, and suddenly freetype seems not to become active. It looks thin and bad again. No log entries that could help.

10110111:
> I did a few more runs, and suddenly freetype seems not to become active. It looks thin and bad again. No log entries that could help.

Try following [this post](https://stackoverflow.com/a/36204635/673852) and getting rid of the command line addition.

gzotti:
Ah, thanks. It seems that the command line parsing of a coma-separated list of options `-platform windows:altgr,fontengine=freetype` does not work. Also giving separated `-platform  windows:...` options fails. The `altgr` option is essential on many international keyboards, though. 

So, what works is a file qt.conf right next to the .exe which contains:
```
[Platforms]
WindowsArguments = fontengine=freetype,altgr

```

This means: 

- On Windows builds, this files has to be created/provided (cmake?)
- It must be deployed in installs (cmake/inno?)
- The `data/*.iss.*` files must be freed from `-platform windows:...` options

10110111:
> Yes, that works. Fonts are a bit narrower now. Right: 23.2, left: this PR

The +SEP and +SCP items look different in your two windows. Is there a good explanation, or is it a bug in this PR?

gzotti:
Uh, yes. Apps are RemoteSync'ed. No explanation, looks like a bug.

alex-w:
Please check https://github.com/Stellarium/stellarium/commit/186febd4d67b53946ce200bb4e69a0a0450b7086

10110111:
What do these +SEP and +SCP represent? How do I enable them?

alex-w:
> What do these +SEP and +SCP represent? How do I enable them?

SCP = South Celestial Pole; SEP = South Ecliptic Pole

Please check poles in Markings tab

10110111:
I have them, and from the code I see that the cross is rendered as a sprite, so this shouldn't be related to this PR.

@gzotti Do you reproduce this lack of the crosses in `master`?

gzotti:
Crosses are fine in master. But I still have the flood of 

`D:\StelDev\GIT\stellarium\src\core\StelApp.cpp:910: OpenGL error: 1281 (GL_INVALID_VALUE)
`

gzotti:
For what it's worth, crosses now also show in this PR.

10110111:
> But I still have the flood of

Would be useful if you could find the point where this error occurs (e.g. via apitrace).

> crosses now also show in this PR

OK, is it OK to merge then? @alex-w has pushed the changes to `qt.conf`, so it should work now.

alex-w:
Error 1281 should not be happening after rebasing

gzotti:
Can qt.conf also be created in the temporary build directory? I usually run a test build from there, without installation. Currently I need to create this manually.

10110111:
> Error 1281 should not be happening after rebasing

Where do you see this error?

gzotti:
The 1281 is in master, not in this PR. This seems OK to merge. We could however really remove the -t option (including SUG deletions).

10110111:
> We could however really remove the -t option

OK, will do it separately.

alex-w:
> Can qt.conf also be created in the temporary build directory? I usually run a test build from there, without installation. Currently I need to create this manually.

It's possible, but I can't do the solution right now - after 2 weeks only. Please fill separate report for this issue for memory.

PR #3342, opened 2023-07-25
Nomenclature plot fixes

gzotti:
### Description

This introduces an option to encircle craters and satellite features (which are usually also craters) with ellipses.

I also just slightly simplified the code.
 

Fixes # (issue)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11
* Graphics Card: Geforce 3070Ti
* Qt 6.5.1

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2043, opened 2021-11-17
Fix issue #1842

A-j-K:
### Description

The change is to invert the logic calling setFlagOrbits() which seemed to be the wrong way around.

Fixes #1842

### Screenshots (if appropriate):
n/a

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

### How Has This Been Tested?

Tested on Windows 10 Visual Studio before and after the change.

**Test Configuration**:
* Operating system: Windows 10
* Graphics Card:  n/a

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #3348, opened 2023-08-02
Fix: orbit details

### LINKED ISSUE #1842: Unchecking "Only orbit for selected object" still hides other planets' orbits
This seems like a bug.

Displaying orbits, and unchecking "Only orbit for selected object" still hides other planets' orbits if one planet is selected (for example, selecting Jupiter will hide every planet's orbit except Jupiter's).

Tested with the latest public beta on a Windows 10 machine, but IIRC this is an old bug.

gzotti:
### Description

As discussed in #2043, the sequence of selection around orbit details is inconsistent and buggy. 
I separated flag setting from the actual selection (which involves 4 flags and an optional selection).
We can discuss yet another finetuning option "with moons" which would influence whether the orbital system of moons would be highlighted even if the flag for only "major planets" or "only selected object" is set and the selected object is a major planet. 

Also, as first commit, I introduce a (currently hidden and manually editable only) option StelApp.flagImmediateSave. If set, this can be used to store settings immediately. The intention is to allow storing detail settings in the GUI, like orbit drawing details, DSO catalog selection or type filtering, without having to save these globally (and thereby maybe accidentally overwrite and globally store other settings that were active only during this session). My personal preference would be to get rid of the global save entirely, but I see resistance. Maybe we can offer a global switch on  "immediate save" vs "global save", when all settings have been adjusted over the next few months. 

I will amend the second commit by removing commented code and optionally adding the "with moons" option. 

After this has been merged, PR #3342 will be adjusted accordingly. 

Fixes #1842 (supersedes PR #2043)

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Geforce 3070Ti (irrelevant)

## Checklist:

### DIFF
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -2635,7 +2635,7 @@ void SolarSystem::reconfigureOrbits()
 	// 1 1 1 0   only selected planet and orbits of its moon system
 	// 1 1 1 1   only selected SSO if it is a major planet
 
-	if (!flagOrbits || (flagOrbits&&flagIsolatedOrbits&&(!selected || selected==sun)))
+	if (!flagOrbits || (/* flagOrbits && */ flagIsolatedOrbits&&(!selected || selected==sun)))
 	{
 		for (const auto& p : qAsConst(systemPlanets))
 			p->setFlagOrbits(false);
@@ -2645,73 +2645,17 @@ void SolarSystem::reconfigureOrbits()
 	if (!flagIsolatedOrbits)
 	{
 		for (const auto& p : qAsConst(systemPlanets))
-			if (!flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly && (p->getPlanetType()==Planet::isPlanet || (p->parent && p->parent->getPlanetType()==Planet::isPlanet) )))
-				p->setFlagOrbits(true);
-			else
-				p->setFlagOrbits(false);
+			p->setFlagOrbits(!flagPlanetsOrbitsOnly || (p->getPlanetType()==Planet::isPlanet || (p->parent && p->parent->getPlanetType()==Planet::isPlanet) ));
 		return;
 	}
 	else // flagIsolatedOrbits && selected
 	{
 		// Display only orbit for selected planet and its moons.
 		for (const auto& p : qAsConst(systemPlanets))
-			if (   (p==selected && (  !flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly && p->getPlanetType()==Planet::isPlanet ) )
-				|| (p->getPlanetType()==Planet::isMoon && p->parent==selected ) ))
-				p->setFlagOrbits(true);
-			else
-				p->setFlagOrbits(false);
+			p->setFlagOrbits(   (p==selected && (  !flagPlanetsOrbitsOnly ||  p->getPlanetType()==Planet::isPlanet )
+					 || (p->getPlanetType()==Planet::isMoon && p->parent==selected ) ));
 		return;
 	}
-
-/*
-	if (!selected || selected==sun)
-	{
-		if (flagPlanetsOrbitsOnly)
-		{
-			for (const auto& p : qAsConst(systemPlanets))
-			{
-				if (p->getPlanetType()==Planet::isPlanet)
-					p->setFlagOrbits(true);
-				else
-					p->setFlagOrbits(false);
-			}
-		}
-	}
-	else if (getFlagIsolatedOrbits()) // If a Planet is selected and orbits are on, fade out non-selected ones
-	{
-		if (flagPlanetsOrbitsOnly)
-		{
-			for (const auto& p : qAsConst(systemPlanets))
-			{
-				if (selected == p && p->getPlanetType()==Planet::isPlanet)
-					p->setFlagOrbits(b);
-				else
-					p->setFlagOrbits(false);
-			}
-		}
-		else
-		{
-			for (const auto& p : qAsConst(systemPlanets))
-			{
-				if (selected == p)
-					p->setFlagOrbits(b);
-				else
-					p->setFlagOrbits(false);
-			}
-		}
-	}
-	else
-	{
-		// A planet is selected and orbits are on - draw orbits for the planet and their moons
-		for (const auto& p : qAsConst(systemPlanets))
-		{
-			if (selected == p || selected == p->parent)
-				p->setFlagOrbits(b);
-			else
-				p->setFlagOrbits(false);
-		}
-	}
-	*/
 }
 
 void SolarSystem::setFlagIsolatedOrbits(bool b)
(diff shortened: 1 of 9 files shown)


src/core/modules/SolarSystem.cpp
```diff
@@ -3287,8 +3343,13 @@ bool SolarSystem::getFlagPermanentOrbits() const
 
 void SolarSystem::setOrbitsThickness(int v)
 {
-	Planet::orbitsThickness=v;
-	emit orbitsThicknessChanged(v);
+	if (v!=Planet::orbitsThickness)
+	{
+		Planet::orbitsThickness=v;
+		if (StelApp::getInstance().getFlagImmediateSave())
+			StelApp::getInstance().getSettings()->setValue("Int", v);
```
alex-w, line 3350: Wrong parameter?

gzotti: Of course. Not sure where this typo comes from.

src/core/modules/SolarSystem.cpp
```diff
@@ -2668,14 +2611,120 @@ void SolarSystem::setNumberIsolatedTrails(int n)
...
+	{
+		for (const auto& p : qAsConst(systemPlanets))
+			p->setFlagOrbits(false);
+		return;
+	}
+	// from here, flagOrbits is certainly on
+	if (!flagIsolatedOrbits)
+	{
+		for (const auto& p : qAsConst(systemPlanets))
+			if (!flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly && (p->getPlanetType()==Planet::isPlanet || (p->parent && p->parent->getPlanetType()==Planet::isPlanet) )))
```
10110111, line 2648: `!flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly &&...)` is redundant: the second check for the flag always yields `true`. After you remove it, you can also get rid of two pairs of parentheses, yielding this simpler condition:
```
if (!flagPlanetsOrbitsOnly || p->getPlanetType()==Planet::isPlanet ||
    (p->parent && p->parent->getPlanetType()==Planet::isPlanet))
```

src/core/modules/SolarSystem.cpp
```diff
@@ -2668,14 +2611,120 @@ void SolarSystem::setNumberIsolatedTrails(int n)
...
+				p->setFlagOrbits(true);
+			else
+				p->setFlagOrbits(false);
+		return;
+	}
+	else // flagIsolatedOrbits && selected
+	{
+		// Display only orbit for selected planet and its moons.
+		for (const auto& p : qAsConst(systemPlanets))
+			if (   (p==selected && (  !flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly && p->getPlanetType()==Planet::isPlanet ) )
```
10110111, line 2658: Same redundancy of the second check of `flagPlanetsOrbitsOnly` here.

src/core/modules/SolarSystem.cpp
```diff
@@ -2668,14 +2611,120 @@ void SolarSystem::setNumberIsolatedTrails(int n)
...
+				|| (p->getPlanetType()==Planet::isMoon && p->parent==selected ) ))
+				p->setFlagOrbits(true);
```
10110111, line 2660: Since you're basically passing the result of the condition check, you can change
`if(cond) set(true); else set(false);` to `set(cond);`.

src/core/modules/SolarSystem.cpp
```diff
@@ -2668,14 +2611,120 @@ void SolarSystem::setNumberIsolatedTrails(int n)
...
+	// O S I P
+	// 0 X X X   NONE
+	// 1 0 1 X   NONE
+	// 1 X 0 0   ALL
+	// 1 X 0 1   all planets only
+
+	// 1 1 1 0   only selected planet and orbits of its moon system
+	// 1 1 1 1   only selected SSO if it is a major planet
+
+	if (!flagOrbits || (flagOrbits&&flagIsolatedOrbits&&(!selected || selected==sun)))
```
10110111, line 2638: Ah, same redundancy with `flagOrbits` here as the ones below.

src/core/modules/SolarSystem.cpp
```diff
@@ -2668,14 +2611,120 @@ void SolarSystem::setNumberIsolatedTrails(int n)
...
+		for (const auto& p : qAsConst(systemPlanets))
+			p->setFlagOrbits(false);
+		return;
+	}
+	// from here, flagOrbits is certainly on
+	if (!flagIsolatedOrbits)
+	{
+		for (const auto& p : qAsConst(systemPlanets))
+			if (!flagPlanetsOrbitsOnly || (flagPlanetsOrbitsOnly && (p->getPlanetType()==Planet::isPlanet || (p->parent && p->parent->getPlanetType()==Planet::isPlanet) )))
+				p->setFlagOrbits(true);
```
10110111, line 2649: Same comment here about passing a Boolean condition.

src/core/modules/SolarSystem.cpp
gzotti: Yes, indeed. This is currently written mostly documentarily.

src/core/modules/SolarSystem.cpp
gzotti: Agreed. I just followed the existing style, but we can change that.

gzotti:
What about functionality? Shall moons for the planets be co-selected when "planet only" is active? Shall the single selected planet also show its moons' orbits? Or should we deal with the moons with yet another flag? Shall a selected object like the Moon (or other planet moons) show its orbit even if "only planet orbits" and "selected only" are active?

Atque:
What about a whole new setting planet + moon orbit?

gzotti:
This is what I am asking: Show moons automatically (like now), or only if allowed by another switch?

src/core/modules/SolarSystem.cpp
10110111: My concern here is that it makes it harder to understand the condition, not only because of an extra check, but also due to multiple nested levels of parentheses.

Atque:
> This is what I am asking: Show moons automatically (like now), or only if allowed by another switch?

Oh yeah, sorry. A new flag, I would prefer.

gzotti:
@alex-w , @10110111 your opinions/wishes/comments on that?

10110111:
My wish is that what's written on the checkbox, that would be seen in the scene, without surprises. If the checkbox says only selected planet, then only the planet's orbit would be visible. So I guess this would mean a radiobutton:

ⵙ selected planet
ⵔ selected planet and its moons

gzotti:
I dislike the `StelApp::getInstance() ...` verbosity. 
You mean something shorter like a static `StelApp::immediateStore("option", value)` which queries the flag and only stores if `StelApp.flagImmediateSave` has been activated? 

Note that for now all these these settings are also still stored in ConfigDialog, so if immediate saving is not active, behaviour is unchanged.

gzotti:
@alex-w , @10110111 any more comments or objections?

10110111 (approved):
Seems to work as expected.

alex-w (approved):

Issue #1650, opened 2021-05-08
Bug in rendering orbits


alex-w:
By chance I found this question and I cannot figure out why.

![stellarium-000](https://user-images.githubusercontent.com/83047344/117544331-88dea980-b053-11eb-81f9-a1f7050aedd6.jpg)
![stellarium-001](https://user-images.githubusercontent.com/83047344/117544338-8da35d80-b053-11eb-88a6-0651e39ed504.jpg)
![stellarium-002](https://user-images.githubusercontent.com/83047344/117544342-909e4e00-b053-11eb-908f-1bf9d7547b38.jpg)

I was standing on Ganymede, with both "atmosphere" and "ground" off. When I selected Jupiter and let the orbit show, it came to me like these. 

I have no idea why the revolving orbit can twist in such a weird manner.  Maybe I misunderstood these kind of "orbits," but what on earth are they symbolizing?

_Originally posted by @coctree52 in https://github.com/Stellarium/stellarium/discussions/1649_

gzotti:
The orbits are displayed as 360-gons. The planet is forced to be a vertex of this figure, but in the closeup seen from its moons, the two legs of the 360-gon which include Jupiter show very distinctly as straight edges. The 358 other edges form a visible curve because they are nicely in the distance. 
Technically it is not a bug as the implementation correctly shows the 360-gon. Somebody with too much time can still add improvements, like splitting those two legs into much smaller parts.

10110111:
I haven't looked into the implementation yet, have things to do in real life.

PR #3371, opened 2023-08-16
Smooth orbit lines (esp. for comets)

gzotti:
### Description

Drawing orbit lines from 360 time steps along the orbit works well for ellipses with low eccentricity only. 
For highly elongated comet orbits, the perihel area is covered only with a few vertices, which causes a polygonal, usually vastly undersampled appearance. With the time step approach, the situation was remedied a bit by not plotting the entire orbit, but only the part which is signified by the orbit_good value in ssystem_minor.ini. 

This PR provides a solution that constructs the entire orbit for ellipses with eccentricities e<0.999, by stepping over the eccentric anomaly. This looks generally much better and allows the estimation where in the outer solar system the object has its aphel. There still remain a few "polygon orbits" when they appear close to the observer. 

Fixes # (issue)

stellarium-094

### Type of change

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Geforce (irrelevant)

## Checklist:

### DIFF
--- a/data/ssystem_1000comets.ini
+++ b/data/ssystem_1000comets.ini
@@ -7,11 +7,6 @@
 [67pchuryumovgerasimenko]
 absolute_magnitude=11
 albedo=1
-color=1.0, 1.0, 1.0
-coord_func=comet_orbit
-dust_brightnessfactor=1.5
-dust_lengthfactor=0.4
-dust_widthfactor=1.5
 name=67P/Churyumov-Gerasimenko
 orbit_ArgOfPericenter=12.7758
 orbit_AscendingNode=50.1497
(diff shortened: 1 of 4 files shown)


10110111:
Is the screenshot the state before this PR or after?

gzotti:
It's the new state. Previously, only the orbit_good days were shown for the orbits, although most can now be shown entirely. Yes, there are still a few edges visible, those are close to the observer. We can increase to 720 or 1440 steps per orbit to remedy that somewhat (at probably high CPU cost), but an n-gon always has vertices...

10110111:
How is the second commit related to this PR?

10110111:
> We can increase to 720 or 1440 steps per orbit to remedy that somewhat (at probably high CPU cost), but an n-gon always has vertices

This way would indeed require too many vertices, the closer the curve and the wider the FoV, the more vertices we'd need. But we need them only in one section, not for the whole orbit. So I think a better approach would be to generate the orbit already in the screen coordinates (which we do need to obtain anyway in `drawOrbit()`), and sample it smartly, something like this:

1. Generate a few points, say, 10.
2. For each pair of points measure angle and distance between them **on screen**.
3. If some quality function of angle and distance is too low, add a point with eccentric anomaly taken as average between the two.
4. Stop when all lines have sufficient quality.

gzotti:
I did not intend to fix #1650. This here provides a significant improvement, though.

alex-w (approved):
Well, good improvement

10110111 (approved):

Issue #3427, opened 2023-09-25
23.3: crash upon exit


DarthGandalf:
### Expected Behaviour
It should exit cleanly

### Actual Behaviour
Downloaded 1 files (0 kbytes) in a session of 41.33 sec (average of 0 kB/s + 0 files from cache (0 kB)).
LandscapeMgr: Clearing cache of 0 landscapes totalling about  0 MB.
malloc(): unsorted double linked list corrupted
[1]    22581 IOT instruction  stellarium

### Steps to reproduce
1. Launch Stellarium
2. Wait until it starts
3. Move mouse to lower edge of the screen
4. Click Quit button

### System
* Stellarium version: 23.3, qt5
* Operating system: Gentoo Linux, amd64
* Graphics Card: GIGABYTE Radeon RX 6700 XT Gaming OC 12G, x11-drivers/xf86-video-amdgpu-23.0.0
* Screen type (if applicable): 1920x1080

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/12719494/log.txt)

Stacktrace: [gdb.txt](https://github.com/Stellarium/stellarium/files/12719545/gdb.txt)

DarthGandalf:
Might be related to #3391 but I don't do anything special with any script

10110111:
I often notice this message, without any scripts. But doesn't always appear, I can't predict whether it will appear on any launch. Can you reproduce it reliably?

DarthGandalf:
Yes, so far it happened every time I launched stellarium since upgrade to 23.3 - five times or so, mostly while debugging this crash

10110111:
Could you try running it under Valgrind?

DarthGandalf:
Doesn't show such error under valgrind

Downloaded 1 files (0 kbytes) in a session of 673.214 sec (average of 0 kB/s + 0 files from cache (0 kB)).
==26533== Warning: set address range perms: large range [0x45753000, 0x579a0000) (noaccess)
==26533== Warning: set address range perms: large range [0x5d896000, 0x7e8c2000) (noaccess)
LandscapeMgr: Clearing cache of 0 landscapes totalling about  0 MB.
StelGLWidget destroyed
==26533== 
==26533== HEAP SUMMARY:
==26533==     in use at exit: 491,216 bytes in 3,624 blocks
==26533==   total heap usage: 58,744,970 allocs, 58,741,346 frees, 30,950,473,774 bytes allocated
==26533== 
[...]

PR #3448, opened 2023-10-05
Use StelMainView as the parent for all message boxes and file dialogs

10110111:
### Description

In many cases the separate-window dialogs like message boxes and file open/save/etc. dialogs are given no parent in Stellarium. This leads to them looking very different in style and color scheme from the rest of the Stellarium GUI. The most striking difference is the light color scheme of the dialogs contrasting with the dark colors of Stellarium. Example:

Screenshot_2023-10-05_22-17-08

This patch changes almost all places that use `QMessageBox` and `QFileDialog` to pass `&StelMainView::getInstance()` as the parent, following the example of other places (and sometimes even the same files in other lines!). Only the message boxes used to tell about OpenGL errors are not parented since in this case this might be useless. The result looks like this:

Screenshot_2023-10-05_22-16-24

### DIFF
--- a/plugins/Satellites/src/gui/SatellitesDialog.cpp
+++ b/plugins/Satellites/src/gui/SatellitesDialog.cpp
@@ -1477,7 +1477,7 @@ void SatellitesDialog::updateTLEs(void)
 	}
 	else
 	{
-		QStringList updateFiles = QFileDialog::getOpenFileNames(nullptr,
+		QStringList updateFiles = QFileDialog::getOpenFileNames(&StelMainView::getInstance(),
 									q_("Select TLE Update File"),
 									StelFileMgr::getDesktopDir(),
 									"*.*");
@@ -1591,7 +1591,7 @@ void SatellitesDialog::savePredictedIridiumFlares()
 
 	QString defaultFilter = QString("(*.%1)").arg(defaultExtension);
 	QString dir = QString("%1/iridium_flares.%2").arg(QDir::homePath(), defaultExtension);
-	QString filePath = QFileDialog::getSaveFileName(nullptr, q_("Save predicted Iridium flares as..."), dir, filter, &defaultFilter);
+	QString filePath = QFileDialog::getSaveFileName(&StelMainView::getInstance(), q_("Save predicted Iridium flares as..."), dir, filter, &defaultFilter);
 
 	int count = ui->iridiumFlaresTreeWidget->topLevelItemCount();
 	int columns = iridiumFlaresHeader.size();
--- a/plugins/SolarSystemEditor/src/gui/ManualImportWindow.cpp
+++ b/plugins/SolarSystemEditor/src/gui/ManualImportWindow.cpp
@@ -190,7 +190,7 @@ void ManualImportWindow::selectTextureFile(QLineEdit * filePathLineEdit)
 		supportedFormats.append(QString("*.%1").arg(QString(format)));//It's a wee bit long...
 	}
 	QString fileFilter = QString("Texture files (%1)").arg(supportedFormats.join(" "));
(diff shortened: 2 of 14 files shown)


gzotti:
Looks great! I hope to have a closer look on Saturday.

alex-w:
I see native dialogs in Windows/macOS. As far as I remember, the non-native style of dialogs was caused problems in Linux Mint.

10110111:
> I see native dialogs in Windows/macOS. As far as I remember, the non-native style of dialogs was caused problems in Linux Mint.

What do you mean by native style? X11/Wayland-based desktops don't have any native style, each GUI toolkit uses its own. And what problems did you have? I'm using Ubuntu 20.04 and there are no problems here.

Moreover, some dialogs are already parented to `StelMainView`, so apparently they didn't cause any problems.

alex-w:
> > I see native dialogs in Windows/macOS. As far as I remember, the non-native style of dialogs was caused problems in Linux Mint.
> 
> What do you mean by native style? X11/Wayland-based desktops don't have any native style, each GUI toolkit uses its own. And what problems did you have? I'm using Ubuntu 20.04 and there are no problems here.
> 
> Moreover, some dialogs are already parented to `StelMainView`, so apparently they didn't cause any problems.

Native look for operating system UI. The issue (crash?) was happened around Qt 5.5-5.6 and it was caused from Qt/GTK+ bridge. Probably this is not actual anymore.

alex-w (approved):

gzotti:
Tried on Windows: No visual change. 
Tried WSL: seems to work, but... 

some observations (maybe unrelated): when first running, I had no wayland plugin installed and a warning. The fullscreen was black, only black. I could not F11 to window mode. Edited the config.ini, then it seemed to work. In exit, there was a Segfault. 

Then I installed qt6-wayland. It works windowed, but now the window is not resizable ( I have read about decoration issues with wayland a few days ago, so maybe that's expected now?) But no Segfault at exit. 

If those observations are unrelated, I think it's OK. I hope window resizing on wayland (if that is what's used on WSL now) will be re-established. Or some alternative Qt plugin recommended for WSL.

10110111:
> wayland plugin installed

Is Wayland the default window system in WSL? I thought it would be X11... I think Wayland support still has a long road towards stability even on "real" Unices.

> In exit, there was a Segfault.

This may be #3427.

> If those observations are unrelated, I think it's OK.

Shouldn't be related. In any case, we've already had some dialogs with non-null parent, I've just made this set more complete in this PR.

gzotti:
> > wayland plugin installed
> 
> Is Wayland the default window system in WSL? I thought it would be X11... I think Wayland support still has a long road towards stability even on "real" Unices.

I did nothing to configure a particular GUI plugin, and never bothered about the syntax to try other plugins. Just running WSL/U22.04.3 LTS.

 
> > In exit, there was a Segfault.
> 
> This may be #3427.

Maybe, but no other error message.

10110111:
In any case, you can just try this commit and the preceding one, and if no regression is noticed, this should be OK.

gzotti:
Yes, unrelated, more observations of `-platform wayland` (apparently now a new default) vs `-platform xcb`: Switching from windowed to fullscreen on wayland shows a black screen, then (on alt-tabbing away), 
`qt.qpa.wayland: Wayland does not support QWindow::requestActivate()`
I can resize the window now, although the mouse cursor does not change. 

xcb: just works. 

Uhm, now I just had a Segfault on exit with xcb :-( :-(. During use, I did not call any system panels (color picking, file dialog, ...)
```
StelGLWidget destroyed
Segmentation fault
```
And now Segfault on exiting the wayland run. 

Going back one commit: No segfault. Going to this commit:  the segfault seems no longer reproducible, and I also have one on exiting emacs when launched into background. Weird.

So, I think this spurious segfault may also be unrelated. OK for me.

PR #3450, opened 2023-10-08
Update SpoutLibrary

hmartinez82:
### Description
Spout2 was very old. The version was from 2016! For the Windows on ARM support a newer version of Spout2 is required.

This adds a new optional CMake parameter to tell the configuration that it's building for ARM64. Unfortunately, the MSVC toolset does not set CMake's x-compiling variables for us to detect so this new optional parameter is required. If you imagine some other way, more automatic, to detect x-compilation I'm all ears :)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
All unit tests pass

## Checklist:


alex-w (approved):
Thanks!

hmartinez82:
How do I use / test Spout?

alex-w:
Please read Stellarium User Guide

gzotti:
Are there any functional changes in Spout2 since 2016? Or is "running on Win64/ARM" the only change?

hmartinez82:
> Please read Stellarium User Guide

It's working \o/
image

Also, in the screenshot above, Stellarium is running natively as ARM64 and the receiver is running emulated, which means that even existing receivers will work if they are still x64 and not native.

hmartinez82:
> Are there any functional changes in Spout2 since 2016? Or is "running on Win64/ARM" the only change?

I'll need help from @leadedge to answer this.

leadedge:
There have been many changes since 2016 but fundamentally it works in the same way. The [SpoutLibrary example](https://github.com/leadedge/Spout2/blob/master/SPOUTSDK/SpoutLibrary/SpoutLibraryExample/src/ofApp.cpp) shows more recent functions but the existing methods can still be used. [SpoutSettings](https://github.com/leadedge/Spout2/releases) should be run initially to set registry keys.

hmartinez82:
@leadedge SpoutSettings was only required for SpoutPanel. But fi SpoutPanel is not used then running SpoutSettings once is not required? Correct? At least during my testing I did not need running SpoutSetting for receivers to recognize senders.

leadedge:
@hmartinez82 - You are right that a receiver will work without any sender selection because it will detect the sender currently active. If SpoutSettings, or SpoutPanel itself, is run just once, the path to SpoutPanel is recorded in the registry so that receivers can read the path to open it.

SpoutSettings sets up several other registry keys in "Computer\HKEY_CURRENT_USER\SOFTWARE\Leading Edge\Spout". If they are not set, defaults are used by senders and receivers. This will be OK for Stellarium on an Arm machine.

If Stellarium is updated for an x86 machine, and a user has previously opened an older settings program (then called SpoutDXmode), keys for "DirectX9" and "Memoryshare" may be set .  If there are problems, the user can run SpoutSettings which will erase these old keys and has diagnostics that are useful to trace the cause.

alex-w:
Looks like we should add some enhancements of description / notifications into Stellarium User Guide for Spout2.

leadedge:
@alex-w - I had a quick look at the user guide. The link to the Spout home page will lead to a download and it contains basic readme files. The SpoutSettings program itself has multiple help buttons.

gzotti:
@leadedge @alex-w I was not even aware of the SpoutSettings program. I will add some more description to the Guide. Else this PR is OK for me.

gzotti (approved):
Does not break Win64 build, OK for me.

PR #3414, opened 2023-09-16
Modernize Mesa library

gzotti:
Retry of #3409

### Description

This is intended to fix problems around outdated Mesa V11 libraries which come with Qt.
It provides OpenGL 4.5 CoreProfile and would allow to run advanced features (OpenGL3.3) on really old systems, 
or in some circumstances where Stellarium needs to run in a VM without GPU.

A first way is replacing the default opengl32sw.dll by another build of Mesa 20 from https://fdossena.com/?p=mesa/index.frag

Further, if users prefer, prepare program for user upgrade with current Mesa3D builds, and add instructions to SUG.

This also finally removes the deprecated "safe mode" CLI switch.

Fixes #2803

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update
- [x] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11
* Graphics Card: Geforce 3070Ti (but this explicitly runs on CPU!)
* CPU: Intel Core i7-12700. 

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (header file)
- [x] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #3452, opened 2023-10-09
Publish Windows on Arm build

hmartinez82:
Draft PR to discuss what's needed for creating the Windows on ARM installer. This requires https://github.com/Stellarium/stellarium/pull/3450 to be merged first!

### Description

The installer itself can be either x64 or x86 since what matters are the binary files it installs and they will be ARM64.
Here is what I was unable to find out:
- I noticed that OpenSSL is copied from a Stellarium repository. Does Stellarium use OpenSSL directly or is it all done via QtNetwork? Would be ok to just copy the two DLLs from AppVeyor directly instead? AppVeyor does have OpenSSL 3.1 -> https://www.appveyor.com/docs/windows-images-software/#tools . Where do these (https://github.com/Stellarium/stellarium-data/releases/tag/qt-5.6) OpenSSL DLLs come from?

- I ran `qtdiag` in an AppVeyor build. (See https://ci.appveyor.com/project/hernancmartinez/stellarium/builds/48227004#L68). I was trying to find out if AppVeyor's build of Qt was using OpenSSL or Schannel as its QtNetwork SSL backend. But here lies the issue: AppVeyor does not have OpenSSL compiled for Windows on ARM and I saw in qtdiag's output that it's using OpenSSL in its Qt's x64 build. If AppVeyor's build of Qt ARM64 is using OpenSSL, where are the OpenSSL binaries? (I can't run ARM64 qtdiag.exe in AppVeyor!) I asked them this at this post https://twitter.com/hmartinez82/status/1711263056549023835 . I don't know how to get in touch with AppVeyor and their Public Forum is not letting me create a new account! If someone here in the Stellarium project has a way to ask them that, please ask it for me!

- Where is the copying of SpoutLibrary.dll to the output folder? In my x64 builds the CMake `INSTALL` target is not copying SpoutLibrary.dll to the `bin` folder even when `ENABLE_SPOUT` is enabled. When I try to run Steallarium.exe I get the `SpoutLibrary.dll was not found` error message until I manually copy it to the bin folder.

- I never used Inno Setup before. So I may need some help, for instance, where are the installer scripts?


.appveyor.yml
```diff
@@ -20,26 +20,26 @@ max_jobs: 5
...
-    qtver: 5.12
-    qtbin: msvc2017_64
-    msvcname: Visual Studio 15 2017 Win64
-    cmake_args:
-  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
-    qtver: 5.12
-    qtbin: msvc2017
-    msvcname: Visual Studio 15 2017
-    cmake_args:
+  # - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2019
```
hmartinez82, line 23: This of course will be uncommented once the PR is ready.

.appveyor.yml
```diff
@@ -49,6 +49,7 @@ before_build:
   - set SIGNING=false
   - set INNSPATH=C:\Program Files (x86)\Inno Setup 5;C:\Program Files (x86)\Inno Setup 6
   - set PATH=C:\Qt\%qtver%\%qtbin%\bin;%INNSPATH%;%PATH%
+  - C:/Qt/6.5/msvc2019_64/bin/qtdiag.exe
```
hmartinez82, line 52: This will be removed once we solve the OpenSSL mystery

src/CMakeLists.txt
```diff
@@ -589,7 +589,7 @@ IF(APPLE)
 	  )
      ENDIF()
      # workaround for https://bugreports.qt.io/browse/QTBUG-105984
-     IF(ENABLE_MEDIA AND (${QT_VERSION} VERSION_GREATER_EQUAL "6.4.0"))
+     IF(ENABLE_MEDIA AND (${QT_VERSION} VERSION_EQUAL "6.4.0"))
```
hmartinez82, line 592: QTBUG-105984 was fixed in 6.4.1 and 6.5.0

src/CMakeLists.txt
```diff
@@ -611,10 +611,13 @@ IF(WIN32)
           SET(VIRTUAL_KBD_SETTING "--no-virtualkeyboard")
      ENDIF()
 
-     IF("${QT_PATHS}" STREQUAL "")
-          SET(_qt_paths "")
+     IF(STELLARIUM_BUILD_ARM64)
+          cmake_path(APPEND QT_PATHS ${QT6_INSTALL_PREFIX} "bin" "qtpaths.bat")
```
hmartinez82, line 615: Now that we have STELLARIUM_BUILD_ARM64 we can locally generate the QT_PATHs value without the need of explicitly set by the user

alex-w:
>I never used Inno Setup before. So I may need some help, for instance, where are the installer scripts?

it's generated by cmake (see skeleton in cmake directory) - I've added support arm64 packages for Windows into InnoSetup in master

alex-w:
> I noticed that OpenSSL is copied from a Stellarium repository. Does Stellarium use OpenSSL directly or is it all done via QtNetwork? Would be ok to just copy the two DLLs from AppVeyor directly instead? AppVeyor does have OpenSSL 3.1 -> https://www.appveyor.com/docs/windows-images-software/#tools . Where do these (https://github.com/Stellarium/stellarium-data/releases/tag/qt-5.6) OpenSSL DLLs come from?

See https://slproweb.com/products/Win32OpenSSL.html

hmartinez82:
> > I noticed that OpenSSL is copied from a Stellarium repository. Does Stellarium use OpenSSL directly or is it all done via QtNetwork? Would be ok to just copy the two DLLs from AppVeyor directly instead? AppVeyor does have OpenSSL 3.1 -> https://www.appveyor.com/docs/windows-images-software/#tools . Where do these (https://github.com/Stellarium/stellarium-data/releases/tag/qt-5.6) OpenSSL DLLs come from?
> 
> See https://slproweb.com/products/Win32OpenSSL.html

I wonder why they all it "VERY EXPERIMENTAL". OpenSSL was one of the first libraries to be ever compiled for Windows on ARM64 all the way back in 2019. I can build it locally, I mean those two DLLs.

But still. Does Stellarium use OpenSSL directly? Or all network communication is done via QtNetwork?

alex-w:
> But still. Does Stellarium use OpenSSL directly? Or all network communication is done via QtNetwork?

via QtNetwork, but winqtdeploy do not attach OpenSSL libraries, so, we should packed it by hands

hmartinez82:
> > But still. Does Stellarium use OpenSSL directly? Or all network communication is done via QtNetwork?
> 
> via QtNetwork, but winqtdeploy do not attach OpenSSL libraries, so, we should packed it by hands

Awesome, so now we need to find out if Qt 6 for Win ARM-64 that AppVeyor has uses OpenSSL or Schannel.

src/CMakeLists.txt
alex-w: OK

alex-w:
> > > But still. Does Stellarium use OpenSSL directly? Or all network communication is done via QtNetwork?
> > 
> > 
> > via QtNetwork, but winqtdeploy do not attach OpenSSL libraries, so, we should packed it by hands
> 
> Awesome, so now we need to find out if the Qt 6 for Win ARM-64 that AppVeyor has, uses OpenSSL or Schannel. If it uses Schannel then we don't have to copy the OpenSSL DLLs in the ARM64 build.

I don't have a Windows/arm64 platform, so, I cannot install SLP packages (W/arm64) and prepare archive for including arm64 libraries into installer. Do you have an arm64 device?

hmartinez82:
> > > > But still. Does Stellarium use OpenSSL directly? Or all network communication is done via QtNetwork?
> > > 
> > > 
> > > via QtNetwork, but winqtdeploy do not attach OpenSSL libraries, so, we should packed it by hands
> > 
> > 
> > Awesome, so now we need to find out if the Qt 6 for Win ARM-64 that AppVeyor has, uses OpenSSL or Schannel. If it uses Schannel then we don't have to copy the OpenSSL DLLs in the ARM64 build.
> 
> I don't have a Windows/arm64 platform, so, I cannot install SLP packages (W/arm64) and prepare archive for including arm64 libraries into installer. Do you have an arm64 device?

Yes, I have a Windows Dev Kit and a personal laptop. Can you ask AppVeyor if their build of Qt for Windows ARM 64-bit uses OpenSSL or Schannel?

alex-w:
> Yes, I have a Windows Dev Kit and a personal laptop. Can you ask AppVeyor if their build of Qt for Windows ARM 64-bit uses OpenSSL or Schannel?

According to https://www.appveyor.com/docs/windows-images-software/#tools AppVeyor don't have OpenSSL for arm64 devices

> What's SLP ?

See https://slproweb.com/products/Win32OpenSSL.html

hmartinez82:
> > Yes, I have a Windows Dev Kit and a personal laptop. Can you ask AppVeyor if their build of Qt for Windows ARM 64-bit uses OpenSSL or Schannel?
> 
> According to https://www.appveyor.com/docs/windows-images-software/#tools AppVeyor don't have OpenSSL for arm64 devices

**Correct.** That's I have been saying all along! They don't have OpenSSL for arm64, so how can it be possible that their build of Qt for arm64 *requires* OpenSSL for arm64?I need to know if they used Schannel, not OpenSSL, for QtNetwork!
> 
> > What's SLP ?
> 
> See https://slproweb.com/products/Win32OpenSSL.html

alex-w:
No OpenSSL for arm64 packages for first step. What about documentation for Qt6 for Windows/arm64?

hmartinez82:
Qt6 for Windows/arm64 says that you can use OpenSSL for Windows on ARM64. I already compiled it for ARM64 dozens of times, it's supported since 2018. The issue is knowing if AppVeyor compiled Qt with OpenSSL for Arm64 or they just used Schannel.
 
 But you are right, I'll change the build script to build the installer without OpenSSL, let's see what happens

hmartinez82:
@alex-w All those screenshots of Stellarium on Windows ARM64 in my machine were without OpenSSL, because I compiled Qt with Schannel, not OpenSSL.

hmartinez82:
The AppVeyor configuration of the ARM64 build of Qt is broken 😑
The publish build got all the way to the end, but it failed to create the `qtstuff` project because ` windeployqt` did not copy the files.
This is working in my machine, but not in AppVeyor. I found out why:
I was looking at the content of  AppVeyor's `qtpaths.bat` file. It's wrong. It's hardcoded to the wrong value:
```
type C:\Qt\6.5.2\msvc2019_arm64\bin\qtpaths.bat
@echo off
\Users\qt\work\install\bin\qtpaths6.exe -qtconf "%~dp0\target_qt.conf" %*
``` 
As you can see above, AppVeyor's `C:\Qt\6.5.2\msvc2019_arm64\bin\qtpaths.bat` is showing `\Users\qt\work\install\bin\qtpaths6.exe`, but that path does't not exist. I should be `C:\Qt\6.5.2\msvc2019_64\bin\qtpaths6.exe` 

I'm opening a Support ticket with AppVeyor. Until that's fixed windeployqt, for Qt's ARM-64 build, is totally broken :(

src/CMakeLists.txt
hmartinez82: And welp. Workaround neeeded.

gzotti:
> Question: Do we need libEGL.dll/libEGLv2.dll for Qt 6 builds? I see those files for in the installed folder

No, these (and D3Dcompiler_47.dll) are necessary only for Qt5 when running ANGLE mode. A Qt6 build should include opengl32sw.dll to be able to run software OpenGL (Mesa) mode.

alex-w:
I think we may skip mesa for Windows/arm64 at first step or we should prepare binaries for all Windows architectures

hmartinez82:
@gzotti @alex-w Sure thing. I'll try to work on that after this is merged. I just need to know does one build that opengl32sw.dll

alex-w (approved):

alex-w:
@hmartinez82 please see https://github.com/Stellarium/stellarium/pull/3414/

hmartinez82:
@alex-w @gzotti One important fact, but also a fun one.

- There's no native OpenGL in Windows on ARM. There's only support for DirectX 11 or 12
If it's an emulated app then Microsoft lets DirectX 9 apps run with a translation DLL that is only supported for emulated apps, if an app is native it must be Direct 11 or 12
- If the app must use real OpenGL then Microsoft provides, for free, this Compatibility Pack https://apps.microsoft.com/detail/9NQPSL29BFFF. In other words, we have to somehow let Windows on ARM users know that this Microsoft Store package must be installed.

The fun fact is that were talking about using Mesa's LLVMPipe to have a software renderer. So I was looking into that and found the page for Gallium D3D12 driver. I think that the "Compatibility Pack" that Microsoft provides is that just Gallium driver: https://docs.mesa3d.org/drivers/d3d12.html

hmartinez82:
Given that all Windows on ARM devices that were launched since 2018 use Snapdragon CPUs with the same Adreno GPU family I think that this Compatibility Package will always work as the OpenGL implementation. But you are right, eventually we should provide opengl32sw.dll for Windows on ARM too.

gzotti:
If it is guaranteed that all Win/arm64 systems have the d3d12 compatibility, then requiring this compatibility pack should be enough. You may add another round of tests in the startup `StelMainView::processOpenGLdiagnosticsAndWarnings()` The opengl32sw.dll is for hardware without even d3d9 compatibility (for Qt5) or for rare cases like VMs on servers without proper GPU. In this case, also the --mesa-mode in CLIProcessor should be excluded for arm builds.

PR #3463, opened 2023-10-14
Remove unused global scaling ratio

10110111:
This notion was introduced in 2013 in c3742e3a3f1403f3afb264c2f9dd0134962b4894, and the corresponding setter is never called. Given the current support for HiDPI scaling via the Qt-supplied mechanism of device pixel ratio and lack of the query of the global scaling ratio in many cases where the pixel ratio is queried, I think this global scaling needlessly complicates the logic, so should be removed.

### DIFF
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -1365,7 +1365,7 @@ void S3DRenderer::drawCoordinatesText()
 	const int fontSize=debugTextFont.pixelSize();
 	const QFontMetrics fm(debugTextFont);
 
-	const float devicePixelscaling = static_cast<float>(altAzProjector->getDevicePixelsPerPixel())*StelApp::getInstance().getGlobalScalingRatio();
+	const float devicePixelscaling = static_cast<float>(altAzProjector->getDevicePixelsPerPixel());
 	float screen_x = altAzProjector->getViewportWidth()  -  240.0f*devicePixelscaling;
 	float screen_x_cs=altAzProjector->getViewportWidth()-10.f-devicePixelscaling*qMax(240, painter.getFontMetrics().boundingRect(gridName).width());
 	float screen_y = altAzProjector->getViewportHeight() -  60.0f*devicePixelscaling;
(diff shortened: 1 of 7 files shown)


10110111:
Another approach may be to get rid of the direct use of the device pixel ratio, replacing it with this global scaling factor, and set this factor to the device pixel ratio at the initialization. This would let us roll out a custom way of handling scaling (which I actually would like to implement, because Qt's implementation sucks with its various artifacts and other shortcomings).

alex-w (approved):

gzotti:
This scaling problem is super annoying. If this global scaling ratio is never used, sure, get rid of it. Else, take care to keep scaling working on all 3 systems with all current flavours (W7/10/11; X11/wayland [does it make a difference?]; whateverMacDoes; see Qt docs and how Qt5/6 evolves to accommodate OS changes), and also take care of multi-screen setups with screens at potentially different scaling ratios. I am confined to one (large) screen where 125% is recommended, so am not too much help on difficult cases. Can test (W10/W11)+WSL(U22.04) though.

10110111:
> I am confined to one (large) screen where 125% is recommended, so am not too much help on difficult cases.

Does Windows support virtual splitting similar to FakeXRandR? Your help will definitely be needed, because I don't have any Windows newer than 7.

gzotti:
I know roughly about xrandr, but not about fakes. What would virtual splitting do?
I have seen on another W10 system (no more in my reach) two HDMI screens (4k plus some older sub-FullHD) with different scaling. I can currently test laptop+external large screen.

10110111:
> What would virtual splitting do?

It creates fake outputs, simulating several smaller monitors covering the viewport of your single large one. So you could e.g. maximize a window, and it would only expand to, say, a third of the total screen space, and then do the same for other windows on other virtual monitors.

gzotti:
Windows does have something like screen splitting where you can assemble 2-6 windows in a regular split pattern. But then you can drag the borders between those regions. I sometimes test things on 2 or 3 instances connected via RemoteSync, or run one Stellarium window plus a webbrowser for RemoteControl. The screen scaling would still globally apply to the physical screen.

PR #3468, opened 2023-10-22
Make keyboard modifiers for mouse motions exact

10110111:
### Description
Currently one can set a custom marker with e.g. <kbd>Shift</kbd>+<kbd>Ctrl</kbd> modifier, which contaminates such combos rendering them impossible to use for other purposes. These commits make the modifiers exact, i.e. for custom markers no keys except <kbd>Shift</kbd> must be pressed, so e.g. <kbd>Shift</kbd>+<kbd>Ctrl</kbd>+click doesn't set a custom marker, while <kbd>Shift</kbd>+click does. Similarly for time dragging: only <kbd>Ctrl</kbd> triggers it, not when other modifiers are pressed.

### Type of change

- [ ] Housekeeping

### DIFF
--- a/src/core/modules/CustomObjectMgr.cpp
+++ b/src/core/modules/CustomObjectMgr.cpp
@@ -58,8 +58,12 @@ void CustomObjectMgr::handleMouseClicks(class QMouseEvent* e)
 {
 	StelCore *core = StelApp::getInstance().getCore();
 	Vec3d mousePosition = core->getMouseJ2000Pos();
+	// Make sure the modifiers are exact, not just "Shift is pressed", but "Shift is pressed while Ctrl and Alt aren't"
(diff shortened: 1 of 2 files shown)


src/core/modules/CustomObjectMgr.cpp
```diff
@@ -68,7 +72,7 @@ void CustomObjectMgr::handleMouseClicks(class QMouseEvent* e)
 	}
 	// Shift + Alt + RightClick -- Removes all custom markers
 	// Changed by snowsailor 5/04/2017
```
gzotti, line 74: This comment could go away then. Also below, in line 79/83.

gzotti (approved):
Looks good to me, thanks.

PR #3474, opened 2023-10-26
GUI scalability and styling improvements

10110111:
These commits prepare the GUI for transformation into a more scaling-friendly version. In particular, I'm planning to get rid of all the hard-coded pixel-based sizes that already work badly with increased font sizes, and replace them with proper `QLayout`-based size management.

My main concern, why I'm making a PR for this change, is that something might get broken or changed in an undesirable way (such as the GPS button that became flat, which is fixed by the 4th commit).

There's also a good side effect: at least on Linux/X11 the color picking buttons are now flat-colored, showing the exact color they are set to instead of the confusing desaturated gradient as there was before.

### DIFF
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -311,40 +311,47 @@ QTabBar QToolButton {
 	background-color: rgb(31, 31, 31);
 }
 
-
 /*
- * QPushButton
+ * QPushButton and QToolButton
  */
 
+QToolButton {
+	padding: 2px;
+}
+
+
 QPushButton {
+	padding: 2px 8px;
+}
+
+QPushButton, QToolButton {
 	border: 1px solid rgb(0, 0, 0);
 	border-radius: 2px;
-	padding: 2px 8px;
 	color: rgb(31, 31, 31);
         font-size: 100%;
 	background-color: qlineargradient(x1: 0, y1: 0, x2: 0, y2: 1, stop: 0 rgb(183, 184, 185), stop: 1 rgb(111, 113, 114));
 }
 
-QPushButton:disabled {
+QPushButton:disabled, QToolButton:disabled {
         color: rgb(61, 61, 61);
         border: 1px solid rgb(61, 61, 61);
 	background: qlineargradient(x1: 0, y1: 0, x2: 0, y2: 1, stop: 0 rgb(150, 150, 153), stop: 1 rgb(120, 120, 120));
 }
 
-QPushButton:flat {
+QPushButton:flat, QToolButton:flat {
 	border: none; /* no border for a flat push button */
 	background: transparent;
 }
 
-QPushButton:hover {
+QPushButton:hover, QToolButton:hover {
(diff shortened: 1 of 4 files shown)


src/gui/LocationDialog.cpp
```diff
@@ -175,7 +175,7 @@ void LocationDialog::createDialogContent()
 	ui->gpsToolButton->setText(q_("Get location from GPS"));
 #endif
 	connect(ui->gpsToolButton, SIGNAL(toggled(bool)), this, SLOT(gpsEnableQueryLocation(bool)));
-	ui->gpsToolButton->setStyleSheet(QString("QToolButton{ background: gray; }")); // ? Missing default style?
+	ui->gpsToolButton->setStyleSheet("");
```
alex-w, line 178: Apparently, this line should be removed

src/gui/LocationDialog.cpp
```diff
@@ -975,7 +975,7 @@ void LocationDialog::resetGPSbuttonLabel()
 #else
 	ui->gpsToolButton->setText(q_("Get location from GPS"));
 #endif
-	ui->gpsToolButton->setStyleSheet(QString("QToolButton{ background: gray; }"));
+	ui->gpsToolButton->setStyleSheet("");
```
alex-w, line 978: Apparently, this line should be removed

10110111: While you were right about the creation of dialog content, here the style must be actively reset after having been set to a red or green color.

alex-w (approved):

data/gui/normalStyle.css
```diff
@@ -758,20 +765,26 @@ QScrollBar::left-arrow {
...
 QSlider {
 	background: none;
+	height: 14px;
 }
 
 QSlider::groove:horizontal {
 	border: none;
-	height: 4px; /* the groove expands to the size of the slider by default. by giving it a height, it has a fixed size */
-	background: qlineargradient(x1: 0, y1: 0, x2: 0, y2: 1, stop: 0 rgb(0, 0, 0), stop: 1 rgb(54, 54, 54));
+	background: qlineargradient(x1: 0, y1: 0, x2: 0, y2: 1,
```
gzotti, line 773: where is this used? ( cannot find a GUI slider that looks differently now. )

10110111: This was the point: to avoid any changes (i.e. regressions) at the 100% scale, and at the same time improve the looks when all `px` sizes are multiplied by some factor. This is not used anywhere yet, but I have a WIP branch, [scalable-gui](https://github.com/10110111/stellarium/tree/scalable-gui), where this is important (in this branch [one commit](https://github.com/10110111/stellarium/commit/e6ff0b81e64420dba30e765ead4514f2e0ef72e9) sets a hard-coded 150% factor, which lets me test all this).

gzotti: OK, no further questions. The new color buttons are excellent, thanks!

10110111: > no further questions

Does this imply you approve the PR?

gzotti: Yes, there is just no "review" button visible after first commenting :-o . Approved.

PR #3480, opened 2023-10-29
Title bar deduplication

10110111:
This PR removes the common part of all dialogs, which sometimes appears to be different while it shouldn't, and replaces it with a single separately defined widget: `TitleBar`. This change is supposed to preserve the look of all dialogs, though the layout of the title bar is now simplified a bit under the hood, which might lead to a couple pixel shift of the close button (I hope it's negligible).


plugins/Satellites/src/gui/SatellitesImportDialog.cpp
```diff
@@ -96,9 +96,9 @@ void SatellitesImportDialog::createDialogContent()
 		connect(gui, SIGNAL(flagUseKineticScrollingChanged(bool)), this, SLOT(enableKineticScrolling(bool)));
 	}
 
-	connect(ui->closeStelWindow, SIGNAL(clicked()),
-	        this, SLOT(close()));
-	connect(ui->TitleBar, SIGNAL(movedTo(QPoint)),
+	connect(ui->titleBar, &TitleBar::closeClicked,
```
gzotti, line 99: You could place these on one line each IMO.

plugins/SolarSystemEditor/src/gui/manualImportWindow.ui
```diff
@@ -18,106 +18,10 @@
...
-          <width>40</width>
-          <height>20</height>
-         </size>
-        </property>
-       </spacer>
-      </item>
-      <item>
-       <widget class="QLabel" name="stelWindowTitle">
-        <property name="text">
-         <string notr="true" extracomment="The title of the window will be set during runtime">Define an object</string>
```
gzotti, line 81: Should this string not be copied over to the new TitleBar.title string?

plugins/SolarSystemEditor/src/gui/mpcImportWindow.ui
```diff
@@ -28,107 +28,11 @@
...
-          <width>40</width>
-          <height>20</height>
-         </size>
-        </property>
-       </spacer>
-      </item>
-      <item>
-       <widget class="QLabel" name="stelWindowTitle">
-        <property name="text">
-         <string extracomment="The title of the window will be set during runtime">Import data</string>
```
gzotti, line 91: ditto here, copy to new TitleBar.string?

gzotti (changes requested):
That's a great simplification, thanks. 
But please update class info in StelDialog.hpp about using BarFrame

plugins/SolarSystemEditor/src/gui/manualImportWindow.ui
10110111: My bad, didn't notice there was something beside the comment.

10110111:
> please update class info in StelDialog.hpp about using BarFrame

Actually, I now think the `BarFrame` as a separate entity is no longer useful. But I suppose external plugins might depend on it for their dialogs. Can I just remove `BarFrame` (merging the functionality into `TitleBar`), or should I leave a type alias?

gzotti:
> > please update class info in StelDialog.hpp about using BarFrame
> 
> Actually, I now think the `BarFrame` as a separate entity is no longer useful. But I suppose external plugins might depend on it for their dialogs. Can I just remove `BarFrame` (merging the functionality into `TitleBar`), or should I leave a type alias?

Good question. You see the age of this code. I had thought you had extended/modified/replaced the BarFrame already (just a named Widget in my understanding) into a dedicated class. Now I saw it's still there as base class of TitleBar.  

Given that "external plugins" (in my understanding dynamic plugins built separately) has been a broken feature on Windows at least since 2010 (so I never saw them), I cannot answer whether anybody depends on it, but it seems there is just now revived activity around the topic (Linux only?). Which external/dynamic plugins do you know? Also @alex-w ?

In any case, this architecture/usage description must be updated.

10110111:
> Which external/dynamic plugins do you know?

I don't know of any. The recent discussion about a failure of `dynamic_cast` is the first time I heard about such a thing and even about the template project.

gzotti:
I just see a slight difference: Title bar font. In some dialogs (View, Location, Time) it used to be boldface. It's good to have same style for all, but given it's a window title, maybe used boldface globally? (Or @alex-w, please your opinion?)

10110111:
> In some dialogs (View, Location, Time) it used to be boldface

Actually, the CSS was supposed to set it (even before this PR). But this failed to be achieved, which is why some dialogs may not have had this setting applied (others had it duplicated in the `*.ui` files). The problem is that the corresponding CSS entry reads `font-style: bold;` instead of the correct `font-weight: bold;`. I'll fix it now.

gzotti (approved):
Fine for me now, thanks!

Issue #3481, opened 2023-10-29
23.3: Text User Interface missing


gzotti:
Preparing for a presentation in a planetarium next week where we are limited to using a downloadable installer. I just realized that the TUI plugin is missing!

### Expected Behaviour

Activate TUI plugin and use it

Configure date and minimal object info being displayed around the circumference

### Actual Behaviour

TUI plugin not included 
-->No date/object info for the auditorium.

TUI has been silently marked as candidate for removal and taken away from 23.3 without warning. 
I agree it is archaic and ill-maintained. But this just reflects our (3) availability inside or access to planetaria. It has important features after all (these info lines and sky background color twiddling for artists). I was about to add some long-missing features (property use to flip those info bits). But we cannot use my laptop there, so I am confined to a release (or beta installer).

### Steps to reproduce
Install 23.3. Try to find TUI plugin.

### System
* Stellarium version: 23.3
* Operating system: all

PR #3483, opened 2023-10-29
TUI fixes

gzotti:
### Description

After my surprise about #3481 I decided to push this branch a few days earlier than anticipated.

This is not a complete fix/rethinking of the TUI plugin, just a few updates.
Keyboard action toggles for
- show time
- show object info

Still missing is any type of config GUI, at least with those two flag toggles.

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: Geforce, but irrelevant 

## Checklist:

### DIFF
--- a/plugins/TextUserInterface/src/TextUserInterface.cpp
+++ b/plugins/TextUserInterface/src/TextUserInterface.cpp
@@ -575,6 +575,8 @@ void TextUserInterface::init()
 
 
 	currentNode = m1;
+
+	addAction("actionShow_TUI_dateTime", N_("Text User Interface"), N_("Toggle TUI date&time"), this, "flagShowDateTime", "Ctrl+Alt+T");
 }
 
 /*************************************************************************
@@ -588,8 +590,10 @@ void TextUserInterface::loadConfiguration(void)
 	font.setPixelSize(conf->value("tui/tui_font_size", 15).toInt());
 	tuiDateTime = conf->value("tui/flag_show_tui_datetime", false).toBool();
 	tuiObjInfo = conf->value("tui/flag_show_tui_short_obj_info", false).toBool();
-	tuiGravityUi = conf->value("tui/flag_show_gravity_ui", false).toBool();
+	setTuiGravityUi(conf->value("tui/flag_show_gravity_ui", false).toBool());
 	color = Vec3f(conf->value("tui/tui_font_color", "0.3,1,0.3").toString());
+	StelCore *core=StelApp::getInstance().getCore();
+	connect(core, SIGNAL(flagGravityLabelsChanged(bool)), this, SLOT(setTuiGravityUi(bool)));
 }
 
 /*************************************************************************
--- a/plugins/TextUserInterface/src/TextUserInterface.cpp
+++ b/plugins/TextUserInterface/src/TextUserInterface.cpp
@@ -589,8 +589,8 @@ void TextUserInterface::loadConfiguration(void)
 	Q_ASSERT(conf);
 
 	font.setPixelSize(conf->value("tui/tui_font_size", 15).toInt());
-	tuiDateTime = conf->value("tui/flag_show_tui_datetime", false).toBool();
-	tuiObjInfo = conf->value("tui/flag_show_tui_short_obj_info", false).toBool();
+	setTuiDateTime(conf->value("tui/flag_show_tui_datetime", false).toBool());
+	setTuiObjInfo(conf->value("tui/flag_show_tui_short_obj_info", false).toBool());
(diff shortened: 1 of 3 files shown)


gzotti:
Problem: I need a Windows install package with TUI tomorrow evening. What was the incantation?

gzotti:
Problem 2: The Regular expressions to remove the HTML decoration in StelObject::postProcessInfoString() is still not good, and it shows in the result. What to update, and where else does it appear (will something else break instead?)

10110111:
> Problem: I need a Windows install package with TUI tomorrow evening. What was the incantation?

`[publish]` in commit message.

10110111:
> The Regular expressions to remove the HTML decoration in StelObject::postProcessInfoString() is still not good, and it shows in the result.

What shows in the result? What is the input?

gzotti:
If you run this, 
configure hotkey to toggle object info, then  just select e.g. Jupiter and see the green insert. 
Go into planetarium settings (F2/Tools) and activate gravity labels and disc viewport to see what I need to deliver.

A date should be visible to the visitors. That works. But the object info shows HTML extras. (`<table style='...'>`)

Ah -- you also need to activate F2/Extras/Use formatting.

10110111:
Well, the only tag I see slipping through is `<table style='...'>`, which isn't handled by any of the regexes. The one relevant to tables, `tableRe2`, doesn't match the `style` attribute.

10110111:
The table with a style can be removed with the following change:
```diff
diff --git a/src/core/StelObject.cpp b/src/core/StelObject.cpp
index a1008d44e8..6fb81e9cc3 100644
--- a/src/core/StelObject.cpp
+++ b/src/core/StelObject.cpp
@@ -961,6 +961,8 @@ void StelObject::postProcessInfoString(QString& str, const InfoStringGroup& flag
     static const QRegularExpression brRe2("<br(\\s*/)?>\\s*");
     static const QRegularExpression tdRe("<td(\\w*)?>");
     static const QRegularExpression tableRe2("<table(\\w*)?>");
+    static const QRegularExpression tableRe3("<table style='[^']*'>");
+    static const QRegularExpression tableRe4("<table style=\"[^\"]*\">");
     str.replace(tableRe, "<table");
     // chomp trailing line breaks
     str.replace(brRe, "");
@@ -977,6 +979,8 @@ void StelObject::postProcessInfoString(QString& str, const InfoStringGroup& flag
         str.replace("<td>", "");
         str.replace("</tr>", "\n");
         str.replace(tableRe2, "");
+        str.replace(tableRe3, "");
+        str.replace(tableRe4, "");
         str.replace("</table>", "");
     }
     else if(!(flags&NoFont))
```

gzotti:
> Well, the only tag I see slipping through is `<table style='...'>`, which isn't handled by any of the regexes. The one relevant to tables, `tableRe2`, doesn't match the `style` attribute.

Yes, same conclusion here.  The "\\w" is "word characters", right? Maybe spaces and quote signs should be added.  I tried `"<table(['\\\"\\w\\s]*)?>"`,  that was not enough.

10110111:
> I tried "<table(['\\\"\\w\\s]*)?>", that was not enough.

Right, because the style also contains dots, semicolons etc. It's just easier to handle it as "an opening quote, anything that's not a closing quote, a closing quote", for two variants of the quotes. You can put it into a single regex by using alternation.

gzotti:
Yes, that works great! Thanks a lot!

gzotti:
Another one: 
`str.replace("<td>", "");`

 Is this not covered by tdRe? And probably also here any style element like align should be removed.

10110111:
> Is this not covered by tdRe?

Yes, it is. But again, what valid HTML tag does this regex represent aside from `<td>`?

gzotti:
`<td>` may also include style. See StelObject::getSolarLunarInfoString(). I will check other occurrences and deal with them the same way.

gzotti:
Ah yes, this could also match `<tdfoo>`. Not really good, but in effect it is only used in the HTML we are creating.

10110111:
> Ah yes, this could also match `<tdfoo>`.

Well, this is the only kind of thing that this regex could match. There's a space missing between the words and the tag name. If you do use explicit style-matching regexes as you do in the new PR, this regex is just useless.

gzotti:
I replaced the indeed useless `\\w` now with `\\s` to catch simple cases of just added space. Currently it looks good.

10110111:
Now it makes sense. But the `(\s*)?` is redundant and is equivalent (unless you really want to capture) to `\s*`.

gzotti:
Right, it's only one element.

PR #3486, opened 2023-10-30
Make a dedicated widget for color buttons

10110111:
This commit creates a dedicated widget for color picking buttons: `ColorButton`. This makes it possible to:
1. theme them all via a single CSS entry,
2. get rid of the `setProperty()` and `sender()` kludges, replacing them with data members and `this` respectively.


alex-w (approved):
Thanks!

PR #3489, opened 2023-11-01
Remove fixed sizes from Config dialog .ui file

10110111:
Most of the sizes are replaced with normal QLayout-based positioning, and a few more essential ones are moved into normalStyle.css for greater flexibility in customization.

This change is required to make it possible to resize the whole GUI when font size changes. Such scaling will let us go away from the Qt's kludge of virtual pixels, whose problem is not only that it's a kludge leading to inconsistencies in the code, but also that it leads to some blurring of text in particular situations and coarsening of mouse input events.

I tried to preserve the look of the dialog as much as possible, with one exception: the sizes of `scriptListWidget` are configured similarly to those of `pluginsListWidget`, so that the split between the list and the description panels is not in the middle, but on the left side, see the screenshots below.

I'm going to change some other dialogs in a similar way, and add a function that will multiply all pixel-based sizes in `normalStyle.css` by a scale factor. To make it all work, I propose some rules that will help maintain scalability of the GUI:
 * Whenever you want to align some widgets, don't set fixed/minimum/maximum sizes — use `QLayout` classes, in particular, `QGridLayout` is quite good for this;
 * If you must set a size in pixels (like for the `pluginsListWidget`), do it via `normalStyle.css`. This will keep these values accessible to the code that will be able to adjust them as required.

#### Old _Scripts_ tab
6a

#### New _Scripts_ tab
6b

### DIFF
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -1034,6 +1034,21 @@ QTextBrowser#deltaTAlgorithmDescription {
 	background: transparent;
 	padding: 0;
 	margin: 0;
+	max-height: 175px;
+}
+
+QListWidget#scriptListWidget {
+	min-width: 190px;
+	max-width: 190px;
+}
+
+QListWidget#pluginsListWidget {
+	min-width: 180px;
+	max-width: 180px;
+}
+
+QCommandLinkButton#getStarsButton {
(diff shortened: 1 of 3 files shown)


alex-w (approved):
Thanks!

PR #3490, opened 2023-11-03
Scale the pixel sizes in the stylesheet according to GUI font size

10110111:
This is another step into the direction of scaling the GUI without kludges. The style itself is still not perfect (namely, the spacing is not scaled yet), but this is already an improvement, see the following screenshots.

#### Original 2× font size

Screenshot_2023-11-03_11-28-57

#### This PR 2× font size

Screenshot_2023-11-03_11-29-50

#### Original 1× font size at 200% via Qt scale factor

Screenshot_2023-11-03_11-31-32

#### Original 1.5× font size

Screenshot_2023-11-03_11-39-00

#### This PR 1.5× font size

Screenshot_2023-11-03_11-38-14

#### Original 1× font size at 150% via Qt scale factor

Screenshot_2023-11-03_11-37-24

### DIFF
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -409,11 +409,18 @@ QCheckBox:disabled {
 }
 
 QCheckBox::indicator,
-QListWidget::indicator,
-QListView::indicator,
 QGroupBox::indicator {
 	width: 16px;
 	height: 16px;
+	margin: 0;
+	background: none;
+}
+
+/* FIXME: these should configured the same way as QCheckBox and QGroupBox
+ * indicators, but there's a problem with spacing discussed at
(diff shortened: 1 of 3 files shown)


alex-w (approved):

gzotti (approved):
seems ok, thanks!

10110111:
Just noticed a blocker, posted [here](https://stackoverflow.com/q/77422984/673852). You can see the problem in _Surveys_ list in the _View_ dialog.

10110111:
I added a workaround that will prevent scaling of `QListView` indicators, so that the text is still fully legible. This doesn't look great, but legibility is more important than prettiness, so this will do for now.

PR #3541, opened 2023-12-09
Fix crash in StelTexture destructor on exit

### LINKED ISSUE #3427: 23.3: crash upon exit

10110111:
On deletion of `StelApp` `StelTextureMgr` is deleted, after which some instances of `StelTexture` are deleted. But `StelTexture` destructor logs VRAM usage via `StelTextureMgr`, accessing a no longer existing object.

This commit turns the nonstatic raw pointer to texture manager into a static weak pointer (`QPointer`), so that it was possible to check whether texture manager still exists before trying to access it. Making it static is supposed to prevent useless copying of this smart pointer while the texture manager is effectively a singleton.

Fixes #3427

### DIFF
--- a/src/core/StelTexture.cpp
+++ b/src/core/StelTexture.cpp
@@ -39,13 +39,19 @@
 # define GL_TEXTURE_MAX_ANISOTROPY 0x84FE
 #endif
 
-StelTexture::StelTexture(StelTextureMgr *mgr)
-	: textureMgr(mgr)
+QPointer<StelTextureMgr> StelTexture::textureMgr;
+
+StelTexture::StelTexture()
 {
 }
 
 StelTexture::~StelTexture()
 {
+	// Don't try to access texture manager or OpenGL context if texture
+	// manager has been deleted, because at this point we are shutting
+	// down, and the OpenGL context may have also been deleted.
+	if (!textureMgr) return;
+
 	if (id != 0)
 	{
 		/// FS: make sure the correct GL context is bound!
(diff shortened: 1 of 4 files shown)


gzotti:
Good find!

However, it may not have been all to solve. It still reported a crash.

image

10110111:
The new version should fix this too.

alex-w:
It's still crahes on `glIsTexture()` line

10110111:
Did you update? It shouldn't even reach this line any more.

alex-w:
> Did you update? It shouldn't even reach this line any more.

It was for previous code, let me check it again

alex-w (approved):
Everything OK now, thanks!

gzotti:
Confirmed also in W11. I wonder why it takes minutes to shutdown the application when running a debug session. But finally it came to an end without error.

PR #3585, opened 2024-01-10
[CI] unit testing on not x86 architectures

alex-w:
### Description
This pull request contains GitHub actions for unit testing on popular not x86 architectures - armv7 (32-bit ARM), aarch64 (64-bit ARM), riscv64 (64-bit RISC-V) - for Qt5-based and Qt6-based builds at Ubuntu 22.04. Of course, the building of source within QEMU'ed dockers is slow in comparison to host systems, so we may enable it via special keyword in git messages (an optional unit testing).

Important note: Stellarium is configured without ShowMySky and QtWebEngine support

### Type of change

- [x] Housekeeping

### DIFF
--- a/.github/workflows/ci-more.yml
+++ b/.github/workflows/ci-more.yml
@@ -11,13 +11,12 @@ on:
 
 jobs:
   # CI on Linux (Qt5)
-  # Note: cmake in U20.04/armhf (armv7) is broken
   ci-linux-qt5:
     strategy:
       matrix:
-        arch: [aarch64, riscv64]
+        arch: [armv7, aarch64, riscv64]
     name: "Linux (${{ matrix.arch }}; qt5)"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     if: "!contains(github.event.head_commit.message, '[ci skip]') && !contains(github.event.head_commit.message, '[skip ci]') && !contains(github.actor, 'transifex')"
 
     steps:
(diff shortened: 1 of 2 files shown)


gzotti:
Do I understand that now it is configured to run at each commit that is not from transifex and that not deliberately disables the test by "[ci skip]"/"[skip ci]" commit messages? Maybe invert to only run (as indicated) if explicitly wanted. ("[ci test-extra]" or such?) (Of course, all these extra messages that trigger special CI actions (e.g. "actually build a downloadable installer package artifact") should be listed somewhere. I cannot find them now.)

alex-w:
> Do I understand that now it is configured to run at each commit that is not from transifex and that not deliberately disables the test by "[ci skip]"/"[skip ci]" commit messages? 

Yes.

> Maybe invert to only run (as indicated) if explicitly wanted. ("[ci test-extra]" or such?) 

Yes, we can invert logic and run unit tests when explicitly wanted.

> (Of course, all these extra messages that trigger special CI actions (e.g. "actually build a downloadable installer package artifact") should be listed somewhere. I cannot find them now.)

"[ci skip]" and "[skip ci]" are standard keywords for many CI environments, but "[publish]" is an own invention, of course. Well, we can add them in wiki.

alex-w:
Hint: https://github.com/Stellarium/stellarium/wiki/Git-Contributor-Workflow#handling-commit-messages

gzotti (approved):
Looks good to me. I cannot test whatever riscv64 produces, though.

alex-w:
> Looks good to me. 

Wait, do you want to the current logic (unit tests for every commit) or inverted logic - unit tests for selected commits?

> I cannot test whatever riscv64 produces, though.

But I do not prepare a downloadable packages for RISC-V (or ARM) devices :)

PR #3599, opened 2024-01-19
[CI] unit testing in FreeBSD and OpenBSD systems

alex-w:
### Description
This pull request added support to build and run unit tests in FreeBSD and OpenBSD systems.

Important part of this PR is updated unit tests for JavaScripting (The deep investigation of test fails is needs for FreeBSD/Qt5 CI). 

### Type of change
- [x] Housekeeping


10110111:
Why do you disable ShowMySky here?

alex-w:
> Why do you disable ShowMySky here?

FreeBSD has binary package for Qt5-based ShowMySky only. At first attempt, building ShowMySky in FreeBSD/Qt6 was failed :(

Let's enable it again and please check the results.

10110111:
> make[2]: make[2]: don't know how to make _deps/showmysky-qt6-build/check\. Stop

I guess this is because of the spaces in `check version update` target. Try using the latest `master` of CalcMySky, I've pushed a potential fix for this.

alex-w:
> I guess this is because of the spaces in `check version update` target. Try using the latest `master` of CalcMySky, I've pushed a potential fix for this.

You can freely change this branch/PR to test your changes

alex-w:
OK, CalcMySky/ShowMySky can be 0.3.2 now.

alex-w:
OK.

NetBSD: no exist binary package of MesaLib for NetBSD 9.x and it's not possible to build it from source code (via pkgsrc) due limitations for VM, so, the critical dependency is missing - no OpenGL, no Stellarium.

DragonflyBSD: Stellarium built without ShowMySky and Lens Distortion Estimator plugin, because binary packages of dependencies for these features is too old. Building from source code is failing (probably GCC 8.4 is too old for it).

alex-w:
OK.

Solaris: version 11.4 was released in 2018 and it has too old packages (cmake 3.9), so, I removed this CI

alex-w:
@gzotti @10110111 please review the PR

src/StelLogger.cpp
```diff
@@ -214,11 +214,7 @@ void StelLogger::init(const QString& logFilePath)
 	dmesg.waitForStarted();
 	dmesg.waitForFinished();
 	const QString dmesgData(dmesg.readAll());
-	#if (QT_VERSION>=QT_VERSION_CHECK(5, 14, 0))
```
gzotti, line 217: Not sure why you removed the branches. I think it may have to be reversed though. Qt::SkipEmptyParts was introduced in Qt5.14. (https://doc.qt.io/qt-5/qt.html#SplitBehaviorFlags-enum). Is Q_OS_BSD4 guaranteed to build on Qt5.14+?

src/scripting/V3d.hpp
```diff
@@ -122,6 +122,8 @@ public slots:
 	//! Formats a comma-separated string in angle brackets
 	//! @todo currently this seems to to nothing. JS print output shows "V3f(address)". Use toVec3f() to show the values formatted by Vec3f's method.
 	Q_INVOKABLE QString toString() const {return QString("[%1, %2, %3]").arg(m_x).arg(m_y).arg(m_z);}
+	//! Formats a hex string
```
gzotti, line 125: Please provide better documentation. What is this for, HTML color formatting?

src/scripting/V3d.cpp
```diff
@@ -63,6 +63,14 @@ V3f & V3f::operator =(const V3f &v)
 	return *this;
 }
 
+QString V3f::toHex() const
+{
+	return QString("#%1%2%3")
+		.arg(qMin(255, int(m_x * 255)), 2, 16, QChar('0'))
```
10110111, line 69: This should be rounded instead of truncating.

gzotti:
I wonder how many kWh per commit we are burning... And how much of that is just "install environment" before even compilation... :-o

src/StelLogger.cpp
alex-w: You can ignore this part, because in the branch "refactoring/logger" I've rewriting the part of logger for getting hardware info without calling CLI tools.

src/scripting/V3d.hpp
alex-w: Yes, I've just added missing function (copy + paste + fix types)

src/scripting/V3d.cpp
alex-w: You're right!

gzotti (approved):
I see no problems now.

10110111:
Some builds fail. Is it expected?

10110111:
Why did you have to edit the tests in this PR?

alex-w:
> Some builds fail. Is it expected?

it's something wrong in memory with Qt5/QtScript - I didn't checked it in deep

alex-w:
> Why did you have to edit the tests in this PR?

wrong limits, so, some parts of tests never running before

PR #3614, opened 2024-01-30
Updating StelLogger

alex-w:
### Description
StelLogger previously obtaining many of hardware info through calling CLI tools, and it was caused persistent problem in some environments. Currently, these parts of code rewriting to using platform-specific API/code and format of data has unification form now. Of course, we can expand collected data. 

Information for compiler name and version is coming from CMake now, and all related C++ code was removed.

The code is tested in these operating systems: Windows, macOS, Linux (Ubuntu and Raspberry OS), FreeBSD, OpenBSD, NetBSD, OpenIndiana, Haiku.

### Type of change

- [x] Housekeeping

### How Has This Been Tested?
1. Run Stellarium
2. Open Help windows, Log section 
3. See upper 10-12 lines of log

## Checklist:


alex-w:
OK, I'll install OpenBSD and run Stellarium tonight

alex-w:
OK, OpenBSD is done

alex-w:
@gzotti @10110111 @worachate001 please test and review this PR

src/StelLogger.cpp
```diff
@@ -44,193 +78,352 @@ void StelLogger::init(const QString& logFilePath)
...
+	}
 
-	HKEY hKey = Q_NULLPTR;
-	DWORD dwType = REG_DWORD;
-	DWORD numVal = 0;
-	DWORD dwSize = sizeof(numVal);
+	// memory info
+	struct sysinfo memInfo;
+	sysinfo (&memInfo);
+	long long totalRAM = memInfo.totalram;
```
10110111, line 280: This should be unsigned. And maybe better to use a sized typedef, like `uint64_t`.

src/StelLogger.cpp
```diff
@@ -44,193 +78,352 @@ void StelLogger::init(const QString& logFilePath)
...
+	sysctlbyname("machdep.tsc_freq", &freq, &len, nullptr, 0);
+	writeLog(QString("Processor speed: %1 MHz").arg(freq/1000000));
+
+	int ncpu = 0;
+	len = sizeof(ncpu);
+	sysctlbyname("hw.ncpu", &ncpu, &len, nullptr, 0);
+	writeLog(QString("Processor logical cores: %1").arg(ncpu));
+
+	// memory info
+	int64_t totalRAM = 0;
```
10110111, line 318: This should be unsigned.

src/StelLogger.cpp
```diff
@@ -44,193 +78,352 @@ void StelLogger::init(const QString& logFilePath)
...
 	#else
-	QStringList systemLines = systemData.split('\n', QString::SkipEmptyParts);
+	mib[1] = HW_PHYSMEM;
 	#endif
-	for (int i = 0; i<systemLines.size(); i++)
-	{
-		// hardware overview
-		if(systemLines.at(i).contains("Model", Qt::CaseInsensitive))
-			writeLog(systemLines.at(i).trimmed());
+	int64_t totalRAM = 0;
```
10110111, line 356: This should be unsigned.

src/StelLogger.cpp
```diff
@@ -44,193 +78,352 @@ void StelLogger::init(const QString& logFilePath)
...
+	// Idea and some code is catching from Haiku OS (MIT license)
+	uint32 topologyNodeCount = 0;
+	cpu_topology_node_info* topology = NULL;
 
-		if(systemLines.at(i).contains("VRAM", Qt::CaseInsensitive))
-			writeLog(systemLines.at(i).trimmed());
-	}
+	get_cpu_topology_info(NULL, &topologyNodeCount);
+	if (topologyNodeCount != 0)
+		topology = new cpu_topology_node_info[topologyNodeCount];
```
10110111, line 383: Better use a smart pointer here, e.g. `std::unique_ptr`.

src/StelLogger.cpp
```diff
@@ -44,193 +78,352 @@ void StelLogger::init(const QString& logFilePath)
...
+	writeLog(QString("Processor name: %1 %2 @ %3").arg(get_cpu_vendor_string(cpuVendor), get_cpu_model_string(platform, cpuVendor, cpuModel), clockSpeed));
+	writeLog(QString("Processor speed: %1 MHz").arg(frequency));
+
+	system_info hwinfo;
+	get_system_info(&hwinfo);
+
+	writeLog(QString("Processor logical cores: %1").arg(hwinfo.cpu_count));
 
+	// memory info
+	int totalRAM = round(hwinfo.max_pages*B_PAGE_SIZE/1048576.0 + hwinfo.ignored_pages*B_PAGE_SIZE/1048576.0);
```
10110111, line 425: `int` is not future-proof. In principle a 64-bit system may have up to 2<sup>64</sup> bytes of RAM, which is 2<sup>44</sup> megabytes—not representable in a 32-bit `int`.

src/StelLogger.cpp
alex-w: Why? I don't think using std::unique_ptr is better here

10110111: Because generally it's better to avoid the need to worry about `delete` or `delete[]`, and let the compiler (destructor) do this for you. In fact, I didn't think trough the case here, the best would be to use `std::vector<cpu_topology_node_info>`, since this would show the semantics better.

alex-w: `get_cpu_topology_info` method can work with standard pointers, but not `std::unique_ptr` nor `std::shared_ptr` - I don't want to touch a haiku kernel.

10110111: You don't need to touch the dependencies. Both `unique_ptr` and `shared_ptr` (which would be a worse choice) provide a `get()` method that returns a raw pointer that you can pass to any C API without giving up ownership. Same for Qt's smart pointers (since Qt 5.11).
If you use `std::vector` instead, then the method to get a pointer is `std::vector::data()` or simply `&myvector[0]`.

10110111 (approved):

PR #3644, opened 2024-02-26
Faster OBJ import

### LINKED ISSUE #3620: Use fast_float library

Sometimes loading large files, esp. large landscape OBJ in scenery3D, takes a long time.

The fast_float library  https://github.com/fastfloat/fast_float provides float/double parsing with much higher throughput than stdlib. We should incorporate that library and change the OBJ loader to use it. 

Currently we just have to sit and wait. 

**Additional context**

Maybe also ssystem_minor.ini parsing would become faster, to be felt esp. with a large number of objects.

gzotti:
### Description

Loading large OBJ files takes too long, easily more than a minute. With this change loading can be improved.

I had assumed it is the string manipulation, splitting/parsing/conversion to floats. Ruslan found the true cause.
To sum this up, the worst offender was using the QRegularExpression instead of simple space separator.

As a small change, I think this can be squashed and merged. 

Fixes #3620 

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Load a large scene, compare loading times. 

**Test Configuration**:
* Operating system: Windows 10/11, Ubuntu 22.04LTS
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/StelOBJ.cpp
+++ b/src/core/StelOBJ.cpp
@@ -384,7 +384,7 @@ bool StelOBJ::parseFace(const ParseParams& params, const V3Vec& posList, const V
 
 	// Find all slashes without allocating memory, including allocation for storing
 	// the result (we assume that the output vector has a reasonable amount of space reserved)
-	const auto findSlashes = [](const QStringRef& line, std::vector<int>& slashes)
+	const auto findSlashes = [](const QStringView& line, std::vector<int>& slashes)
 	{
 		slashes.clear();
 		for(int n = 0; n < line.size(); ++n)
--- a/src/core/StelOBJ.cpp
+++ b/src/core/StelOBJ.cpp
@@ -382,58 +382,62 @@ bool StelOBJ::parseFace(const ParseParams& params, const V3Vec& posList, const V
 	//note: the indices start with 1, this is fixed up later
 	#define FIX_REL(a, list) if(a<0) {a += list.size()+1; }
 
+	// Find all slashes without allocating memory, including allocation for storing
+	// the result (we assume that the output vector has a reasonable amount of space reserved)
+	const auto findSlashes = [](const QStringRef& line, std::vector<int>& slashes)
+	{
+		slashes.clear();
+		for(int n = 0; n < line.size(); ++n)
+			if(line[n] == QLatin1Char('/'))
+				slashes.push_back(n);
+	};
+
+	thread_local std::vector<int> slashes;
+	slashes.reserve(4);
 	//loop to parse each section separately
 	for(int i =0; i<vtxAmount;++i)
 	{
-		if(!params[i+1].contains('/'))
+		const auto& vertStr = params[i+1];
+		findSlashes(vertStr, slashes);
+
+		switch(slashes.size())
 		{
-			//no slash, only position
-			CHK_MODE(1)
-			CHK_OK(posIdx = params[i+1].toInt(&ok));
-			FIX_REL(posIdx, posList)
-		}
-		else
-		{
-			//split on slash
-			#if (QT_VERSION>=QT_VERSION_CHECK(6,0,0))
-			ParseParams split = params.at(i+1).split('/').toVector();
-			#else
-			ParseParams split = params.at(i+1).split('/');
-			#endif
-			switch(split.size())
-			{
-				// case 1 is handled separately above
-				case 2: //single slash, vert/tex
-					CHK_MODE(2)
-					CHK_OK(posIdx = split.at(0).toInt(&ok));
+			case 0:
+				//no slash, only position
+				CHK_MODE(1)
+				CHK_OK(posIdx = vertStr.toInt(&ok));
+				FIX_REL(posIdx, posList)
+				break;
+			case 1: //single slash, vert/tex
+				CHK_MODE(2)
+				CHK_OK(posIdx = vertStr.mid(0, slashes[0]).toInt(&ok));
+				FIX_REL(posIdx, posList)
+				CHK_OK(texIdx = vertStr.mid(slashes[0] + 1, vertStr.size() - (slashes[0] + 1)).toInt(&ok));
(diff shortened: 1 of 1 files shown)


10110111:
Do you have an example of a large file to load?

10110111:
OK, from my experiments with Callanish's OBJ file (and Qt 5.15), here's what I got on this branch:

* Full file loads in 40 s
* Stubbing out `parseVec3` from the `"v"` handling reduces reading time to 34 seconds
* Stubbing out `parseFace` from the `"f"` handling in addition to `parseVec3` from the `"v"` handling reduces reading time to 24 seconds
* Resetting the code and stubbing out everything after `QString line = stream.readLine().trimmed()` reduces reading time to 1.6 s
* Allowing `line.splitRef(separator, Qt::SkipEmptyParts)` to run and stubbing everything after it bumps reading time to 24 s.
* Replacing the `separator` regex with a `QChar(' ')` in addition to the above stubbing reduced reading time to 3.8 s.
* Doing only the replacement of the separator after code reset results in total loading time of 16 s, which alone is quite an improvement over the initial 40 s.

So, looks like the first thing to optimize is the regular expression. Particularly, the separator as used, `\s`, is too general for the OBJ file. Just a simple space with skipping of empty parts should be sufficient:

```diff
diff --git a/src/core/StelOBJ.cpp b/src/core/StelOBJ.cpp
index fb8f19627d..f0fda5eb54 100644
--- a/src/core/StelOBJ.cpp
+++ b/src/core/StelOBJ.cpp
@@ -814,8 +814,7 @@ bool StelOBJ::load(QIODevice& device, const QString &basePath, const VertexOrder
 
     VertexCache vertCache;
     CurrentParserState state = CurrentParserState();
-    static const QRegularExpression separator("\\s");
-    separator.optimize();
+    static const QChar separator(' ');
 
     int lineNr=0;
 
```

The parsing of the numbers may actually not even be so important, the real bottlenecks may be elsewhere.

gzotti:
Great! Thanks for experiments. My second idea (after this) was to skip the high-level line splitting entirely and read the lines into char* directly. Probably the new "space only" regexp can go into master for immediate implosion in 24.1. I will test with some of my large scenes.

10110111:
> skip the high-level line splitting entirely and read the lines into char* directly

I'm afraid the next major bottleneck is not in file IO. And the splitting here is done via `QStringRef`, which doesn't allocate anything in the heap.

> the new "space only" regexp

The point is that it's no longer a regexp, just a fixed char (see the diff above).

10110111:
I pushed the regex-to-char change.

I also added one more commit, which is not quite complete—in a way. Namely, it improves loading of vertex-only faces while keeping the other combinations of vertex, texture, normal unchanged. Ideally, these should also be changed.

And again, this shows that real bottlenecks are not in number parsing, so I think we can safely remove your first commit.

10110111:
Now I've pushed the full handling of all cases. Please check how it behaves with textured and normal-mapped models if you have any.

gzotti:
Hopefully later this week, thank you!

10110111:
BTW, the `vertCache` is another source of slowdown. I don't think it should even exist in this code, because normally the model should already be optimized to avoid duplication of vertices, and if it does duplicate them, this may have been done on purpose by the exporting software (e.g. to optimize for GPU cache locality).

gzotti:
Next short view. Hm, the QStringRef is part of Qt5compat which I was happy to get rid of before 1.0. But Qtdocs recommend it explicitly for speeding up parsers etc, just this case. I hope this module will not fall away in 2 years.
Have you tried also with QStringView? It compiles with Qt6.6, but I did not install Qt5compat to make comparisons against QStringRef so far.

10110111:
> the QStringRef is part of Qt5compat which I was happy to get rid of before 1.0

Well, `QString::splitRef` is also part of this API, and the calls to it are still present in current `master`. This is why I continued using it. Ideally, indeed, I'd just use `QStringView`, which is basically the same as `QStringRef`.

10110111:
Oh, actually I'm wrong: this API is only used in the Qt5-specific branches. OK, I'll try to switch to `QStringView` in Qt6 parts, though I really dislike this dependence on Qt version in such a platform-independent code.

gzotti:
Did you notice any performance change? If not, probably all occurrences of QStringRef can be switched to QStringView (to remove the `#ifdef Qt6/Qt5` branching), but I think in my edits towards Qt6 I came across some cases where I needed to even move to QString. And probably we should use QStringViews more frequently in many other cases where the string content doesn't change.

10110111:
> probably all occurrences of QStringRef can be switched to QStringView

Unfortunately, `QStringView` in Qt5 lacks some methods that are present in `QStringRef`, so we'll have to keep this conditional compilation as is for now.

gzotti:
OK, I hoped for that. Just the description of QStringRef was so focused on highlighting the use for more efficient parsing.

gzotti:
Current state: 42s vs. 70s when loading another large scene. Already a big win. 

Loading Callanish fresh after program start:
23.4: 34s
This: 22s
This with from_chars changes reverted: 21s. 


OK, it seems I should revert my attempts, and we could finish this already for 24.1.

10110111:
Did you check textured scenes?

gzotti:
Yes, with textures and with simple materials. Looks identical to the previous.

10110111:
I've pushed the relevant changes to `master`, reshaping the commits to make the history cleaner. So I suppose this PR can be closed.

What I've noticed just now is that in the Qt6 version loading takes 3× as long as in the Qt5 one. In both cases my changes improve the times, but still Qt6 version appears to be similarly slower.

ElonVampire:
收到！

gzotti:
OMG! What is it then? QVector/QList were more or less merged in Qt6. Does that make a difference?
OK, we can call the branch closed, but if you have an idea, please feel free to further improve.

PR #3655, opened 2024-03-04
Further optimization of OBJ loading

10110111:
### Description

It appears that `QString::toInt()` has become *much* slower in Qt 6.4.1 compared to Qt 5.15.4. For this reason I switched to `std::from_chars` for ints. Some more improvement comes from doing this for floats. Additionally, I removed the use of `QString` for parsing, so that it all now works with usual `char` units, which is more friendly to `std::from_chars` calls.

The results are as follows in the test where I simply call `StelOBJ::load()` from `main` to avoid any variability due to multithreading:

* Qt 6.4 : 16.5 s -> 3.4 s
* Qt 5.15: 4.9 s  -> 3.9 s

In the real-life scenario of loading in background during rendering the final times of loading are about 12 s (both in Qt5 and Qt6) down from about 30 s in Qt6 and 16 s in Qt5.

The only problem is that our macOS CI seems to use a relatively old compiler whose C++ standard library doesn't implement `std::from_chars` for floating-point types, and this breaks the build. One fix is to switch to a newer compiler if possible, and another is to employ `fast_float` library when the C++ standard library doesn't provide these functions.

### DIFF
--- a/src/core/StelOBJ.cpp
+++ b/src/core/StelOBJ.cpp
@@ -188,9 +188,21 @@ bool StelOBJ::parseFloat(const ParseParams &params, float &out, int paramsStart)
 		qCWarning(stelOBJ)<<"Additional parameters ignored in statement"<<params;
 	}
 
-	bool ok;
-	out = params.at(paramsStart).toFloat(&ok);
-	return ok;
+	const auto& xQStr = params[paramsStart+0];
+	// We'll do the allocationless equivalent of QString::toLocal8Bit(). The buffer will contain
+	// non-terminated Latin1 version of the contents of the input QString.
+	char string[1024]; // There can't be too many chars in a number; if there are, just fail
+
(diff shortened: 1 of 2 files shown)


gzotti (approved):
Yess!! 42s->33s is a further great improvement, thank you. I tested with textured, untextured and a mixed (partly textured, partly simple material) scene. Now, @alex-w , is there a better compiler for the CI for Mac?

alex-w (approved):
We can switch to macOS 13 (clang 15) for CI, but I see strange troubles for unit tests for these instances…

P.S. Or we can switch to experimental images with macOS 14.

alex-w:
> Would that influence required minVersion of macOS? 

It's depended on version of Qt and target platform. Currently, we use macOS 11.0+ for Qt6-based packages, but I've built these packages in macOS 14.

PR #3657, opened 2024-03-06
Fall back to fast_float when C++ stdlib doesn't provide from_chars for floats

### LINKED ISSUE #3656: std::from_chars for floating-point chars
### Expected Behaviour
Complete build Stellarium in systems with Clang/LLVM, like macOS or FreeBSD

### Actual Behaviour
Building Stellarium in macOS is broken currently, because `std::from_chars` is [not implemented yet](https://libcxx.llvm.org/Status/Cxx17.html#note-p0067) for floating-point chars in Clang

### System
* Stellarium version: master
* Operating system: macOS

10110111:
Fixes #3656.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1038,6 +1038,18 @@ IF(DOXYGEN_FOUND)
      ADD_CUSTOM_TARGET(apidoc ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} COMMENT "Generate the doxygen documentation into the doc directory.")
 ENDIF()
 
+include("cmake/Check-from_chars.cmake")
+IF(NOT FROM_CHARS_WORKS)
+    CPMFindPackage(NAME fast_float
+                   GIT_REPOSITORY https://github.com/fastfloat/fast_float
+                   VERSION 6.1.0
+                   EXCLUDE_FROM_ALL yes)
+    GET_TARGET_PROPERTY(fast_float_INCLUDE_DIRECTORIES
+                        FastFloat::fast_float INTERFACE_INCLUDE_DIRECTORIES)
+    INCLUDE_DIRECTORIES(${fast_float_INCLUDE_DIRECTORIES})
+    ADD_DEFINITIONS(-DUSE_FAST_FLOAT)
+ENDIF()
+
 ########### Top level include directories ###########
 # This will be used for all compilations in sub-directories
 INCLUDE_DIRECTORIES(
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -1038,6 +1038,18 @@ IF(DOXYGEN_FOUND)
      ADD_CUSTOM_TARGET(apidoc ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} COMMENT "Generate the doxygen documentation into the doc directory.")
 ENDIF()
 
+IF((${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 11.1) OR
+   (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 14.0))
```
alex-w, line 1042: CMake using name AppleClang for Apple-provided Clang - probably MATCH will be better here

CMakeLists.txt
```diff
@@ -1038,6 +1038,18 @@ IF(DOXYGEN_FOUND)
      ADD_CUSTOM_TARGET(apidoc ${DOXYGEN_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/Doxyfile WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR} COMMENT "Generate the doxygen documentation into the doc directory.")
 ENDIF()
 
+IF((${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 11.1) OR
+   (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang" AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 14.0))
```
alex-w, line 1042: Hmm... For Clang no need part `AND ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 14.0` IMHO

10110111: We do need it. Clang does implement these functions since v14, if we believe your reference. But AppleClang might not, or it has some other version numbering. This is what we need to find out.

alex-w: Please see logs for macOS and FreeBSD CI

alex-w: > We do need it. Clang does implement these functions since v14, if we believe your reference. But AppleClang might not, or it has some other version numbering. This is what we need to find out.

I see notification for `to_chars` method only

gzotti (approved):
I cannot help with Apple, just approving formally. I am sure you will do the right things.

alex-w:
It's OK for me

PR #3664, opened 2024-03-12
Add /utf-8 for MSVC++

10110111:
This change will let us write e.g. `QChar(u'°')` instead of `QChar(0x00b0)`, which would make the code more self-documenting.

Our code style still doesn't allow this, but this is a secondary thing to adjust (and I think it should be adjusted to allow such literals, because this will increase maintainability and readability). This PR is concerned with the technical side.

MSVC++ 2019 apparently has it enabled by default, but 2017 doesn't. GCC and Clang already use UTF-8 source charset by default, so this patch covers the last bit.


alex-w (approved):
Thanks!

gzotti (approved):
Sure, good idea!

PR #3721, opened 2024-04-20
Localize solar eclipse file name

10110111:
The default output file name for eclipse maps will now depend on locale.

### DIFF
--- a/src/gui/AstroCalcDialog.cpp
+++ b/src/gui/AstroCalcDialog.cpp
@@ -4225,7 +4225,7 @@ void AstroCalcDialog::generateKML(const EclipseMapData& data, const QString& dat
 	using EclipseType = EclipseMapData::EclipseType;
 
 	stream << "<?xml version='1.0' encoding='UTF-8'?>\n<kml xmlns='http://www.opengis.net/kml/2.2'>\n<Document>" << '\n';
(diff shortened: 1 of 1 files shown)


src/gui/AstroCalcDialog.cpp
```diff
@@ -5043,18 +5043,17 @@ void AstroCalcDialog::saveSolarEclipseMap()
...
-	int year, month, day;
-	StelUtils::getDateFromJulianDay(JDMid, &year, &month, &day);
-
 	core->setUseTopocentricCoordinates(savedTopocentric);
 	core->update(0);
 
 	// Use year-month-day in the file name
-	const auto eclipseDateStr = QString("-%1-%2-%3").arg(year).arg(month,2,10,QChar('0')).arg(day,2,10,QChar('0'));
+	const auto eclipseDateStr = localeMgr->getPrintableDateLocal(JDMid);
+	const auto fileBaseName = qc_("Solar eclipse %1",
```
alex-w, line 5051: The context data is very long. Maybe just comment for translators and short context (or without context)

10110111: How would you formulate this in a shorter way, keeping it unambiguous?

10110111: OK, I've dropped this, since the simple concatenation is already used in the caption inside KML.

alex-w (approved):
Thanks!

PR #3730, opened 2024-04-30
Rename sky culture change signal for better consistency

10110111:
This change makes the names of all the `Q_PROPERTY` manipulation methods consistent (include the "ID" part), and this will also let me later add another signal that will pass `StelSkyCulture` structure instead of the string.

I'm not sure if this signal is used in any non-obvious ways like e.g. in scripting API, so please review this.


alex-w (approved):
Thanks! Switching sky cultures in scripts work as expected.

PR #3735, opened 2024-05-08
Use single lines between table cells

10110111:
This looks cleaner IMO. See the screenshots:

### Original

Screenshot_2024-05-08_15-29-39

### After this patch

Screenshot_2024-05-08_15-28-39


alex-w (approved):
Very nice!

PR #3747, opened 2024-05-18
Move some solar eclipse computations from AstroCalcDialog

10110111:
This reduces the size of AstroCalcDialog.cpp by 15% while localizing much of solar eclipse map generation and some other related computations in another file.


alex-w (changes requested):
Good improvement, thanks!

But please add new file (`src/core/SolarEclipseComputer.cpp`) into `po/stellarium/POTFILES.in` also and remove extra empty lines.

10110111:
The empty line at the end? Or is there some more?

alex-w:
> The empty line at the end? Or is there some more?

See CodeFactor report please

alex-w (approved):

src/core/SolarEclipseComputer.hpp
```diff
@@ -0,0 +1,170 @@
+/*
+ * Stellarium
+ * 
+ * Copyright (C) 2015 Alexander Wolf
+ * Copyright (C) 2016 Nick Fedoseev (visualization of ephemeris)
```
gzotti, line 5: Maybe reduce Nick? Or is some of his content from AstroCalc in this file?

alex-w: +1 - Yes, the new file doesn't have Nick's code

alex-w: Hmm... my name should be removed from new files too

gzotti (approved):
Yes, good idea to split this huge class.

src/core/SolarEclipseComputer.hpp
10110111: OK, removed Nick and Alex.

10110111:
Let's wait until the CI finishes, after which, if successful, I'll merge.

src/core/SolarEclipseComputer.hpp
gzotti: Sorry to be nitpicky, but now it looks strange. Worachate did most of this and should be first, I refactored a little bit (more possible).

10110111: Is this OK? (Also removed "(Eclipses)" since the whole file is for eclipses now.)
```
 * Copyright (C) 2022 Worachate Boonplod
 * Copyright (C) 2022 Georg Zotti
 * Copyright (C) 2024 Ruslan Kabatsayev
```

PR #3757, opened 2024-05-30
Fix consistency of different lines in solar eclipse maps

10110111:
The original version of the eclipse maps has some ugly artifacts at the locations where the limit lines connect, like rise-set limit vs the penumbra limits, which should touch but not cross, while they instead either cross or don't reach each other (see the screenshots below). This happens because of a few issues with the implementation:

 * In some computations Earth flattening is taken into account in not quite accurate way;
 * In the computation of penumbra limits the fact that a single JD may imply more than one point of the curve results in loss of solutions and thus too short curves.

This PR reimplements some of these functions, trying to avoid any dubious approximations and also avoiding the procedures given in the reference book that don't yield satisfactory results.

There are a few caveats, which you can read in the commit messages, but to me the result looks better than it was before.

#### Old

Screenshot_2024-05-30_23-29-27

#### New

Screenshot_2024-05-30_23-29-40

#### Old

Screenshot_2024-05-30_23-30-00

#### New

Screenshot_2024-05-30_23-30-10

### Type of change

- [ ] Housekeeping

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/SolarEclipseComputer.cpp
+++ b/src/core/SolarEclipseComputer.cpp
@@ -26,10 +26,13 @@
 #include "StelCore.hpp"
 #include "sidereal_time.h"
 
+#include <utility>
 #include <QPainter>
 
 namespace
 {
+
+double sqr(const double x) { return x*x; }
 
 static const QColor hybridEclipseColor ("#800080");
 static const QColor totalEclipseColor  ("#ff0000");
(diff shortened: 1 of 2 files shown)


10110111:
If you are going to merge this, please don't squash. These commits are big enough to be kept separate.

alex-w (approved):
Thanks!

gzotti:
Thanks. While the shown cases look indeed much better and seem more correct, note that not all lines meet at expected places, there are sometimes unexpected intersections in the begin/end regions. See the discussions in Meeus, Morsels (1-3?).

10110111:
> note that not all lines meet at expected places

Maximum in the horizon and penumbra limits must join by construction (they follow the same equation, just with different constraints). Same for shadow limits and rise-set limit curve, since they both denote the limits for the shadow, just for different stages of shadowing.

Indeed, the rise-set limit intersects the maximum in horizon curve at odd places, which is to be expected since they are quite unrelated geometrically.

Anyway, I followed the geometrical logic while implementing this, rather than try to fit the curves, so this nice result is just a side effect of correctness.

10110111:
Here are some examples of unexpected intersections that are still correct:

#### 2025-03-29



#### 2024-10-02

gzotti:
Yes, this is what I meant. I just wanted to start compiling to check that despite lack of time. I also like the good docs! OK, Looks good to me!

PR #3761, opened 2024-06-03
Add name of NGC 5139 from discoverer Edmond Halley

silas1037:
Well, still some to be confirmed. So Edmond Halley named ω Centauri as dorso Equino Nebula, right?

The latin spell seems to be "dorso equino nebula" rather than "dorfo equino nebula", and it means Horseback Nebula?

"in dorso equino nebula" or "dorso equino nebula"?

Ptolemy calls ὁ ἐπὶ τῆς ῥάχεως ἐφώσκομενος (“the one shining on the back”), perhaps he discovered its non-stellar nature.

image

image

### DIFF
--- a/nebulae/default/names.dat
+++ b/nebulae/default/names.dat
@@ -133,6 +133,7 @@
 #      - Huygens region, Gentil… (region names of Great Orion Nebula); http://www.astrosurf.com/patricio/DS/M42-topografia.htm
 #      - The Kite Cluster (NGC 1664); https://observing.skyhound.com/archives/dec/NGC_1664.html
 #      - NGC 3310 - Bow and Arrow Galaxy; http://astroa.physics.metu.edu.tr/~umk/ccd_wshop_adelman/Astronom/GALAXY/NGC3310.HTM
+#      - NGC 5139 - dorso Equino Nebula; (1714). An Account of Several Nebulae or Lucid Spots Like Clouds, Lately Discovered among the Fixt Stars by Help of the Telescope. Philosophical Transactions of the Royal Society of London, 29(338-350), 390–392. doi:10.1098/rstl.1714.0046
 #      - NGC 6559 - Chinese Dragon Nebula; https://noirlab.edu/public/images/gemini0509b/ https://skyandtelescope.org/astronomy-news/astro-image-in-the-newsdragon-in-the-sky/
 #      - Ced 19D - Electra Nebula; http://www.irida-observatory.org/CCD/VdB20_23/VdB20_23.html
 #      - IC 423 - Teardrop Nebula; Sky and Telescope February 2017 (2017S&T...133b..75A)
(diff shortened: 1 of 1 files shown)


silas1037:
Of course, many modern nicknames are derived from these astronomers' descriptions! For example, the Dumbbell Nebula was proposed to be created based on John Herschel's description of its dumbbell-like history, and historical nicknames that have not been discovered should certainly be valued. We should use a standard to evaluate whether they are appropriate to be adopted. The Horseback Nebula is very vivid in its constellation position. As for whether Halley used the perspective of nicknames to call it, further discussion is needed.
As for the Greek issue, perhaps only you can interpret it more easily, it is too difficult for me, a native Chinese speaker.

gzotti:
> Of course, many modern nicknames are derived from these astronomers' descriptions! For example, the Dumbbell Nebula was proposed to be created based on John Herschel's description of its dumbbell-like history, and historical nicknames that have not been discovered should certainly be valued. We should use a standard to evaluate whether they are appropriate to be adopted. The Horseback Nebula is very vivid in its constellation position. As for whether Halley used the perspective of nicknames to call it, further discussion is needed.

Yes, Dumbbell is from 19th-century observers like Herschel and like most "classical" and well-known DSO names has to do with the _appearance_ in a telescope. Horsehead, Owl, Cirrus, Ring, Pinwheel, Sombrero, ... I am less aware of good _positional_ names, apart from Orion and Andromeda nebulae (ok, yes, Triangulum galaxy is also not because of its shape :). You will most likely find Latin positional descriptions (that are IMHO no names) for many objects in author catalogs and atlases from the 17th-19th centuries, similar in style to Ptolemy's star list.



> 
> Moreover, this nickname will be placed after the ω Cen name, which would not have much impact on most users who were focused on the top-priority nickname. Moreover, there were already too many randomly created nicknames in the entire database.

I fully agree there are awful names in our list which I have never seen in use. I'd like to have a feature to "disable unwanted sources", esp. those stupid new names from people who just attach new names to known objects with well-established names. 

> As for the Greek issue, perhaps only you can interpret it more easily, it is too difficult for a native Chinese speaker.

Sorry, I forgot. Must dig out my dusty old dictionary...

10110111:
> Ptolemy calls ὁ ἐπὶ τῆς ῥάχεως ἐφώσκομενος

On the first glance I'd rather recognize the letters as "ὁ ἐπὶ τῆς ἆ νώτα ἐχφύσεως". The χ may actually be a κ though since it's the way it often looks in italics. [Some googling](https://www.google.com/search?q=%22it+is+in+the+centaur+that+which+ptolemy+calls%22) suggests that the text might be "ὁ ἐπὶ τῇ νωτα έκρισεως", but all these results are from Google Books, so are likely an OCR output. Some other results from the link above denote the same as "Star in the Excreſcence on the Back" (with Google's OCR mistaking "ſ" for "f", just as the OP mistook "dorſo" for "dorfo"), which may help interpreting the meaning.

gzotti:
Ah, yes, the strange sign is just a flipped alpha with accents when the next word is "nota". The kappa is IMHO undoubtedly kappa, no chi. (and chi-phi probably don't like to be adjacent, but it's too long ago to remember.) 

Yes, the long s glyph "ſ" was also used in English 18th century print as we see here, and remains visible today as initial part of the German sharp-s character ß (which can be upcased as SS or SZ).

PR #3775, opened 2024-06-21
Allow disabling StelAudioMgr at runtime

### LINKED ISSUE #3725: Stellarium waiting several seconds on 'Audio' at startup
When starting Stellarium it shows a splash screen showing what is loading and then it starts Stellarium normally.
The issue is that in the splash screen I see 'Initializing audio...' which takes several seconds. But In don't have audio (and why does a planetarium app need sound ?) so this is completely useless.
### Expected Behaviour
No audio loading.

### Actual Behaviour
Loading Audio at startup takes several seconds.

### Steps to reproduce
Start Stellarium on Ubuntu without sound drivers.
Screenshot 2024-04-27 at 11 01 19

### System
* Stellarium version: 24.1 installed with apt package manager in Ubuntu.
* Operating system: Ubuntu 22.04.4
* Graphics Card: ? (it is not a graphics issue)
* Screen type (if applicable): 1366x768 (it happens at any resolution)
* 
### Logfile 
`stellarium 2>&1 | tee err2` results in attached file.
[err2.log](https://github.com/Stellarium/stellarium/files/15137696/err2.log)

gzotti:
### Description

We had a complaint (#3725) that Stellarium takes some extra (unspecified) time to initialize on a system without sound hardware (or disabled).  While sound hardware has been generally available on consumer systems of the past 25 years, there may be circumstances, e.g. on Virtual Machines. This branch brings in yet another configuration option to avoid initialisation. The simple alternative is to build Stellarium from source without media support.

Fixes #3725

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Running scripts/test/audiotest.ssc. When audio system is disabled, observe text output. 

I have not yet tested videos with sound output! And I cannot test a physical system without sound hardware.

**Test Configuration**:
* Operating system: Win10pro
* Graphics Card: irrelevant

## Checklist:

- [] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -563,7 +563,8 @@ void StelApp::init(QSettings* conf)
 
 	// Init audio manager
 	SplashScreen::showMessage(q_("Initializing audio..."));
-	audioMgr = new StelAudioMgr();
+	const bool audioOK = !(qApp->property("onetime_no-audio").toBool());
+	audioMgr = new StelAudioMgr(audioOK && confSettings->value("audio/enabled", true).toBool());
 
 	// Init video manager
 	SplashScreen::showMessage(q_("Initializing video..."));
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -561,14 +561,22 @@ void StelApp::init(QSettings* conf)
 	toasts->init();
 	getModuleMgr().registerModule(toasts);
 
-	// Init audio manager
-	SplashScreen::showMessage(q_("Initializing audio..."));
+	// Init audio manager. When media support is disabled,
+	// we silently create dummy managers.
+#ifdef ENABLE_MEDIA
 	const bool audioOK = !(qApp->property("onetime_no-audio").toBool())
 			&& confSettings->value("audio/enabled", true).toBool();
+#else
+	const bool audioOK = false;
+#endif
+	if (audioOK)
+		SplashScreen::showMessage(q_("Initializing audio..."));
 	audioMgr = new StelAudioMgr(audioOK);
 
 	// Init video manager
+#ifdef ENABLE_MEDIA
 	SplashScreen::showMessage(q_("Initializing video..."));
+#endif
 	videoMgr = new StelVideoMgr(audioOK);
 	videoMgr->init();
 	getModuleMgr().registerModule(videoMgr);
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -563,12 +563,13 @@ void StelApp::init(QSettings* conf)
 
 	// Init audio manager
 	SplashScreen::showMessage(q_("Initializing audio..."));
-	const bool audioOK = !(qApp->property("onetime_no-audio").toBool());
-	audioMgr = new StelAudioMgr(audioOK && confSettings->value("audio/enabled", true).toBool());
+	const bool audioOK = !(qApp->property("onetime_no-audio").toBool())
+			&& confSettings->value("audio/enabled", true).toBool();
+	audioMgr = new StelAudioMgr(audioOK);
 
 	// Init video manager
 	SplashScreen::showMessage(q_("Initializing video..."));
-	videoMgr = new StelVideoMgr();
+	videoMgr = new StelVideoMgr(audioOK);
 	videoMgr->init();
 	getModuleMgr().registerModule(videoMgr);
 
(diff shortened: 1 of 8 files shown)


alex-w (changes requested):
Well, what about video support? I guess one config option to suppress both - audio and video support will be better…

gzotti:
I wonder what piece of hardware or Linux subsystem one has to remove/disable to accidentally not being able to show a (silent) video. No graphics card? No screen? And you are running the rest of Stellarium successfully?

As commented, I have not yet looked into the case of disabling sound output for videos with sound. Can do that only on the other PC.

alex-w:
Well… log in #3725 has info about missing gstreamer’s plugin by the fact and gstreamer is not our direct dependency. OK, you’re suppressed audio playback and missing gstreamer’s parts for audio, but Qt using gstreamer for audio channels in video also…

alex-w (approved):

alex-w:
I hope all videos will be silent…

gzotti:
Please do not merge yet!
> As commented, I have **not yet looked into the case of disabling sound output for videos with sound**. Can do that only on the other PC.

gzotti:
gstreamer is some Qt dependency. I assume it gets called on creating a `new QAudioOutput()`.

gzotti:
Someone can test this now on a sound-less PC or VM. How does it behave attempting to play videos with sound? 
I am very unsure about the solution in the Qt5 path, I don't see a clear way to disable accessing some audio component. Here videos always were problematic anyhow. The one user who needs a Qt5-based soundless system, just build without media support!

src/core/StelApp.cpp
```diff
@@ -563,12 +563,13 @@ void StelApp::init(QSettings* conf)
 
 	// Init audio manager
 	SplashScreen::showMessage(q_("Initializing audio..."));
```
alex-w, line 565: Hmm, probably these messages ("Initializing audio..." and "Initializing video...") shouldn't be displayed when features are suppressed...

gzotti: After another look, AudioMgr and VideoMgr are initialized with these messages even when MEDIA support is disabled. Just that the functions do nothing but write explanatory info to the logfile. The next message shows reasonably fast so that I don't think we need a change. The message helps to identify where a problem occurred.

alex-w: In one hand you are right, but in other hand we should suppressing these messages when media support is not exist in the binary package or one message when audio playback is disabled.

src/core/StelVideoMgr.cpp
```diff
@@ -36,19 +36,21 @@
 #endif
 
 
-StelVideoMgr::StelVideoMgr() : StelModule()
+StelVideoMgr::StelVideoMgr(bool withAudio) : StelModule()
 {
 	setObjectName("StelVideoMgr");
+	audioEnabled = withAudio;
```
10110111, line 42: This belongs in the initializer list.

src/core/StelAudioMgr.cpp
```diff
@@ -24,10 +24,11 @@
 
 #include <QMediaPlayer>
 
-StelAudioMgr::StelAudioMgr()
+StelAudioMgr::StelAudioMgr(bool enable)
 {
+	enabled = enable;
```
10110111, line 29: This belongs in the initializer list.

gzotti:
Now I just need confirmation about whether the actual changes work on a (deliberately) crippled Linux system. Videos play silently on Windows, but i don't remove my audio drivers for testing. I also did not test Qt5, sorry to be short in time.

10110111:
> but i don't remove my audio drivers for testing

You can just disable a device in the Device Manager.

10110111:
I get the following errors when building on a Windows VM (MSVC++ 17.9.34728.123; `ENABLE_MEDIA=0`):
```
C:\Users\User\source\repos\stellarium\src\core\StelAudioMgr.cpp(194): error C2511: 'StelAudioMgr::StelAudioMgr(void)': overloaded member function not found in 'StelAudioMgr'
C:\Users\User\source\repos\stellarium\src\core\StelAudioMgr.hpp(31): note: see declaration of 'StelAudioMgr'
...
C:\Users\User\source\repos\stellarium\src\core\StelVideoMgr.cpp(39): error C2614: 'StelVideoMgr': illegal member initialization: 'audioEnabled' is not a base or member
```

PR #3780, opened 2024-06-25
Fix custom user dir option.

gzotti:
### Description

In our Google forum https://groups.google.com/g/stellarium/c/js2v-tkERIw/m/RfdXb0MKAwAJ we have an issue reported about what happens if you give --user-dir option on the command line. 
When dir is relative and does not exist, it fails to create the directory. Some other errors seem to follow then. 

This PR should fix that.

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Start Stellarium from the command line, Provide --user-dir ppp.
Earlier: fail to create ppp
Now: creates ppp and uses it.

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:

- [x] My changes generate no new warnings (only those intended...)

### DIFF
--- a/src/core/StelFileMgr.cpp
+++ b/src/core/StelFileMgr.cpp
@@ -91,6 +91,15 @@ void StelFileMgr::init()
 		qFatal("Error: cannot create user config directory: %s", e.what());
 	}
 
+	// There is still a chance that userDir is just a relative path!
+	QFileInfo fi(userDir);
+	if (fi.isRelative() && fi.isDir())
+	{
+		qDebug() << userDir << "looks like a relative path";
+		userDir=fi.absoluteFilePath();
+		qDebug() << "userDir now " << userDir ;
+	}
+
 	// OK, now we have the userDir set, add it to the search path
 	fileLocations.append(userDir);
 	
@@ -515,7 +524,7 @@ void StelFileMgr::makeSureDirExistsAndIsWritable(const QString& dirFullPath)
 	{
 		// The modules directory doesn't exist, lets create it.
 		qDebug() << "Creating directory " << QDir::toNativeSeparators(uDir.filePath());
-		if (!QDir("/").mkpath(uDir.filePath()))
+		if (!QDir("/").mkpath(uDir.absoluteFilePath()))
 		{
 			throw std::runtime_error(QString("Could not create directory: " +uDir.filePath()).toStdString());
 		}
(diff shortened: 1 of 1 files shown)


src/core/StelFileMgr.cpp
```diff
@@ -515,7 +524,7 @@ void StelFileMgr::makeSureDirExistsAndIsWritable(const QString& dirFullPath)
 	{
 		// The modules directory doesn't exist, lets create it.
 		qDebug() << "Creating directory " << QDir::toNativeSeparators(uDir.filePath());
-		if (!QDir("/").mkpath(uDir.filePath()))
+		if (!QDir("/").mkpath(uDir.absoluteFilePath()))
```
10110111, line 527: Why are we even using `/` as the origin? If we had simply constructed `QDir` as `QDir{}`, it'd use current directory and handle this case automagically.

gzotti: I had hoped QDir had a static method mkpath(), but no. So I just copied what was there. The important change here is the absoluteFilePath(). Maybe QDir().filePath() would have been another solution.

10110111: > The important change here is the absoluteFilePath()

This is the point. You introduced a bunch of code just to work around a mistake that consists of three chars. If you simply remove the `"/"` instead of this commit, I'm pretty sure things will work as expected, since `QDir::absoluteFilePath()` also works relative to CWD.

gzotti: I exchanged one method that was not enough by another which works. The other changes attempt to catch further problems. 

I had just understood the QDir("/") as basically a dummy dir object, and the absoluteFilePath() as the solution that fixed the reported issue.   If it really matters, I can alternatively exchange one constructor by the default constructor, change back the other method call, and perform a sequence of tests if that also works. But I can test this only in early July, I am off for a conference.

Paul-Gilmartin: If I create an empty directory and supply a --user-dir
relative to the directory path relative to the directory
in which I issue the "stellarium" command, stellarium
uses it, partially popullat4es it, then fails attempting
to create additional content.

I consider that dangerous because it appears unpredictable
where Stellarium tries to create content if user-directory is
not absolute

It would be better if Stellarium simply failed on a relative path.

Even better if it used realpath() or an equivalent utility to qualify
a relative path relative to the starting directory and used the
resulting value consistently thereafter.

Thanks,
gil

gzotti: This change is supposed to create and/or use a directory even if only a relative name is given, and configure UserDir with an absolute path, avoiding further issues. Have you tested this branch and does it still misbehave on your system?

The comment left by the original developer of this path check implies that in case of missing subdir, it should be created.

10110111 (approved):
Given what I've read in the original report (failure to create files and subdirectories by relative path in an existing user directory), I now think this change makes sense.

src/core/StelFileMgr.cpp
Paul-Gilmartin: I like "configure UserDir with an absolute path, avoiding further issues."

Is Linux likewise affected?

Thanks,
gil

alex-w (approved):

PR #3794, opened 2024-07-03
Parallelize ephemeris computation.

### LINKED ISSUE #1427: show comet locations regardless of magnitude
Someone thought s/he saw a comet at a given location, so I wanted to check this after loading ALL known comets.

Unfortunately, comets are filtered and the user has to zoom in to find a comet (the find feature is useless in this use case). The user needs an option to override the magnitude and show all comets regardless of their actual brightness, similar to how satellites can be shown even when they would be invisible.

A related issue is that comet orbits can be shown, but without the location of the comet.

gzotti:
### Description

Following hints from
https://developer.nvidia.com/blog/developing-accelerated-code-with-standard-language-parallelism/
this branch should update and improve performance on all current CPUs by using standard C++ parallelizing instructions.
It seems `std::transform_reduce()` is perfect for the typical ephemeris sequences. 

This is an ongoing process as I learn how to use this. The good thing is that this is still standard C++17, no extensions/extra libraries required.

EDIT: C++ std:: parallelism is only as good as the compiler it implements. There is no (or hardly any) free threading magic involved. So, loops mostly remain loops. However, this branch then focused on at least splitting the actual SSO computations into as few parallel threads. With many objects, there is a real benefit, good enough to finally want to display thousands of asteroids with little icons. This happens to 

Fix #1427

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win10/11
* Graphics Card: Irrelevant at this stage. Optional continuation towards Cuda possible.

## Checklist:


src/core/planetsephems/precession.cpp
```diff
@@ -51,7 +54,7 @@ static double c_psi_A=0.0, c_omega_A=0.0, c_chi_A=0.0, /*c_p_A=0.0, */ c_epsilon
 
 static const double arcSec2Rad=M_PI*2.0/(360.0*3600.0);
 
-static const double PQvals[8][5]=
+static const std::vector<std::vector<double>> PQvals=
```
10110111, line 57: Is this really required? `std::vector<std::vector<double>>` is less efficient in terms of memory locality than `double[N][M]`, since every subvector is located in a separate part of the heap. If you really want it to be a C++ container, you should use `std::array<std::array<double,5>,8>`, which will, aside from keeping the data together, also make the sizes compile-time constants.

Similarly for the subsequent instances of such change.

src/core/planetsephems/precession.cpp
```diff
@@ -412,36 +416,55 @@ void getNutationAngles(const double JDE, double *deltaPsi, double *deltaEpsilon)
...
+//		}
+		std::pair<double, double> dEpsDPsi = std::transform_reduce(
+					std::execution::par,
+					nut2000Btable.begin(),
+					nut2000Btable.end(),
+					std::pair<double, double>{0.0,0.0},
+					[](std::pair<double, double>sum, std::pair<double, double>addon){
+						return std::pair<double, double>{sum.first+addon.first, sum.second+addon.second };
+						},
+					[=](struct nut2000B nut){
```
10110111, line 452: Is it really necessary to accept `nut` by value, and a non-const one?

alex-w:
Please update including headers in unit tests also

src/core/planetsephems/precession.cpp
```diff
@@ -412,36 +416,55 @@ void getNutationAngles(const double JDE, double *deltaPsi, double *deltaEpsilon)
...
+//			deltaPsi+=(nut->A + nut->Ap*t)*sinTheta + nut->App*cosTheta;
+//			deltaEps+=(nut->B + nut->Bp*t)*cosTheta + nut->Bpp*sinTheta;
+//		}
+		std::pair<double, double> dEpsDPsi = std::transform_reduce(
+					std::execution::par,
+					nut2000Btable.begin(),
+					nut2000Btable.end(),
+					std::pair<double, double>{0.0,0.0},
+					[](std::pair<double, double>sum, std::pair<double, double>addon){
+						return std::pair<double, double>{sum.first+addon.first, sum.second+addon.second };
```
10110111, line 450: `return std::make_pair(sum.first+addon.first, sum.second+addon.second);` will reduce verbosity.

alex-w:
Hmm… macOS/FreeBSD using Clang/LLVM by default and this is not a good.

Details:
* https://en.cppreference.com/w/cpp/17
* https://github.com/llvm/llvm-project/issues/57898
* https://github.com/emscripten-core/emscripten/issues/18476

gzotti:
> Hmm… macOS/FreeBSD using Clang/LLVM by default and this is not a good.

Unfortunate. The NVidia post read as though this was a standard C++17 feature. Is there an `#ifdef` flag to detect LLVM/Clang? If all else fails, those could compile the older version.

10110111:
> The NVidia post read as though this was a standard C++17 feature.

It is. But, as with `std::from_chars<float>` and company, not all compilers support it even though they claim to support C++17.

> Is there an #ifdef flag to detect LLVM/Clang?

It's a bad idea to decide on the basis of which compiler it is, since it may get support for this starting at some future version. This should instead be tested via a CMake test (see `cmake/Check-from_chars.cmake` for an example), and then the code could decide accordingly.

> If all else fails, those could compile the older version.

Or you could try omitting `std::execution::par,` parameter (hiding it behind a macro), so that the unparallelized version would be used. E.g.
```cpp
#if HAS_STD_EXECUTION_PAR
# define STD_EXECUTION_PAR_COMMA std::execution::par,
#else
# define STD_EXECUTION_PAR_COMMA
#endif
...
std::transform_reduce(STD_EXECUTION_PAR_COMMA
                      mycontainer.begin(),
...
```

alex-w:
Probably it can be using by adding experimental headers (see https://github.com/emscripten-core/emscripten/issues/18476)

gzotti:
Of course, testing a compiler for support of this feature would be better than excluding the compiler family. A cmake test looks useful. Where then best define the preprocessor macro?
Currently I am looking into the various ephem files to see what can be converted easily. Although hard to see, I think I can see a tiny speedup already when running side-to-side with stock 24.2. I expect more benefit on weaker multicore CPUs though (Raspi4 etc) when finally the other cores get some work to do.

The ELP82 and VSOP87 would benefit most, obviously, but they are a bit more complicated than I remembered.

10110111:
> Where then best define the preprocessor macro?

If it's used in more than one file, then you could put it into `src/core/StelUtils.hpp`. Or, if it's only ever useful for `planetsephems`, then to some common file in this subdir, like `ephem_common.hpp`.

> I expect more benefit on weaker multicore CPUs though (Raspi4 etc) when finally the other cores get some work to do.

Actually, I wouldn't expect much speedup given how many work items there are (a few dozen for each loop). At least if the parallelization is implemented by multithreading, since there'll be the overhead of thread launching and management. So all this work might be in vain.

Maybe this should be parallelized in another way: e.g. computation of one loop in one thread, another one in another etc.

And, BTW, why did you choose this particular point for the optimization? Did your profiler tell you this is the bottleneck?

gzotti:
No, some other work guided me to that info about the existing standard built-in language feature to allow parallel execution of parts of our typical ephemeris number crunching which so far all happens on one core. From the description it looks as if the compiler would add underlying code to distribute work to other cores, without doing things manually with OpenMP or so.  Given it does not cost much to try it (no other libraries, no cmake extras, or so I had thought... :-), I just started this branch. Some of the code is >20 years old! 

I am not familiar enough with profiling to say where to best improve from the call statistics.  Of course, we may have other bottlenecks as well. Another important place would be also to speedup the minor body Kepler solutions in parallel. But I start from here, little by little..

10110111:
BTW, just tried a simple test, and it appeared that the code for e.g. `std::transform` doesn't differ much whether I use `std::execution::par` or not. See the test: https://godbolt.org/z/b8qW945G8. Try changing `#if 1` to `#if 0`, and compare the resulting assembly. There's not even a call to anything thread-related.

At least these are the results for GCC and CLang. MSVC appears to be broken at that website.

gzotti:
I have difficulties with marssat. How to define the start/end iterators for the fixed-size arrays (which are of unequal length for the two moons), or the underlying C style arrays that can be reached with `data()` with array length separately stored? The data can be rearranged of course.

10110111:
> How to define the start/end iterators for the fixed-size arrays (which are of unequal length for the two moons), or the underlying C style arrays that can be reached with data() with array length separately stored?

Not sure what you mean (and which loop you're trying to replace), but pointers are as good iterators as those returned by `std::begin(someContainer)` and `std::end(someContainer)`. E.g.
```cpp
double myarr[234];
// library-defined iterators (pointers in practice)
const auto startIt = std::begin(myarr);
const auto endIt = std::end(myarr);
// or explicit pointers
const auto startItP = myarr;
const auto endItP = myarr + std::size(myarr);
```

If you have a pointer to some dynamic array and a length, you can try using C++20 with `std::span` (this one works similarly to `std::string_view`, but is not limited to `char`).

gzotti:
I have come to some forum that mentioned span, but we are limited to C++17.
Please see marssat.cpp lines 360-371. This should replace LL 351-358. I am just unable to formulate the begin/end iterators.

10110111:
Well, as I said above, these pointers
```cpp
const struct MarsSatTerm *const begin = bp->lists[j].terms;
const struct MarsSatTerm *p = begin + bp->lists[j].size;
```
are already good iterators. Particularly, `begin` is the "begin" iterator, while `p` at its initial value is the "end" iterator.

Not sure though why the original iteration is being done backwards. Maybe you need to follow this—if, e.g. the terms are being summed from the smallest in magnitude to the largest to avoid too rapid accumulation of errors. (In this case you do want to employ `std::span` with its `rbegin` and `rend` methods—or use [`std::make_reverse_iterator`](https://en.cppreference.com/w/cpp/iterator/make_reverse_iterator) if you want to stay on C++17.)

gzotti:
Thanks a lot, that worked. I was close, but a near miss is also a miss... 

Yes, the original iteration is done backwards to sum up the small factors first.  This is a general recommendation to avoid rounding errors also found in Meeus, Astr.Algorithms. Not sure how much it really matters, though. Something to consider fixing when the forward iteration works. 

Please bear with me, I am very unfamiliar with std:: stuff, having started with Qt/Stellarium as first (and so far almost only) C++ contact. How to apply the reverse iterator here? (I think given other constraints on other systems C++20 is no option here, but how would I apply span?)

10110111:
A few simple examples of printing a C-style array ([play with it online](https://coliru.stacked-crooked.com/a/15303da153ddead9)):
```cpp
#include <span>
#include <vector>
#include <iterator>
#include <iostream>

int main()
{
    const double arr[] = {1.3, 2.5, 7.8, 9.3, -3.5};

    // Normal forward iterator (could use range-based for,
    // but using explicit iterator for illustration)
    for(auto it = std::begin(arr); it != std::end(arr); ++it)
        std::cout << *it << "\t";
    std::cout << "\n";

    // Backwards iteration via std::make_reverse_iterator (note
    // the end-begin order instead of begin-end)
    for(auto it = std::make_reverse_iterator(std::end(arr));
        it != std::make_reverse_iterator(std::begin(arr)); ++it)
        std::cout << *it << "\t";
    std::cout << "\n";

    // Backwards iteration via std::span
    const std::span sp(arr, std::size(arr));
    for(auto it = sp.rbegin(); it != sp.rend(); ++it)
        std::cout << *it << "\t";
    std::cout << "\n";
}
```

gzotti:
Thanks for the examples. Ah, error on my side: As I understood the description, the whole idea of parallelizing assumes unknown sequence order, as long as the reduce operation (here: summing up) is commutative. This should mean, partial sums may form in arbitrary order.

Uh, elp82 and vsop87 are really much harder cases.

10110111:
> This should mean, partial sums may form in arbitrary order.

This is just one more reason to prefer a higher-level parallelization (i.e. on the level of calling each computation function in a separate thread) than try to parallelize a few dozen iterations in a single computation. This will not only better hide the threading overhead, but also avoid the numeric issues. And, BTW, this will then actually speed up the computation on all platforms, regardless of how the execution policy is implemented (on libstdc++, apparently, it's just a stub, see [this comment](https://github.com/Stellarium/stellarium/pull/3794#issuecomment-2208475274)).

gzotti:
Yes, indeed. This was now a 1-2 day trial where potential speed-ups still have to be verified on more platforms. (Hopefully RasPi4 and a semi-modern "4 efficiency-cores only" notebook this weekend.) It seems this current implementation is still all on one core, but can probably use more registers at the same time and is just that tiny bit more efficient.  The NVidia blog mentions a massive speedup with their own compiler and hardware, which we don't have.  Somewhere in the docs I have seen the std authorities are preparing a future std::execution::cuda, which reads like a very simple change may improve things further in a few years. (Of course we need to allow both, CPU and CUDA-where-available, somehow.)

Someone else could look into distributing VSOP87 (look into the notes in vsop87.c from about 2006(!) about extending the static buffers for possible multithreading), Pluto and ELP82 into 10 worker threads, and creating a ~10-20-thread pool for minor bodies.  Alternatively, if available, use CUDA for animating ~1.500.000 minor bodies.  Also the thousands of satellites could be parallelized. But this all needs much larger reconfiguration.

10110111:
> still all on one core, but can probably use more registers at the same time

Just look at the disassembly with `std::execution::par` and without it. The compiler has access to all the registers the platform provides, and the policy won't tell it anything new about the platform. Whenever a compiler can vectorize a loop (in a SIMD fashion), it will already do this. If it can't, one can try

 * rewriting the loop in a more vectorization-friendly form (the best option),
 * using intrinsics (platform-dependent),
 * using `std::experimental::simd` from C++20 (which says for itself that it's experimental, and may be a stub like `std::execution::par`).

> and is just that tiny bit more efficient.

This tiny bit is going to be quite elusive, and not worth the reduction in readability and vectorizability. Note that the `transform_reduce` form is not vectorization-friendly, since it hides parallelizable instructions inside functions (lambdas), so you can't expect that it will result in a better SIMD vectorization.

> Alternatively, if available, use CUDA for animating ~1.500.000 minor bodies.

There are OpenCL and OpenGL 4.3+ compute shaders, which are more cross-platform. But GPU computation (including the NVIDIA-specific CUDA) requires a lot better understanding of the GPUs to get any performance improvement. From my recent experiments for CalcMySky, it's quite hard (at my current level) to
 * beat `glGenerateMipmap` for a simple summation of all texels in a texture,
 * write a code that will work correctly for all the GPUs (e.g. I have a code that works on Intel but not NVIDIA, which I still have to debug), and still be fast.

gzotti:
I tested on the "4 efficiency cores only" notebook (Win11, MSVC2022). Alright, any potential impact is not detectable in frame rate when running this and 24.2 side-by-side with the same features activated. 

On RPi4 (Ubuntu 24.4) I now face a black screen/window only, both with weekly beta (v24.2.187-1871641) and with this build. Sigh...

gzotti:
I have now turned to the obvious candidate for multithreading: updating the planet positions with a large minor planets file. It seems to work already for the JPL DE solutions (with a mutex, unfortunately, as it reads a file). The VSOP is documented as not thread-safe, so this default path is currently broken, but I should be able to parallelize that without mutex.
I can observe higher CPU load on a 8-core CPU now as (hopefully) some load is better distributed away from the main thread. Of course, it is useless to have a 1000-entry thread pool. It raises CPU load to almost max but likely just for switching overhead. Need to test effect on framerate on a modern ~20-thread CPU.

10110111:
Meanwhile you can measure the time taken for the things you replaced with `blockingMap` before and after the change. This will let you see whether the difference is significant compared to time taken per frame (don't forget to raise the minimum and maximum fps settings).

10110111 (commented):
Regarding the refactoring:

I'd keep these methods non-static and move the `SolarSystem* ssystem` variable obtained via `GETSTELMODULE` in these methods to the data members.
Also, why did you remove the `private` specifier? These variables are not public API of this class and should remain private.
And finally, why does this have to go into the parallelization PR? This change is very separate.

gzotti:
> Regarding the refactoring:
> 
> I'd keep these methods non-static and move the `SolarSystem* ssystem` variable obtained via `GETSTELMODULE` in these methods to the data members. 

clangd suggested that these can actually be made static. It is correct as of now, but can be changed at the next refactoring. 

> Also, why did you remove the `private` specifier? These variables are not public API of this class and should remain private. 

clangd complained that this (deleted) private repeats "private" specified for the methods above and should better be deleted. AFAICS the variables stay private. 

> And finally, why does this have to go into the parallelization PR? This change is very separate.

Indeed. I sometimes see clangd hints and just follow them regardless of branch.

gzotti:
Interesting observation: on my i7-12700 gaming laptop, I have about 95 (this) vs. 85 (V.24.2) fps when running side-by-side with RemoteSync on a big monitor with 37 vs 8 threads (number slightly variable), and CPU load is even lower than before. FPS is not stable, however, so sometimes 85 vs 65, or 107 vs 96, 103 vs 86 or so. This branch is almost always faster here. I switched off everything but planets. Zooming into Jupiter or Saturn seems to suppress the advantage, then they are almost equal. Ah, the text rendering seems responsible for the "balancing slowdown".

gzotti:
A bit less clear, but on the 4-core small notebook I see 30(this) vs.28 (24.2).

10110111:
The ultimate test for your optimization is, I think, the computation of the KML maps of solar eclipses. There we have lots of `core->setJD(JD); core->update(0);` calls, which, in my understanding, recompute all these ephemeris.

gzotti:
> The ultimate test for your optimization is, I think, the computation of the KML maps of solar eclipses. There we have lots of `core->setJD(JD); core->update(0);` calls, which, in my understanding, recompute all these ephemeris.

Now I need to instrument master with the same timing code... hold on.
OK: Timing without the debug IO: Master 19.874. This branch: 20.140. Hmm, ugly result. Several runs, similar result, master is a tiny bit faster in this case. 

Has master been improved since 24.2?

calcSolarEclipseBessel() includes the topographic switching which also causes additional 2 forced core updates per call. Better ensure geocentric mode before calling this method (and maybe add an Assert in debug build). 

I have felt for some time now that for some computations (e.g., eclipses, RTS) we should utilize other "mini computing cores" that should be callable in another thread and should include only the relevant bodies. Currently the StelCore is a singleton, and the forced update is a super inefficient hack. But for computing eclipses, we can ignore all the thousands of minor bodies, planetary moons and the other planets.

10110111:
> Has master been improved since 24.2?

I don't think so. Are you testing in Release mode?
Anyway, the numbers being almost the same tells us that this isn't even a bottleneck—or that something was done wrong in your optimization attempts.

gzotti:
I wonder then why I saw ~5-10% better fps before. It was less than expected, OK, but still looked like an improvement.
A bad bottleneck certainly is the core update(0) forcing hack of 2000 SSOs when all we need are sun, earth and moon.

gzotti:
The largest real restructuring was in SolarSystem::computePositions(), where I use QtConcurrent::blockingMap() three times instead of three sequential loops over all (potentially many thousands) SS objects so far executed in the main thread. This call is itself called in every StelCore::update(). QtConcurrent utilizes a QThreadPool (the number of threads depends on the actual hardware and OS scheduler) to do the actual work. I had to make those other changes to allow parallel computation, which messed up each other before the fixes. What I could revert now are the first commits in this branch, using higher-level std calls instead of traditional C style loops. 

Doing all these astronomical ephemeris calculations used to be the difficult part in traditional position astronomy. People are asking why only ~12% of CPU (sounds like 1 core to me) is used on an 8-core PC, so this was the obvious candidate to try. If thread setup is costlier than a whole planet position, we don't need it. And I don't know if any manual thread management could make it any better.  Sorry, I have never learned to "profile a large program in six easy steps" or so. I know valgrind exists on Linux, and I could run it on my 2014-NUC (Core i5-5xxx). I know GDB exists. Where is the 2-page/30 minutes manual for those? Or where have all the students gone?

10110111:
> where I use QtConcurrent::blockingMap() three times instead of three sequential loops over all (potentially many thousands) SS objects so far executed in the main thread.

Well, then something is definitely wrong if you don't get any improvement—or this is not actually taking a significant time.

> I know valgrind exists on Linux

Valgrind is not a profiler, and it will slow the program down to a crawl, which would affect the timings (even relative ones) anyways.

> I know GDB exists.

This one also won't help to profile.

The one I use relatively often is sysprof (also Linux-specific), which is a sampling profiler, thus not requiring one to instrument the program. You just start sampling at the beginning of some long procedure (e.g. eclipse map creation) and then stop when it ends. Make sure to have debug info enabled in release builds for this. And then look at the resulting call tree with timings.

gzotti:
https://valgrind.org/: 
> There are Valgrind tools that can automatically detect many memory management and threading bugs, and profile your programs in detail.

Whatever, I should really do other things. :-(

gzotti:
Probably I should not run one thread per object, but just a few extra threads that each could handle a part of the objects.

10110111:
I just tried setting a breakpoint at `CalcVsop87Elem` that you tried to optimize previously, and... it is called extremely rarely! Optimizing it is simply useless. And I'd expect that computation of ephemeris is generally a rare thing to do, unlike something actually often called from `StelCore::update`. Particularly, from my measurements, `SolarSystem::computePositions` indeed takes some time (16.6% during eclipse map computation), with its callee `Planet::computePosition` being the most prominent place to spend time in (9.3% in eclipse map computation), so this may be worth optimizing.

gzotti:
> I just tried setting a breakpoint at `CalcVsop87Elem` that you tried to optimize previously, and... it is called extremely rarely! Optimizing it is simply useless. And I'd expect that computation of ephemeris is generally a rare thing to do, unlike something actually often called from `StelCore::update`. Particularly, from my measurements, `SolarSystem::computePositions` indeed takes some time (16.6% during eclipse map computation), with its callee `Planet::computePosition` being the most prominent place to spend time in (9.3% in eclipse map computation), so this may be worth optimizing.

CalcVsop87Elem should be called at least 3x8 times per frame. And to me it was always (since 1989) considered an example of "severe number crunching". Allowing parallel execution should be beneficial. Of course, as PCs became more powerful than in the 1990s, it is less relevant, and the threading overhead compensates for at least most of it. I now am distributing SolarSystem::computePositions() onto just a few threads, and will check on a 12-core/20 thread PC now. Note that you need at least the 1000 comets file added to have a relevant number of SSOs.

10110111:
> CalcVsop87Elem should be called at least 3x8 times per frame.

If I enable the breakpoint after the launch is completed, it will hit only after 10 simulated days pass. Then it will be called several times, and then again rarely.

10110111:
> Note that you need at least the 1000 comets file added to have a relevant number of SSOs.

Oh, I guess this is why I don't get this. Do comets need more frequent recomputations of ephemeris?

gzotti:
Ah, with some better understanding of the undocumented VSOP87 code... OK. Only every 10 days. Then it's indeed lightweight, only EllipticToRectangularA() 8x3/frame. 
Objects on Kepler orbits are computed with a more compact data set. Just Kepler's equation and a few transformations. But there are potentially thousands of objects.

gzotti:
And yes, eclipse map in under 10 seconds. And this branch and 24.2 linked with RemoteSync has now 108 vs. 75 fps. My SolarSystem has a bit more than 2000 objects, and the system distributes the load on 9 extra threads for this. According to task manager, CPU load is even lower. 

The Intel VTune in general indicates terrible performance with many blocked threads, however.

gzotti:
stellarium-033
Finally, about 4fps for 12000 visible Solar System objects (including 1000 comets file and the first 10.000 MPC objects.) I added another level of artificial planet markers for that. This utilizes a Core i7-12700, distributing at least the pos calculation on 20 threads. Whatever else blocks the program I cannot identify in VTune. Is it the whole Qt widget framework? Win11 Task manager shows 5% CPU load from Stellarium, Resource monitor lists 23 threads in total, and Firefox (with tens of tabs in the background) has higher CPU load. 

Even more objects could probably be shown with a compute shader. However, we should find out what is really blocking the program.

gzotti:
with a bit more balancing thread count, I finally get better CPU utilisation for a 20-core CPU. With the 12.000 objects, I still have >30fps vs 17fps for 24.2, running without stars, atmosphere, gridlines. 
Now, this doubles framerate while utilizing 16 more cores. Sure, position is only a little part of the program, but it shows the potential in this time of tens of CPU cores. We need to localize more places for moving stuff to separate threads.

src/core/planetsephems/l12.cpp
```diff
@@ -667,20 +667,20 @@ static std::array<struct sat, 4>SATS = {{
...
-	double elem[6];
-	double val[5] = {0.0};
-	double xv[2][3];
+	const struct sat *sat = &SATS[body];
+	thread_local double elem[6];
+	thread_local double val[5] = {0.0};
+	thread_local double xv[2][3];
 
-	const double t = jd - 2433282.5;
+	thread_local const double t = jd - 2433282.5;
```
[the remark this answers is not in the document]
gzotti: Huh? I don't mean static, of course. Thanks for pointing this out. 
I had two rounds with VTune which seemed to me to indicate blocking in L12. I could also do the same as in VSOP, using arrays of 4 temp. vars each.

gzotti:
Something is not fully stable yet. :-(

gzotti:
This seems to work well now, but needs more testing. On the small 4-core notebook there is a slight increase utilizing one extra thread, while on other systems I utilize up to 4. Also the sorting-per-distance of SSO before each frame drawing could be split, and VTune actually reports function names that indicate its use.

gzotti:
Now there is a GUI to play with number of extra threads. With 12.000 objects, no atmo, no grids, no ground, I can increase fps from about 22 to 35 with 4-6 extra threads, but it hardly gets higher with 10 more. When plotting the 12.000 as asteroid marker, I can increase from 3.6 to over 5fps. So it's definitely useful, but there is room for improvements elsewhere.

gzotti:
https://github.com/user-attachments/assets/f153fe6b-918e-4349-8a3f-d90234a0d068

12.000 objects now about 8-10fps. Not sure how to further improve.

gzotti:
Loaded another 10.000 asteroids (Takes several minutes via SS editor, but eventually finishes). Avoid wide-angle views, but zooming in a bit gives a nice view of what floats around in the field. However, probably the first 10.000 asteroids are enough for principal visualisation.

gzotti:
stellarium-003
17.642 objects: 1000comets, MPCAT:10.000, MPCORB:TNOs, Centaurs, SDOs. Interesting pseudo-clustering, but presumably just an observation effect. It is really slow now. Further ideas?

10110111:
Can you give instructions on how to enable this many objects so that I could test this?

> Interesting pseudo-clustering, but presumably just an observation effect.

By observation effect do you mean the effect of uneven detalization of the catalogs?

gzotti:
You simply use the Solar System Editor plugin. 
- "use add objects from ini file" to load the 1000 comets. 
- Then use the "import from MPC" feature to add the first 10.000 numbered asteroids, 
- and the TNOs, centaurs, SDOs. 

In the list offered, just "select all", and import. It may take minutes to load those amounts. (Maybe here is the next thing to speed up? But it's a one-time loading only.)

Then, in the SSO tab, activate Minor Planet Markers (Yes, I should rename to Minor Body Markers!). It is very clear that the actual drawing is costly. Zoom in to get higher framerate. (Did I mess up anything that makes the GPU part inefficient? I grabbed most based on StelSkyDrawer.) Switch to equatorial mode, disable atmosphere, and set higher speed to see swarms of minor bodies glide along the ecliptic. See opposition effects. See the main belt asteroids speed in front of the SDOs. (We can discuss the colors, of course. These are hardcoded for a start.)

gzotti:
> By observation effect do you mean the effect of uneven detalization of the catalogs?

Yes. I assume those objects are discovered by campaigns that depend on available survey instruments, weather, and (mostly) the placement of Earth, Sun, Moon, and Milky Way. TNOs have been found only recently. Given their slow pace and similar orbital characteristics (distance a>30AU, low e), the clusters will take a long time (decades) to disperse. In the meantime, when they have cleared their "regions of discovery" (off Milky Way, presumably), new TNOs will be found there.

gzotti:
I can see in the VTune flame graph with a SolarSystem with 27701 objects (1000comets, MPCAT 20.000 plus TNO/Centaurs/SDOs) that now the duplicate calls to compute magnitude etc. are gone. However, there are many malloc/free entries which must be related to the StelPainter. The overall algorithm to draw the solar system with all bells/whistles is:

```
sort objects by distance
From far to near:
    if parts of 3D sphere are on screen:
        draw orbit
        Use a StelPainter to:
             draw planet hint and label
            make an entry into separate array to draw asteroid mark. If buffer is full, flush marks.
        Draw 3D sphere involving a series of cases each requiring a differently set StelPainter.
    else:
        draw orbit (if so configured)
Flush remaining asteroid marks.
```

I can see the actual position/magnitude calculation is now only a tiny fraction of the loop. The problem are the temporary allocations/releases of StelPainter's private buffers. With 25.000 SSOs this really is a problem. Would it be possible to just use one painter that is allocated before the loop and hand that down to the various calls, frequently reconfiguring (assign different projection, or whatever), but not malloc'ing/deleting the arrays of at least 2 StelPainters per object? Or would applying a Z buffer and not caring for distance beforehand be possible? Then we won't even need the ordering (not sure about semitransparent planet rings though), could have separate loops for hints/labels, asteroid markers, and actual 3D planets, each using just one StelPainter. This might also solve the ancient Orbit overdraw bug (#82).

10110111:
> actually, who frees the arrays?

Which arrays do you mean? I see `smallCircleVertexArray` and `smallCircleColorArray` as `static` members, but they are static, so they are destroyed only when the program is terminating (and constructed when it's starting).

gzotti:
StelPainter's constructor:
```
StelPainter::StelPainter(const StelProjectorP& proj)
	: QOpenGLFunctions(QOpenGLContext::currentContext())
	, glState(this)
	, vao(new QOpenGLVertexArrayObject)
	, verticesVBO(new QOpenGLBuffer(QOpenGLBuffer::VertexBuffer))
	, indicesVBO(new QOpenGLBuffer(QOpenGLBuffer::IndexBuffer))
```

what is `new `should be `delete`d, right? Or what am I missing? And I assume these 'new's take their time for allocations.

10110111:
These pointers are smart, they delete their pointees in the destructors.

10110111:
As for optimizing their creation and deletion, yes, they could possibly be reused. But I'd make it (and some of the recent commits) separate, because this PR is growing unreviewable.

gzotti:
OK. What I can do to complete this PR:
- undo a few of the first changes, they were my misunderstanding of NV marketing news while not really speeding up anything with standard compilers. (Esp. revert the backwards summation.)
- clean up what is now only commented away
- describe the news in the User Guide

The current changes (last 6 commits or so) are a speedup in themselves as they avoid many computations that were called for each solar system object. Further optimisation of the SolarSystem drawing by better taking care of StelPainter internals or re-use their existing buffers can go into another round (which I'd happily leave to you or other masters of OpenGL). As you can see, I am interested in getting as many asteroids as possible. I always thought it was the position computation. No, it's the one-by-one drawing. Further ideas immediately emerge, like a slider that could enable/hide asteroids by number (works only for numbered asteroids, obviously) or absolute_magnitude H or discovery date (if available), or selection of asteroids by type.

10110111:
> undo a few of the first changes

I'd prefer them to be completely removed from the patch sequence, to avoid cluttering the git-blame output.

And generally, this PR should be concentrated on parallelization, otherwise we'll either get a huge final squashed commit, or a lot of different half-baked states in the history. The last few patches that are quite self-sufficient should be applied separately.

gzotti:
What the best way to do that? Make two new branches/PR (1: threads; 2: avoid multiple computations) and cherry-pick? We never were that picky about having each commit in a perfect working state before.

10110111:
> We never were that picky about having each commit in a perfect working state before.

This PR is huge, and I suspect it to have broken something due to the complexity of the changes (threading is quite error-prone generally, especially when you need to communicate between threads). It would be unfortunate to get git-bisect point to a half-baked state or to a huge single commit. And it's also easier to review clean commits one by one.

> What the best way to do that? Make two new branches/PR (1: threads; 2: avoid multiple computations) and cherry-pick?

I'd make a separate branch, initially as a copy of this one, and then do a `git rebase -i master` in both branches, removing unneeded commits, editing others as appropriate, and squashing some when it makes sense.

gzotti:
Superseded by #3847

PR #3802, opened 2024-07-08
Bundled QtCompress

alex-w:
### Description
Restore support the building Stellarium with bundled `qtcompress` library in additional to Qt-provided library for manage ZIP archives. Using bundled `qtcompress` library provide binary packages of Stellarium without `qt6-base-abi` dependency in our PPA.

Fixes #3793 (discussion)

### Type of change

- [x] Housekeeping


10110111:
What are the cases when one would want to use the Qt-provided version? Are the maintainers of, say, Linux distributions supposed to always use the Qt-provided version?

alex-w:
> What are the cases when one would want to use the Qt-provided version? Are the maintainers of, say, Linux distributions supposed to always use the Qt-provided version?

I think for source-based Linux distros, Qt-provided version will be better solution, but maintainers should have known best way for their distros.

alex-w:
Any news for review?

10110111:
I'm not completely convinced that we want to use the Qt-provided code in this case, since it's simply unclean to rely on implementation details. OTOH, this flag is already an improvement, since we can now disable this reliance.

If all the issues with mixing different licenses are clear (which I'm not competent enough to check), I don't have any additional objections.

10110111:
I wouldn't bet on it. The request was added 15 years ago at the times of Qt 4.5, and nothing changed since.

PR #3828, opened 2024-07-29
fix the name in find query

mosaicthej:
The .config installed by github's fast_float, makes the library to be exposed as FastFloat, not fast_float.

So current situation is that no matter if the library exist or not, it pulls from github remote anyways.

Also updated the version to be 6.1.3 as it's the newest release.

### Description

Fixes # (issue)

The current CMakeLists would pull fast_float from github anyways irrespective if the library is installed or not.

I changed the NAME field to FastFloat so it would pick up if previously has been installed.

### Type of change
- [x] Housekeeping

### How Has This Been Tested?
Tested with a clean install of fast_float library, then build stellarium. When change the NAME, cmake won't pull from remote.

**Test Configuration**:
* Operating system: 
*  - OS: Gentoo Linux x86_64 
*  - Kernel: 6.10.0-gentoo-x86_64 
* Graphics Card: Intel Raptor Lake-P [Iris Xe Graphics], 13th Gen Intel i9-13900H (20) @ 5.200GHz 

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1048,9 +1048,9 @@ ENDIF()
 
 include("cmake/Check-from_chars.cmake")
 IF(NOT FROM_CHARS_WORKS)
-    CPMFindPackage(NAME fast_float
+    CPMFindPackage(NAME FastFloat
                    GIT_REPOSITORY https://github.com/fastfloat/fast_float
-                   VERSION 6.1.0
+                   VERSION 6.1.3
                    EXCLUDE_FROM_ALL yes)
     GET_TARGET_PROPERTY(fast_float_INCLUDE_DIRECTORIES
                         FastFloat::fast_float INTERFACE_INCLUDE_DIRECTORIES)


10110111:
Does it work if the user has version 6.1.0 installed?

mosaicthej:
> Does it work if the user has version 6.1.0 installed? 

Probably not? it probably just pull in 6.1.3 from remote then, like what we used to always have. But in this way everyone would be having 6.1.3 with stellarium. 

CMakeList's VERSION specifics the **minimum** version, so even if i entered 6.1.0, for people having 6.1.3 it would still work (i just tested on my system, it worked). 

That was not the case with the main issue, the main issue is the CamelCase not snake_case for the fast_float (I know it's confusing, the package installed in system is FastFloat but the repo is called fast_float lol).

Do we want to use 6.1.0 (and higher) or use 6.1.3 (and higher)

https://github.com/fastfloat/fast_float/releases

From the release notes, there shouldn't be any ground-breaking change, I guess this commit might be something useful? https://github.com/fastfloat/fast_float/pull/167

I just think it's also a good habit to keep it with the current version, maybe I should do in 2 commits as it's 2 different purposes.

Let me know what you think

10110111:
Yes, I was just wondering if this is because of some name change (which some projects do, unfortunately), or simply because of my mistake. I see the following [here](https://github.com/fastfloat/fast_float/blob/d7417618f93d2c47e9bbde561510f9fc8bafe003/CMakeLists.txt#L31), which makes me wonder.
```
add_library(fast_float INTERFACE)
add_library(FastFloat::fast_float ALIAS fast_float)
```

alex-w:
Hmm… I found a similar [patch for FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/astro/stellarium/files/patch-CMakeLists.txt)

DarthGandalf:
> I see the following [here](https://github.com/fastfloat/fast_float/blob/d7417618f93d2c47e9bbde561510f9fc8bafe003/CMakeLists.txt#L31), which makes me wonder.

That defines how the target name will look like after successful `find_package`. But `find_package` itself looks at the filenames, which in the case of fast_float [are set](https://github.com/fastfloat/fast_float/blob/d7417618f93d2c47e9bbde561510f9fc8bafe003/CMakeLists.txt#L55-L57) to `FastFloat`.

10110111:
Now that I think of it, the change of the name shouldn't have bothered me too much, since `fast_float` library is just a temporary workaround until all the target machines' toolchains implement `std::from_chars` (see 389dbd5dff553e31a408b16bbc9fc8e3b39e535e). Thus, we can just pull the latest version and follow whatever naming it defines, without much regard to what version is installed on user's machine (it's likely that none is even available from the repos of such old operating systems).

@DarthGandalf If you can confirm that this PR fixes the issue for you, I think we can merge it.

DarthGandalf:
@10110111 I had opened #3949 which does essentially the same, without seeing this PR first :) This version should work too.

alex-w (approved):

Issue #3148, opened 2023-03-31
Zooming sometimes leads to a hang for a second or two


10110111:
### Steps to reproduce
1. Make sure that initial zoom level is on the order of 80°
2. Launch Stellarium during the night (or disable atmosphere after launching)
3. Scroll deeply at once to zoom quickly into about 10°

### Expected Behaviour
Smooth video, maybe with some loading messages but without stuttering.

### Actual Behavior
A progress bar appears in the bottom-left corner, first partially out of the frame, waits about a second, then on a subsequent frame it's completely visible and says "Loading 'Nebulae'" for another second, then the progress indicated increases, and soon after this the video resumes.

### System
* Stellarium version: 3a96310b84bf1b2e9de4f00c8b075ce0c148baf1
* Operating system: Ubuntu 20.04
* Graphics Card: Intel UHD Graphics 620
* Screen type (if applicable): FullHD

### Logfile
[log.txt](https://github.com/Stellarium/stellarium/files/11124605/log.txt)

### Config file
[config.ini.gz](https://github.com/Stellarium/stellarium/files/11124632/config.ini.gz)

10110111:
The call to `loaderThreadPool->setMaxThreadCount()` in `StelTextureMgr::StelTextureMgr` appears to be the reason of this problem. Apparently, 8 threads on my machine take too much CPU time and bandwidth, so that the animation stops working for some time. Replacing the `QThread::idealThreadCount()` with `4` removes all lags here, so that the progress bar shows growing percentage while the animation, albeit less smooth, still continues, making the app usable.

10110111:
My CPU is Intel(R) Core(TM) i5-8265U CPU @ 1.60GHz, which has 4 cores and 2 threads/core.

> Or maybe, (my idealThreadCount is now 20), make it min(16, idealThreadCount()/2) ?

I'm not so sure how this will behave on other CPUs, including very high-end machines. Will reducing the number of threads slow down loading noticeably?

gzotti:
I am "guilty" for 984951b3c04c25deed630679ec23bf81a794fe85 where I raised this number from 4 to 16, given the current generation of CPUs. I had expected a little speedup by loading more in parallel if possible, which was however almost unnoticed. The short stall was there before and after, and I did not notice a difference there, but yes, the loading bar vanished faster. Now, if actually starting the threads cause the stall and starting more threads causes longer stall, we should revert to fewer threads. 

[...]

PR #3847, opened 2024-08-09
Threaded planet computation (and a few more speedups)

### LINKED ISSUE #1427: show comet locations regardless of magnitude

gzotti:
### Description

This is a revamped version of #3794 which was too experimental and is hereby superseded. (The attempted trial of using std::transform_reduce() showed useless.)

All SolarSystem object (SSO) computations so far were run on the main thread. This branch allows splitting the SSO position loop onto several threads. Likewise, the far-to-near sorting can be split. This is useful in case of thousands of objects. Given the complexity of drawing commands, there is not much use for more than 4 additional worker threads even for 25.000 objects. The additional threads can be configured or even suppressed in GUI.

Further, using the Intel VTune profiler, I have identified a series of sometimes surprising inefficiencies addressed in the later commits of this PR. 

With 27.701 SSO objects with some default setting (Preetham atmosphere, a few gridlines, stars, landscape), this improves framerate from about 6 (V24.2) to about 18. 

A new feature that is possible by this is a display of symbolic markers for minor bodies. These are plotted regardless of accuracy of orbital elements, and just provide an impression of distribution of the minor bodies. Color coding is applied per pType. The colors are currently hardcoded. However, drawing those many objects, despite adding code similar to StarMgr for mass drawing, is slow. I hope another developer can optimize StelPainter so it could be re-used instead of frequent buffer alloc/delete. 

 
Fixes #1427 (not intended, just happened...)

stellarium-003
### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Load many SSOs with the SSEditor. At least the first 10.000 asteroids and 1000comets.ini file. Then find new options in View/SSO and Config/Tools tabs. 

**Test Configuration**:
* Operating system: Win11 home on i7-12700H
* Graphics Card: Geforce RTX 3070 Ti Laptop GPU

Will test on smaller notebook, older Win10 PC and RasPi4.

## Checklist:


10110111 (commented):
Isn't c1302640804ec257a5f0f9be7d2edde316b88dd7 no longer needed?

src/core/planetsephems/jpleph.cpp
```diff
@@ -578,20 +581,21 @@ int DLL_FUNC jpl_state(void *ephem, const double et, const int list[14],
...
 	struct jpl_eph_data *eph = static_cast<struct jpl_eph_data *>(ephem);
 	unsigned i, j, n_intervals;
-	uint32_t nr;
 	double *buf = eph->cache;
 	double t[2];
 	const double block_loc = (et - eph->ephem_start) / eph->ephem_step;
 	bool recompute_pvsun;
 	const double aufac = 1.0 / eph->au;
 
+	mutex.lock();
```
10110111, line 590: `QMutexLocker` would be useful here. Using it you'll no longer need to call `unlock` before every return, the destructor will to it for you.

gzotti:
> Isn't [c130264](https://github.com/Stellarium/stellarium/commit/c1302640804ec257a5f0f9be7d2edde316b88dd7) no longer needed?

The std::sort also accepts this parameter.

gzotti:
Test on small notebook: 30fps (V24.2) --> 45fps. Quite a gain!
Test on Raspi4: Still only black window like last time I tried (i.e. black screen is unrelated). I can save screenshots, though. Sorry, I have no time to follow up on this.

src/gui/StelGuiItems.cpp
```diff
@@ -774,15 +774,18 @@ void BottomStelBar::updateText(bool updatePos, bool updateTopocentric)
 	}
 
 	QString planetName = core->getCurrentLocation().planetName;
-	QString planetNameI18n;
+	if (planetName!=planetNameEnglish)
+	{
```
10110111, line 778: This change leaves indentation broken.

gzotti: Ah yes. I was just so stunned by the effect of this single change that I just committed. Will fix that and rebase -i .

gzotti:
It's unfortunately not just the long ssystem_minor.ini. Does master run on your RPi4 with Ubuntu 24.4? I only see the black screen reported earlier.  Activating wayland in raspi-config makes it only worse, Stellarium fails to launch. I have not tried RaspberryOS lately, though.

To see a large ssystem file, just use the SSEditor and load the first 10.000 numbered asteroids from MPCAT. (Loading takes minutes but eventually succeeds. Another place to add QFuture magic?) I succeeded with adding the "interesting" minor bodies as well. 17.000 more. But you can drink a few coffees while waiting for successful loading.

alex-w:
> It's unfortunately not just the long ssystem_minor.ini. Does master run on your RPi4 with Ubuntu 24.4? I only see the black screen reported earlier. Activating wayland in raspi-config makes it only worse, Stellarium fails to launch. I have not tried RaspberryOS lately, though.

Hmm... please run via CLI:
`MESA_GL_VERSION_OVERRIDE=3.0 MESA_GLSL_VERSION_OVERRIDE=130 QT_QPA_PLATFORM=xcb ./stellarium --opengl-compat`

gzotti:
Yes, right, I have just found that again this morning, and it works. 
Compiling however had shown again the occurrence of "please build with -fPIC". I don't know where to configure this. CXX_OPTS? It is required for ShowMySky, which I have disabled for this test now. @10110111 ?
Also, on Linux/GCC there are still several compiler warnings which MSVC does not show. Does any of you care about those?

10110111:
> Compiling however had shown again the occurrence of "please build with -fPIC". I don't know where to configure this. CXX_OPTS?

Ideally, Qt should provide this via their CMake config. I'm not sure what I'm doing wrong.

alex-w:
RaspberryOS 64bit (10k+ solar system bodies, Qt6-based builds):
[master] Solar System Observer ~ 1.2 FPS
[this branch] Solar System Observer ~ 4.2-6.8 FPS (1 extra thread)

10110111:
Do the markers enabled by the new checkbox represent all the objects added as you described [here](https://github.com/Stellarium/stellarium/pull/3794#issuecomment-2268843692)? I don't reproduce any serious reduction of performance with 0 extra threads (Core i5-8265U with built-in GPU), and for some reason I only get the green markers but not the white ones closer to the center. What do these white markers represent?

gzotti:
```
colorMap={
				{isAsteroid,     Vec3f(0.35, 0.35, .35 )},
				{isPlutino,      Vec3f(1   , 1   , 0   )},
				{isComet,        Vec3f(0.25, 0.75, 1   )},
				{isDwarfPlanet,  Vec3f(1   , 1   , 1   )},
				{isCubewano,     Vec3f(1   , 0   , 0.8 )},
				{isSDO,          Vec3f(0.5 , 1   , 0.5 )},
				{isOCO,          Vec3f(0.75, 0.75, 1   )},
				{isSednoid,      Vec3f(0.75, 1   , 0.75)},
				{isInterstellar, Vec3f(1   , 0.25, 0.25)},
				{isUNDEFINED,    Vec3f(1   , 0   , 0   )}};
```

The thin white ones are the overwhelming number of "ordinary" asteroids, to be expected mostly in the main belt between Mars and Jupiter, and the Jupiter Trojans. Pure white is for the few dwarf planets. SDOs and Sednoids are green. Someone might elaborate on the actual differences, and propose better colors.

Did you load 15.000 objects or more to put a little stress to the SSO computation? (The logfile will likely show many complaints, but will give the number of loaded objects.  Maybe your CPU can go very fast with 1 core, so the effect would cancel out?

10110111:
> Did you load 15.000 objects or more to put a little stress to the SSO computation?

I followed [your list](https://github.com/Stellarium/stellarium/pull/3794#issuecomment-2268843692) of things to add, and now I get this in the log:
```
Loaded 6879 Solar System bodies from  "/home/ruslan/.stellarium/data/ssystem_minor.ini"
Solar System now has 6962 entries.
```

So apparently, I've missed something. How can I check what's missing?

alex-w:
> So apparently, I've missed something. How can I check what's missing?

Adding first 10000 asteroids via Solar System Editor plugin...

10110111:
> Adding first 10000 asteroids via Solar System Editor plugin...

Hmm, I surely had done that step... Could this not have been saved?..

alex-w:
> Hmm, I surely had done that step... Could this not have been saved?..

The log should contain the report for result of adding extra asteroids - I have no problem for adding thousands asteroids, but it need a lot of time...

gzotti:
My final solar system had the first 10.000, second 10.000, plus the "interesting" objects. in total about 27.000 objects. Indeed the loading was terribly slow, in this time the program seemed to be unresponsive, but it was finished an
hour later.  (Maybe earlier, but I did something else. ---This would be a place for QFuture and progressbar stuff, but it is a rare call after all.)

Much of the overall speed gain in this (renewed) branch was found from studying profiling diagrams. I am new to this field, but was able to track down and remove a few expensive operations. As a general pattern, if you need GETSTELMODULE in a frequently run method (update, draw, ...), make it static. If you often need to retrieve strings and there is a chance they have not changed, cache them if possible. These fixes helped a lot.

Seeing profile runs with 0 and with 4 extra threads it seems indeed much of the gain from running extra threads is lost by the synchronizing waits which are only even shown when running those threads. If somebody has an idea what to do here, it would clearly help. Still, I consistently see improved framerate with just a few extra threads. 

However, now I think (also talking to colleagues) the frequent alloc/free in the many short-lived StelPainter instances may be what currently slows it down most when the object count is large. If we want a fluent SolarSystem with many objects, we should do something about the actual drawing, re-use StelPainter's internal buffers or whatever. The Planet::draw() may need to be rearranged. Maybe using a "modern" (float) Z buffer can be an option, this would also fix the orbit overdrawing.

gzotti:
Hm, the low framerate replicates mine. You see the 10-15% gain with a few threads. Not sure why 7 is shown slightly worse than 3, but it shows that SSO computation is just a tiny fraction of the per-frame operations. 

Here are the "flame graphs" from VTune made last week in a terrestrial scene without landscape, atmosphere or gridlines. Capture time was about 20 seconds.  (Sorry for my shaky lines, mouse drawing is not my profession. I discovered the arrow tool only later...)

0 extra threads:

2024-08-22 12_16_16-Intel VTune Profiler
This shows the time spent in various methods stacked on top of the calling functions. 
Most of the QOpen[GL*] functions shown to be called in the StelPainter sections are constructing/destructing the temporal structures. 
Solar system positions make up about 20% of frame time. About 5% are the call to sort the planets by distance before drawing. A similar amount of time is spent with MinorPlanet::getEnglishName, but I think this one cannot be avoided. Time for computing magnitudes and extincted magnitudes have been largely reduced and are no longer labelled. Some frame sync (?) function is visible on the right. 

4 extra threads:
2024-08-22 12_01_39-Intel VTune Profiler
The planet position computation fraction has apparently shortened, but some ugly new WaitForSingleObjectEx is visible in lower left which steals most of the gained time. I cannot say if this is caused by the necessary Mutex introduced, or whether there is some other unavoidable reason for this time loss. Mutex is IIRC now only required in the JPL file reading functions, I have to run that again with VSOP. If it's the Mutex effect, it takes a disproportional fraction of time, occurring only for 8 planets opposed to ~15.000 KeplerOrbit runs. If you have other ideas what can cause this, I am willing to learn!

gzotti:
This may be caused by that annoying long wait shown at left, and that fact you mentioned earlier, that planet position computation is only a small part of the whole loop. 

If it takes, say, 40% of the frame time single-threaded, in the optimal case 4 threads could do the job in 10% of the original frame time which is however reduced by 30%, right?  That would be 1/7 or 14% of the new frame time. 14%*400+86%*100=142% CPU.

If it takes, say, 80% of the frame time single-threaded, in the optimal case 4 threads could do the job in 20% of the original frame time which is however reduced by 60%, right?  That would be 1/2 or 50% of the new frame time. 50%*400+50%*100=250% CPU.

There is always the drawing part that runs single threaded. This should be made as fast as possible by (possibly) re-using StelPainter buffers or other tricks.

Your fps degradation jump between 3 and 4 extra threads (and 6/7) is interesting. Maybe here the behaviour of the multithreaded std::sort changes? (I should also try if that is really useful...)

Or wait, on an 8-core machine, I think thread pool size would currently be set to 4 only (select any object and read the DEBUG auxiliary string in the object info!) so this may explain a conflict/wait for free pool thread in the thread pool. 

My own parallelisation starts n-1 threads which compute every nth planet each, and then the main thread computes the rest. When the main thread is done, it waits for the others which meanwhile should be completed or nearly so. We must do that three times for accurate positions (light time correction etc.). I am just using the available QThreadPool (`QtConcurrent::run()`). Maybe manual thread handling would be more efficient after all to get rid of these long ugly orange sync waits (if that is an effect of using the QThreadPool)?

10110111:
> There is always the drawing part that runs single threaded.

There should be no drawing given that I've turned off everything except the GUI (to see the frame rate). And sysprof tells me that in such a mode at least 50% of time is spent in `SolarSystem::computePositions` (checked with 0 extra threads). Not sure where the rest is, but stubbing out this function results in frame rate > 100FPS, which clearly indicates that almost all the time is spent there.

The stubbing out I did looked like the following, to make sure all the initialization is done before beginning to avoid the computations:
```diff
diff --git a/src/core/modules/SolarSystem.cpp b/src/core/modules/SolarSystem.cpp
index aa9a6ed3cf..65f4998404 100644
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -1522,6 +1522,9 @@ bool SolarSystem::loadPlanets(const QString& filePath)
 // The order is not important since the position is computed relatively to the mother body
 void SolarSystem::computePositions(double dateJDE, PlanetP observerPlanet)
 {
+    // XXX: debug
+    static int timesDone = 0;
+    if(++timesDone > 100) return;
     StelCore *core=StelApp::getInstance().getCore();
     const bool withAberration=core->getUseAberration();
     // We distribute computing over a few threads from the current threadpool, but also compute one stride in the main thread so that this does not starve.
```

gzotti:
Yes, and... sorry, but what do you want to check by this? Of course, not drawing planets means more relative time is spent in computing them. And if you don't compute them, you may have insane framerates, or you will see more calls for GUI items or DSO images not previously listed. 
I have just found another few percent of QString avoidance potential...

10110111:
> what do you want to check by this?

What I learned from this is virtually all the time is spent in computing planet positions when I turn everything off by typing the cheat code `gaspimq` after launching Stellarium. Then, if I, after adding one more thread, don't get 200% CPU usage and (approximate) doubling of frame rate, then something is parallelized suboptimally, and this is not related to drawing at all.

gzotti:
When you practically disable (exit early) SolarSystem::computePositions(), you prevent the extra threads from ever becoming configured and active, later in the method. 
There are 2 things to go after: 
- find out what causes the long sync waiting and fix if it can be avoided. Currently this steals much of the gain.
- Begin optimizing StelPainter. This has not been touched so far, apart from doing batch rendering of asteroid markers.  Creating and destroying about 3 StelPainters per object is costly when we have hundreds or even thousands of objects. This can be done in a later step.

10110111:
> When you practically disable (exit early) SolarSystem::computePositions(), you prevent the extra threads from ever becoming configured and active, later in the method.

Well, anything that could be configured has been configured in the first 100 iterations which are not affected. Moreover, I didn't really care about threading in this experiment. The test was simply to determine whether the method is the only CPU hog, and it proved to be. Now we can focus solely on it, expecting linear dependence on the number of threads, ideally (until we hit hardware limits, of course).

gzotti:
To understand this right, you have launched with a certain number of extra threads in config.ini and just run and measure the first 100 frames? No time to even press gaspimq IMO. I don't get it. It is evident from above graphics that the current solution has its issues (orange WaitForSingleObjectEx bars), where much of the gain is lost again. Now the question is what can be done to avoid (or at least mitigate) the wait.

10110111:
> you have launched with a certain number of extra threads in config.ini and just run and measure the first 100 frames?

No, the idea was to localize the CPU hog experimentally. Too often do profilers give strange results, and your words that something could be taking some additional time thus preventing linear dependence of frame rate on parallelism also made me doubt this. Now that I'm convinced that the only CPU hog currently, after gaspimq, is `computePosition`, I can focus on it and its callees. That's all. I'm now experimenting with different approaches, all inside this function.

gzotti:
Ah, ok. I just switched everything interesting off (esp. all the gridlines) and saw an insane 473fps. (That is, with default solar system from 24.2.) Without Milky Way, 505fps (of course, a black screen is not too useful...). Local installation of DSS: almost 400fps. Unfortunately, this ended in a crash after a minute (1 of 3 runs now).

src/core/modules/SolarSystem.cpp
```diff
@@ -1608,7 +1610,7 @@ void SolarSystem::computePositions(double dateJDE, PlanetP observerPlanet)
 					const auto planetPos = it->data()->getHeliocentricEclipticPos();
 					const double lightTimeDays = (planetPos-obsPosJDE).norm() * (AU / (SPEED_OF_LIGHT * 86400.));
 					Vec3d aberrationPush(0.);
-					if (withAberration && (observerPlanet->englishName!="Earth" || it->data()->englishName!="Moon"))
+					if (withAberration && (observerPlanetIsEarth || it->data() == getMoon()))
```
alex-w, line 1613: Inverted logic…

src/core/modules/SolarSystem.cpp
```diff
@@ -1683,7 +1685,7 @@ void SolarSystem::computePositions(double dateJDE, PlanetP observerPlanet)
 				const auto planetPos = it->data()->getHeliocentricEclipticPos();
 				const double lightTimeDays = (planetPos-obsPosJDE).norm() * (AU / (SPEED_OF_LIGHT * 86400.));
 				Vec3d aberrationPush(0.);
-				if (withAberration && (observerPlanet->englishName!="Earth" || it->data()->englishName!="Moon"))
+				if (withAberration && (observerPlanetIsEarth || it->data() == getMoon()))
```
alex-w, line 1688: Inverted logic again…

src/core/modules/SolarSystem.cpp
gzotti: oops, yes...

10110111:
My latest commit now yields the following figures:
|Extra threads| Frame rate, FPS|
|:-----------:|:--------------:|
| 0 | 26.7 |
| 1 | 34.3 |
| 2 | 39.7 |
| 3 | 42.7 |
| 4 | 38   |

I'm still not sure why there's a limit on 4 additional threads though while I have 8 virtual CPUs.

gzotti:
Code looks good! Yes, I also had thought of placing that in one longer loop to avoid 2/3 of the waits. Must test and profile...

main.cpp, line 473, sets ThreadPool size to 3 on your 8-core CPU. This is because the Textureloader (with its private pool) can also use a bunch of threads, and I did not want to overschedule the CPU, esp. after the first tests it appeared more than 3 are rarely beneficial. They would usually not run at the same time of program use, though. You could try to set that to idealThreadCount() (i.e., number of virtual CPUs).

gzotti:
VTune fails. "Assertion failed: thread_manager_impl:1321: (threads_count != (tpss_u32_t)(-1)) : . Please contact the technical support.". 
Without parallel sort, it works again, the flame graph (8 extra threads, 27.000 SSOs) is:
2024-08-23 00_47_29-Intel VTune Profiler

This shows the parallel sort also had a high overhead, the right wait in the earlier graphs. A sync overhead (orange, left) is still there and is as long as the actual threaded runs, I cannot say why. The string compares seen earlier and removed have vanished. Now sort is the worst single offender (18%), and Planet::draw() (30% of frame time) is dominated by MinorPlanet::getVMagnitude (6%), and as said StelPainter constructor (6%) and destructor (7%). 

Will continue on weekend.

gzotti:
Uh, sorry. I had excluded most plugins. Will fix the mess ASAP.

10110111:
My new commit results in the following frame rates after I turn off all the drawing:

|Extra threads| Frame rate, FPS|
|:-----------:|:--------------:|
| 0 | 30 |
| 1 | 43 |
| 2 | 53 |
| 3 | 56 |

For some reason, even when I comment out the code limiting the number of threads to 4, I still get no improvement after going beyond +3 threads setting.

gzotti:
Good!

We are using the global threadpool, the size of which is auto-controlled by QtConcurrent (?) and the OS. I had a debug string attached to the StelObjectMgr which you unfortunately have taken out which gave some threading info. It was visible when any object was selected. 

I see 9.7/13.3/15/15.8/16.6/16.4/16.6/...17.5 with 0...16 threads on my 20-core CPU.  So yes, diminishing revenue with more than 4 extra threads. Startup message says threadpool size 20.  FPS is dancing too much to detect a clear falloff. This is with 27.000 objects, not plotted, just computed. Plotting Markers bogs down to <5fps.

When I switch off planets (P), I have 19/34/46/53/60/64/67/70/75/...102. 

With latest detection&avoidance of string comparison I even have 20fps. String operations seem really slow. Maybe we should start checking whether QStringView could be helpful in some places. 

For unknown reasons, VTune now does not like threaded sort (invalid trace file). When I disable this, I see only 8 fps and faster, with sort by far the largest time eater. The more threads are running, the longer the orange WaitForSingleEx. This is truly annoying. I had suspected RemoteControl messages colliding, but without Remote* plugins it's the same. 

When profiling with 16 extra threads and no planet drawing (85fps from the profiler), the orange wait is more than 50%, but overall the CPU is displayed to be finally better utilized. 

Having no other experience with VTune or a comparable program to profile, I don't know what actually to expect. The orange wait bar looks frustrating, but I don't have a clue what triggers it.

10110111:
Can you somehow find out where that `WaitForSingleObjectEx` is called from? It seems that the only waiting point you've introduced with this PR is the mutex in JPL ephemeris, but I don't even catch the entry into `jpl_init_ephemeris` (even adding `abort()` at the beginning of this function doesn't change anything), so apparently this file is not even active.

Another source could simply be `QFuture::waitForFinished`, but this should take quite small time, since we execute the final block in the main thread, spending the same if not longer time as the pooled threads before waiting.

10110111:
Oh, actually, from your screenshots it's rather obvious that this waiting is the expected waiting inside the pooled threads: idle waiting for a task. The question is now how you are doing the profiling: do you turn off all the items to draw (`gaspimq`)? If yes, then I wonder what can be taking additional time making the threads idle for half the profiling time.

gzotti:
Yeah, that's the big problem. I don't see any place in VTune that would show where it is called. I think it is hidden in the inner workings of QtConcurrent. I just see the call to `QThreadPoolThread::run` is time-split into `QtConcurrent::RunFunctionTaskBase<void>::run` (which calls the actual processing) and `QWaitCondition::wait`, the latter calling (again?) `QWaitCondition::wait`, calling `QWaitConditionPrivate::wait`, which then calls `WaitForSingleObjectEx`.

Yes, the mutex in JPL file is only ever active for the 8 planets and only of you use JPL algorithms (DE4** options and extra files). Likely in 8 out of 15...25.000 or so objects it is hard to find. I cannot notice a speed difference w.r.t. VSOP87. 

Yes, I had expected the sync wait for the futures to be just number_of_threads*wait(), which should not be so much. However, it may indicate some auto-blocked access to a contested resource that is blocked per-object, not per-thread. 

In my original attempts, before I settled on using just a handful of threads and splitting load, I had used QtConcurrent::map() which could take a Threadpool and modifier functions to work on each entry in the systemPlanets list, auto-scheduling distribution to the available threads. I cannot say whether this could avoid the wait, but it's worth a new try. 

I have detected one other point of optimization, though, will commit that in a few minutes.

10110111:
> However, it may indicate some auto-blocked access to a contested resource that is blocked per-object, not per-thread.

No, from the graphic call stack it's obvious that the wait happens from a call outside `SolarSystem::computePosition` and its caller lambda, so it's definitely the wait for a task. And from that same screenshot I actually see that there's a `StelPainter` actively used, so the drawing is on, which then explains why the pooled threads have to wait.

src/core/modules/SolarSystem.cpp
```diff
@@ -1686,7 +1687,7 @@ void SolarSystem::draw(StelCore* core)
 	}
 
 	// And sort them from the farthest to the closest. std::sort can split this into parallel threads!
-	std::sort(STD_EXECUTION_PAR_COMMA
+	std::sort(//STD_EXECUTION_PAR_COMMA
```
10110111, line 1690: Was this intended for this commit?

gzotti:
So should I test with really everything switched off, deep-sky photos, milky way, zod. light? I see those when I zoom into the remaining narrow column of QEventLoop::processEvents...StelApp::update.

src/core/modules/SolarSystem.cpp
gzotti: Yes, temporarily, The auto-threaded sorting for some reason currently kills the profiler, and I think it's irrelevant for our discussion now. Error message a few sentences above.

10110111:
> So should I test with really everything switched off, deep-sky photos, milky way, zod. light?

Well, I'm currently concerned only with `SolarSystem::computePositions`, and this is why I'm testing without drawing any elements. It would be useful to see the dissection of time spent there to understand what limits the improvement currently.

I'm still struggling to find out why there's such an inefficiency in thread count vs frame rate. It stays even when I manually increase time taken per thread by 10× (making the loop repeat 10 times). Looks as if something is blocking somewhere, maybe because of reading some file? Are any files read on every computation of position if JPL data are not in use?

10110111:
One of your latest commits broke Observability plugin, I get lots of errors like this:
```
/home/ruslan/src/stellarium/plugins/Observability/src/Observability.cpp: In member function ‘void Observability::updateSunData(StelCore*)’:
/home/ruslan/src/stellarium/plugins/Observability/src/Observability.cpp:774:41: error: no matching function for call to ‘Planet::computePosition(double&, Vec3d)’
  774 |                 myEarth->computePosition(yearJD[i].second, Vec3d(0.));
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

gzotti:
If you don't have JPL stuff enabled, I am not aware of any file read. VSOP data are compiled in. I need to find (again) a tool to check files opened by programs. 
New profile with fully black screen.

2024-08-24 10_50_24-Intel VTune Profiler

There are still painters active, though, from non-drawn Skycultures and Asterisms. Here a zoom into the StelApp::draw column. 

2024-08-24 10_57_44-Intel VTune Profiler zoomDraw

gzotti:
Grmpf, done again... OK. Yes, I build without most plugins for now, also to avoid side-effects.

gzotti:
Tried to convert to QtConcurrent::blockedMap(). The net result is that main thread is waiting for the worker pool threads to complete. It works and makes the code simpler to read. But the long percentage of waits persists. No progress :-(

10110111:
> It works and makes the code simpler to read.

I wonder if the performance differs, since there are multiple obvious ways to map a function on a range (e.g. use blocks as I did, or call each new function in a freed thread; the latter way I think would have more overhead when the blocks of the former approach are of almost equal sizes).

gzotti:
I had now suspected that both my "strided" and your "block" approach would even see the systemPlanet list as contested resource when not using map(). Fortunately not. I cannot see any notable performance difference of map(), but could commit what I did if you want.

10110111:
f0b4ab87037b1ef1ec819b3f9d8c2cf9b69b525a applied the `blockingMap` change, so now there's no way to choose parallelization level.

gzotti:
Sorry, reverted. I think observerPosFinal is not needed (and the mapping function is IIRC only taking one argument). 
To test, just comment out path (1) and activate path (2).

10110111:
> I think observerPosFinal is not needed

Yes, I somehow didn't even think of updating the reference-captured variable. But you haven't set `obsPosJDE` to be captured by reference, so now the code has different semantics.

gzotti:
You mean, a &obsPosJDE in the capture list? (I am still shaky with that lambda argument when it's not [=]). Does it change anything in result? (It's used read-only.)

10110111:
> Does it change anything in result? (It's used read-only.)

Yes, when it's `[obsPosJDE]`, it's copied by value as a data member of the lambda. When it's `[&obsPosJDE]`, it's referenced by a reference data member, so the change to the variable after the lambda is created is reflected on the subsequent executions of the lambda.

10110111:
Hmm, I now think that this whole updating of `obsPosJDE` might be not completely correct. At least, this seems to be semantically different from `master`. Particularly, `lightTimeDays = (planetPos-obsPosJDE).norm...` was previously using the initial value of the observer's position, and only for `computeTransMatrices` was the new value obtained. Now the new value is being used for all the planets and trans-matrices.

gzotti:
OK, the remaining string comparison is indeed much lighter now: Previous run (max fps about 85-92)  zoomed on the self-running threads:
image

New solution:
image

Overall image:
image

This is with 16 extra threads, nothing but GUI is drawn. Reported 53.6% of CPU time waiting for the Unknown. :-(
image

Top of Bottom-Up view. 
image

10110111:
The focus of the profiler on this waiting time may be misleading. What if you apply the following patch? It might lead to a bit lower performance due to constant creation and deletion of threads, but at least there'll be no waiting noticed by the profiler (except the `join()` call).
```diff
diff --git a/src/core/modules/SolarSystem.cpp b/src/core/modules/SolarSystem.cpp
index fdf84ef492..737e1771d4 100644
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -47,6 +47,7 @@
 #include "AstroCalcDialog.hpp"
 #include "StelObserver.hpp"
 
+#include <thread>
 #include <algorithm>
 #include <execution>
 
@@ -1606,19 +1607,19 @@ void SolarSystem::computePositions(StelCore *core, double dateJDE, PlanetP obser
                 processPlanet(planets[i]/*, observerPosFinal*/);
         };
 
-        QList<QFuture<void>> futures;
+        std::vector<std::thread> threads;
         const auto totalThreads = availablePoolThreads+1;
         const auto blockSize = systemPlanets.size() / totalThreads;
         for(int threadN=0; threadN<totalThreads-1; ++threadN)
         {
             const int indexMin = blockSize*threadN;
             const int indexMax = blockSize*(threadN+1)-1;
-            futures.append(QtConcurrent::run(loop, indexMin,indexMax));
+            threads.emplace_back(loop, indexMin,indexMax);
         }
         // and the last thread is the current one
         loop(blockSize*(totalThreads-1), systemPlanets.size()-1);
-        for(auto& f : futures)
-            f.waitForFinished();
+        for(auto& t : threads)
+            t.join();
 
         // (2) INSTEAD: Blocking Map
         //QtConcurrent::blockingMap(systemPlanets, processPlanet);
```

gzotti:
I have tried to find info on VTune and multithreading. Few but intersting videos. In a "pro" edition of that tool apparently code can be augmented by extra labels. I don't have that. 
The framerate is abysmal, 16 instead of 85 or so, but maybe thread re-use would be an option against that. Or what can be done?

For some reason profile trace loading took ages. 
image

Also here are long unknown functions colored as "system". In another graph the "Effective CPU Utilization" dropped from 12% to 1.0%.  (Forget the colored "Idle-Poor-OK" bar, limits are self-configurable. )

image

10110111:
> The framerate is abysmal, 16 instead of 85 or so

Wow, I knew that in Windows threads aren't lightweight, but I didn't expect so much inefficiency.

> maybe thread re-use would be an option against that

Well, thread pool is used exactly for this purpose.

The chart looks quite crazy. Particularly, `Thrd_join` taking so much time while `func@0x1402...` is not seen in any thread except main is definitely a sign of something going wrong. Maybe the profiler somehow fails to identify these threads?..
Anyway, looks like the experiment didn't work well...

gzotti:
> Hmm, I now think that this whole updating of `obsPosJDE` might be not completely correct. At least, this seems to be semantically different from `master`. Particularly, `lightTimeDays = (planetPos-obsPosJDE).norm...` was previously using the initial value of the observer's position, and only for `computeTransMatrices` was the new value obtained. Now the new value is being used for all the planets and trans-matrices.

Ouch. Yes, I see sidereal time (and equivalently, planet rotations) have changed.

gzotti:
Here is a clip from a thread view. Running 17.000 objects, witout any drawing, even GUI switched off (~120fps):
grafik

With GUI (~120fps):
grafik

With GUI and planets (~20fps):
grafik

With GUI, planets and AsteroidMarkers (~7fps):
grafik

The main thread (top) is always busy. Good. The planet workers are busy at intervals of frame_time. There is one thread (bottom) that does something else just before the next set of planet computations. It appears to point into nvogl64.dll, i.e. NVidia driver, so this has to do with the actual drawing. That means, maybe the excessive waits we saw earlier are just the threads waiting for the next time they are called to compute planets, and we are already as good as it gets, at least concerning solar system object positions. (Apart from the new bug described above.)

10110111:
What is the main thread doing while the worker threads are waiting? I thought there's almost nothing to do, at least for so long.

gzotti:
Probably this is the module::update() stuff. Updating Faders etc. Note that many modules are "active" even when they don't draw. ... No they are not even listed in the top 30.
Zooming in on the main thread with nothing visible on screen:
grafik

I would say, most of that relates to the main thread's share of the solar system computation, and then OpenGL context stuff. I am surprised to see a function QRhiGles2::beginFrame. I am certainly not running ANGLE on Qt6.

10110111:
Ah, I didn't notice that it's just 5 ms. Then it seems OK for Qt to take this time (though I'm not completely sure it should take even this long).

> I am surprised to see a function QRhiGles2::beginFrame. I am certainly not running ANGLE on Qt6.

It's Qt's internals, Qt may be using GLES2 for drawing.

10110111:
I've done some experiments on the efficiency of parallelization, and here's what I found out. First, my initial attempts at making sense what's going on were impeded by CPU frequency scaling. Once I found a way to fix the frequency, things became much more reproducible.

Now, it appears that the more time the threaded loop takes, the closer the parallelization efficiency to ideal, see the chart below ("base frame rate" is the one in single-threaded mode, "real code" means the current unaltered branch). This should mean that there's a noticeable overhead in sending the function to the threads ~and waiting for completion~ (looks like 1.5 ms per thread in my experiment; commenting out waiting for completion doesn't change much).

So I guess we shouldn't look for any additional causes of this inefficiency, it's just a normal overhead that should be expected.

fps

gzotti:
YAY!
grafik
This is again 27.000 SSO, AsteroidMarker drawing. The percentage of StelPainter constructor/destructor is much smaller now.

10110111:
> The percentage of StelPainter constructor/destructor is much smaller now.

What about the framerate?

gzotti:
Hard to say. So much has changed in the last 3 days. It did not feel so much faster (5->6fps?), but then I realized my memories were with parallelized sort. Now, without atmosphere, with parallel sort and no gridlines I am at around 10fps. So, with 27.000 markers it finally becomes usable again! However, probably 17.000 objects (first 10.000 numbered, plus "interesting" ones from MPC) is the better compromise. 
I had started to configure a private ThreadPool for SolarSystem, just like for StelTextureMgr. What do you think?

10110111:
> I had started to configure a private ThreadPool for SolarSystem, just like for StelTextureMgr. What do you think?

What would the advantage of this be compared to using the global pool?

gzotti:
Maybe particular control about thread numbers. Why do we have an own Threadpool for textures? Generally, would it be wiser to just configure "take whatever number can be run without conflict", and not even have a GUI setting for that?

10110111:
> Why do we have an own Threadpool for textures?

Well I wasn't the one who implemented it, but from what I can see now, one advantage there could be to have more threads than CPUs in case when the loading is I/O-bound. This doesn't always seem to be the case though (due to #3148, as noted in the sources).

> Generally, would it be wiser to just configure "take whatever number can be run without conflict", and not even have a GUI setting for that?

I don't think it's the best choice, given the seriously diminishing returns on at least my CPU. I think for the beginning it might be OK to use 1 additional thread as a default (which is rather efficient as my chart above says), and let the user choose up to the hardware limit. But, OTOH, the default install of Stellarium doesn't have so many SSOs, so such split will likely reduce performance by adding the overhead of threading without any noticeable improvement. Thus, I guess the default should be 0. 

Also, this computation seems to be kind of a "main" computation, i.e. required for each frame, rather than a side activity as in the texture loading, so maybe the global pool is the better place for it.

gzotti:
OK. I will then make the GUI-available maximum extra cores "idealThreads"-1. 
Now we must still fix the broken position logic...

10110111:
> Now we must still fix the broken position logic

If you restore the variable you eliminated, this will be fixed. You could even make it reference-captured instead of passing it every time.

gzotti:
Comparing against 24.2: 10fps with thousands of asteroid markers down to mag25 vs. 6.6fps without any markers. This seriously has paid off!

alex-w (approved):
Thanks for great work!

P.S. Are you sure that phrase “Minor Planet Markers” correct?

gzotti:
Ah yes, I wanted to change that to "Markers for Minor Bodies" to include comets. And SUG changes are not complete. Currently running overnight test on laptop, and want to test 2 days on RPi with at least 5000 obs. Now, 17.000 obs, regular scene (stars, landscape, azimuths, atmosphere, a few lines), asteroid markers: 13.6fps. I think master with default ssystem_minor.ini was slower before. 

@10110111 when I ran no visible modules, as I had 600fps and more (and 400fps 2 days earlier), I experienced a few crashes. Any ideas?

10110111:
> I experienced a few crashes. Any ideas?

It would be useful to at least have stack traces...

gzotti:
I can send my crash data to Microsoft in the notification box, but still have to find where on my system they might be hidden away from the nosy user. (OK, Google will know the answer...)

gzotti:
The number of extra threads should be saved immediately (I much prefer immediate storage!). However, on startup the upper value of the spinbox is reset to the number of max. threads in the pool (IIRC) minus one. On an 8-core PC, this should be 7. It has happened to me that my "19" was reset to 16, not sure why. Maybe idealThreadcount-1 would be better.

10110111:
It doesn't even appear in `config.txt` for me.

gzotti:
It seems Satellites are the next candidate. regular program: 150fps. With Satellites on: 42fps. This first change pushed to about 44fps, not too much. With all these stupid sat constellations pushing the number of sats into the 50.000s, computation could be again put on separate threads, and there are certainly other places for optimization. Not necessarily in this branch, though.

alex-w:
Is std::execution used anywhere still?

gzotti:
Yes, as said, the std::sort which we use before the planets can be drawn uses it.

alex-w:
> Ah yes, I wanted to change that to "Markers for Minor Bodies" to include comets.

 Don’t forget text for action also…

gzotti:
Satellites: Couldn't resist trying. Using a QtConcurrent::blockingMap() (which just uses the global threadpool), this is now at about 80fps. Not committed yet. Now, should I just add one more commit and satellites are also done, or do I make the final commits (SUG, action name), and then wait until after merge and only then commit the satellite changes?

Raspberry Pi 4/RaspberryOS on smallish screen: Master:  12fps. This: 12fps. Hmm... That is, with stock ssystem_minor.ini. Then I loaded the "unusual" objects from MPC. I just added all. Bad idea: 42.000 objects. (We must crucially add the number of objects in the downloaded file to the SSE GUI!)  
My little RPi4 masters this with 3 extra threads, 111% CPU load, and 2.3fps. 24.2: 1.36fps. Of course, wrong platform for so many objects (I had hoped the file held ~5000 obs). My PC with these 42.000 "unusual" SSOs: 20fps@15 extra threads, 16@3, and about 7fps with markers.

10110111:
I guess this PR can be a playground from which we'll extract good changes into `master` one by one after they prove useful and flawless enough. Then rebase, and review further. Otherwise this huge change set is simply unmanageable for review.
As an example, I think the StelPainter change can be pushed to `master` now, just as the `(QString)` -> `(const QString&)` change.

alex-w:
Latest changes is not compatible with Qt5 :(

gzotti:
Yes, I have seen the CI messages. I would just revert to the previous solution for Qt5. BTW I always see FreeBSD CI is internally broken.

alex-w:
> BTW I always see FreeBSD CI is internally broken.

Nope: https://github.com/Stellarium/stellarium/actions/runs/10586733002

gzotti:
>extract good changes into master one by one

This is a now fully rebased sequence of improvements, with later fixes (apart from the last 2) squashed into the major contributions. Some have been retained separately to show the sequence history of improvements. If you want to reverse order of applications, please do it, but honestly I don't see the point. Overnight tests on RPi and my small notebook (13hrs so far) are OK. I will try to re-provoke and stackdump a 650fps crash with procdump for you.

alex-w:
@gzotti "SolarSystem Extra Threads" should be renamed IMHO

gzotti:
Any Suggestions?

Atque:
I get frequent, seemingly random crashes, using this branch. It doesn't help to limit the FPS.

alex-w:
> I get frequent, seemingly random crashes, using this branch. It doesn't help to limit the FPS.

Platform and hardware?

Atque:
> Platform and hardware?
OS: Windows 10 home 64-bit
CPU: Ryzen 7 3700X
GPU: RX 5700 XT (driver 24.8.1)

Built using Qt 6.7.2.

gzotti:
Uh, that's bad news then. Yes, I can provoke crashes when using the default ssystem_minor.ini (i.e. not running thousands of objects), switching everything visible off, at framerates of 500 and more (reported max: 678fps), and enabling more than 12 or so additional threads. 

Does it crash also with fewer than e.g. 4 extra threads, or even zero extra threads? Does it crash when you have imported e.g. the first 10.000 numbered asteroids from MPC (so there is actually need for distributing workload)? I have neither Ryzen nor any AMD GPU in my reach, and have again fundamentally too little time in the next weeks, so cannot test this.

Atque:
Okay this is weird. It seems a lot more stable with more SSO's. Does this make sense?

gzotti:
My only guess is too many GPU calls in too little time. But I cannot figure it out myself. My other, less powerful devices have run their 2-day tests without a glitch. Please report on the suggested tests. Does "a lot more stable" just mean rarer but still occasional crashes, or does it work stable? Is it stable with 0 extra threads? There is still a distributed sorting call if 0 extra threads are selected. Also, satellites are meanwhile distributed, without user choice. If it still crashes, please disable (unload) satellites plugin, and try again.

Atque:
Still occasional crashes, unfortunately.  Using many SSO's (10,000 e.g.) it is more stable while panning around and time-lapsing, but pressing AstroCalc will almost always lead to a crash (it seems the Positions tab specifically is fragile).

I have not managed to provoke a crash using 0 extra threads.

alex-w:
> Still occasional crashes, unfortunately. Using many SSO's (10,000 e.g.) it is more stable while panning around and time-lapsing, but pressing AstroCalc will almost always lead to a crash (it seems the Positions tab specifically is fragile).

I can confirm crash when AstroCalc/Position tool is open and the button to updating list is pressed

gzotti:
Too bad. Back to draft.

alex-w:
probably the issue in calling `rts = planet->getRTSTime()`

gzotti:
FYI, I cannot produce a crash in AstroCalc/Positions with 10.000 asteroids and 13 extra threads. "Update Positions" takes 10 ugly seconds of blocking wait, yes, but else works. 

Is there now a crash for you with this function and no extra threads? The intermediate solution would be for users to check and activate only if it works on their systems, allowing more users to participate, and maybe (dreaming) one will come in the next 4 years and find the issue.

gzotti:
@Atque @alex-w can you produce a crash when using VSOP87 position calculation only? (Disable DE430 etc in the config panel)

alex-w:
I can reproduce crash with VSOP87 - I didn’t check it with DExxx

10110111:
@alex-w Can you reproduce it under the debugger? If so, could you post a stack trace?

alex-w:
> @alex-w Can you reproduce it under the debugger? If so, could you post a stack trace?

I can't reproduce crash in Windows, I'll try it in linux tonight at home.

Atque:
I can produce a crash in Windows with the first 10,000 asteroids loaded, using the AstroCalc/Positions tab and 15 threads. Not with 0 threads though.

VSOP87 vs DE44X doesn't matter.

alex-w:
> > @alex-w Can you reproduce it under the debugger? If so, could you post a stack trace?
> 
> I can't reproduce crash in Windows, I'll try it in linux tonight at home.

Ouch... I can reproduce crash when extra threans are 1 or more.

image

10110111:
Interesting, apparently `p` becomes empty at some point. The only possible point for this seems to be the sorting, but even this sounds unrealistic. What if you remove `STD_EXECUTION_PAR_COMMA` part from the call?

gzotti:
But sorting takes place in draw(). In the next cycle, in computePositions(), computation will recount systemPlanets. 

The call p->getHeliocentricEclipticPos() cascades down to the Solar position. It should be initialized to 0/0/0 in the constructor, and updated to be zero by get_sun_helio_coordsv(). Is there still any chance for a problem?

E.g., clangd warns of a possible problem in Planet.cpp:L2044: Use of memory after it is freed. I have seen such warnings in several places. Do we need to take them more seriously?

10110111:
> The call p->getHeliocentricEclipticPos() cascades down to the Solar position

From the looks of the stack trace, the problem here happens at the point of dereferencing `p`. If you can show the disassembly around the crash point (*Debug*→*Operate by instruction* in QtCreator), this could clarify what's going on.

> clangd warns of a possible problem in Planet.cpp:L2044: Use of memory after it is freed.

Looks like a false positive. Nothing pointed to by `pos` is freed here.

gzotti:
I am afraid I am seeing another issue now: The overall algorithm earlier had 3 distinct loops. First, all planets and their moons receive an approximate position w.r.t. their orbital centres (Sun for planets, the respective planet for moons). From this, a first solution for heliocentric position can be derived, from which light time is found. The second loop repeated computations with light time correction, and a third loop further refined the solution. 
I am afraid now it's no longer guaranteed that planet moons will reply their positions w.r.t. the heliocenter, when their planets themselves may not have been computed or updated yet. So, I am afraid we should revert to the 3-loop solution. Each of the loops can of course be multithreaded.

I don't see however where this may be cause of our current problem.

10110111:
From what I see in `Planet::computePosition`, the aberration push is not used for the computation, it's just saved. Maybe you mean the calls to `RotationElements::updatePlanetCorrections`? These could be special-cased somehow without breaking into three loops. Such a breaking will degrade performance, so if there's a simple enough way to avoid this, it should be avoided.

gzotti:
No, line 1561 (see Alex' screenshot) computes lightTime from the distance from Planet to observer. However, finding the distance from heliocentric planet position for a moon requires (by iteration through the parent bodies) that the moon's planet has already been computed. In a threaded run I am afraid this is currently no longer guaranteed. Either we must deal with the 8 planets (or at least Earth...Neptune, and yes, Pluto (and do we have other minor bodies with moons?)) before doing the rest, or find some other solution.

gzotti:
@alex-w I started to cherry-pick to get at least the non-threading changes into 24.3. It seems you changed SSE in master and here, cherry-pick has conflicts. The changes are not related to this branch, so what is the "right" version?
grafik

alex-w:
> @alex-w I started to cherry-pick to get at least the non-threading changes into 24.3. It seems you changed SSE in master and here, cherry-pick has conflicts. The changes are not related to this branch, so what is the "right" version?

Hmm… the code should be identical - please use HEAD as “right” version

alex-w:
> Did you put the latest version also in master? Then I can ignore these commits.

Yes, I put all changes for SSE from this branch into master as one commit, because these changes are important.

gzotti:
OK, so I will better "rebase -i" and skip them.

gzotti:
OK, I cherry-picked most speedups from this branch which were not in the series of developing threading support to master. Now, I could try to recreate the original 3 separate loops.

gzotti:
There could be another strategy: While loading the solar system, create yet another list which holds pointers to Planet objects once they are used as parent. Then, apply the planet computation to those first, and only then to the others. This circumvents the problem with unknown planet positions when computing Moons first. It just adds additional complexity. @10110111 your ideas?

10110111:
I think we need to do something with the third loop first. It consumes a lot of additional time, and if we split the computation into multiple loops, this will also add to the threading overhead.

My idea is to first, outside of this PR, get rid of the third loop by modifying the second one. Particularly, as I understand (correct me if I'm wrong), the idea of it is to correct for the fact that the object has moved away from or towards the observer at the time of the computation (`dateJDE`), so the distance we compute first represents the current value at time `dateJDE`, while it should instead represent the distance at the moment of emission of the light that we are to draw.

While this iteration looks like a sensible approach, we can achieve a good approximation if we take object's velocity and solve for the position it likely had at the time of scattering of the currently observed light. The curvature of its trajectory is most likely negligible, so we can ignore the acceleration of the object. E.g. suppose the observer is on Pluto in aphelion. During the time it takes for light from Mercury to reach the observer (about 6 hours), Mercury will have rotated by about 1° in its orbit. It's 1% of deviation from the linearly extrapolated trajectory (i.e. Δx/Δy~0.01 for initial velocity along y), which I think is pretty acceptable.

Not completely sure of the objects with high eccentricities though, this may need some checking. If the acceleration appears to not be negligible, we could try computing it in addition to velocity, and hopefully this will still be much faster than adding the third loop.

gzotti:
The general approach of the original algorithm was: 
Round 1: Compute all positions at "time". 
Round 2: compute heliocentric position, and from this, distance, to find light time. Recompute positions at "time-lightTime"
Round 3: Refine, but only if needed. The third call to the actual Planet::coordFunc may already be avoided when the lighttime difference to the second round was small enough. So, the third loop should not cause headaches.

One principal problem lies somewhere else, as described above:

getHeliocentricEclipticPos() computes the heliocentric position of the object by traversing and accumulating positions of the parent. If the parent has not been computed already, the result is wrong, which is why these calls used to come after completion of computePosition() loops. This is a principal problem which may be excluded by first completely computing the few objects: Earth, Mars, Jupiter, Saturn, Uranus, Neptune, Pluto, and then the current observer planet (which may be a moon orbiting one of the planets). And possibly some other minors with moons which could all be identified and collected into a new list during loadPlanets. Obviously these few would be computed in the main loop before threading starts, and have to be excluded in those threaded loops.

It is a problem that may even out when running for a few frames, because the solar system is sorted by distance between drawing, and planets won't usually move a lot between frames, so usually the motion between Jupiter and Ganymed will hardly be noticeable, but it's a principal problem. Maybe running a script, centering on a moon like Ganymed, with time stopped, and setTime() to months in the future, will reveal the error, when Jupiter's moons suddenly refer to Jupiter's old position when computing heliocentric position before Jupiter has been computed.

However, I still have a hard time to trigger a crash in search for your lost p. Yesterday I was running a build "Release with debug info" with the debugger. Half an hour, computing and displaying AstroCalc ephemerides, running black window at 400fps (it's not 670 in the debugger...),  just did not fail to work. What do I have to press or do to trigger a crash?

10110111:
> Round 3: Refine, but only if needed.

But why refine — this is my point. We could compute it well enough from the beginning, using easily available velocity of the object.

> One principal problem lies somewhere else, as described above

Yes, I understood it when you described it for the first time. I just think that if we remove the third iteration, the solution with two separate loops might be good enough. The computation of parent planets before the others seems error prone, especially if we take into account possible future changes.

> Yesterday I was running a build "Release with debug info" with the debugger.

What if you launch the same build separately and only then _Attach_ your debugger to the already running process? Sometimes the debugger can disable some things (e.g. ASLR) which may prevent the crash due to excessively reproducible environment.

gzotti:
We don't need to bother with the third loop, it is likely only called rarely to begin with. The problem happens before the second, because the planets with moons must have been computed before the moons report their positions and distance for the very first time at the beginning of loop2. 
Trying to remember the history of this, yes, the first team had disabled speed information, I was lucky to find and restore that later. I am not 100% sure today all objects deliver speed, though. If you want to fix position by evaluating `pos3=pos2+(light_time_loop3-light_time_loop2)*speed2`, this may be possible and ok, will be faster than `computePosition(JDE-light_time_loop3)` and should be in practice indiscernible from `computePosition(JDE-light_time_loop3)`. But again, pos2 must have been computed for ALL planets with moons before the first moon goes into loop3 (your abbreviated or previous full position calculation).

OK, I could attach to the running program from the Debug menu. Black window fluctuating, 410, 550, 610fps. AstroCalc. Ephemeris lines. Redo. No crash. Which function to call to trigger a crash? And what is ASLR?

10110111:
> it is likely only called rarely to begin with

How so? There's no condition to skip it, be it partially or completely.

> If you want to fix position by evaluating `pos3=pos2+(light_time_loop3-light_time_loop2)*speed2`

No, my idea was to solve <b><i>r</i></b><sub>hist</sub>+<b><i>v</i></b>|<b><i>r</i></b><sub>hist</sub>|/c=<b><i>r</i></b><sub>now</sub> for <b><i>r</i></b><sub>hist</sub>, where <b><i>r</i></b><sub>now</sub> is the position computed from the first loop, and <b><i>v</i></b> is the full heliocentric velocity, again, taken after the first loop.

> No crash. Which function to call to trigger a crash?

Do you even reproduce it without a debugger? If not, then there's no point in trying to do it under the debugger.

> And what is ASLR?

Address Space Layout Randomization, a technique that makes some exploits harder to implement.

gzotti:
> > it is likely only called rarely to begin with
> 
> How so? There's no condition to skip it, be it partially or completely.

SolarSystem line 1577 announces: 
```
			// The next call may already do nothing if the time difference to the previous round is not large enough.
			p->computePosition(obs, dateJDE-lightTimeDays, aberrationPush);

```
and in Planet::computePosition(StelObserver*, ...) you find in line 1781:

`	if (fabs(lastJDE-dateJDE)>deltaJDE) [{coordfunc(...)}]`

where deltaJDE is an object-own distance-dependent estimate how much time may elapse so that position should actually be recomputed. If light times in loops 2 and 3 differ by less, coordFunc in loop3 is skipped.

> 
> > If you want to fix position by evaluating `pos3=pos2+(light_time_loop3-light_time_loop2)*speed2`
> 
> No, my idea was to solve rhist+v|rhist|/c=rnow for rhist, where rnow is the position computed from the first loop, and v is the full heliocentric velocity, again, taken after the first loop.

You will get the full  `Planet::getHeliocentricEclipticVelocity()` of a moon only after its planet knows it. Therefore those few planets with moons must have been computed first. Again. The current solution does not warrant that.
 
> > No crash. Which function to call to trigger a crash?
> 
> Do you even reproduce it without a debugger? If not, then there's no point in trying to do it under the debugger.

The whole thing is (fortunately, given the current logical bug) not merged because @Atque and @alex-w reported crashes when using something in AstroCalc. I would like to catch such a crash to see where and why it happened. You saw earlier that a planet pointer may suddenly be empty. I just cannot reproduce that. I now ran 500+fps black window overnight, 8 hours, all fine. The only errors that were added to the log since yesterday were the flood of 

>  OpenType support missing for "Arial", script 66
>  OpenType support missing for "MS UI Gothic", script 66
>  OpenType support missing for "SimSun", script 66 etc.

which have annoyed me for years now. But they are unrelated to this PR.

> > And what is ASLR?
> 
> Address Space Layout Randomization, a technique that makes some exploits harder to implement.
TNX.

10110111:
> Again. The current solution does not warrant that.

I realize this. I'm still thinking in terms of possible splitting of the loops, after which this should be possible (or do I get you wrong?). Just want to avoid the unnecessary (AFAICT) third loop.

> I now ran 500+fps black window overnight, 8 hours, all fine.

@alex-w @Atque Could you re-check with the current state of this branch if the crash still occurs?

alex-w:
I can reproduce the crash.

Steps to reproduce:
1) Open AstroCalc/Positions tool
2) Select "Solar system objects"
3) Press button "Refresh" 2-3 times

10110111:
> I can reproduce the crash.

Can you reproduce it in Debug mode? If yes, please post the stack trace.

alex-w:
> Can you reproduce it in Debug mode? 

Yes

> If yes, please post the stack trace.

Not a stack trace, but a "quick trace"...
image

gzotti:
Finally! After about 13 updates...
grafik
The variables in top-right are "inaccessible".
Other threads that point to our code:
grafik

grafik
grafik
The other threads point into disassembly, many are just waiting.

10110111:
Hmm, so @gzotti seems to have caught the culprit by the tail. The third screenshot shows that, apparently, `systemPlanet` was in the process of being reallocated. Thus, I think the following is in order:
```diff
diff --git a/src/core/modules/SolarSystem.cpp b/src/core/modules/SolarSystem.cpp
index a8b94ddc01..c9de5fad54 100644
--- a/src/core/modules/SolarSystem.cpp
+++ b/src/core/modules/SolarSystem.cpp
@@ -1599,7 +1599,7 @@ void SolarSystem::computePositions(StelCore *core, double dateJDE, PlanetP obser
         const Vec3d observerPosFinal = observerPlanet->getHeliocentricEclipticPos();
 
         // Threadable loop function for self-set number of additional worker threads
-        const auto loop = [&planets=systemPlanets,processPlanet,
+        const auto loop = [&planets=std::as_const(systemPlanets),processPlanet,
                    observerPosFinal](const int indexMin, const int indexMax)
         {
             for(int i = indexMin; i <= indexMax; ++i)
```

Please check if it stops crashing with this patch.

alex-w:
WOW! I can't reproduce the crash anymore :)

Atque:
Indeed seems solved. Great work!

gzotti:
> I realize this. I'm still thinking in terms of possible splitting of the loops, after which this should be possible (or do I get you wrong?). Just want to avoid the unnecessary (AFAICT) third loop.

It may be possible to just refine the second position with knowledge of the speed and new light time, OK. Or maybe with your way.

But I just realized even in the first (L.1597) there may be a danger. In
```
		processPlanet(observerPlanet, Vec3d(0.));
		observerPlanet->computeTransMatrix(dateJD, dateJDE);

```
there is a possibility of error when current observerPanet is a moon. The "carrier" planet will not be at the correct position, and the moon will thus have the same error.

gzotti:
I am about to commit a comparative solution. All 4 solutions available for comparison, following config.ini:[devel]/compute_positions_algorithm:

0: single-threaded original 3-loop
1: BlockingMap 3-loop
2: strided 3-loop
3: partitioned 1-loop (Ruslan's solution which I fear has various problems with moons)

With default ssystem_minor.ini and some particular config, I see about 0:160-210 fps; 1: 150-170 (!? too much overhead?),  2: 180-230fps; 3: 200-260fps

We could merge this sequence of changes (I would just remove a few recent experiments) in master for 24.3 and try to fix the known problems in path (3) until 24.4. Currently I default the ini key to 2, going for the strided 3-loop solution. 

I will test now with RemoteSync runs.

alex-w:
Good. I will wait your commit and merge to publish RC2

src/core/modules/SolarSystem.cpp
```diff
@@ -1543,88 +1544,317 @@ void SolarSystem::computePositions(StelCore *core, double dateJDE, PlanetP obser
...
-			const auto update = &RotationElements::updatePlanetCorrections;
-			if      (p->englishName==L1S("Moon"))    update(dateJDE-lightTimeDays, RotationElements::EarthMoon);
-			else if (p->englishName==L1S("Mars"))    update(dateJDE-lightTimeDays, RotationElements::Mars);
-			else if (p->englishName==L1S("Jupiter")) update(dateJDE-lightTimeDays, RotationElements::Jupiter);
-			else if (p->englishName==L1S("Saturn"))  update(dateJDE-lightTimeDays, RotationElements::Saturn);
-			else if (p->englishName==L1S("Uranus"))  update(dateJDE-lightTimeDays, RotationElements::Uranus);
-			else if (p->englishName==L1S("Neptune")) update(dateJDE-lightTimeDays, RotationElements::Neptune);
-
-			if(p != observerPlanet)
+			std::function<void (int)> plCompLoopOne = [=](int offset){
```
10110111, line 1674: No need in `std::function`, just keep it a lambda (`auto`). It will be converted into a Qt's callable anyway, so `std::function` is just an unnecessary overhead.

gzotti: According to QtCreator's tooltip, auto resolves to this type. I don't like auto, the written type is IMO clearer to read. Or does auto resolve differently on Linux?

10110111: A lambda is never an `std::function`. You can see it if you compile and query its type in a variable viewer or simply check `sizeof` it compared to the case when you've assigned it to an `std::function`. Each lambda has a unique type, while `std::function` is a type-deleting wrapper over any kind of callable: a lambda, a named class with `operator()`, a function pointer, etc.

> the written type is IMO clearer to read

It might be when you know the exact type (and for lambda there's no name for the type, except `decltype(myLambda)` which already needs you to have used `auto` in order to use `myLambda` further). And this is the pitfall where you can get inefficiencies like e.g. doing a copy instead of getting a reference.

gzotti: OK, thanks for the explanation. I change these functions to auto.

gzotti:
No idea why, but this rebase -i was difficult. Commit "fix crash" should have gone just after @10110111 's last commit, but it wouldn't go. We can leave it as-is as document that there was a problem.

alex-w (dismissed):
Great work!

gzotti:
Wait, something is off!

alex-w:
> Wait, something is off!

Where is? I see no problems...

gzotti:
Is Polaris where it should be at ~90-lat? It seems earth's axis is pointing somewhere else. (right: 24.2; left: this.) Embarrassing...
Screenshot (341)

alex-w:
Ouch.... I didn't checked it in this branch :-/

10110111:
Maybe this shouldn't go into 24.3 so late with this potential for breakages.

alex-w:
> Maybe this shouldn't go into 24.3 so late with this potential for breakages.

I fear this part should be postponed to version 24.4

gzotti:
I had not remembered that Ruslan had integrated the final call into his loop. Fixed that in the conservative paths. Now I can run RemoteSync'ed V24.2 and this. 135 vs 73 fps. As far as I can see, result with setting "2" are identical.

alex-w (approved):

PR #3858, opened 2024-08-20
Cast a reddish tint over the landscape when sun is low.

### LINKED ISSUE #3855: Tint landscapes orange and red at sunrise/sunset
A small but visually beautiful tweak would be for landscape panoramas to progressively tint yellow, orange and red once the Sun rises or sets, to adjust for the different coloring of the scenery.

Could also be applied to 3D Sceneries.

Values could possibly be called from the current extinction reddening tweak.

gzotti:
### Description

This may add a yellowish-orange tint to the landscape when the sun is low. 
When sun is -3...0° below horizon, tint is smoothly transitioning.
 Also when atmosphere is off, there is no tint.  

Fixes #3855 (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: (irrelevant)

## Checklist:

### DIFF
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -985,11 +985,13 @@ void S3DRenderer::calculateLighting()
 	}
 	else
 	{
-		//for now, lighting is only white
+		static LandscapeMgr *lmgr=GETSTELMODULE(LandscapeMgr);
+		QVector3D solarTint=lmgr->getLandscapeTint().toQVector();
+		//for now, ambient is only whitish. Directional and specular are tinted with possibly reddened sun.
 		lightInfo.ambient = QVector3D(ambientBrightness,ambientBrightness, ambientBrightness);
-		lightInfo.directional = QVector3D(directionalBrightness,directionalBrightness,directionalBrightness);
+		lightInfo.directional = directionalBrightness*solarTint;
 		lightInfo.emissive = QVector3D(emissiveFactor,emissiveFactor,emissiveFactor);
-		lightInfo.specular = QVector3D(specular,specular,specular);
+		lightInfo.specular = specular*solarTint;
 		lightInfo.torchDiffuse = QVector3D(torchDiff,torchDiff,torchDiff);
 	}
 }
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -975,8 +975,8 @@ void S3DRenderer::calculateLighting()
 	static LandscapeMgr *lmgr=GETSTELMODULE(LandscapeMgr);
 	QVector3D solarTint=lmgr->getLandscapeTint().toQVector();
 	//ambient tries to neutralize sunrise reddening a bit to model collecting light more of the blue sky. Directional and specular are tinted with possibly reddened sun.
-	lightInfo.ambient = ambientBrightness*(QVector3D(0.25,0.25,0.25)+0.75*solarTint);
-	lightInfo.directional = directionalBrightness*solarTint;
+	lightInfo.ambient = ambientBrightness*(QVector3D(0.4,0.4,0.4)+0.6*solarTint);
+	lightInfo.directional = directionalBrightness*QVector3D(powf(solarTint[0],2.5f), powf(solarTint[1], 2.5f), powf(solarTint[2], 2.5f));
 	lightInfo.emissive = QVector3D(emissiveFactor,emissiveFactor,emissiveFactor);
 	lightInfo.specular = specular*solarTint;
 	lightInfo.torchDiffuse = QVector3D(torchDiff,torchDiff,torchDiff);
--- a/plugins/Scenery3d/src/S3DRenderer.cpp
+++ b/plugins/Scenery3d/src/S3DRenderer.cpp
@@ -969,31 +969,17 @@ void S3DRenderer::calculateLighting()
 	//specular factor is calculated from other values for now
 	float specular = std::min(ambientBrightness*directionalBrightness*5.0f,1.0f);
 
-	//if the night vision mode is on, use red-tinted lighting
-	bool red=StelApp::getInstance().getVisionModeNight();
-
 	float torchDiff = shaderParameters.torchLight ? torchBrightness : 0.0f;
 	lightInfo.torchAttenuation = 1.0f / (torchRange * torchRange);
 
-	if(red)
-	{
-		lightInfo.ambient = QVector3D(ambientBrightness,0, 0);
-		lightInfo.directional = QVector3D(directionalBrightness,0,0);
-		lightInfo.emissive = QVector3D(emissiveFactor,0,0);
-		lightInfo.specular = QVector3D(specular,0,0);
-		lightInfo.torchDiffuse = QVector3D(torchDiff,0,0);
-	}
-	else
-	{
-		static LandscapeMgr *lmgr=GETSTELMODULE(LandscapeMgr);
-		QVector3D solarTint=lmgr->getLandscapeTint().toQVector();
-		//for now, ambient is only whitish. Directional and specular are tinted with possibly reddened sun.
-		lightInfo.ambient = QVector3D(ambientBrightness,ambientBrightness, ambientBrightness);
-		lightInfo.directional = directionalBrightness*solarTint;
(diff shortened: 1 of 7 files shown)


alex-w:
The solution is not complete yet...

stellarium-018

Atque:
I can't currently test; how does it work with solar eclipses?

10110111:
The coloration should be a mixture of the illuminance by the sky and by the Sun. This would make the tint change with the solar altitude non-monotonically. Moreover, the amount of reddening would ideally also depend on the angle between the Sun direction and the normal to the surface being lit.

gzotti:
> I can't currently test; how does it work with an eclipsed Sun? I guess we wouldn't want a red-tinted landscape even with a low eclipsed Sun.

The tint is just taken from the solar color. During eclipse, the landscape is first and foremost darkened. I assume a partially eclipsed sun will still be reddened enough to tint the landscape. However, such tests can be done when we are happy about the commonday situation.

gzotti:
Stupid: tint changes only when sun is in frame... Must do that elsewhere.

Atque:
I think the second of the options is the best.
https://github.com/Stellarium/stellarium/commit/4df73d27ce99091bb4e659fa6c9318ebeb8e7a7e#diff-147261ff5c272049b26356d320341d2f4045a49692259c91ce08ee058f389682R2869-R2873

stellarium-296

The one currently in use in this PR is too weak, barely altering the color of the scene even with the sun at the horizon:
stellarium-298

gzotti:
For some reason the tint in Scenery3D appears much weaker than the tint in regular landscapes. I have not found why so far. Of course, Scenery3D has ambient/diffuse/specular components, and ambient is still white (neutral gray) to simulate the all-sky contribution. 

For landscapes, one of the two last solutions looks best to me. I don't want to become too kitschy.

Atque:
You're right, I hadn't even noticed. Indeed one of the two latter looks best.

I suppose we can't simply split them, so that 3D Sceneries use one tint, and 2D landscapes another?

gzotti:
Of course some better solution is needed, therefore this is still a draft.  Now you see where endless late hours of finetuning go... Where are my weekends?

gzotti:
I think behaviour looks OK now. Maybe the color hue itself needs some more work.

alex-w:
Hmm... I see no effect in comparison with version 24.2

Atque:
> Hmm... I see no effect in comparison with version 24.2

The difference is quite subtle compared to the early stages of the PR. Do you see any difference using a snowy landscape panorama, such as Hurricane?

gzotti:
It's just a slight change, as I don't want to become too kitschy. Use e.g. any grassy ground, and see reddening with the sun higher than -3°, max. on the horizon and tint fading away as the sun climbs higher. With Scenery3D, I now had also to tint the ambient term slightly. If you really think it is not strong enough, I can step up one level.

10110111:
Wouldn't it be better to query Atmosphere for illuminance (including chromaticity)?

alex-w (approved):
Thanks!

I see the effect now :)

gzotti:
>Wouldn't it be better to query Atmosphere for illuminance (including chromaticity)? Maybe even split between the direct solar component and the diffuse sky component.

In the long run, sure. This is just a quick first solution. In Scenery3D, someone with enough time could think about global illumination, although we never initially aimed that high. (Earth curvature correction would be more important in its primary field of use.)  

The color of the solar "halo" [yes, it's not a halo in the meteorological sense. It's just a variable name] and reddening of sun and moon should take atmospheric extinction factor into account. (If SMS is used, even spectral differences are possible, we had talked about Mars and many other incredible possibilities once somebody could explain and document the generation parameters of SMS and write a guideline how to create other atmosphere settings, including several real-world examples.) 

For now, this provides a first improvement. I am open for better ones, now or next year.

gzotti:
These large scenes unfortunately cannot be made publicly available, my largest just shown in our paper. Once they extend to 80km or so, a difference is notable, and it depends on the application whether it is to be regarded relevant. 
Scenes are usually modelled from UTM-based DEM grids. They are unfortunately flat by design. The importer would probably have to convert to planetocentric Cartesian coordinates (double prec.), and during use, find a local vertical, and always rotate the scene so that this local vertical is "up", and true north is "north". However, this needs double precision math which GPUs still don't like very much. But we digress here, it's off-topic to this small-cosmetic PR.

gzotti:
> Wouldn't it be better to query Atmosphere for illuminance (including chromaticity)? Maybe even split between the direct solar component and the diffuse sky component.

@10110111 where could I ask the atmosphere for chromaticity? If this is not available now, I think this PR should be OK as-is, but as soon as atmosphere can deliver such data, we can further improve here.

10110111:
> where could I ask the atmosphere for chromaticity?

There's no such API currently.

> If this is not available now, I think this PR should be OK as-is

Well it doesn't look realistic to me. The colors look a bit weird, particularly for Guereins the grass becomes too red (I'd expect more of an orange tint than red), and this redness continues into the twilight—why?

gzotti:
It just looked bad to just cut off at solar altitude 0 as originally planned. So, now I fade it out 0...-3°. Tell me your favourite solar altitude (or transition altitude) for the cutoff. 
The red in early twilight is best arguable when there is some haze or clouds in the sky (which yes, we don't show). Maybe I was also biased, using Preetham sky during tests. 

Can you quote your favourite RGB values (1, G, B) for max discoloration?

gzotti:
Just tried with SMS. Looks even better. Horizon is reddish (sun side)/pink (opposite) with the sun just below horizon. Anywhere in-between the color that is applied on the grass.

10110111:
> It just looked bad to just cut off at solar altitude 0 as originally planned.

The illuminance due to the setting sun compared to the illuminance by the diffuse sky glow decreases with decreasing solar elevation. Especially so for the horizontal surfaces. So the tinting should have a maximum somewhere between 0° and the upper cutoff, not at 0°.

> The red in early twilight is best arguable when there is some haze or clouds in the sky (which yes, we don't show).

Or the afterglow, but we don't simulate it either (though in real life it has a profound effect on the twilight color of mountains).

> Can you quote your favourite RGB values (1, G, B) for max discoloration?

I don't necessarily think the issue is in the RGB, rather in the way the tinting is applied (see above). In the absence of afterglow or red clouds we shouldn't have too much redness after the sun has set.

10110111:
Note though that in 3D scenery the reddening of the sunlight should be maximum at 0° (or rather -0.25°), since there, IIUC, the sunlight is separate from ambient skylight.

gzotti:
There is no upper cutoff. Reddening goes with airmass. The lower cutoff is currently set arbitrarily to -3°, and looks pretty good IMHO. Definitely better than 0. 

Scenery3D: You are cordially invited to improve my attempts. Yes, the ambient term should ideally receive its information from the atmosphere. This is not possible yet. Therefore I have to mix in some of the reddened light, or else the 3D surfaces are not perceivably colored. As soon as the atmosphere can be asked for global chromaticity, this can of course be improved. So, we can have this effect as simple painting effect now, or this PR has to wait until it can immediately be presented physically correct in X years, or have the simple solution now AND the better solution whenever atmosphere API is available. 

For a test, setting transition to -1° only. The message is "there is reddening around sunrise". Too much then depends on real-world conditions which we fail to simulate in any case. 

@10110111 , @alex-w, @Atque your votes please.

alex-w:
This is acceptable solution for me at the moment

10110111:
> or have the simple solution now AND the better solution whenever atmosphere API is available.

I'm worried that when the physically-correct solution appears, people will complain that the effect is not as noticeable as it used to be. And then, if we revert to this ad hoc implementation in response, this will mean that we prefer an exaggerated drawing to a simulation.

And from my experiments, the effect on horizontal surfaces is indeed very subtle.

> this PR has to wait until it can immediately be presented physically correct in X years

The API to query illuminance is not so hard to create, it can be done in a few weeks.

Atque:
I also think this is good enough. Not everything needs to be physically correct to be usable.

gzotti:
I added a bit more green to the color, does not change too much. The effect is noticeable during 20 minutes of sun near horizon, but not too strong. Good enough now, but as everywhere, there will be room for improvement based on atmosphere tuning later. A cloudy sky (whenever this may come) will certainly also have to influence colours.

PR #3898, opened 2024-09-17
Cache UTC offset

gzotti:
Finding UTC offset is surprisingly costly but is executed per-frame, shown in VTune. Caching the result per-minute saves at least part of it. 

This needs some testing. 

### Description

Fixes # (issue)

2024-09-16d getCommonInfoString_UTCoffset

With this caching: 
2024-09-16e cachedUTCoffset

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/StelCore.cpp
+++ b/src/core/StelCore.cpp
@@ -1351,8 +1351,6 @@ void StelCore::moveObserverTo(const StelLocation& target, double duration, doubl
 	emit locationChanged(getCurrentLocation());
 }
 
-QCache<size_t, qint64> StelCore::utcOffsetCache;
-
 double StelCore::getUTCOffset(const double JD) const
 {
 	// This method takes a significant amount of time. Try to cache a few entries.
--- a/src/core/StelCore.cpp
+++ b/src/core/StelCore.cpp
@@ -1413,6 +1413,7 @@ double StelCore::getUTCOffset(const double JD) const
 	// Complete the cache object and hash it.
 	u.loc= & const_cast<StelLocation &>(loc);
 	u.tz = & const_cast<QTimeZone&>(tz);
+	// TODO FIXME ERROR: We must hash the contents, not the pointer!
 	size_t dateLocHash=qHash(&u);
 
 	qint64 shiftInSeconds;
--- a/src/core/StelCore.cpp
+++ b/src/core/StelCore.cpp
@@ -1351,40 +1351,55 @@ void StelCore::moveObserverTo(const StelLocation& target, double duration, doubl
 	emit locationChanged(getCurrentLocation());
 }
 
+QCache<size_t, qint64> StelCore::utcOffsetCache;
+
 double StelCore::getUTCOffset(const double JD) const
 {
-	int year, month, day, hour, minute, second;
-	StelUtils::getDateTimeFromJulianDay(JD, &year, &month, &day, &hour, &minute, &second);
+	// This method takes a significant amount of time. Try to cache a few entries.
+	// All these elements can influence UTCOffset, so we must include them in the cache.
+	// Presumably, offset does not change during a minute, so we can exclude seconds here.
+	struct UTCOffsetHashData {
+		int year;
+		int month;
+		int day;
+		int hour;
+		int minute;
+		//int second;
+		StelLocation *loc;
+		QTimeZone *tz;
+	} u;
+	int second;
+	StelUtils::getDateTimeFromJulianDay(JD, &u.year, &u.month, &u.day, &u.hour, &u.minute, &second);
 	// as analogous to second statement in getJDFromDate, nkerr
-	if ( year <= 0 )
+	if ( u.year <= 0 )
 	{
-		year = year - 1;
+		u.year = u.year - 1;
 	}
 	//getTime/DateFromJulianDay returns UTC time, not local time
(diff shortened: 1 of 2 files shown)


src/core/StelCore.hpp
```diff
@@ -315,8 +316,13 @@ class StelCore : public QObject
 	const StelLocation& getCurrentLocation() const;
 	//! Get the UTC offset on the current location (in hours)
 	//! N.B. This is a rather costly operation. Re-use where possible!
+	//! In addition, this is now cached with 1-minute granularity.
 	double getUTCOffset(const double JD) const;
+private:
+	// Cache is over qHash(&UTCOffsetHashData).
+	static QCache<size_t, qint64> utcOffsetCache;
```
10110111, line 323: Why static?

src/core/StelCore.cpp
```diff
@@ -1395,83 +1410,98 @@ double StelCore::getUTCOffset(const double JD) const
 		qWarning() << "Invalid timezone: " << tzName;
 	}
 
-	qint64 shiftInSeconds = 0;
-	if (tzName=="system_default" || (loc.planetName=="Earth" && !tzValid && !QString("LMST LTST").contains(tzName)))
+	// Complete the cache object and hash it.
+	u.loc= & const_cast<StelLocation &>(loc);
```
10110111, line 1414: This is some ugliness here. Why do you keep the object by `const` reference and then violate this constness?

src/core/StelCore.hpp
gzotti: If I make this a per-object variable, it breaks many const-qualified functions. If there will ever be more than one core, they would just share the results. Or do you see problems with this?

src/core/StelCore.cpp
```diff
@@ -1351,40 +1351,55 @@ void StelCore::moveObserverTo(const StelLocation& target, double duration, doubl
...
 
+QCache<size_t, qint64> StelCore::utcOffsetCache;
+
 double StelCore::getUTCOffset(const double JD) const
 {
-	int year, month, day, hour, minute, second;
-	StelUtils::getDateTimeFromJulianDay(JD, &year, &month, &day, &hour, &minute, &second);
+	// This method takes a significant amount of time. Try to cache a few entries.
+	// All these elements can influence UTCOffset, so we must include them in the cache.
+	// Presumably, offset does not change during a minute, so we can exclude seconds here.
```
10110111, line 1360: Looks like an unsafe assumption to me. What is the cost of keeping the seconds?

src/core/StelCore.cpp
gzotti: This is what various warnings made me do.

10110111:
What is the performance benefit from this? Is it even measurable in any way except by using a profiler?

src/core/StelCore.cpp
gzotti: updates every second, not every minute.

gzotti:
> What is the performance benefit from this? Is it even measurable in any way except by using a profiler?

I just wondered what takes so long just displaying InfoText. The commonInfoString was a surprisingly costly part of it, and here this seemingly trivial UTC lookup appears to be far from trivial.

src/core/StelCore.cpp
10110111: Well, this is the wrong way to read the warnings then :P They are supposed to make you improve the code, not equip it with kludges.

src/core/StelCore.hpp
10110111: > If I make this a per-object variable, it breaks many const-qualified functions.

This is exactly why `mutable` keyword was introduced: one of the uses for it is implementation of transparent caches.

> If there will ever be more than one core, they would just share the results.

Hmm, this doesn't seem like a likely situation, but in this way this might make sense, though there's some strangeness in the way you implement the cache that I can't tell whether it does.

src/core/StelCore.cpp
```diff
@@ -1395,83 +1410,98 @@ double StelCore::getUTCOffset(const double JD) const
 		qWarning() << "Invalid timezone: " << tzName;
 	}
 
-	qint64 shiftInSeconds = 0;
-	if (tzName=="system_default" || (loc.planetName=="Earth" && !tzValid && !QString("LMST LTST").contains(tzName)))
+	// Complete the cache object and hash it.
+	u.loc= & const_cast<StelLocation &>(loc);
+	u.tz = & const_cast<QTimeZone&>(tz);
+	size_t dateLocHash=qHash(&u);
```
10110111, line 1416: Are you hashing a pointer to a local variable?..

gzotti: Did I misunderstand the qHash() docs? Does this only hash the pointer, not its contents? 
OK, that's not what I wanted... :-o
Do I need to write my own qHash function for the struct?

10110111: Yes, you need a custom hash function for every custom data type. [This](https://stackoverflow.com/q/2590677/673852) may be of help.

src/core/StelCore.hpp
gzotti: OK, I forgot about mutable.

PR #3899, opened 2024-09-17
Fix linking with <execution> which may require libtbb

### LINKED ISSUE #3896: After threaded Solar System PR merge linking on Linux g++-11 fails
I get the following error on Ubuntu 20.04, g++ 11.4.0-2ubuntu1~20.04:
```
/usr/include/tbb/task_arena.h:449: undefined reference to `tbb::interface7::internal::task_arena_base::internal_current_slot
```

If I comment out `#include <execution>`, linking succeeds. Also note that I get this:
```
-- Checking that std::execution::par is supported by the C++ library - NO
```
This is despite my compiler supporting this functionality. It just needs to link against `libtbb` (via `-ltbb` linker option) for linking to succeed.

10110111:
This should fix #3896.


gzotti (approved):
Still compiles and runs on W11. If it fixes Linux problems, sure, approved!

alex-w:
This is a fix for GCC

PR #3921, opened 2024-09-28
Fix code scanning alert no. 20: Unsafe jQuery plugin

alex-w:
Fixes [https://github.com/Stellarium/stellarium/security/code-scanning/20](https://github.com/Stellarium/stellarium/security/code-scanning/20)

To fix the problem, we need to ensure that the `options.of` input is always treated as a CSS selector and not as HTML. This can be achieved by using the `jQuery.find` method, which interprets the input strictly as a CSS selector. This change will prevent any potential XSS vulnerabilities arising from interpreting the input as HTML.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/plugins/RemoteControl/webroot/js/jquery-ui.js
+++ b/plugins/RemoteControl/webroot/js/jquery-ui.js
@@ -907,9 +907,7 @@ $.fn.position = function( options ) {
 	var atOffset, targetWidth, targetHeight, targetOffset, basePosition, dimensions,
 
 		// Make sure string options are treated as CSS selectors
-		target = typeof options.of === "string" ?
-			$( document ).find( options.of ) :
-			$( options.of ),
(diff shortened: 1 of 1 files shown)


gzotti:
We can update our own imported modules 1/year or so, but try not to diverge from upstream.

gzotti:
I am still not clear on this. Is there any tool to test whether this line of code is even executed when RC is active?

alex-w:
> I am still not clear on this. Is there any tool to test whether this line of code is even executed when RC is active?

I am not very familiar with javascript and I don’t know special tools for testing JS code.

> EDIT: We use jquery-ui 1.14.0. The current 1.14.1 still includes the flagged code. Should we just report that to the jquery-ui project?

Probably yes

PR #3987, opened 2024-11-25
Immediate-store all of view dialog

gzotti:
### Description

This continues a series of changes which aims at optionally avoiding the requirement to press "save settings".
All features in the View settings dialog and Configuration should be immediate-store now.

The Language and font settings are probably the only settings that crucially needs to be protected by an explicit store so they are not messed up on a panic exit. All other settings are as easy to reset as they were set. 

In the end of this branch, the "save settings" button will be disabled when the corresponding "immediate storage" option in the config dialog has been activated. 

It is to be discussed whether "save languages" and "save fonts" buttons should only be available in this new mode. Else they would cause no harm. 

Plugins usually have their own "save settings" logic. They may be modified later. 

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:


src/core/StelSkyDrawer.cpp
```diff
@@ -924,3 +926,177 @@ double StelSkyDrawer::getWorldAdaptationLuminance() const
...
+{
+	refraction.setPressure(static_cast<float>(mbar));
+	StelApp::immediateSave("landscape/pressure_mbar", mbar);
+	emit atmospherePressureChanged(static_cast<double>(refraction.getPressure()));
+}
+
+// These are to find out the best sky parameters. Program feature for debugging/expert mode.
+// These will be connected from AtmosphereDialog and forward the settings to SkyLight class.
+
+void StelSkyDrawer::setFlagDrawSunAfterAtmosphere(bool val){
```
alex-w, line 1073: Formatting issues

src/core/modules/MilkyWay.cpp
```diff
@@ -83,6 +83,7 @@ void MilkyWay::init()
 	setFlagShow(conf->value("astro/flag_milky_way").toBool());
 	setIntensity(conf->value("astro/milky_way_intensity",1.).toDouble());
 	setSaturation(conf->value("astro/milky_way_saturation", 1.).toDouble());
+	setColor(Vec3f(conf->value("astro/milky_way_color", "1.0,1.0,1.0").toString()));
```
alex-w, line 86: Color for texture?

gzotti: Yes. It was always available. For the old texture it was even necessary to deviate from white to counteract the strong blue hue. While the current strong recommendation is to leave it white, the color can be used for personal preference or even artistic reasons. I am just unsure whether to even enable this setting (and similar for Zod. light, and, while I am at it, even the background color else only settable via TUI) in the GUI.

alex-w (changes requested):
Looks good, but 2 obvious fixes in GUI are needed:

gzotti:
OK for the SC panel. For the other, it has become a bit convoluted over the last years, historically grown. I understand the top-left options (IIRC the same since ~2010) were options most likely used in an actual planetarium setup (hence the group title): Distortion, disc viewport, gravity labels. For explanations in a planetarium (and actually also regular screens), the "select single constellations" was there as well, but it better fits to the SC panel. The Auto-zoom-out (yes, still open issue...) is also an option apparently (I never used it, so missed its relevance, sorry) invented for and most critically used in a planetarium. The rest of the tab looks like a collection of "I didn't know where to place it, so I added it here". 

Decimal degrees and Az. from south could both move to the Extras/Info details tab. (Just have two columns in the top area.)
Nutation/aberration could likewise move there, or move to the view settings/Sky, and Topocentric correction could move to the view settings/SSO (Or also Extras/Info tab). Shall I try how this looks?

Another thing is the font selection which is still hidden by default and needs a manual edit to config.ini. Shall we show it to all at beginners level?

alex-w:
> Decimal degrees and Az. from south could both move to the Extras/Info details tab. (Just have two columns in the top area.)

Both options touching not only information section, but yes, them can be moved to the Extras/Info tab.

> Nutation/aberration could likewise move there, or move to the view settings/Sky, and Topocentric correction could move to the view settings/SSO (Or also Extras/Info tab).

Well, View/Sky and View/SSO are bloated already, but you’re right - these options can be moved there.

> Shall I try how this looks?

Yep.

> Another thing is the font selection which is still hidden by default and needs a manual edit to config.ini. Shall we show it to all at beginners level?

I fear it may be very painful for beginners…

gzotti:
> > Decimal degrees and Az. from south could both move to the Extras/Info details tab. (Just have two columns in the top area.)
> 
> Both options touching not only information section, but yes, them can be moved to the Extras/Info tab.
> 
> > Nutation/aberration could likewise move there, or move to the view settings/Sky, and Topocentric correction could move to the view settings/SSO (Or also Extras/Info tab).
> 
> Well, View/Sky and View/SSO are bloated already, but you’re right - these options can be moved there.

Indeed Sky and SSO were already large. I left them in the Config dialog, but moved to Info as well. These 3 options involve position refinements, but also of course the provided information. In light of current work on the star catalog, topocentric flag could also switch parallax observance for stars, or this will need another flag later.

I edited the .ui XML directly to also clean up the meanwhile 18 (or so) rows in the Info dialog layout. Maybe the currently only used colums 1 and 3 should also be edited back to 0 and 1?

The mag/squareArcMin and "use symbol" settings are AFAICS used only for DSO. Might be moved there, but the current shortened strings are also OK. (Can they be combined into 1 line?)

> > Another thing is the font selection which is still hidden by default and needs a manual edit to config.ini. Shall we show it to all at beginners level?
> 
> I fear it may be very painful for beginners…

OK, I left the default-hidden font panel as it was.

alex-w:
This is OK:
stellarium-010

> > Well, View/Sky and View/SSO are bloated already, but you’re right - these options can be moved there.
> 
> Indeed Sky and SSO were already large. I left them in the Config dialog, but moved to Info as well. These 3 options involve position refinements, but also of course the provided information. In light of current work on the star catalog, topocentric flag could also switch parallax observance for stars, or this will need another flag later.

This is not OK:
stellarium-011

Topocentric coordinates, nutation and aberration touches not only information box - they touches rendering the sky, so, Extras/Info is bad place for all of them. It may be accepatble for moving to View/SSO tab, but definitely to to Extras/Info.

Topocentric coordinates, nutation and aberration should be moved back to Tools tab, IMHO.

> The mag/squareArcMin and "use symbol" settings are AFAICS used only for DSO. Might be moved there, but the current shortened strings are also OK. (Can they be combined into 1 line?)

The current view for these lines are OK for me.

gzotti:
OK, next try.

alex-w (approved):

PR #3982, opened 2024-11-20
Fix proper motion equation in Star1/2 and Pulsar

henrysky:
### Description
This PR fixes the issue of reading proper motion for the main catalog and issue of errorous calculation of RA/DEC with proper motion (equation is wrong).

My understanding of the past issue regarding proper motion is that the pmra/pmdec values in the catalog can not be decoded properly(?). But it seems to me the catalog proper motion are simply pmra and pmdec in the unit of 0.1 mas/yr. I've picked a few HIP stars and seem to check out.

Now this PR simply read the pmra/pmdec from the catalog and use it to calculate RA/DEC at any date. IMO this is a good temporary solution until new catalog arrives.

I kept the extra proper motion data in the repository but they are not used to display nor calculation.

Fixes #85 (issue), #1061 (partially)

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [ ] Housekeeping

### How Has This Been Tested?

Nutation, Abberation and atmosphere are turned off!!

Compared to SIMBAD on high proper motion stars Proxima Centauri and Barnard's star.

On the issue of Proxima Centauri from #1061, here are the comparison

On 1 Jan 2000
|Source (1 Jan 2000)|RA|DEC|Diff|
|---|---|---|---|
|Gaia DR3|14h29m42.94|-62d40m46.16| 0.00s / 0.00s|
|Stellarium 24.3 & PR|14h29m47.59|-62d40m52.50| 4.64s / 6.34s|

so Proxima Centauri is off by RA/DEC +4.64s/+6.34s to begin with between Stellarium catalog and Gaia DR3 regardless if proper motion is done correctly.

on 22 Apr 2020
|Source (22 Apr 2020)|RA|DEC|Diff|
|---|---|---|---|
|Gaia DR3|14h29m31.74|-62d40m30.48| 0.00s / 0.00s|
|Stellarium This PR (corrected with offset manually)|14h29m31.68|-62d40m30.66|-0.06 / 0.16s|
|Stellarium This PR|14h29m36.32|-62d40m37.00|5.84s / 6.52s|
|Stellarium 24.3 (corrected with offset manually)|14h29m32.33|-62d40m17.76|1.85s / -12.72s|
|Stellarium 24.3|14h29m36.97|-62d40m24.10|5.23s / -6.38s|

[...]

Issue #3832, opened 2024-07-30
Objects not selectable at higher zoom level


StarFlea:
Hello folks!

To select an object (e.g. star, asteroid, ...) in Stellarium, you simply click on it. But at a higher zoom level, the same object is not selectable anymore.

Example: Pluto - At this level there is no problem. Pluto is selected
![Pluto_1](https://github.com/user-attachments/assets/3337cf16-d043-4193-a10c-1a06c6ec1df5)

But if you zooming in, Pluto is not selected by mouseclick on it but the star on the right. You can click furthermore, but Pluto is unreachable...
![Pluto_2](https://github.com/user-attachments/assets/15ded254-9767-464b-8104-2b694bfb84ec)

The same problem exists with other objects, e.g. stars. At a higher zoom-level you have not to click directly to the object, but a little bit away to select it.

Version 24.2.197-623f996 [HEAD]
Based on Qt 5.12.6
Windows : 10

Best regards!
StarFlea

gzotti:
Duplicate of https://github.com/Stellarium/stellarium/issues/3733

Issue #3733, opened 2024-05-04
Click to select a star at high zoom gets the wrong star


GrandfatherJ:
Issue noticed at high zoom with FOV < 0.4 and all star catalogues loaded.

### Expected Behaviour
Click on a star to select it and view its details.

### Actual Behaviour
A nearby star is selected instead of the star clicked on.

### Steps to reproduce
1/ With star catalogues loaded so that stars to 16th mag centre on Mimosa in the Southern Cross and zoom so that the FOV is 0.2 degrees or less. To the left of Mimosa is a 11.45 mag white star. 
2/ Click on the star. 
3/ Notice that the 15.5th mag star nearby is selected instead

The correct star can be selected by clicking on the opposite side of the correct star about the same distance as the incorrectly selected star is.

![image](https://github.com/Stellarium/stellarium/assets/64131943/88513c3a-db36-42b1-8f1f-8e534217aba4)

### System
* Stellarium version: stellarium-24.1-qt6-win64.exe
* Operating system: Windows 11 Home, 23H2
* Graphics Card: NVidia, RTX4070, 31.0.15.3210
* Screen type (if applicable): 2560x1600 Lenovo DisplayHDR

### Log file
[log.txt](https://github.com/Stellarium/stellarium/files/15206016/log.txt)

Atque:
I can reproduce this but only with aberration enabled.

GrandfatherJ:
It may be related to catalogue search order. In my playing around faint stars seem to be preferred. In the image below I clicked on the bright star below and slightly left of the selected star. The issue doesn't happen with any of the other bright stars in the screen shot.

![image](https://github.com/Stellarium/stellarium/assets/64131943/a5c9ae36-7417-4443-b003-815d84780b10)

Note that aberration is enabled by default. I guess with the default catalogue only this may not happen at all, or there are so few close bright stars it's not much of a problem. Complete pain when it does happen though!

gzotti:
Yes, the anti-aberration of the clicked point may need some improvement. However, even with A. disabled, sometimes some stars just don't like to be clicked, so there may even be more to fix. Combine this with the need to re-investigate the catalog algorithms altogether for slight misbehaviour.  For now you describe what to do: click a bit to the other side.

GeoffClements:
Just installed Stellarium to support my physics degree and came across this issue.

I'm running version 24.02 as a flatpak on a KDE Neon machine (Ubuntu 22.04 lts).

[...]

Issue #353, opened 2018-09-19
Binary Star : Mean Proper Motions and True Motions


alex-w:
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1271270

Source: https://sourceforge.net/p/stellarium/discussion/278769/thread/e5092e67/

One of this continuing issues that I find with Stellarium is that moving into the distant future or sitant past shows the proper motions and true of binary stars, like Alpha Centauri, through space.

[Other examples are; Acrux, Gamma Centauri, Mu Verorum, Delta Velorum, y Centauri [not letter Gamma] or Beta Muscae. Northern examples, Delta Cygni, Alpha Piscium, Epsilion Hydra, The worst example is 61 Cygni, which at cuurrent date has the stars already too far ap.art

Here, in case of Alpha Centauri, A and B stars have different proper motions, using the HIP data, and these are displayed as two different diverging stars, flying apart, which simply get wider and wider as separate stars. (All these stars above do too!)

Alpha Centauri is the principle example, as it is the nearest star with very high proper motions. Using Stellarium as a planetarium for the past or future sky shows very poor educational value if the stars are not behaving as they should.

Exampled, for Alpha Cen A (HIP 71683) have individual proper motions pmRA= −3679.25 and pmDec.=+473.67 mas per year and Alpha Centauri B (HIP 71681) as −3614.39 and +802.98 mas per year.

The mean of these motions should adopted as a first approximation, I.e. When combined as −3646.82 and +638.25 mas per year. Hence, the proper motions for both stars would be better expressed as this value.

Usually, there are inherent problems with the companions proper motions, as their individual deviations in the directions they are travelling are not constant due to the binary system's orbital motion. Nominally, the true measure of the proper motion should be when both stars are travelling together (a line perpendicular to position angle of the direction the system is moving. The quoted proper motions here are not under this condition. [There are serious doubts on the given proper motion in declination for the B-star, and it could be easily argued that the A star motion should apply to both. Reasoning to explain this is not easy to say in a few words - at least in this expression of a problem.]

There are more serious issues than proper motions here, as the differing radial velocities of both stars. This changes over the 79.9 year orbital period, whose velocity varies by the orbital eccentricity / true distance the stars are apart. It is the mean or average radial velocity that is important here, being 25.1+/-0.3 km per second.

[...]

PR #3992, opened 2024-11-30
New star catalog with Gaia DR3

### LINKED ISSUE #18: Excess star near 119 Tau
Original report: https://sourceforge.net/p/stellarium/discussion/278769/thread/42b17b79/

> Hello, I am still using version 0.12.5
> 
> Last night I took an image of the area around 119 Tau and noticed there was a mag 9.7 star on Stellarium which was not on the photo. Simbad confirms there is no star in that position. So is it possible to delete a false star to avoid confusion? I include a screenshot.
>
>tkmo3m

I can confirm the issue in the star catalog.

### LINKED ISSUE #85: Stellar Proper motions are not right
### Bug Description
I noticed in Stellarium 14.1beta (and back to Version 12) there is a problem with the proper motions.

If you look at Alpha Centauri, the future motion is moving northwards at a PA of about 345 degrees, however that is not possible. If the HIP2 parallax is -3679.25 and +473.67 mas/year, the position angle should be about 277.3 degrees, calculated as tan-1(-3679.25/473.67) in Quadrant 4. I.e. It should be heading just little north of Beta Centauri, not northwards!.

Yet if you select the star HIP 57367 (11.5 magnitude) in Musca, which has an equally large proper motion but is in the opposite direction. I.e. +2663.91 and -347.38 mas/ yr, the position angle is +97.4 degrees.

The motion in Stellarium is seemingly wrong for Alpha Centauri but appears roughly correct for HIP 57367.

I also checked , being roughly correct.

Arcturus, pm. -1093.99, -2000.16 mas/yr, moving in PA 208.7 deg.
Sirius , pm. -546.01, -1223.07, PA 204.8 deg.
Barnard's Star -798.71, 10337.2 mas/yr, PA 353.6 deg.

1) Please check the calculations when adding proper motions.
2) Could someone point out the C++ file that does these proper motions calculation?
3) Are the proper motion calculated independently of the precession, as some stars move as arcs against the background stars in Stellarium rather than straight lines? (I.e. Follow the motion of HIP 57367.)

Alpha Centauri is a critical star because of its brightness and being the closest star, and educationally is important to many who us Stellarium. It is critical that something is done to fix these apparent errors.

Kind regards and appreciation to its solution,
And hope this is useful.

Andrew

I use Mac OS X 10.7.5, ATI Radeon HD 4670 256MB graphics

Note 1 : it is possible there is something wrong in solving the tan ambiguity here, where the PA is being calculated incorrectly.

Note 2 : As previously noted, SIMBAD gives the following proper motions;

Alpha Centauri : -3608, 686 mas/yr [PA 280.8 deg]
http://simbad.u-strasbg.fr/simbad/sim-id?Ident=alpha+cen&NbIdent=1&Radius=2&Radius.unit=arcmin&submit=submit+id

Alpha Centauri A : -3679.85, 473.67 mas/yr [PA 277.3 deg]
http://simbad.u-strasbg.fr/simbad/sim-id?Ident=%403396054&Name=*%20alf%20Cen%20A&submit=submit

Alpha Centauri B : -3614.39, 802.98 mas/yr [PA 282.5 deg]
http://simbad.u-strasbg.fr/simbad/sim-id?Ident=%403396297&Name=*%20alf%20Cen%20B&submit=submit

As the stars are labelled individually, the two components appear to diverge and are moving wider apart, rather than towards their common mutual shared motion.

The HIP2 file should be modified to either Alpha Centauri's mean motion of -3608, 686 mas/yr [PA 280.8 deg], or the mean motion of AB -3648.62, 638.33 mas.yr (HIP2) [PA 279.9 deg]. This number should be assigned to both the A and B stars in Stellarium's HIP2 star file.

### Original report with detailed dicussion
https://bugs.launchpad.net/stellarium/+bug/1527856

### LINKED ISSUE #221: Arcturus-data
Greetings,
I think in Stellarium is not correct Arcturus star description, at first Arcturus isn’t double star but red giant star, and I think visual magnitude isn’t correct too because Stellarium says magnitude −0,15 but I found magnitude − 0,05.
arcturus

source of my information: https://en.wikipedia.org/wiki/Arcturus

Thank you for answer.

### LINKED ISSUE #348: Arcturus disappears when zoomed in on in the far future
_Original report by qam1:_ https://bugs.launchpad.net/stellarium/+bug/648214

Way, way in the future the star Arcturus disappears when zoomed in on (FOV <16). It starts happening after the year 37000 but only on somedays and then further in the future it happens all the time, One date in particular, March 24th 60011 17:38UTC when there's a Mars occultation of Arcturus (Yes Really) it is ruined because Arcturus disappears and Mars just occults a blank space. Alpha Centauri/Rigel Kent does the same thing but so far every other star I looked at seems OK.

### LINKED ISSUE #362: cannot select θ1 Ori B which is the B star in the orion trapezium
_Original report by Richard Bruckner:_ https://bugs.launchpad.net/stellarium/+bug/1195285

cannot select ?1 Ori B which is the B star in the orion trapezium

the other 3 stars in the trapezium are easily selected by clicking on them at any zoom level

win 7 64 bit 0.12.2 dev7

### LINKED ISSUE #363: wrong position angle & dist for 61 Cygni binary
_Original report by Robert Toop:_ https://bugs.launchpad.net/stellarium/+bug/1299412

The position angle for 61 Cyg A to B appears to be 90 degrees; should be 152 degrees.
The separation appears to be 48 arcsec; should be 31.5 arcsec.
Good news is that cranking back the clock in Stellarium 0.12.4 shows proper motion that matches 1916 & 1948 photos in "Burnham's Celestial Handbook."

I love this program, but until now I assumed ALL star positions were based on Hipparchos satellite. Is there a problem with close doubles in general?

### LINKED ISSUE #364: Color index for some stars is wrong
_Original report by Nick Fedoseev:_ https://bugs.launchpad.net/stellarium/+bug/1405467

Color index for some stars is wrong, they usually look too red.
Actual index of the selected star at the picture is about ~1.5, while 3.50 is shown.

Perhaps, this can be useful: such extreme index value corresponds to integer value of 127 before conversion (maximum allowed for the field in the data structure)

### LINKED ISSUE #369: Star magnitudes for 46 Vir and 2 Com
_Original report by s. Mohammad Mozaffari:_ https://bugs.launchpad.net/stellarium/+bug/1733320

Please check the star magnitudes. I am occasionally confronted with these errors: 1) 46 Vir is assigned a magnitude of 8.5 (!), while it is of 5.99 in the BSC and a bit more than 6 in Hipparcos. 2) Also, check 2 Com.

The Stellarium is assumed to be widely used by historians of ancient and medieval astronomy for a readily and easily finding of astronomical data. So, such errors should be taken seriously.

### LINKED ISSUE #398: minor star catalog issue
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1644191

Original report: https://sourceforge.net/p/stellarium/discussion/278769/thread/d30cbd79/

Hi, there,
Star J2000.0 23h33m29.27s +89d11'38.5" does not exist or misplaced. And seems like it's that neighbor of star 23h24m4.85s + 89d11'24.1" . Please look into it.
I'm not sure it's in stars_3_1v0_3.cat or stars_4_1v0_1.cat.
Thanks!
my tweet 1 - https://twitter.com/kensenjiha/status/801262105995284480
my tweet 2, with DSS photo - https://twitter.com/kensenjiha/status/801270494276091908

### LINKED ISSUE #399: Two stars shown where there should only be one
_Original report:_ https://bugs.launchpad.net/stellarium/+bug/1655226

Original report by anonymous: https://sourceforge.net/p/stellarium/discussion/278769/thread/4b162039

Hi,
The star V400 Cep (HIP 1041, HD826) should only be one star, in one location, but, a lttile further over it is repeated. If you click on either, it says it's the same star, but the "second one" is in the wrong spot. (This is in the latest version, 0.15.1)
I noticed this when I was trying to add my own image for NGC40, (The bow-tie nebula, Caldwell2), and found this erroneous star where it didn't belong.
Can someone tell me how to delete this incorrect star? Seems it is embedded in the binary file, that can't be edited.
Thanks!
Also, I am happy to share my manually plate-solved NGC 40 image. (that took hours - longer than it took to get the image). It's not great, but still better than the empty space that was there, and I got the other stars to line up very well, doing it by trial-and-error (with much help from stellariumtextures).

### LINKED ISSUE #677: Stellarium shows an non-existing star
I use stellarium to star-hop my way around the night sky with a telescope and as such rely on its star maps being accurate. Yesterday I discovered that a star close to 24 Canum Venaticorum, which can be seen in stellarium, does not in fact exist! This is especially egregious since the supposed star in question is a relatively bright magnitude 10 star. 

I am attaching to figures with the issue: 
* stellarium_error_observation: A processed camera picture showing the star field around 24CNv as captured by me yesterday night 
* stellarium_error_app_screenshot: A screenshot from the application. Notice how the star which is selected is nowhere to be found in the actual night sky. Also, note that other stars do match up closely between the application and observation so I definitely looking at the right place in the sky!

I guess this might be an error in the star catalog rather than the application proper, but I did not know where else to report this bug.  

stellarium_error_app_screenshot
stellarium_error_observation

### LINKED ISSUE #777: This bright star doesn't really exist.
This imaginary star obscures M102. Its magnitude is 5.85.

star

### LINKED ISSUE #1016: Effects about globular clusters
image

image

This kind of effect of GCL in Stellarium is common. It does not match visible forms or photographes, and may also causes some misleading. The causes may come from star catalogs and the particularity of global cluster. I consider ripping stars in filed as one of feasible measures.

### LINKED ISSUE #1032: A cluster-like object may not exist?
I found a cluster-like object around HR6692 (HIP 88012) in Sgr when I was making a DSO hunting plan. However it doesnot get any DSO label. The cluster-like object looks like below. The star catalog for Stellarium is "Star catalogues, v2.0" (extended data).
image

The object looks like Jewel Box and NGC3293 in FOV = 0.29°.

As I removed "Star catalogues, v2.0", the same field looks like below. The same confuse may not exist in Standard Version of Stellarium.
image

I checked same filed in Sky Safari and did not found the same struct.

The size of the "cluster-like object" is 0°2'30''. I searched in SIMBAD, and got the result ranked as Mag V as below. 

image

http://simbad.u-strasbg.fr/simbad/sim-coo?Coord=17+58+39.04983636+-28+45+32.7174625&Radius=3&Radius.unit=arcmin&submit=submit+query

### LINKED ISSUE #1061: Wrong Coordinates for High Proper Motion Stars (eg. Proxima)
I measured the location of Proxima Centauri at 2020-04-22 12:57:47 UTC and found it at 14:29:31.87 -62:40:31.28 (J2000). Yet Stellarium predicted 14:29:36.97 -62d40’24.1’’ (J2000) at that time. This is a ~36 arcsec difference. I have checked nearby (low proper motion) stars' location and they perfectly agree with Stellarium's prediction.

I'm not familiar with how Stellarium calculates high proper motion stars' location, and I have limited ability so I could only report this single case as an example. Hope some expert would know why!

Cheers
stellarium

---LOG Head---
2020-04-23T01:31:24
Operating System: Darwin 18.2.0 x86_64
Compiled using Clang 9.0.0
Qt runtime version: 5.9.8
Qt compilation version: 5.9.8
Addressing mode: 64-bit
Model Name: MacBook Pro
Model Identifier: MacBookPro11,4
Processor Name: Intel Core i7
Processor Speed: 2.2 GHz
Number of Processors: 1
Memory: 16 GB
Chipset Model: Intel Iris Pro
VRAM (Dynamic, Max): 1536 MB
/Applications/Stellarium.app/Contents/MacOS/stellarium 
 -----------------------------------------------------
[ This is Stellarium 0.19.3 - https://stellarium.org/ ]
[ Copyright (C) 2000-2019 Fabien Chereau et al.       ]
 -----------------------------------------------------

### LINKED ISSUE #1189: Star catalog failure?
Line arranged stars are shown next to Antares. They are about 12-15 magnitude. I cannot see them in a photographic star atlas.

image

### LINKED ISSUE #1192: Mystery star near M51
### Expected Behaviour
Don't show a magnitude-6 star near M51 where none exists in a picture I took recently.

### Actual Behaviour
Stellarium shows an unnamed magnitude-6 star near M51 where none exists in a picture of the sky I took recently in an attempt to record M51, and where the Hipparcos catalog (via [VizieR](https://vizier.u-strasbg.fr/viz-bin/VizieR-4)) doesn't list any stars within 15 arcmin.  The unnamed star shows up already with only the default star catalogs installed (i.e., Configuration / Extras says "Get catalog 5 of 9").

My picture: mystery-star-picture

My picture shows stars down to about magnitude 10, so a magnitude-6 star should be very obvious.  The three stars HIP 66116 - HIP 65768 - HIP 66004 near the location of the mystery star have magnitudes near 7 and are clearly visible, yet the mystery star is not to be seen.

Screenshot of Stellarium showing roughly the same FOV and showing the star's properties: mystery-star-stellarium

Note in particular that the star is listed by Stellarium to have a noticeable proper motion.  If the star was really some part of M51 mislabeled as a star then you'd expect it to have no proper motion, M51 being way too far away to show a noticeable proper motion.

### Steps to reproduce
In Stellarium, go to coordinates indicated in screenshot.  Observe star.

### System
* Stellarium version: 0.20.2
* Operating system: Windows 10 with the latest updates installed.
* Graphics Card: NVidia Geforce 940MX
* Laptop screen.

### LINKED ISSUE #1541: beta hydris (HIP 2021) proper motion discrepancy with Hipparcos data
### Expected Behaviour

https://hipparcos-tools.cosmos.esa.int/cgi-bin/HIPcatalogueSearch.pl?hipId=2021 gives **2220.12**

### Actual Behaviour
Stellarium 0.20.4:  **2648**

### Steps to reproduce

1. find HIP 2021

### System
* Stellarium version: 0.20.4

### Logfile 

(External ref: http://fer3.com/arc/m2.aspx/Obscure-Nautical-Almanac-star-FrankReed-jan-2021-g49481)

### LINKED ISSUE #1659: Wrong data for TYC 4280-1203-1
A star TYC 4280-1203-1 in Stellarium has wrong data for photometry and proper motion (errors in NOMAD catalog).

Original report: https://astronomy.ru/forum/index.php/topic,190855.msg5351677/topicseen.html#msg5351677

Russian text:
>Да, редкостный глюк Стеллариума! Для этой звезды еще и неправильно указано собственное движение! Поставьте дату на 50, 100 или 200 лет вперед-назад и увидите, что эта "маленькая звёздочка" слева от HIP 115691 смещается относительно других звезд. Звезд с большим собственным движением в радиусе 10 минут вокруг HIP 115691 нет.
>На самом деле звезда на снимке asm - это TYC 4280-1203-1 с координатами R.A. = 23 26 25.31, Decl. = +61 06 31.8 на эпоху 2000.0 и собственным движением (-1.687 -1.660) миллисекунды в год, или 1" по обеим координатам за 600 лет. Величины у нее должны быть V=10.52 и B=10.60 (по каталогу Tycho), а Stellarium дает 11.30 и (B-V)=0.00.
>
>Ошибка, очевидно, в файле stars_4_1v0_2.cat. Чтобы отследить происхождение ошибки, надо написать разработчикам Стеллариума (из наших - Александру Вольфу). Если удастся понять, откуда взялись ошибочные данные по этой звезде, возможно, будут выявлены и другие аналогичные случаи. Тогда можно будет их устранить.
>
> Откуда ошибка, я уже понял: врет каталог NOMAD, который дает для этой звезды V=11.34 и пропёртый моцион (223.2 -336.7) миллисек в год. Что с этим делать - большой вопрос.

### LINKED ISSUE #1724: A nameless star near α Lyrae with wrong apparent magnitude data
### Expected Behaviour
This star at 18h36m57.59s +38d54m09.3s(J2000.0),near the α Lyrae.
I have searched in DSS,WISE and 2MASS data base.Found the apparent magnitude of this star is about +15.

(In WISE,the cyan ring show the star) 
图片

(Invisible in my photograph,the red ring show the place of the star.)
图片

### Actual Behaviour
In Stellarium,the apparent magnitude of this star is +8.10，not suitable with reality.
图片

### Steps to reproduce
Input 18h36m57.59s +38d54m09.3s(J2000.0) in Stellarium can find this star.

### System
* Stellarium version: <Stellarium 0.20.1>
* Operating system: <Windows 10 Professional 20H2>
* Graphics Card: <NVIDIA GeForce GTX1060 461.72>

### LINKED ISSUE #2786: False stars in database
### Expected Behaviour
m33-2

### Actual Behaviour
m33-stellarium

There is some unnamed star with 7.45 magnitude in Stellarium in the center of M33. But there is no such star in reality.

### System
* Stellarium version: 0.22.2

### LINKED ISSUE #3108: Including conversion of spherical plane into star's propermotion calculations

### LINKED ISSUE #3136: On big zoom mouse clicks are off
### Expected Behaviour
When I click on a star, it's selected

### Actual Behaviour
Some stars click fine, but some are not. To select the left star I need to click on empty space (pun intended) to the left of it
https://youtu.be/L5-CGJaUoNk

I tried to disable "atmosphere visualization", without it sky moves somewhat (I see the same stars in different place on screen), but it still misclicks

### Steps to reproduce
* Open Stellarium
* Zoom by *a lot*
* Try to click on stars

### System
* Stellarium version: 23.1
* Operating system: Gentoo linux amd64
* Graphics Card: Gigabyte Radeon RX 6700XT, x11-drivers/xf86-video-amdgpu-22.0.0:0
* Screen type (if applicable): Resolution 1920x1080, no scaling

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/11094530/log.txt)

### LINKED ISSUE #3315: Telescopic asterisms are broken
### Expected Behaviour
Render telescopic asterisms

### Actual Behaviour
Telescopic asterisms cannot be created due problem for finding reference stars

### System
* Stellarium version: 23.2
* Operating system: macOS 13.4.1

### Logfile 
[log.txt](https://github.com/Stellarium/stellarium/files/11969593/log.txt)

henrysky:
### Description

Still draft and writing. See discussion in #3982

Major Changes:
- Star catalogs are built upon Hipparcos and Gaia DR3. Every stars brighter than magnitude ~6.5 should be included for visual purposes. In total having ~220 millions stars down to ~18 mag.
- New star catalogs format and codes to allow proper astrometry calculation. RA, DEC, PMRA, PMDEC, Parallax, Radial Velocity (and apperant magnitude) are calculated and displayed at epoch T for stars that have all required data.
- RA/DEC coordinates are now stored at ICRS epoch J2016 (so one does not need to propagate large amount of Gaia stars back to J2000 but Hippracos stars are forwarded to J2016), there is now epoch information at the header of each catalog so Stellarium can check if the epoch is the one expected. 
- Magnitude and color system remains the same as the old catalogs, meaning now V-band magnitudes for majority of the stars are estimated from Gaia G-band magnitude and G_bp-G_rp color with about ~0.03 mag scattering (see https://gea.esac.esa.int/archive/documentation/GDR3/Data_processing/chap_cu5pho/cu5pho_sec_photSystem/cu5pho_ssec_photRelations.html). B-V color are more challenging if one simply estimating from G-mag and G_bp-G_rp color as shown. As a result, the values in the new catalog are synethetic photometry computed from Gaia low resolution spectra (~0.03 mag scattering compared among Hipparcos stars).
- `Star1` in the catalog now has object types field and will be displayed among the field `Type` on top of existing information. Meaning of the display in its current state can be found at https://simbad.cds.unistra.fr/Pages/guide/otypes.htx.
- `Star1` that travel outside of their own zone in the far future/past are handled by pre-computing which stars will have this behavior and put them in a global zone (an extra zone for each geodesic level). So stars like Arturus can be displayed properly in the far future.
- Some stars being difficult to be selected are fixed.
- Search by Gaia source ID are added (both via SIMBAD and offline), so the star with that Gaia source ID will be selected instead of a custom object
- Updated documentation regarding the star catalogs and astrometry accuracy.

I will upload the code to generate the new catalog here: https://github.com/henrysky/stellarium_star_catalogs. Give me a few days to organize.

To-do list for Henry:
- [ ] Release source code

Fixes #18
Fixes #85
Fixes #221
Fixes #348
Fixes #362
Fixes #363
Fixes #364
Fixes #369
Fixes #398
Fixes #399
Fixes #677
Fixes #777
Fixes #1016
Fixes #1032
Fixes #1061
Fixes #1189
Fixes #1192
Fixes #1541
Fixes #1659
Fixes #1724
Fixes #2786
Fixes #3108
Fixes #3136
Fixes #3315

Many issues related to missing stars, false stars, incorrect information, astrometry inconsistent with other softwares...

### Type of change

- [x] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: MacOS 15
* Graphics Card: Apple M3

## Checklist:


alex-w:
@henrysky I've uploaded your catalogs into our GH page - https://github.com/Stellarium/stellarium-data/releases/tag/stars-3.0 - you may fix the URL's in JSON config for stars.

But (!) why this JSON config is not updating automatically anymore in users side?

alex-w:
Ouch... the `StarWrapper1::getObjectTypeI18n()` is should be updated too to re-enabling translation the types of stars

gzotti:
Before I start testing: Is it possible to have both catalogs installed and start one or the other branch without problems? (What about installations of several versions of Stellarium using the same user data dir?) I cannot risk messing up anything before later next week.

alex-w:
> @henrysky I've uploaded your catalogs into our GH page - https://github.com/Stellarium/stellarium-data/releases/tag/stars-3.0 - you may fix the URL's in JSON config for stars.

Damn! GH can't uploading file over 2GB, so, we should change URL for part 8 (https://dss.stellarium.org/stars-3.0/stars_8_2v0_2.cat) or change the range between magnitudes inside of parts and get a more parts for catalog.

alex-w:
> Before I start testing: Is it possible to have both catalogs installed and start one or the other branch without problems? (What about installations of several versions of Stellarium using the same user data dir?) I cannot risk messing up anything before later next week.

Yes, but automatic updating JSON data in users directory should be fixed for it

alex-w:
Checksum for part 8 in JSON file should be updated too

alex-w:
OK, this PR probably fix many related to star catalogs bugs and we should revised all of them: #18, #85, #221, #348, #362, #363, #364, #369, #398, #399, #677, #777, #1016, #1032, #1061, #1189, #1192, #1541, #1659, #1724, #2786

gzotti:
2GB! Note that some users may still use FAT32 (or other size-limiting FS?) for whatever reasons. File sizes should be kept under 2GB.

10110111:
> Note that some users may still use FAT32

I doubt they'd use such large catalogs if they are limited by such file systems.

alex-w:
> I doubt they'd use such large catalogs if they are limited by such file systems.

Previously part 8 has size 0,5GB

gzotti:
This is part of my original concern about memory footprint. How much to load on which systems may depend on system capabilities. Some users may prefer to keep the current lightweight catalog. 4GB RAM is still very common. (Not sure about 1-2GB apart from the RPi platform and maybe other SBCs usable as telescope computers.)

henrysky:
> @henrysky I've uploaded your catalogs into our GH page - https://github.com/Stellarium/stellarium-data/releases/tag/stars-3.0 - you may fix the URL's in JSON config for stars.

Thanks! I have updated the JSON config for stars

> [edit] Please fix line `static const int StarCatalogFormatVersion = 12;` in file `StarMgr.cpp`

Fixed in the branch

> Ouch... the StarWrapper1::getObjectTypeI18n() is should be updated too to re-enabling translation the types of stars

Fixed in the branch but the SIMBAD 3-letters object type code covering stars, set of stars, quasar, galaxies.... probably need to convert to words (can use the one provided on SIMBAD too). Parts of the logic inside ``StarWrapper1::getObjectType`` can also merged to this object type column in the `Star1` catalog. Any opinion?

> Damn! GH can't uploading file over 2GB, so, we should change URL for part 8 (https://dss.stellarium.org/stars-3.0/stars_8_2v0_2.cat) or change the range between magnitudes inside of parts and get a more parts for catalog.

I think I will break the >15 mag catalogs into slightly smaller magnitude range to make it more flexible to donwload/storing them/

> This is part of my original concern about memory footprint. How much to load on which systems may depend on system capabilities. 

I don't think stellarium uses much more memory with the new star catalog compared to 24.3 both with all star catalogs at least on my computer.

Atque:
I see a few stars lack proper motion and distance (Caph, Navi and Phecda for instance). Possibly more.

I guess there was an error compiling the catalog?

stars/default/defaultStarsConfig.json
```diff
@@ -86,8 +86,8 @@
 			"count": 138.4,
 			"magRange": [16.5, 18],
 			"sizeMb": 0,
```
alex-w, line 88: Please fix size of file

alex-w:
> I think I will break the >15 mag catalogs into slightly smaller magnitude range to make it more flexible to donwload/storing them.

Good solution IMHO

alex-w:
> Fixed in the branch but the SIMBAD 3-letters object type code covering stars, set of stars, quasar, galaxies.... probably need to convert to words (can use the one provided on SIMBAD too). Parts of the logic inside StarWrapper1::getObjectType can also merged to this object type column in the Star1 catalog. Any opinion?

I've added cosmetic fix for star types. Adding support the SIMBAD's 3-letters object type codes for each object is not possible at the moment without many refactoring. Adding "humar readable" explanation is possible, but not need at the moment IMHO (see morphological description for DSO objects).

alex-w:
> I see a few stars lack proper motion and distance (Caph, Navi and Phecda for instance). Possibly more.

Absolute magnitudes for stars also seems very weird - for example alpha Cassiopeia has -12.02 now

henrysky:
> I see a few stars lack proper motion and distance (Caph, Navi and Phecda for instance). Possibly more.

Ah, good catch. There is some logic flaws in the cross-matching between Hipparcos and Gaia. All stars dim enough to be observed by Gaia but bright enough to have no Gaia pmra/pmdex/parallax will have this issue. I am figuring out a solution.

> Absolute magnitudes of stars also has very weird values - for example alpha Cassiopeia has -12.02 now

The parallax used to compute absolute magnitude is 100 times too small (but the parallax is correct in the catalog and correctly displayed). I will fix it now.

alex-w:
Ouch... please check spectral classes of stars also

Atque:
> Just started a first build. Very promising! However, see e Eri (82 Eri, HIP15510). Going backwards around November -30070 (yes, I know, very early. Just an arbitrary example, I have seen a few similar ones) the star vanishes. I have not traced a reason. In the current Zone catalog, a star may leave its zone triangle and vanish, maybe it's the same phenomenon.

I have seen this phenomenon in many other software. It seems difficult to solve.

sunshuwei:
Very much looking forward to Gaia replacing Hipparcos.
Will other catalog (HD, HR, SAO) numbers be added to the stars not covered by the Hipparcos catalog in the future? Gaia basically covers these stars.
And with the Gaia star chart, can sky culture use non Hipparcos stars?

alex-w:
> Very much looking forward to Gaia replacing Hipparcos. 

Gaia not replacing Hipparcos yet, only NOMAD and partially Tycho 2.

alex-w:
> Ouch... please check spectral classes of stars also

Yes, most stars has wrong spectral classes

alex-w:
I see many stars from Gaia catalog with distance, but without absolute magnitudes - why?

alex-w:
> I simply forget to add absolute magnitudes display to Star2 because parallax is a newly added information to them. (edit: actually I've thought of adding Absolute Magnitude as commented in the code, I was thinking about extinction because many of them might have high extinction so will mess up absolute magnitude (because apperant magnitude will be lower). If you think absolute magnitude is useful sure I can add them too)

Yes, absolute magnitude for stars with parallax/distance is important (at least for didactic purposes)

henrysky:
> Yes, most stars has wrong spectral classes

Spectral class of stars being wrong should be fixed. There is a slight reduction in number of stars in the first few catalogs because now HIP stars and their companions are better matched with Gaia DR3 counterpart so much less stars are drawn on top of each other (because they are being the same star).

> Yes, absolute magnitude for stars with parallax/distance is important (at least for didactic purposes)

Absolute magnitudes and distances are added now only if less than 20% parallax uncertainty. But parallax are displayed as long as there are measurements.

> Hopefully the numebr of stars in this case in small and won't bring much performance impact. I can quickly later to give you a number estimate.

I've estimated those numbers, going back to -100000 and going forward to 100000. ~13% of stars will leave their own zone in star catalog 0,  14.6% of stars will leave their own zone in star catalog 1 and 22.8% of stars will leave their own zone in star catalog 2.

> I think this is the reason for this version of star catalogs too. One solution I am thinking of is to create a level -1 Zone (the whole sky is a single zone) such that we put those HIP stars in the level -1 zone...

I think splitting the large star catalogs of the same geodesic grid level into a few smaller files and having a "-1 zone" require changes in quite a few places, which at this rate might be worth to switch from geodesic grid to HEALPix which will allow efficient searching of a specific Gaia ID because each Gaia ID are encoded with the info for HEALPix index of level 12 and lower (i.e., given a gaia ID, you know which zone they are going to be in at a certain level<12). But we will be looking at a timeline of a few months probably.

alex-w:
> Spectral class of stars being wrong should be fixed. There is a slight reduction in number of stars in the first few catalogs because now HIP stars and their companions are better matched with Gaia DR3 counterpart so much less stars are drawn on top of each other (because they are being the same star).

But files are changed and checksums should be updated in JSON file too to avoid problems.

> Absolute magnitudes and distances are added now only if less than 20% parallax uncertainty. But parallax are displayed as long as there are measurements.

OK, good solution!

alex-w:
@henrysky I see in the log now:
```
Loading star catalog: /opt/stellarium/share/stellarium/stars/default/stars_3_1v0_5.cat - error - file format is bad.
```

henrysky:
Actually the performance is not that bad if every stars in the first 3 catalogs are checked if they are inside the screen (instead of by zone). The brute-force check is turned on if it is more than 3000 years away from 2016. You guys can check the latest commit (just a temporary solution, the code/logic sure can be better).

Atque:
> Actually the performance is not bad if every stars in the first 3 catalogs are checked if they are inside the screen (instead of by zone). The brute-force check is turned on if it is more than 3000 years away from 2016. You guys can check the latest commit (just a temporary solution without modifying the star catalogs, the code/logic sure can be better). Stars like Arcturus/Barnard's star can now be properly displayed in the future/past even when zoomed in.

This commit keeps stars selectable in the distant future or past too.

I see a few double stars only show PM for one component, and the other (bearing only a Gaia designation) being completely stationary (Porrima, Sabik).

gzotti:
On stars leaving their Zone: Are there stars migrating by more than one adjacent zone? It may be enough to just check all visible zones plus their immediate neighbors. (Or expand "neighbors" further if this is not enough.)

henrysky:
Hi all, I think I have found the root cause of #3832 and #3733 and have pushed a commit to fix that issue in this branch. The root cause is abbrevation is incorrectly apply when searching solar system object leading to abbrevation needed to be applied to `cleverfind()` function which leads to a wrong mouse click pixel location on screen which lead to some stars are unselectable (especially in those sky region most affected by abbrevation). I believe I have fixed the root cause and you can select easily and correctly for plutos and all its moons as well as stars k Cru cluster star.

On the other hand, I have added functionality to allow search by Gaia ID (via both SIMBAD and offline). So if a particular stars with a Gaia ID in the star catalog, that star will be selected instead of a custom object.

I have also added a global zone at each geodesic level for those stars that move more than 1 zone in -110000 to 110000 instead of brute-forcing. @gzotti you are right that only stars moved more than 1 zone need to be taken care of because border zones are checked but not border of border zones. Number of stars that are going to move more than 2 zones are much smaller (less than 1% of the stars in the first 3 catalogs)

alex-w:
Hmm... a quick test and I got a crash dumps in linux (revison [d8a7bc7](https://github.com/Stellarium/stellarium/pull/3992/commits/d8a7bc751f71280ac2aff878f2b46a2ebbab5ffd))

henrysky:
How about with the latest [8b91971](https://github.com/Stellarium/stellarium/pull/3992/commits/8b91971596038a04a1a6d1cb74a52be13f24ee5a)? Works on my MacOS. Other star catalogs (those not included in stellar out of the box) are not going to work because of an exra zone added as a global zone for each geodesic level, I am working on those currently.

alex-w:
No crash anymore!

stars/default/defaultStarsConfig.json
```diff
@@ -1,93 +1,93 @@
 {
-	"version": 13,
+	"version": 14,
```
alex-w, line 2: Did you changed the version number in C++ code?

henrysky: I just recognized this and force pushed to bump the version.

alex-w: Hint: `static const int StarCatalogFormatVersion`

henrysky: I will change that in the morning or you can push the change😅 i have turned off my laptop to go to sleep. I will make a checklist for myself to keep track of what action need to be taken each time I update catalogs and JSON.

alex-w:
Hmm... The search tool cannot found Gaia stars.

alex-w:
I doing not all checks and tests yet, but the pull request is mature almost - of course related changes in the User Guide are needed. I think we may unmark “draft” for PR.

Probably the PR may be targeted to release 24.4.

10110111:
> Probably the PR may be targeted to release 24.4.

To the nearest release? No! This PR has diff stats of +8395/-243319 lines, but hasn't been tested by anyone but a few people. Its merge shouldn't happen before the nearest release.

gzotti:
I just see a crazy flashing: Center Arcturus, date -84796-4-30 (not sure how much it matters), fov 55-65. Else happy it is still visible!
Yes, it's too early to enable this, but it looks promising. Needs more testing and working out the side effects. In far past/future framerate seems to go down considerably, etc.

alex-w:
@10110111 you're right - just producing the binary packages for this PR should be enough for wide testing by community.

OK, the target milestone is 25.1 now

alex-w:
> I just see a crazy flashing: Center Arcturus, date -84796-4-30 (not sure how much it matters), fov 55-65. Else happy it is still visible! 

Yes, I can see flashing too - and I see the very fast movement the 39 Ser star at flashing moment

gzotti:
Against which "reference solution" can we actually test the new proper motion? Is there anything besides CdC? How accurate is orbitsimulator.com?

Atque:
> Against which "reference solution" can we actually test the new proper motion? Is there anything besides CdC? How accurate is orbitsimulator.com?

AFAIK Orbitsimulator is very good, but sadly its coordinate system is simplified, so only relative positions between stars can be used.

henrysky:
I have updated the star catalogs to solve various issues reported by you guys so far. Extra catalogs are the same so no need to be updated so far.

> Hmm... The search tool cannot found Gaia stars.

@alex-w Searching Gaia stars is working fine for me. Are you sure you have input a valid Gaia source ID (Gaia DR3 xxxxxxx....)? Like `Gaia DR3 4472832130942575872`.

> > I just see a crazy flashing: Center Arcturus, date -84796-4-30 (not sure how much it matters), fov 55-65. Else happy it is still visible!
> 
> Yes, I can see flashing too - and I see the very fast movement of the 39 Ser star at flashing moment

@gzotti @alex-w I cannot reproduce this issue but I did found the root cause for the problem of 39 Ser. 39 Ser is a star in a 6.0-7.5 mag star catalog where it gets brighter than catalog minimum magnitude in the past and cause issue by looking up drawing paramaters outside of `rcmag_table` range. The issue has been fixed by constructed `rcmag_table` brighter than each star catalog minimum magnitude (HIP 16757 actually get way brighter in the past). Can you check if the latest commit has fixed the problem? If not can you attach a video of the issue?

> Needs more testing and working out the side effects. In far past/future framerate seems to go down considerably, etc.

@gzotti Low framerate in far past/future issue might have caused by recalculating magnitude and then check EVERY single stars in catalog 0/1/2 to make sure stars that get much brighter in the past/future get drawn instead of not surviving magnitude cutoff. Now stars that get much brighter (delta mag is <-0.3, meaning the stars can get more than 0.3 mag bright at some point) in the past/future are put in the global zone in each geodesic level. So only a fraction of the stars are checked (i.e., those in the global zone) and make sure they are drawn in the past/future even zoomed out but not those stars in normal zones. Stars that get much dimmer in the past/future stay in their zone because the issue is bright stars potentially not get drawn but not dim stars get drawn (they will still be drawn according to their recomputed magnitude at that year NOT catalog value if information avaliable).

You can see if the latest commit run smoother in far past/future.

alex-w:
The flashing is gone in latest commit

alex-w:
> Searching Gaia stars is working fine for me. Are you sure you have input a valid Gaia source ID (Gaia DR3 xxxxxxx....)? Like `Gaia DR3 4472832130942575872`.

Please see screenshots:
stellarium-023
stellarium-024

So, probably, `getGaia()` gives wrong identifiers

henrysky:
> So, probably, `getGaia()` gives wrong identifiers

I have just fixed the issue because getGaia() has a return type of `long` which has no guranttee the integer is 64bit, now I've switched to `int64_t` and should have fixed the issue.

alex-w:
> Against which "reference solution" can we actually test the new proper motion?

Not a reference solution, but very good start point for tests - https://articles.adsabs.harvard.edu/pdf/1994QJRAS..35....1M (the link from issue #353)

alex-w:
> > So, probably, `getGaia()` gives wrong identifiers
> 
> I have just fixed the issue because getGaia() has a return type of `long` which has no guranttee the integer is 64bit, now I've switched to `int64_t` and should have fixed the issue.

This is strange, but when I select any Gaia stars in search tool I got a selected Rigel

henrysky:
> This is strange, but when I select any Gaia stars in search tool I got a selected Rigel

Searching Gaia stars get you to Rigel? That is weird, Rigel does not even has a Gaia source id for being too bright!

> Against which "reference solution" can we actually test the new proper motion? Is there anything besides CdC? How accurate is orbitsimulator.com?

I think the "reference solution" can be IAU SOFA library: http://www.iausofa.org/ with astrometry related tool documentation: http://www.iausofa.org/2023_1011_C/sofa/sofa_ast_c.pdf but I have never tried it. `astropy` is a good alternative but I have never been in a situation before where very precise RA/DEC are needed so need to see how good they are for this task but I would assume `astropy` is good since I often use it in my research for other purposes.

gzotti:
It would be great to set up a few representative stars (any random mag10-12 stars, and a few outstanding "runners" [Barnard, ...]) into an automated unit test. Work out reference data with e.g. astropy, and see how close your solution comes to that, in J2025 and say e.g. &pm;5000 and &pm;10.000 years. (Or even more? What are the guarantees for astropy regarding long-time simulation?)

stars/default/stars_hip_cids_0v0_0.cat
alex-w, line 1: Please update JSON also

henrysky:
I have updated the star catalogs again. A main update is I've put stars that are brighter than 6.5 mag who will cross at least ONE zone in -100,000 to 100,000 years into the global zone (instead of crossing at least two zone). I've found the level 0 triangles are so big that you can zoom enough to NOT have border triangles checked so some stars still disappearing in the future and in the past. A minor update is I have put ``Ross 248`` into the third star catalog even thought it is very dim now because it is an interesting object which will be the closest star to us in ~40,000 years. **If you think there are interesting objects deserving full 6D astrometry propagation that are dimmer than magnitude 10.5 and does not have a HIP numbers, you can comment below with its Gaia source ID so I can put them in the bright star catalogs.** For dim star catalogs, only level 5 catalog is changed and is here: https://www.astro.utoronto.ca/~hleung/shared/stellarium_new_catalog/trial4/stars_5_1v0_4.cat

I have also added astrometry test on 2 stars at +/- 5,000 years as well as +/- 50,000 years compared to astropy, asserting deviation in those epoch is within 36 arcseconds for RA/DEC, 1 mas for parallax, 5 mas/yr for PMRA/PMDEC, and 2km/h for radial velocity. The main issue is not the algorithm in stellarium is wrong or we are too simple to not take relativistic effects/light traveling time into account, but small numerical error for very high proper motion stars like Barnard's star actually is signaficant like 50,000 years from now. So I think this is as good as we can do in terms of limited computational resources but Don't expect <1 arcsecond precision 50,000 years from now.

I have also started to update the docs. Let me know if there are other places I need to update besides the two I am working on (Appendix C star catalogs and Appendix F astrometry accuracy).

gzotti:
36 arcsecs (or even an arcmin) difference in 50ky sounds perfectly adequate to me. We just should avoid real systematic errors like it may have been before. 
My main interest is the naked-eye sky (Arcturus &friends not vanishing by Zone error, &alpha; Cen not flying apart, ...), I cannot name favourite mag 10 runners without own investigation. Will Ross 248 be visible to the naked-eye then?

henrysky:
I Don't think we have systematic errors on RA/DEC anymore. ``Ross 248`` is not gonna be naked eyes visible (maxed at ~9.8 mag) but it is going to be our closest stars for about 9,000 years covered by stellarium date range (https://en.wikipedia.org/wiki/Ross_248), something maybe of interest for educational purposes to show how close it can get.

I am also marking this PR ready for review unless there is other existing issue that can potentialy be fixed with changing the catalog format. I do not have any more features to add and I think I have fixed bugs reported here so far.

Atque:
We ought to make sure all Bayer and Flamsteed designations are added in the new catalog. For example, 41 Ari is only refered to as its Gaia designation (and lacks PM).

bild

alex-w:
> We ought to make sure all Bayer and Flamsteed designations are present in the new catalog. For example, 41 Ari is only refered to as its Gaia designation (and lacks PM).

It's a task for next step (in my todo already :) ), not for current branch

henrysky:
> We ought to make sure all Bayer and Flamsteed designations are present in the new catalog. For example, 41 Ari is only refered to as its Gaia designation (and lacks PM).

41 Ari was cross-matched to a wrong star near the true 41 Ari (if you search 41 Ari, it will point it to another stars with HIP 13209A). I will fix it later to see if we have more bright stars get matched incorrectly.

src/core/modules/Star.hpp
```diff
@@ -178,7 +178,7 @@ struct Star
                                  double & RV,
                                  float    dyrs) const
    {
-      if (getX2() != 0. || getDx2() != 0.) {
+      if (getPreciseAstrometricFlag()) {
```
alex-w, line 181: Unit test is fails now

henrysky: 🥲 I did not think it throught and that is actually a wrong logic. Have reverted the commit back. But still need to solve the issue if a stars is `Star1` having no parallax but with radial velocity in the catalog will result in nan/inf in info string.

alex-w (approved):
Thank you very much!

Atque:
I've been testing, and two stars still have stationary doubles: Xi UMa (Alula Australis) and Porrima. Otherwise, this looks finished to me.

guide/app_star_catalogue.tex
```diff
@@ -45,44 +44,41 @@ \subsection{General Description}%\label{general-description}
...
+\item[\file{stars\_7\_1v0\_3.cat}]
+\item[\file{stars\_8\_2v0\_3.cat}]
 \end{description}
 
 There also exist some control and reference files:
 
 \begin{description}
-\item[\file{stars\_hip\_cids\_0v0\_0.cat}]
-\item[\file{stars\_hip\_sp\_0v0\_3.cat}]
+\item[\file{object\_types\_v0_1.cat}]
```
gzotti, line 61: backslash missing in front of third underscore.

guide/app_star_catalogue.tex
```diff
@@ -209,32 +204,50 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
-                                               \file{stars\_hip\_component\_ids.cat} file. The value of this field turns out to be 
-											   the line number in the file - 1\\%\midrule
-x0             &  4 & int                & 4 & position of the star relative to the central point in the star's zone, in axis 1\\%\midrule
-x1             &  8 & int                & 4 & position of the star relative to the central point in the star's zone, in axis 2\\%\midrule
-b\_v           & 12 & unsigned char      & 1 & magnitude level in B-V colour. This value refers to one of 256 discrete steps 
-                                               in the magnitude range for the file\\%\midrule
-mag            & 13 & unsigned char      & 1 & magnitude level in the V-I colour. This value refers to one of 256 discrete steps 
-                                               in the magnitude range for the file\\%\midrule
-sp\_int        & 14 & unsigned short int & 2 & index into an array of spectral type descriptions 
+gaia_id        &  0 & long long          & 8 & Gaia source id\\%\midrule
```
gzotti, line 234: must be gaia\_id for typesetting

guide/app_star_catalogue.tex
```diff
@@ -209,32 +204,50 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+x0             &  8 & int                & 4 & Cartesian x position of the star on an unit sphere at the catalog epoch scaled by $2\times10^9$\\%\midrule
+x1             & 12 & int                & 4 & Cartesian y position of the star on an unit sphere at the catalog epoch scaled by $2\times10^9$\\%\midrule
+x2             & 16 & int                & 4 & Cartesian z position of the star on an unit sphere at the catalog epoch scaled by $2\times10^9$\\%\midrule
+dx0            & 20 & int                & 4 & First component of the star's proper motion vector at the catalog epoch scaled by $1\times10^3$\\%\midrule
+dx1            & 24 & int                & 4 & Second component of the star's proper motion vector at the catalog epoch scaled by $1\times10^3$\\%\midrule
+dx2            & 28 & int                & 4 & Third component of the star's proper motion vector at the catalog epoch scaled by $1\times10^3$\\%\midrule
+b\_v           & 32 & short              & 2 & $B-V$ colour in milli-mag unit.
```
gzotti, line 241: This and the next table line must have \\\\ as linebreak marker.

guide/app_star_catalogue.tex
```diff
@@ -209,32 +204,50 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+mag            & 34 & short              & 2 & $V$-band magnitude in milli-mag unit.
+plx            & 36 & unsigned short     & 2 & parallax of the star in 0.02 mas unit.\\
+plx_err        & 38 & unsigned short     & 2 & parallax error of the star in 0.01 mas unit.\\
```
gzotti, line 244: plx\\_err

guide/app_star_catalogue.tex
```diff
@@ -209,32 +204,50 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+rv             & 40 & short              & 2 & radial velocity of the star in 0.1 km/s unit.\\
+sp\_int        & 42 & unsigned short     & 2 & index into an array of spectral type descriptions 
                                                which is taken from the file \file{stars\_hip\_sp.cat}, the index corresponds to the 
 											   line number in the file - 1\\%\midrule
-dx0            & 16 & int                & 4 & proper motion of the star in axis 1\\%\midrule
-dx1            & 20 & int                & 4 & proper motion of the star in axis 2\\%\midrule
-plx            & 24 & int                & 4 & parallax of the star. To get the actual value, divide by 10000.\\\bottomrule
+obtype		   & 44 & unsigned char      & 1 & object type classification index. The index corresponds to the line number in the file \file{object\_types_v*.cat} - 1.\\
```
gzotti, line 249: > \file{object\\_types_v*.cat}

\file{object\\_types\\_v*.cat}

guide/app_star_catalogue.tex
```diff
@@ -243,14 +256,15 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
 \emph{Name} & \emph{Offset}& \emph{Type} & \emph{Size} &\emph{Description}\\\midrule
-x0   &  0      & int          & 20 bits & position of the star relative to the central point in the star's zone, in axis 1\\%\midrule
-x1   & 20 bits & int          & 20 bits & position of the star relative to the central point in the star's zone, in axis 2\\%\midrule
-dx0  & 40 bits & int          & 14 bits & proper motion of the star in axis 1\\%\midrule
-dx1  & 54 bits & int          & 14 bits & proper motion of the star in axis 2\\%\midrule
-b\_v & 68 bits & unsigned int &  7 bits & magnitude level in B-V colour. This value refers to one of 128 discrete steps 
-                                          in the magnitude range for the file\\%\midrule
-mag  & 75 bits & unsigned int &  5 bits & magnitude level in the V-I colour. This value refers to one of 32 discrete steps 
-                                          in the magnitude range for the file\\\bottomrule
+gaia_id        &  0 & long long          & 8 & Gaia source id\\%\midrule
```
gzotti, line 259: same as above, gaia\\_id and int64

guide/app_star_catalogue.tex
```diff
@@ -209,32 +204,50 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+	\vec{q} &= \left(-\frac{\texttt{x0} \cdot \texttt{x2}}{xy}, -\frac{\texttt{x1} \cdot \texttt{x2}}{xy}, xy\right);\\
+	\mu_{\alpha^*} &= \vec{p} \cdot \left(\texttt{dx0}, \texttt{dx1}, \texttt{dx2}\right); \\
+	\mu_{\delta} &= \vec{q} \cdot \left(\texttt{dx0}, \texttt{dx1}, \texttt{dx2}\right);
+\end{aligned}
+\end{equation}
+
+where $\mu_{\alpha^*}\equiv\mu_{\alpha}\cos{(\delta)}$.
 
 \begin{table}[p]
 \begin{tabular}{llllp{55mm}}\toprule
```
gzotti, line 232: After these changes, the paragraph width could be enlarged to 75mm. (55mm-->75mm)

guide/app_star_catalogue.tex
```diff
@@ -243,14 +256,15 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
-dx0  & 40 bits & int          & 14 bits & proper motion of the star in axis 1\\%\midrule
-dx1  & 54 bits & int          & 14 bits & proper motion of the star in axis 2\\%\midrule
-b\_v & 68 bits & unsigned int &  7 bits & magnitude level in B-V colour. This value refers to one of 128 discrete steps 
-                                          in the magnitude range for the file\\%\midrule
-mag  & 75 bits & unsigned int &  5 bits & magnitude level in the V-I colour. This value refers to one of 32 discrete steps 
-                                          in the magnitude range for the file\\\bottomrule
+gaia_id        &  0 & long long          & 8 & Gaia source id\\%\midrule
+x0             &  8 & int                & 4 & Right ascension $\alpha$ of the star in mas unit.\\%\midrule
+x1             & 12 & int                & 4 & Declination $\delta$ of the star in mas unit.\\%\midrule
+dx0            & 16 & int                & 4 & Proper motion $\mu_{\alpha}$ of the star in $mu$as/year%\midrule
```
gzotti, line 262: This and next 3 lines: again, please use \\\\ as line end marker (The "%\midrule" is a testwise commented-away previous horizontal line and can go away. )

Also, this and next line, use "$\mu$as/year"  (with backslash) to typeset "&mu;as/year"

guide/app_star_catalogue.tex
```diff
@@ -243,14 +256,15 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+dx1            & 20 & int                & 4 & Proper motion $\mu_{\delta}$ of the star in $mu$as/year%\midrule
+b\_v           & 24 & short              & 2 & $B-V$ colour in milli-mag unit.
+mag            & 26 & short              & 2 & $V$-band magnitude in milli-mag unit.
+plx            & 28 & unsigned short     & 2 & parallax of the star in 0.01 mas unit.\\
+plx_err        & 30 & unsigned short     & 2 & parallax error of the star in 0.01 mas unit.\\
```
gzotti, line 267: again plx\\_err. In the last line you can omit the trailing \\\\.

guide/app_star_catalogue.tex
```diff
@@ -259,12 +273,11 @@ \subsubsection{Star Data Records}%\label{star-data-records}
 \begin{table}[htbp]
 \begin{tabularx}{\textwidth}{llllX}\toprule
 \emph{Name} & \emph{Offset} & \emph{Type} & \emph{Size} & \emph{Description}\\\midrule
-x0   & 0       & int          & 18 bits & position of the star relative to the central point in the star's zone, in axis 1\\%\midrule
-x1   & 18 bits & int          & 18 bits & position of the star relative to the central point in the star's zone, in axis 2\\%\midrule
-b\_v & 36 bits & unsigned int &  7 bits & magnitude level in B-V colour. This value refers to one of 128 discrete steps 
-                                          in the magnitude range for the file\\%\midrule
-mag  & 43 bits & unsigned int &  5 bits & magnitude level in the V-I colour. This value refers to one of 32 discrete steps 
-                                          in the magnitude range for the file\\\bottomrule
+gaia_id        &  0 & long long          & 8 & Gaia source id\\%\midrule
```
gzotti, line 276: gaia\\_id

guide/app_star_catalogue.tex
```diff
@@ -259,12 +273,11 @@ \subsubsection{Star Data Records}%\label{star-data-records}
...
+x0             &  8 & int                & 3 & Right ascension $\alpha$ of the star in 0.1 arcsecond unit.\\%\midrule
+x1             & 11 & int                & 3 & Declination $\delta$ of the star in 0.1 arcsecond unit.\\%\midrule
+b\_v           & 14 & unsigned short     & 1 & $B-V$ colour in 25 milli-mag unit offset by $+1$ mag.
```
gzotti, line 279: add \\\\ to end the table line.

src/core/StelObject.hpp
```diff
@@ -332,7 +332,8 @@ public slots:
 	//! @param core
 	//! @param flags
 	//! @param decimals significant digits after the comma.
-	virtual QString getMagnitudeInfoString(const StelCore *core, const InfoStringGroup& flags, const int decimals=1) const;
+	//! @param magOffset magnitude offset to apply (can be due to changing distance)
```
gzotti, line 335: Can you please describe this new parameter in more detail? Or maybe it is described elsewhere?

src/core/modules/Star.hpp
```diff
@@ -22,266 +22,394 @@
...
 // to store much data for bright stars (Star1), but only little or even
-// very little data for faints stars (Star3). Using only 6 bytes for Star3
+// very little data for faints stars (Star3). Using only less bytes for Star3
 // makes it feasible to store hundreds of millions of them in main memory.
 
-
 static inline float IndexToBV(int bV)
 {
-	return static_cast<float>(bV)*(4.f/127.f)-0.5f;
+   return static_cast<float>(bV) * (4.f / 127.f) - 0.5f;
```
gzotti, line 54: indentation looks broken

gzotti:
Actually for @alex-w: I haven't done so in a long time now. After loading a new star catalog it appears the stars are immediately used. At the end of catalog downloads however, the message is something like "All catalogs downloaded, restart Stellarium to load them". I think this last sentence could be removed, right?

gzotti:
Just running on RPi4. Mem footprint is very relaxed (comparable to the old). I don't think I had ever downloaded all star catalogs. Now I start seeing some loss in framerate at level 7 of 9. 12fps is still acceptable, though. I dare going still deeper...
What is the recommended way to remove a star catalog? Or could we just deactivate somehow? I guess only manually, editing starsConfig.json?

And important for comparisons against older versions or if there is just other needs to run older versions: Is it possible to keep all catalogs under one Stellarium user dir? The star catalog's name is always "default" and has been probably since 2006. If there is need for separate catalogs, this new one could be named e.g. hip_gaia, so that only somewhere in code (StarMgr.cpp) the directory name is adjusted from `default` to `hip_gaia`. Note that I don't say we must allow loading the old catalog into new program versions by config, but this is such a significant change that we should think about such mixed legacy installations.

alex-w:
> Just running on RPi4. Mem footprint is very relaxed (comparable to the old). I don't think I had ever downloaded all star catalogs. Now I start seeing some loss in framerate at level 7 of 9. 12fps is still acceptable, though. I dare going still deeper... What is the recommended way to remove a star catalog? Or could we just deactivate somehow? I guess only manually, editing starsConfig.json?

Just remove specific *.cat files - no need manually editing json file.

> And important for comparisons against older versions or if there is just other needs to run older versions: Is it possible to keep all catalogs under one Stellarium user dir? The star catalog's name is always "default" and has been probably since 2006. If there is need for separate catalogs, this new one could be named e.g. hip_gaia, so that only somewhere in code (StarMgr.cpp) the directory name is adjusted from `default` to `hip_gaia`. Note that I don't say we must allow loading the old catalog into new program versions by config, but this is such a significant change that we should think about such mixed legacy installations.

JSON file will be recreated if their version number is not equal to expected one in the binary package - as result you may use one place to storage all possible catalogs ;)

alex-w:
> Actually for @alex-w: I haven't done so in a long time now. After loading a new star catalog it appears the stars are immediately used. At the end of catalog downloads however, the message is something like "All catalogs downloaded, restart Stellarium to load them". I think this last sentence could be removed, right?

Maybe, I just don’t test loading catalogs on-the-fly…

gzotti:
OK, but running e.g. 25.1, 1.2 and 0.15.2 on the same PC (for comparison, linked with RemoteSync or for whatever reasons; or alternating between versions for with/without ASCOM stuff), sharing the same STEL_USERDIR, may trigger conflicts when the second launched overwrites the first's JSON. (And does this mechanism work for old versions like 0.15?) This looks like dangerous design for me, and goes against whatever may have been planned in 2010 when "default" was apparently introduced. @xalioth, @10110111 ?

alex-w:
> OK, but running e.g. 25.1, 1.2 and 0.15.2 on the same PC (for comparison, linked with RemoteSync or for whatever reasons; or alternating between versions for with/without ASCOM stuff), sharing the same STEL_USERDIR, may trigger conflicts when the second launched overwrites the first's JSON. (And does this mechanism work for old versions like 0.15?)

Overwriting is possible after comparing versions in JSON file and in binary package, plus our catalogs is not changed frequently…

> This looks like dangerous design for me, and goes against whatever may have been planned in 2010 when "default" was apparently introduced. @xalioth, @10110111 ?

Well, the `default` section for our catalogs (stars + dso) and some custom catalogs was ambitious idea, but it never implemented and the code for switching between various collections was removed a long time ago…

gzotti:
OK, no need to add switching code. Does a simple change of the catalog path in StarMgr::copyDefaultConfigFile() and StarMgr::init() lead to the suggested separation of "all versions which used the HIP/TYC/NOMAD catalog (named "default" from 2010 to 2024)" and "versions starting with 24.4+ using this new HIP/GAIA catalog". Old versions keep working with their current logic (in theory even catalog bugs may still be fixed, as unlikely as this is), while new versions will just never have the necessity to touch the "default" catalog. If you have no old version installed, the name of the (only) catalog will just be hip_gaia (or whatever we name this now, "v25", just not "default"). Truly incompatible changes should be handled like this, IMHO.

alex-w:
> OK, no need to add switching code. Does a simple change of the catalog path in StarMgr::copyDefaultConfigFile() and StarMgr::init() lead to the suggested separation of "all versions which used the HIP/TYC/NOMAD catalog (named "default" from 2010 to 2024)" and "versions starting with 24.4+ using this new HIP/GAIA catalog". Old versions keep working with their current logic (in theory even catalog bugs may still be fixed, as unlikely as this is), while new versions will just never have the necessity to touch the "default" catalog. If you have no old version installed, the name of the (only) catalog will just be hip_gaia (or whatever we name this now, "v25", just not "default"). Truly incompatible changes should be handled like this, IMHO.

Maybe you're right and string `stars/default` in the code (C++ & CMake) should be changed to `stars/hip_gaia`

alex-w:
> Actually for @alex-w: I haven't done so in a long time now. After loading a new star catalog it appears the stars are immediately used. At the end of catalog downloads however, the message is something like "All catalogs downloaded, restart Stellarium to load them". I think this last sentence could be removed, right?

Just checked right now - no, we shouldn't remove this last sentence, because stars are not loaded from catalogs on-the-fly (at the moment of loading catalog)

gzotti:
RPi3 also works, at least with extra cats stars_4 and stars_5. It downloads stars_6 but initialisation fails due to lack of memory. I did not realize this and happily started downloading even stars_7, then stellarium crashed with std::bad_alloc. But I think RPi3 is just the minimum platform to be used for limited installations, so going to about Pluto magnitude (13.75) should be fine.

src/core/StelObject.hpp
alex-w: OK, this is final point who need to fixation

henrysky:
I am going to fix issue of the two stars brought up by @Atque. @alex-w Should I bump all the star catalogs version to 1.0? Beacuse the new star catalogs are not compatible with the old v0 star catalogs?

alex-w:
> @alex-w Should I bump all the star catalogs version to 1.0? Beacuse the new star catalogs are not compatible with the old v0 star catalogs?

I see no necessary repacking catalogs

src/core/StelObject.hpp
henrysky: I have described more. Basically when the distance of stars change because of proper motion/radial motion, magOffset is used to compute new apparant magnitude.

henrysky:
I have fixed all missing pm/parallax issue mentioned by others. @alex-w Only stars4/5 is changed and is here https://www.astro.utoronto.ca/~hleung/shared/stellarium_new_catalog/trial5/.

src/core/StelObject.hpp
```diff
@@ -332,7 +332,7 @@ public slots:
 	//! @param core
 	//! @param flags
 	//! @param decimals significant digits after the comma.
-	//! @param magOffset magnitude offset to apply (can be due to changing distance)
+	//! @param magOffset magnitude offset to apply to display final apperant magnitude, are used if a star distance has changed in the past/future
```
alex-w, line 335: Maybe _apparent_?

henrysky: done!

alex-w:
> I have fixed all missing pm/parallax issue mentioned by others. @alex-w Only stars4/5 is changed and is here https://www.astro.utoronto.ca/~hleung/shared/stellarium_new_catalog/trial5/.

OK, done!

alex-w:
@gzotti @10110111 any comments?

P.S. A good question - squash or rebase?

10110111:
Some new files here lack the POSIXly-desired trailing newlines, which is indicated by git-diff and by GitHub's diff viewer. These should be added.

> P.S. A good question - squash or rebase?

Are there any testable versions among these commits, so that it would make sense to run, say, git-bisect on them? If yes, I suppose it's better to rebase.

In any case, the changes to documentation (the Guide first of all) should be separated and squashed into a single commit after all the others.

gzotti:
Sorry our ISP was offline for some time in mid-afternoon for whatever reason. Dec.24 overload? (about 15.000 reports at my ISP.)

 The current incremented version number is OK for me. Just the new path was important to allow parallel installations, thanks. 
I just wanted to follow the morphing of our 3-star-linked constellation artwork over the millennia. However it seems the projection is computed only at loading of the respective skyculture. E.g. loading Inuit around 85000AD looks weird. But this is not for this PR. Now SC authors face a new challenge: define the 3-star link with low-PM stars :-)  (Or make distortion part of the show...)

I think there is no half-working in-between version to ever bisect. 24.4 has the HIP/TYC/NOMAD catalog, and this PR switches to HIP/GAIA DR3. For me a rebase&squash would be OK, even if the changes is large. Also, the directory issue may break any bisects.  Of course, later fixes will be separate.  So, for me it's OK.

gzotti (approved):
Cosmetics like the empty line at end should be fixed, I second that (whatever git or CodeFactor report). 

Else, this looks OK for me, many thanks!

gzotti:
I just see that TestAstrometry fails, at least on Windows. RA stays at 0 after searchGaiaIDepochPos(,. ,. RA, ...) @henrysky ?

henrysky:
> I just see that TestAstrometry fails, at least on Windows. RA stays at 0 after searchGaiaIDepochPos(,. ,. RA, ...) @henrysky ?

Fixed in fecb74827cacc0cd12d649f2943df704d117589d, issue is how the testAstrometry finds the star catalog for testing depending where you launch the test, nothing is wrong with the main code that testAstrometry is designed to test.

gzotti:
Still no luck. I build in QtCreator and run the tests just there, so no installation. 

Fail: TestAstrometry::initTestCase
'!files.isEmpty' returned FALSE. (No matching star catalog found at D:/)

My star catalogs from master are in D:\StelDev\GIT\stellarium\stars\hip_gaia3. 
Build directory is D:\StelDev\GIT\build-stellarium-Desktop_Qt_6_8_2_MSVC2022_64bit-Release

Maybe it would work if build dir is inside source dir? 
Most important that it works in the CI, not my system, though...

Issue #3170, opened 2023-04-11
Drag and drop an image, platesolve it and display it with variable transparency above background


void4:
**Is your feature request related to a problem? Please describe.**
It's frustrating to match external images by eye, since they can be of different scales, show features with different brightnesses compared to Stellarium's built in images (or surveys like SDSS) or be located in an unknown location altogether.

**Describe the solution you'd like**
It would be nice if one could drag and drop any astrophotography images into Stellarium, which it would then plate solve (possibly using third party services/code like [astrometry.net](https://astrometry.net/use.html)) and then overlay it over the existing sky background (ideally with a variable transparency slider). 

**Describe alternatives you've considered**
This is probably too complex to be built in to the standard program, but it would be nice if at least such a plugin existed. Only very few Stellaria seem to have this ability. The only one I'm aware of is [WorldWideTelescope](http://worldwidetelescope.org/webclient/?wtml=http%3A%2F%2Fwwtcoreapp-data-app.azurewebsites.net%3A80%2Fwwtweb%2FShowImage.aspx%3Freverseparity%3DFalse%26scale%3D3.050514%26name%3Dimage.png%26imageurl%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F15747254%26credits%3DAstrometry.net%2BUser%2B(All%2BRights%2BReserved)%26creditsUrl%3D%26ra%3D148.849856%26dec%3D68.991582%26x%3D625.5%26y%3D318.9%26rotation%3D-173.17%26thumb%3Dhttps%3A%2F%2Fnova.astrometry.net%2Fimage%2F15747256%26wtml%3Dtrue)

**Additional context**
I think this would be a very useful addition and save many people's time.

mleila1312:
Hello, 

I would be interested in contributing and working on this suggestion. If I used Astronomy.net, there could be a small issue. This library is under GPL 3.0 license however the license of stellarium is 2.0 or above, would that be a problem? 

A solution to this could be to indicate that this particular plug-in is under GPL 3.0+ license, that the entirety of Stellarium (plug-in included) would be under GPL 3.0+, and that Stellarium without the pulg-in is under GPL 2.0+. I have looked for other similar libraries but the ones I found were under GPL 3.0(+) or copyrights.

gzotti:
Thanks for your interest. Look into @10110111 's recently added Lens Distortion plugin, I think it could be enhanced with automatic plate solving.

10110111:
[...]

Issue #75, opened 2018-01-14
Interactive Editor Plugin for nebula images 


sunshuwei:
I want to have a plugin that has the following features:
1. Allows users to easily add a square the nebula image, integrated astrometry.net function, the program can automatically determine the coordinates of four corners.
2. Allow users to upload the nebula image to the server, the official audit to determine whether the incorporation of stellarium procedures.
3. If the user uploads more information, depending on these photos may add new features. Of course, if the server resources are not enough, then can some way to limit the number of uploads?
there are too many people want to contribute their nebula image to Stellarium but do not know the channels and methods.

gzotti:
I think we somehow have wanted this (at least (1)) for years now. Unfortunately I just found even Barry's tools (Guide, section 8.2) are (sadly) no longer available. Any volunteers?

gzotti:
It was a blueprint, Matthew Gates filed it in 2009 (!): 
https://blueprints.launchpad.net/stellarium/+spec/nebula-editor

An interactive tool for creating nebula texture sets. This includes modifying the StelSkyImageMgr class so that it can handle editing of the images sets, and allowing users to create new nebula sets, load images, manipulate their position (RE/Dec/Rotation/distortion), and generate the appropriate index files to save the nebula image set.

The code could be be done in a separated plugin.

The new JSON format allow to create multiresolution images, as described there: http://groups.google.com/group/astronomex/web/multi-resolution-image-webservice

A more advanced but very useful feature would be to provide the interface to an astrometry.net service, which automatically adjust the images position.

Skills needed: C++, Qt, GUI Design, knowledge in astronomy
Difficulty: a self contained mini-project touching various disciplines

mariocannistra:
Hello @gzotti 
Just wanted to let you know that the link to the new JSON format gives a 404 error:
http://groups.google.com/group/astronomex/web/multi-resolution-image-webservice

I also tried applying to join the group but I got "error message #261 - reload" from the form :-(

gzotti:
Above text was copy-pasted from our previous blueprint at Launchpad. This is just to collect these ideas under this issue thread. Details of implementation will be left to whoever takes this challenge.

mariocannistra:
Ah thank you, got it.

gzotti:
In October 2019 we received a pretty good workflow description, now in the Stellarium User Guide.

axd1967:
see also https://github.com/Stellarium/stellarium/issues/3964#issuecomment-2453577206

[...]

PR #4003, opened 2024-12-11
[WIP] Add Custom Nebula Textures Plug-in.

ultrapre:
For amateur astronomers to use their own nebula images as textures.
By inputting photos and Astrometry.net's API to Plate-solve images online, and calculating the four corner coordinates, finally rendering textures of custom nebula images.

The project is work-in-progress, welcome guiding and joining.


plugins/NebulaTextures/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+# This is the cmake config file for the HelloStelModule
```
10110111, line 1: The placeholder module names should be replaced with the actual name.

ultrapre: Thanks.

ultrapre:
@10110111  @alex-w @gzotti Do you have any suggestions for conflicts between custom texture rendering and original texture rendering? Should we let users choose to turn off the original texture and then display the custom texture, or use some algorithm to detect and compare the texture area and then turn off the conflicting texture?

alex-w:
Please check cmake files in other plugins ;)

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.hpp
```diff
@@ -0,0 +1,82 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef NEBULATEXTURESDIALOG_HPP
+#define NEBULATEXTURESDIALOG_HPP
+
+#include "StelDialog.hpp"
+#include <QString>
+
+class Ui_nebulaTexturesDialog;
```
alex-w, line 26: Please change case for name of related file or change the name of class - please check similar parts in other plugins

ultrapre: Sorry but I can't understand what's the problem

gzotti:
I am not sure: Can we handle user catalogs independently? There is probably infrastructure for more, but just one "default" nebula catalog. Maybe users could make (manually) a copy of default and then replace whatever they want to replace. Or they just want to use this welcome plugin to then contribute their own images of so far not included DSOs. Given the very latest developments, you need this plugin to platesolve/fit the image with stars, and then apply some de-starring filter to extract the nebula alone. (Unless image is of a cluster of course.)

ultrapre:
It would be better to have separate user-defined json and image storage and management, and allow users to choose to hide conflicting original textures (by algorithm).

Another question is that the following formula still has a slight position deviation when calculating the four corners based on wcs fits. Does it also involve the calculation of parameters such as A_0_0, B_0_0, AP_0_0, BP_0_0? Maybe it is related to field of view distortion? I just started to get started, so it would be great if you could help me take a look.

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.hpp
ultrapre: ok it's done

gzotti:
Sorry, will be of no help here. Not my field. For fitting, you may want to disable aberration (to show mean positions). Small-angle (telescope) images should usually be OK, for large images it is not enough to have the corner coords, but need to take lens distortion into account.

ultrapre:
Perhaps another use is to allow users to analyze their own images and then locate the position in Stellarium, which would be a very practical function

alex-w:
From the log:
```
gui/NebulaTexturesDialog.cpp:226:70: error: call to member function 'arg' is ambiguous
   QString contentType = QString("multipart/form-data; boundary=%1").arg(boundary);
```

Compilation the qt5-based edition is broken

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
```diff
@@ -0,0 +1,490 @@
...
+   for (int i = 0; i < 19; ++i) {
+      boundary_key += QString::number(QRandomGenerator::global()->bounded(10));  // Generates a random digit (0-9)
+   }
+
+   // Format the boundary as '===============<random_number>=='
+   QString boundary1 = "'===============" + boundary_key + "==";
+   QByteArray boundary = boundary1.toUtf8();
+
+
+   QString contentType = QString("multipart/form-data; boundary=%1").arg(boundary);
```
10110111, line 226: What is this conversion to UTF-8 and back to QString supposed to do? Seems like an unnecessary roundtrip.

ultrapre:
wcslib works well for converting pixel to RA/Dec. Stellarsolver is also good for it. 
I plan to introduce one of them

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
10110111: This question is still relevant. The conversion to UTF-8 byte array and then back to `QString` is completely unnecessary. You can form the string as a `QString`, and only when passing it to `setHeader()` do the single conversion.

ultrapre: thanks，I got it

ultrapre:
Effects:
image
image

ultrapre:
@alex-w @gzotti @10110111 
Sorry, it seems that StelSkyLayerMgr::loadSkyImage is deprecated? The rendering result of the call is different from what I debugged using textures.json. The code of StelSkyLayerMgr and StelSkyImageTile is hard to read and understand for me. Can I ask for your help in the next development of the plugin?

gzotti:
I fully agree this is hard to read and follow. To me it appears even more features were planned but then not completed.

ultrapre:
Do you know how to disable rendering a selected image in the default texture?

gzotti:
Alexander invented some trick to hide M1 before 1054. Maybe you can go a similar way based on other conditions? Or you will have to add a scriptable function setVisible(QString imgName_or_JSONkey, bool visible). Sorry, I am not enough familiar in this section and would have to study the code again myself.

sunshuwei:
It is best to have a UI that can manage the opening or closing of each texture. 
Since the Crab Nebula can be triggered to close under certain conditions, I think perhaps every texture should be able to do the same.

ultrapre:
The display of M1 is not related to the method of manipulating the display separately, as this also uses the filtering criteria in the data

https://github.com/Stellarium/stellarium/blob/3c8d3c448f82848e9d8c1af307ec4cad20f2a9c0/nebulae/default/textures.json#L14

gzotti:
If there is no built-in way and you need it, you may need to add a key "visible" which should default to true. Then add the setVisible() command as outlined above. While you are exploring, everything about minResolution and probably other fields need better documentation IMO.

ultrapre:
Please review the code changes in core.
Does multi-language support only require q_ to wrap the strings that need to be translated?

alex-w:
> Does multi-language support only require q_ to wrap the strings that need to be translated?

Yes, it's enough

sunshuwei:
Could we consider adding an option in the future that is turned off by default, but can be turned on after consulting with the user, giving the image to Stellarium, using CC BY 3.0 or CC BY 4.0 copyright, and then the official can use these images as materials. It is even possible to develop a community in the future, similar to the Creative Workshop in Steam games, where users can freely choose mods provided by others.

ultrapre:
> Could we consider adding an option in the future that is turned off by default, but can be turned on after consulting with the user, giving the image to Stellarium, using CC BY 3.0 or CC BY 4.0 copyright, and then the official can use these images as materials. It is even possible to develop a community in the future, similar to the Creative Workshop in Steam games, where users can freely choose mods provided by others.


This requires server storage and a multimedia community, although uploading to a designated interface is easy to achieve. Anyway, such functionality can be envisioned in the future, as it is not currently required for this standalone software and the first version of the plugin.

sunshuwei:
> > Could we consider adding an option in the future that is turned off by default, but can be turned on after consulting with the user, giving the image to Stellarium, using CC BY 3.0 or CC BY 4.0 copyright, and then the official can use these images as materials. It is even possible to develop a community in the future, similar to the Creative Workshop in Steam games, where users can freely choose mods provided by others.
> 
> This requires server storage and a multimedia community, although uploading to a designated interface is easy to achieve. Anyway, such functionality can be envisioned in the future, as it is not currently required for this standalone software and the first version of the plugin.

Yes, this is not required for the first version.

gzotti:
The current DSO image collection is named "default". I don't know what would have to be done to allow others, i.e. customized lists that could be downloaded elsewhere and installed in a directory next to default. Not sure if it makes really sense to allow user-configuration at the single image level, though. Such collections need to be curated, and it will take a while until, say, 250 images are taken, processed, and configured correctly. I'd like to see alternative collections, though, maybe one collection of sketches through 6-12 inch instruments (to approximate what an observer can expect to see), one with late-19th-century illustrations, early-20th century b/w photos, watercolor and oil paintings, ...

sunshuwei:
> The current DSO image collection is named "default". I don't know what would have to be done to allow others, i.e. customized lists that could be downloaded elsewhere and installed in a directory next to default. Not sure if it makes really sense to allow user-configuration at the single image level, though. Such collections need to be curated, and it will take a while until, say, 250 images are taken, processed, and configured correctly. I'd like to see alternative collections, though, maybe one collection of sketches through 6-12 inch instruments (to approximate what an observer can expect to see), one with late-19th-century illustrations, early-20th century b/w photos, watercolor and oil paintings, ...
Yes, I have always hoped for at least two textures, photographic and visual (with a telescope). This can only rely on the user's ability, there are simple channels, and I think many users are willing to contribute their own strength. 
Visual texture may be a sketch or an approximation of black and white photo processing.
However, the analysis of sketches is difficult and coordinates cannot be obtained through "astrometriy.net".

sunshuwei:
Yes, I have always hoped for at least two textures, photographic and visual (with a telescope). This can only rely on the user's ability, there are simple channels, and I think many users are willing to contribute their own strength.
Visual texture may be a sketch or an approximation of black and white photo processing.
However, the analysis of sketches is difficult and coordinates cannot be obtained through "astrometriy.net".

ultrapre:
The current plugin has implemented this feature, allowing users to directly edit coordinates and render effects in real-time, thereby increasing the number of images that cannot be parsed.

image

ultrapre:
The current plugin has implemented this feature, allowing users to directly edit coordinates and render effects in real-time, thereby adding the number of images that cannot be parsed.

image

For example, [an excellent sketch](https://deepeye.hu/en/drawings/ngc3372.html)  even could be correctly solved and pasted here, but the projection accuracy requirements are too high, so there may still be deviations in the stars. Therefore, I would suggest that hand drawn images need to be preprocessed.

image

ultrapre:
> Yes, sure. I'd first fit (with stars), then remove the stars. Maybe the AI star remover also works here. (It could be wise to also keep "starred" images available. Even as yet another collection?)

For common nebulaes, slight deviations at the arcseconds level are irrelevant. I have been exposed to some deep space photography, and currently using Siri and StarNet++to remove star points is very convenient. However, this process requires a large amount of computing power, so it is obviously better for users to handle it themselves rather than integrating it into Stellarium.

gzotti:
Yes, I did not suggest adding a star remover to Stellarium! Just seeing the new star-free DSOs by @sunshuwei , sure that's an external (optional) process. Just that fitting/finding coordinates certainly is easier *with* stars, even if some are inaccurate.

sunshuwei:
This PR largely solves Issue #75

ultrapre:
Using a local solver for blind solving heavily post-processed deep-sky images doesn't seem to work well, so I am currently not introducing it. In [recommendations](https://nighttime-imaging.eu/docs/master/site/advanced/platesolving/), although the unofficial PlateSolve3-GUI can perform blind solving, it lacks a command-line blind solving invocation, and ASTAP doesn't seem to work for so many samples either.

@alex-w @10110111  The remaining work primarily involves testing the existing code to identify any potential issues. May I ask if you can start reviewing now? Also, is it sufficient to write the operational guide in guide.pdf and the plugin's About page?

alex-w:
> @alex-w @10110111 The remaining work primarily involves testing the existing code to identify any potential issues. May I ask if you can start reviewing now? Also, is it sufficient to write the operational guide in guide.pdf and the plugin's About page?

Yes, instructions (short for About page and detailed for User Guide) for users are required

10110111:
This PR fails to build on Qt 5.12, because `Qt::TextFormat::MarkdownText` was only added in 5.14. I'm not sure whether we want to still be on 5.12, given that this May even 5.15 will be officially discontinued by Qt. @alex-w ?

alex-w:
> This PR fails to build on Qt 5.12, because `Qt::TextFormat::MarkdownText` was only added in 5.14. I'm not sure whether we want to still be on 5.12, given that this May even 5.15 will be officially discontinued by Qt. @alex-w ?

Qt 5.15.2 has serious regression in lrelease/lupdate tools and we cannot compile Stellarium with localization support with Qt 5.15. Of course, dropping Qt5 support will caused problems for telescopes users and owners PC's with Windows 7.

ultrapre:
Ok I'll modify it again to be compatible

alex-w:
@ultrapre please fix issues from CodeFactor report

plugins/NebulaTextures/src/NebulaTextures.cpp
```diff
@@ -43,9 +41,11 @@ StelPluginInfo NebulaTexturesStelPluginInterface::getPluginInfo() const
 	StelPluginInfo info;
    info.id = "NebulaTextures";
    info.displayedName = "Nebula Textures";
-	info.authors = "Stellarium team";
-	info.contact = "www.stellarium.org";
-   info.description = "User Define Nebula Textures.";
+   info.authors = "WANG Siliang";
+   info.contact = "bd7jay@outlook.com";
+   info.description = "The plugin allows users to create and display their own astronomical sky images or even sketches in Stellarium, supporting both online plate solving for coordinate parsing or manual input of coordinates to localize the image, render it, and add it to the custom texture management system.";
```
alex-w, line 46: Please make description translatable (see other plugins for example)

ultrapre:
> @ultrapre please fix issues from CodeFactor report

The results in CodeFactor are a bit strange, as many of the suggestions, like the ones below, do not seem to correspond to any issues at the given line numbers.

>plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp#L316
>Redundant blank line at the end of a code block should be deleted. (whitespace/blank_line)

https://github.com/ultrapre/stellarium/blob/nebTexPlug/plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp#L316

ultrapre:
> @ultrapre please fix issues from CodeFactor report

ok now

gzotti:
> > This PR fails to build on Qt 5.12, because `Qt::TextFormat::MarkdownText` was only added in 5.14. I'm not sure whether we want to still be on 5.12, given that this May even 5.15 will be officially discontinued by Qt. @alex-w ?
> 
> Qt 5.15.2 has serious regression in lrelease/lupdate tools and we cannot compile Stellarium with localization support with Qt 5.15. Of course, dropping Qt5 support will caused problems for telescopes users and owners PC's with Windows 7.

@alex-w Which issues do you see with localisation on 5.15.2? 5.15.0 had the memory problem, but I saw no problems recently.  (does the CI fail?) Yes, in May Qt5 will be obsoleted by Qt, and in October Windows 10, and with it most older PC hardware (at least on the Windows side). As soon as we can adopt Alpaca with Qt6, we should also be able to leave Qt5 behind, maybe after 25.4.

alex-w:
> @alex-w Which issues do you see with localisation on 5.15.2? 5.15.0 had the memory problem, but I saw no problems recently. (does the CI fail?) 

Regression was fixed in Qt 5.15.3/5.15.4 and these version are not available as binary packages for Windows.

> Yes, in May Qt5 will be obsoleted by Qt, and in October Windows 10, and with it most older PC hardware (at least on the Windows side). As soon as we can adopt Alpaca with Qt6, we should also be able to leave Qt5 behind, maybe after 25.4.

I can enable building Stellarium with Qt 5.15.2 for tests. In any way right now we have 2 target groups for Qt5-based Stellarium - owners of telescopes and lecturers at planetariums.

gzotti:
What is different for planetarium use? The missing script pause, or more? I thought the finally working movies (with Qt6 multimedia) are a greater gain than the requirement to split up scripts which could be paused previously.

alex-w:
> What is different for planetarium use? The missing script pause, or more? I thought the finally working movies (with Qt6 multimedia) are a greater gain than the requirement to split up scripts which could be paused previously.

Well, legacy of old scripts :)  We may announce EOL for building Qt5-based binary packages after v25.4 of course

gzotti:
Just installed Qt5.15.2 and built on a 32GB Win10 PC (MSVC2019). Memory footprint while compiling was usually unremarkable, but then yes, again max 10GB with 8 cores when translating planetary-features. Are the CI systems memory limited? Maybe limit to 4 or even just parallel translations if needed?

alex-w:
> Just installed Qt5.15.2 and built on a 32GB Win10 PC (MSVC2019). Memory footprint while compiling was usually unremarkable, but then yes, again max 10GB with 8 cores when translating planetary-features (lconvert still misbehaves, but it's much better than in 5.15.0). 

Please see building [without translations](https://ci.appveyor.com/project/alex-w/stellarium/builds/51313756) and [with translations](https://ci.appveyor.com/project/alex-w/stellarium/builds/51314928) 

> Are the CI systems memory limited? Maybe limit to 4 or even just parallel translations if needed?

AppVeyor has limits - 2 cores and 6 GB RAM.

gzotti:
> P.S. macOS 10.13+ and Ubuntu Linux 20.04+ use Qt5-based packages too

Yes, but they don't have the legacy necessities around ANGLE and ASCOM. And Ubuntu 20.04 has EOL also in May.

alex-w:
Well, I cannot check the working of plugin because it causes a crash at startup in linux - catching an _ASSERT: "z->level==maxGeodesicGridLevel+1"_ in file `src/core/modules/StarMgr.cpp`, line 576

In any way I see a big problem for coding style and code formatting also :(

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.hpp
```diff
@@ -0,0 +1,126 @@
...
+public:
+	NebulaTexturesDialog();
+	~NebulaTexturesDialog() override;
+
+   StelSkyImageTile* get_aTile(QString key);
+
+public slots:
+	void retranslate() override;
+
+
```
gzotti, line 53: please add source documentation at least to the non-trivial functions. What does it do, what are the arguments, ...

gzotti:
> In any way I see a big problem for coding style and code formatting also :(

I don't see it now in the Github view. But @ultrapre did you use our code style file for QtCreator? If not, please install it and reformat as needed (wiki).

ultrapre:
> Well, I cannot check the working of plugin because it causes a crash at startup in linux - catching an _ASSERT: "z->level==maxGeodesicGridLevel+1"_ in file `src/core/modules/StarMgr.cpp`, line 576

Well what happened? I rarely change the core code, some change like move SkyLayerElem and allSkyLayers from private to public type in StelSkyLayerMgr.hpp. But I did not encounter any similar issues as yours during Windows Qt6 testing.


> I don't see it now in the Github view. But @ultrapre did you use our code style file for QtCreator? If not, please install it and reformat as needed (wiki).

Thanks I will check it

alex-w:
> Well what happened?

Please try clean building and run Stellarium without plug-ins data in User Directory

ultrapre:
> > Well what happened?
> 
> Please try clean building and run Stellarium without plug-ins data in User Directory

I only see assert like here: (Windows 11, Qt6)

ASSERT: "dateTime.timeSpec()==Qt::UTC" in file I:\stellarium\src\core\StelUtils.hpp, line 529

image

10110111:
> ASSERT: "dateTime.timeSpec()==Qt::UTC" in file I:\stellarium\src\core\StelUtils.hpp, line 529

You need to rebase to fix this.

ultrapre:
> Well, I cannot check the working of plugin because it causes a crash at startup in linux - catching an _ASSERT: "z->level==maxGeodesicGridLevel+1"_ in file `src/core/modules/StarMgr.cpp`, line 576好吧，我无法检查插件的工作情况，因为它会导致在 linux 中启动时崩溃 - 在文件 `src/core/modules/StarMgr.cpp` 第 576 行中捕获一个 _ASSERT：“z->level==maxGeodesicGridLevel+1”_

@alex-w Which Linux distribution is experiencing this issue? Ubuntu? I did not encounter this bug when compiling the merged code in Win11. I am trying to switch OS for testing again.

alex-w:
Yes, Ubuntu 20.04 at the moment

gzotti:
Startup on Win11 fails in NebulaTexturesDialog::reloadTextures(), first case. Probably you have something in your directories, but for somebody without first-run results, this does not work nicely.

ultrapre:
图片

I fix an overflow, it may be good now. (Ubuntu 22.04 + Qt6)

alex-w:
I tested the branch in Windows and at least it works, but not all textures are rendering (not checked a deep tough) :(

ultrapre:
> I tested the branch in Windows and at least it works, but not all textures are rendering (not checked a deep tough) :(

Is it that the image cannot be solved using astrometry, or is the coordinate information resolved but cannot be rendered?

ultrapre:
The plugin guide has been initially written.

guide/plg_interfaces.tex
```diff
@@ -872,6 +872,63 @@ \section{Solar System Editor Plugin}
...
+
+\subsection{Steps for Online Solving and Texture Management (For Astrophotography)}
+\begin{enumerate}
+    \item \textbf{Prepare an Image}: Use a processed rectangular bitmap file of a deep-sky object, such as PNG, JPG, TIF, or BMP format. It’s recommended to use a descriptive ID or nickname as the filename. While Nova Astrometry does not strictly limit image size, keeping it below 20 MP ensures efficient plate solving. Adjust the image background brightness to suit Stellarium by referencing the nebulae images in the Stellarium \file{nebulae/default} directory. Typical astrophotography images with TAN projections are fine, but avoid images with high field distortion.
+
+    \item \textbf{Open the Plugin}: By default, the plugin opens the “Solve and Add” page. Click the \button{Open File} button to select an image file or paste the file path into the "Image Filepath" text box.
```
alex-w, line 889: Wrong quotes for _Image Filepath_ text

alex-w:
> Is it that the image cannot be solved using astrometry, or is the coordinate information resolved but cannot be rendered?

It was troubles for rendering

alex-w:
OK, I've checked the plugin at linux now and it works without problem, but the GUI of plugin is need to improve and information in the log should be human readable :)

ultrapre:
> > Is it that the image cannot be solved using astrometry, or is the coordinate information resolved but cannot be rendered?
> 
> It was troubles for rendering

Ensure that the image is not mirrored!

ultrapre:
> OK, I've checked the plugin at linux now and it works without problem, but the GUI of plugin is need to improve and information in the log should be human readable :)

Suggestions for GUI? What log?

alex-w:
> What log?

Part of the log:
```
Creating scene FBO with size 2048x1152
"{\"status\": \"success\", \"message\": \"authenticated user: alex.v.wolf@gmail.com\", \"session\": \"9d7ikex3j7ah7jetchzo9wtsp1jujg0v\"}"
"{\"status\": \"success\", \"subid\": 11622137, \"hash\": \"4ee193603050ff8d8b1405c7c1e171c34773ce14\"}"
"{\"user\": 7267, \"processing_started\": \"None\", \"processing_finished\": \"None\", \"user_images\": [], \"images\": [], \"jobs\": [], \"job_calibrations\": []}"
"{\"user\": 7267, \"processing_started\": \"2025-01-26 12:15:12.317675\", \"processing_finished\": \"2025-01-26 12:15:13.819688\", \"user_images\": [11875730], \"images\": [29048028], \"jobs\": [null], \"job_calibrations\": []}"
"{\"user\": 7267, \"processing_started\": \"2025-01-26 12:15:12.317675\", \"processing_finished\": \"2025-01-26 12:15:13.819688\", \"user_images\": [11875730], \"images\": [29048028], \"jobs\": [null], \"job_calibrations\": []}"
"{\"user\": 7267, \"processing_started\": \"2025-01-26 12:15:12.317675\", \"processing_finished\": \"2025-01-26 12:15:13.819688\", \"user_images\": [11875730], \"images\": [29048028], \"jobs\": [12450540], \"job_calibrations\": []}"
"288.49131767"
288.491
"14.4912103439"
14.4912
"800.5"
800.5
"534"
534
"-5.34221405892E-06"
-5.34221e-06
"-0.0002923011779"
-0.000292301
"0.0002923011779"
0.000292301
"-5.34221405892E-06"
-5.34221e-06
"1600"
1600
"1067"
1067
Initializing planets GL shaders... 
```

ultrapre:
# NebulaTextures Plugin Test Report Outline

## 1. Functional Testing
- **Follow guide steps**: ✅ Passed  
- **Different datasets (Astrophotography/Hand-drawn images)**: ✅ Passed  
- **Polar images**: ✅ Algorithm validated  
- **Cross-equatorial and cross-meridian images**: ✅ Algorithm validated  
- **File management**: ✅ No invalid images left  
- **Restore default**: ✅ Passed  

## 2. Negative & Boundary Testing
### 2.1 Invalid Operation Steps
### 2.2 Empty Input & Length Limits
- **Invalid file path**  : ✅ Handled 
- **Multi-language filepath support**: ✅ Supported  
- **Invalid API key**: ✅ Login request handled  
- **Invalid coordinate format**: ✅ Bounded within `-90~90`, `0~360`  

### 2.3 File-related Issues
- **Invalid file type** ✅ (Basic extension filtering, actual format verified via network parsing)  
- **Invalid image** ✅ (Determined via network parsing)  
- **File size limit**: ❓ *No restriction set*  
- **Image mirroring**: ❌ *No image editing functionality, unhandled*  

### 2.4 Network Response Errors
- **Disconnecting network during parsing**: ✅ *Retry limit & cancellation implemented*  

## 3. Performance Testing
### 3.1 Frequent Uploads
- ✅ *UI restriction applied*  
### 3.2 UI Responsiveness
- ✅ **Rapid button clicks**  

## 4. Compatibility Testing
### 4.1 Cross-Platform Testing
- ✅ **Windows 11 21H2**, Qt 6.8.1 MSVC2022 64-bit  

- ✅ **Ubuntu 24.10-desktop-amd64**, Qt latest 

- ✅ **MacOS 14.0**, Qt 6.9.0 Desktop 

image


### 4.2 Installation Testing
- ❌ *Not conducted*  

## 5. Security Testing
- ✅ **API key visibility settings**  
- **Configuration stored in plaintext**: ❌ *No security measures applied*  

## 6. Regression Testing
- ✅ *No issues after rebase*

ultrapre:
Changelog

- [x] GUI changes readable  
- [x] Log output format fix 
- [x] Restore functionality  
- [x] Corrections based on the test checklist  
- [x] Brightness control  
- [x] Double-click centering  
- [x] Cancel solving function 
- [ ] Multi-system testing  


GUI Form

image

image

ultrapre:
> While you are exploring, everything about minResolution and probably other fields need better documentation IMO.

The minResolution at the top of textures.json controls the visibility and FOV relationship of all textures (equivalent to sub textures) within it. Therefore, the minimum resolution of each image does not work.

sunshuwei:
I adjusted the minResolution values using two adjacent images on the celestial sphere, both textures being 1024x1024. TextureA.png was set to 0.02 and textureB.png to 1.0. When zooming in on the screen, textureB.png would appear first. If we swap the minResolution values - setting textureA.png to 1.0 and textureB.png to 0.02 - then textureA.png would appear first instead.

gzotti:
Impressive work! 
Also thanks for documentation. I guess it makes chapter 8.2 obsolete. 

A few usage points: 
- You should probably make a menu button with a clear icon. Going over config dialog/plugins/config each time is cumbersome. 
- I tried with two images. After solving, the lower "Rendering debugger" could be more intuitive: 
-- just one checkbox could be used to "render"/"stop render". Or is there a reason for the two buttons?
-- The Bright/Normal/Dark should emit a signal with new setting, so that a redraw at the correct brightness is initiated. 
-- I don't know what "Disable default textures" should do. The built-in DSO texture is still visible. 

Ah- this is active only after restart of Stellarium! Also the option so the second tab. You could add tooltips or a label to tell the user that newly added images need a program restart to become fully usable. 

Does the brightness control set the maxBrightness entry in the json?

gzotti:
A tiny fix for the User Guide:
image

ultrapre:
> * You should probably make a menu button with a clear icon. Going over config dialog/plugins/config each time is cumbersome.

For Showing only?

> * I tried with two images. After solving, the lower "Rendering debugger" could be more intuitive:
>   -- just one checkbox could be used to "render"/"stop render". Or is there a reason for the two buttons?
>   -- The Bright/Normal/Dark should emit a signal with new setting, so that a redraw at the correct brightness is initiated.
>   -- I don't know what "Disable default textures" should do. The built-in DSO texture is still visible.
> 
> Ah- this is active only after restart of Stellarium! Also the option so the second tab. You could add tooltips or a label to tell the user that newly added images need a program restart to become fully usable.

Good advise, all fixed.

> Does the brightness control set the maxBrightness entry in the json?

Yes.

gzotti:
> > * You should probably make a menu button with a clear icon. Going over config dialog/plugins/config each time is cumbersome.
> 
> For Showing only?

Yes. Most plugin dialogs can be called up via bottom-menu button. Click once to activate, another time to hide. Usually these plugin draw some extras. A separate dialog (usually the respective PluginConfig dialog) can be called via right-clicking. Your plugin also has a drawing aspect and a GUI aspect. Actually, the menu button could toggle "show my own images" while right-clicking could toggle the editor dialog.   

>>...
> > Does the brightness control set the maxBrightness entry in the json?
> 
> Yes.

While I was testing, the underlying (existing) image could not be switched off, so images visually merged but I saw "dark" added some resolution. After restart I could toggle visibilities of the default and my own image, however I had set it to "dark", and all by itself it was too dim now. You  could add a section to the User Guide about the created JSON file in modules/NebulaTextures, where the maxBrightness value (or others, minResolution, ...) could be manually adjusted to peoples' needs. At this point you may know more about these numbers than the current maintainers. Documenting helps not forgetting!  

It is not necessary (but would be just awesome...:) to manipulate the displayed brightness interactively in the GUI. I know, there are always more ideas than time, but some grim ugly rainy winter days can be also spent in useful ways... :-)

ultrapre:
> While I was testing, the underlying (existing) image could not be switched off, so images visually merged but I saw "dark" added some resolution. After restart I could toggle visibilities of the default and my own image, however I had set it to "dark", and all by itself it was too dim now. You could add a section to the User Guide about the created JSON file in modules/NebulaTextures, where the maxBrightness value (or others, minResolution, ...) could be manually adjusted to peoples' needs. At this point you may know more about these numbers than the current maintainers. Documenting helps not forgetting!
> 
> It is not necessary (but would be just awesome...:) to manipulate the displayed brightness interactively in the GUI. I know, there are always more ideas than time, but some grim ugly rainy winter days can be also spent in useful ways... :-)

The existing underlying texture makes it difficult for users to accurately judge brightness, so I plan to set "Disable Default Texture" as checked by default.

The three values in the brightness dropdown are set based on my experience, and they should be sufficient for most texture scenarios. Would allowing users to freely adjust an empirically determined value—one that is inherently difficult to fine-tune—actually make the results less reliable?

Of course, providing a description in the user manual is necessary.

plugins/NebulaTextures/src/NebulaTextures.cpp
```diff
@@ -51,11 +53,12 @@ StelPluginInfo NebulaTexturesStelPluginInterface::getPluginInfo() const
 
 NebulaTextures::NebulaTextures()
 {
-	countRefresh = 0;
-	maxCountRefresh = 2;
 	setObjectName("NebulaTextures");
 	font.setPixelSize(25);
 	configDialog = new NebulaTexturesDialog();
+	flagShowNebulaTextures = false;
```
alex-w, line 59: The variable is not defined

ultrapre: thanks

ultrapre:
Change:

> You should probably make a menu button with a clear icon. Going over config dialog/plugins/config each time is cumbersome.

Added an "Enter" button

used signal mechanisms to refresh the texture drawing during the operation to control region conflicts

completed the disabling of the default texture option during texture testing

gzotti:
Tried another photo. 
I had previously pressed "remember me" for the Astrometry.net APIkey. The entry box was empty, though. I think it should appear filled with the stored APIkey (hidden chars) and checked box "remember me". 
Unfortunately solving failed. Not sure if the service was down or overloaded? Log message is just 

>[NebulaTextures] subid "0"
[NebulaTextures] Failed to get submission status.
[NebulaTextures] Failed to get submission status.
[NebulaTextures] Failed to get submission status.
[NebulaTextures] Failed to get submission status.

Of course I filled the API key as copied from astrometry.net.

ultrapre:
> > [NebulaTextures] subid "0"

The subid should not be "0". This is because you did not set a valid API key.

gzotti:
OK, this works again, thanks. Now, the result was not perfect, and I tried to tweak the corner coordinates. 
- The left and right are inverted. (RA goes from right to left!) 
- Changing the coordinates (with MouseWheel) does nothing on screen. I have to "Stop test" and "Test this texture" to redraw at the new position or with new corners. The texture coordinate spinboxes emit a "valueChanged" signal which you could receive and redraw immediately to make the changing an interactive experience. 
- If user messes around too much, a "Reset to solution" could then place the new image where astrometry.net had estimated.
-  You can utilize our derived GUI class, AngleSpinBox, for more user-friendly coordinates. (And you can make DisplayFormat selectable for everybody's taste.)

alex-w:
@ultrapre is it possible to fix Georg's suggestions up to end of current week?

ultrapre:
> @ultrapre is it possible to fix Georg's suggestions up to end of current week?

Okay I'll find time to finish it as soon, sorry for a little bit busy lately = =

alex-w:
> Okay I'll find time to finish it as soon, sorry for a little bit busy lately = =

No problem, it just need planning the version of Stellarium to including of the plugin - 25.1 or 25.2

alex-w:
@ultrapre any news?

ultrapre:
> @ultrapre any news?

Soon, sorry!

ultrapre:
Reply and new work:

> * The left and right are inverted. (RA goes from right to left!)

Of course, celestial image follows left-east and right-west, so called *Right Ascension*.

> * Changing the coordinates (with MouseWheel) does nothing on screen. I have to "Stop test" and "Test this texture" to redraw at the new position or with new corners. The texture coordinate spinboxes emit a "valueChanged" signal which you could receive and redraw immediately to make the changing an interactive experience.

Good, the new commit approves it.

> * If user messes around too much, a "Reset to solution" could then place the new image where astrometry.net had estimated.

Good, the new commit approves it.

> * You can utilize our derived GUI class, AngleSpinBox, for more user-friendly coordinates. (And you can make DisplayFormat selectable for everybody's taste.)

However, AngleSpinBox seems to be a SpinBox-based for Int but not Double.

gzotti:
The AngleSpinbox can do both, depending on its format settings. The practical thing is you can set the mouse cursor in the minutes or seconds field and use the wheel to adjust values. Try it e.g. in the location panel.

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
```diff
@@ -0,0 +1,1419 @@
...
+	double cosphi = std::cos(dphi * D2R);
+
+	// Calculate celestial longitude
+	double x = sinthe4 - costhe3 * cosphi;
+	double y = -costhe * sinphi;
+	double dlng = R2D * std::atan2(y, x);
+	double lng = CRVAL1 + dlng;
+
+	// Normalize the celestial longitude
+	if (CRVAL1 >= 0.0) {
```
gzotti, line 518: You apply the fmod(lng, 360) below and make sure it's [0...360]. These lines seem redundant.

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
```diff
@@ -0,0 +1,1419 @@
...
+		if (lng < 0.0) {
+			lng += 360.0;
+		}
+	} else {
+		if (lng > 0.0) {
+			lng -= 360.0;
+		}
+	}
+
+	lng = std::fmod(lng, 360.0);
```
gzotti, line 528: you could apply `lng=StelUtils::fmodpos(lng, 360.0);`

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
```diff
@@ -0,0 +1,1419 @@
...
+ */
+void NebulaTexturesDialog::avoidConflict()
+{
+	if (getAvoidAreaConflict()){
+		StelSkyImageTile* cusTile = get_aTile(CUSTOM_TEXNAME);
+		StelSkyImageTile* defTile = get_aTile(DEFAULT_TEXNAME);
+		if(!cusTile) return;
+		if(!defTile) return;
+
+		for (int i = 0; i < defTile->getSubTiles().size(); ++i) {
```
gzotti, line 1251: Here and elsewhere, please adhere to our coding style for braces, these usually go on new lines. 

In the else clause below,  if it's just one statement, it can go without braces.

plugins/NebulaTextures/src/SkyCoords.cpp
```diff
@@ -0,0 +1,212 @@
...
+	double sinphi = std::sin(dphi * D2R);
+	double cosphi = std::cos(dphi * D2R);
+
+	// Calculate celestial longitude
+	double x = sinthe4 - costhe3 * cosphi;
+	double y = -costhe * sinphi;
+	double dlng = R2D * std::atan2(y, x);
+	double lng = CRVAL1 + dlng;
+
+	// Normalize the celestial longitude
```
gzotti, line 108: I meant that this, line 109-117 can go away as it's redundant with the following mod operation. And with fmodpos used in line 119, you can delete lines 120-122 as well.

ultrapre:
Can I submit a paper of NebulaTextures-Plugin to JOSS (Journal of Open Source Software) ? Would you have any suggestions?

gzotti:
Of course! As reference paper about Stellarium, please cite our 2021 JSA paper. (See homepage for link and reference.) You may need to explain celestial coordinates and basic concepts of "platesolving" to the OS community, or what is behind software trickery for the readers from the astronomical fields, it's always a balance to find. Or the experience of how to add something into an existing program.

gzotti:
And don't forget to update your plugin (PR) with a reference link to the paper once it has appeared.

alex-w (approved):
Thanks for the plugin!

gzotti:
Ah, I did not realize it's ready. @ultrapre is it done from your side?

ultrapre:
Okay Please wait some time I will fix it

ultrapre:
Sorry I click request for mistake.

plugins/NebulaTextures/src/gui/nebulaTexturesDialog.ui
```diff
@@ -109,7 +133,7 @@
               <item row="1" column="1">
                <widget class="QLineEdit" name="lineEditApiKey">
                 <property name="echoMode">
-                 <enum>QLineEdit::EchoMode::PasswordEchoOnEdit</enum>
+                 <enum>QLineEdit::Normal</enum>
```
alex-w, line 136: Oops… it’s my fault

ultrapre: So is it recommended to continue using password display?

ultrapre:
Thank you, and please feel free to continue improving the compactness of the GUI. Since I mostly use Qt Creator for editing, it might be difficult for me to produce code that fully matches Stellarium’s style.

gzotti:
The latest change looks like this:
image

The expanding vertical space between the upper spinners should be taken away again, also to keep the dialog size compact. 
While we disagree on the intuitive visual matching of the corners, it is not a stopper for me, but user response will go to you. (I expect mostly praise...) The rest seems Ok for me now.

ultrapre:
> While we disagree on the intuitive visual matching of the corners, it is not a stopper for me, but user response will go to you. (I expect mostly praise...)

I actually agree with the idea of making the UI intuitive, but what I need more are some suggestions and methods for implementation—especially on how to visually align the image or how to adjust the layout system of Qt widgets to more intuitively reflect the actual corner positions. However, I think this is not easy to achieve.

alex-w:
Please check the changes...

10110111:
"RA (Center)" and "Dec (Center)" seem to also have fixed sizes, which don't scale with font settings.

alex-w:
> "RA (Center)" and "Dec (Center)" seem to also have fixed sizes, which don't scale with font settings.

Hmm… I didn’t checked properties for these labels. In any way I’ll add tomorrow tooltips for RA/Dec labels and check theirs properties

10110111:
Just tried the functionality itself, and it appears to be stuck at "Checking job status (1/99)...". It would be good to have a way to restart this check query without cancelling the job, or at least to have a hyperlink where we could see what's wrong.

10110111:
The "ApiKey" text should instead be "API key" in the user-facing strings, particularly in the GUI (the translatable version in the `.cpp` and the placeholder one in `.ui`) and the SUG.

10110111:
### SUG

 - [x] SUG section name "Nebula Textures Plugin User Guide" should be simply "Nebula Textures Plugin" as its siblings.
 - [x] SUG: "so it is recommended to use the texture without stars (de-starring)" — the "de-starring" should rather be "de-star it" to be logically consistent with the preceding sentence.
 - [x] SUG: "For Windows, the path is typically" — it would be more useful to provide the exact path in the About section of the plugin's GUI and refer to it in the SUG.

### Working of the plugin

* Positioning appears to be worse than that of the nebulae included with Stellarium, e.g. `barnard150.png`. Maybe it's inevitable with the plate-solving method used, but it might be a bug.
* Changing the center coordinates doesn't affect the positioning. It would be useful if such an action shifted the image (affecting the corner coordinates), otherwise there's no way to shift it as a whole.
* It would be useful if there was a button to change FoV to twice the image size, in addition to centering on the image.
* Bright/Dark/Normal settings don't change anything for me when the image is zoomed in to be larger than the scene frame.

### Messages

 - [x] "Importing custom textures successfully" should start with "Imported".
 - [x] "WCS download complete. Processing..." and other messages with "WCS" should include some hint for the translator as to what WCS means.
 - [x] Messages like `q_("Solve failed: ") + msg` should instead be like `q_("Solve failed: %1").arg(msg)`, so that the translators can put the placeholder at the right spot for their language. To find these cases you can use the regex `q_\([^)]+\)\s*\+` for git-diff terminal output via `less`, or `q_([^)]\+)\s*+` for `grep`.
 Partially solved, but these remain:
```
+       emit solvingStatusUpdated(q_("Checking submission status") +QString(" (%1/%2)...").arg(retryCount + 1).arg(maxRetryCount));
+               emit failed(q_("Astrometry error.") +" " + json["error_message"].toString());
+       emit solvingStatusUpdated(q_("Checking job status") +QString(" (%1/%2)...").arg(retryCount + 1).arg(maxRetryCount));
```

### Icons

* Looks like the base of `btNebulaTextures-o{n,ff}.png` was copied from some existing icons. It would be best to have a vector source `icons.svg` file similar to the other plugins, see for example `plugins/EquationOfTime/resources/icons.svg` including the notes in its `documentation` layer.

### Coding style etc.

 - [x] `} else {` inside `NebulaTexturesDialog::toggleTempTexturePreview` should be split into `}\nelse\n{`.
 - [x] `if ( (mountMode==StelMovementMgr::MountEquinoxEquatorial) && (fabs(spinLat)> (0.9*M_PI_2)) )` inside `NebulaTexturesDialog::moveToCenterCoord`: too many unneeded parentheses here. Similarly near the end of `NebulaTexturesDialog::gotoSelectedItem`.
 - [x] `#define NT_CONFIG_PREFIX QString("NebulaTextures")` — such local macros shouldn't appear in headers. They should be moved to the implementation file.

10110111:
Also, sections 13.9.1 and 13.9.2 in the SUG have too long titles, the former even doesn't fit on a line in the Table of Contents. It may be better to limit them to e.g. "Online Solving and Texture Management" — removing the "steps for" and the parenthesized parts, and then adding a small paragraph that expresses that what follows are steps for the formerly parenthesized use case.

ultrapre:
Thank you for your thorough review and detailed feedback!

I have improved SUG, Messages, Coding Style.



> *Positioning appears to be worse than that of the nebulae included with Stellarium, e.g. barnard150.png. Maybe it's inevitable with the plate-solving method used, but it might be a bug.*

The plate-solving algorithm is working correctly, but amateur astrophotography images often contain slight projection distortions. In contrast, the default nebula textures integrated into Stellarium were manually calibrated to achieve more precise alignment.

> *Changing the center coordinates doesn't affect the positioning. It would be useful if such an action shifted the image (affecting the corner coordinates), otherwise there's no way to shift it as a whole.*

The center coordinate field was originally intended only for locating the texture in the sky, not for adjusting its position. I actually considered making it non-editable. The center coordinates derived from astrometric solving are typically the most accurate, and shifting them can interfere with previously fine-tuned corner positions. That said, I agree it’s a valid enhancement request. I’ve also thought about implementing other adjustment methods, such as directly modifying the rotation angle or applying scaling transformations. However, texture alignment is a very delicate process, so these tools are not meant to replace precise manual tuning. If the 25.2 release schedule is tight, I want to defer this feature to a future version to avoid introducing potential bugs.

> *It would be useful if there was a button to change FoV to twice the image size, in addition to centering on the image.*

This is also a good optional suggestion and could be introduced in a future patch, as it involves additional calculations and validation.

> *Bright/Dark/Normal settings don't change anything for me when the image is zoomed in to be larger than the scene frame.*

This is related to Stellarium’s rendering behavior — the visual effect usually becomes more noticeable after zooming in.

> Icons

I haven’t fully figured this out yet — I might need some further guidance from you.

10110111:
> The plate-solving algorithm is working correctly, but amateur astrophotography images often contain slight projection distortions. In contrast, the default nebula textures integrated into Stellarium were manually calibrated to achieve more precise alignment.

The texture I was talking about, `barnard150.png`, is one of the textures included with Stellarium, so I assume it was calibrated as you say. Then I'd expect it to be positioned precisely. But maybe it's just that the plate solver doesn't try to achieve very precise positioning and just tells where in the sky to look.

> The center coordinates derived from astrometric solving are typically the most accurate, and shifting them can interfere with previously fine-tuned corner positions.

This is OK for astrophotography, but for the sketches the lack of an ability to simply move the solid image is very inconvenient. But OK, I guess this use case can be considered underdeveloped in this PR and improved on in later ones.

> > Icons
> 
> I haven’t fully figured this out yet — I might need some further guidance from you.

What icons did you base yours on?

gzotti:
Not sure how the platesolving really works and what other result it provides, but I understand it currently just positions (with least-squares fit?), but does not provide hints about radial field distortions by flatteners etc.  So, the image is placed without warping. You can fit the corners, but I then see mismatch in stars closer to center. In some way this should be combined with @10110111 's lens distortion estimator plugin, and the image should then actually be reprojected into a new texture. But this may be far beyond this PR now.

ultrapre:
> What icons did you base yours on?

[btNebulaeBackground-on.png](https://github.com/Stellarium/stellarium/blob/master/data/gui/btNebulaeBackground-on.png)

> So, the image is placed without warping. You can fit the corners, but I then see mismatch in stars closer to center.

This is because Stellarium’s texture rendering mechanism only supports mapping rectangular images using the coordinates of the four corners. However, real astrophotographs often contain optical distortions. In fact, the underlying computation of astrometry is highly precise and supports solving for high-order distortions through WCS (World Coordinate System) data. This allows accurate calculation of celestial coordinates for every pixel in the image. The resulting parameters can be as complex as high-degree polynomial equations.

10110111:
> > What icons did you base yours on?
> 
> [btNebulaeBackground-on.png](https://github.com/Stellarium/stellarium/blob/master/data/gui/btNebulaeBackground-on.png)

Aha, so the source is in `data/gui/icons.svg`. You can copy-paste the corresponding objects and layers from there into a new file (possibly by removing all extras from the original and saving-as a new file), apply your modifications, and save it in the source tree. Then export the two PNG icons as explained in the original source's "documentation" layer.

ultrapre:
@10110111  use what editor?

10110111:
> use what editor?

I use Inkscape.

alex-w:
@10110111 @gzotti @ultrapre any news?

alex-w:
I have question: the currently RA fields has values in DMS, but standard one is HMS, so, should we change the GUI according to standard units?

ultrapre:
Oops, right now I'm exploring how to make SVG icons, but using Inkscape is a bit difficult... Is the purpose of creating this file to preserve the editing space, or...?

ultrapre:
> I have question: the currently RA fields has values in DMS, but standard one is HMS, so, should we change the GUI according to standard units?

I'm afraid this might reduce the precision and cause inconsistent adjustment feedback between RA and Dec.

gzotti:
> > I have question: the currently RA fields has values in DMS, but standard one is HMS, so, should we change the GUI according to standard units?
> 
> I'm afraid this might reduce the precision and cause inconsistent adjustment feedback between RA and Dec.

Isn't this just an AngleSpinBox setting issue? Precision should be enough. But what are users of astrometry.net used to getting, degrees or hours?

alex-w:
> I'm afraid this might reduce the precision and cause inconsistent adjustment feedback between RA and Dec.

Why? The precision of output data is configurable for AngleSpinBoxes

ultrapre:
You are right, I fixed that. Well I mean the adjustment feedback is not the same, we concern coordinates in the plugin when we need to adjust it by scolling, but steps speed is different (x15 times), ~maybe it is not important

ultrapre:
I redesigned the entry Icon and the SVG file. The above discussion should have been responded to. 
As for what kind of elements are suitable to add to the original icon, I thought about ∞, +, ⚙, 🔍, etc. Meaning, aesthetics and placement space are all issues that need to be considered. Finally, I made a decision. I hope this plugin can let users see an infinite creative space.

10110111:
How did you export the PNGs? I don't find shadows or glows in the SVG file, while they are present in the PNGs. Same for boundaries. And the infinity sign stays separate from the `icons` layer...

ultrapre:
> How did you export the PNGs? I don't find shadows or glows in the SVG file, while they are present in the PNGs. Same for boundaries. And the infinity sign stays separate from the `icons` layer...

Sorry, I need to clarify that I was not able to figure out how to use inkscape to directly create the final effect and export this icon, so it also went through the editing process of AI, PS, etc, and the intermediate file is not the svg.  Can any technical support intervene?

10110111:
Have you seen the "documentation" layer in `data/gui/icons.svg`? It's quite detailed. If you still can't, I suppose I'll have to fix it up afterwards.

ultrapre:
> I don't find shadows or glows in the SVG

When I attempt to separate target from icons.svg, they losts. Selecting and deleting a lot of unnecessary elements in inkscape is difficult. I think I need some time to start with other software

10110111:
> Selecting and deleting a lot of unnecessary elements in inkscape is difficult.

Just tried, I can easily select a bunch of elements by rectangle selection and then press <kbd>Del</kbd> to delete them. Just make sure that you've unlocked the necessary layers before this.

10110111:
Anyway, this seems a minor issue to stop from going further with this PR. I'll fix this after merge. Just need to make sure that all my review comments have been addressed.

10110111:
I've updated [my review comment](https://github.com/Stellarium/stellarium/pull/4003#issuecomment-2926407575) with checkboxes that are unchecked next to the points that I haven't seen addressed. Please take care of them.

ultrapre:
> I've updated https://github.com/Stellarium/stellarium/pull/4003#issuecomment-2926407575 with checkboxes that are unchecked next to the points that I haven't seen addressed. Please take care of them.

Thanks, the previous submission missed some changes and are now added along with the new suggestions.

10110111:
> Thanks, the previous submission missed some changes and are now added along with the new suggestions.

Thanks, I've added a couple of comments to your commit. When you fix them, I think this PR will be OK for merging.

guide/plg_interfaces.tex
```diff
@@ -962,7 +963,7 @@ \subsection{Notes}
     \item After adding a texture, you must restart Stellarium to fully integrate the texture into the 
 	module. The same applies when deleting a texture.
     \item To back up your texture library or share it with others, copy the plugin folder from your user 
-	config directory. For Windows, the path is typically:
+	config directory. You can find the exact plugin storage path in the “About” section of the plugin GUI. For Windows systems, this path is usually something like:
```
alex-w: Please use TeX style for quotes

guide/plg_interfaces.tex
```diff
@@ -898,7 +899,7 @@ \subsection{Steps for Online Solving and Texture Management (For Astrophotograph
 	button to select an image file or paste the file path into the “Image Filepath” text box.
 
     \item \textbf{Enter API Key}: Paste your Astrometry API key into the corresponding text box (password 
-	mode by default). If you don’t have one, click \href{https://nova.astrometry.net/api_help}{\textcolor{blue}{\underline{Astrometry API key}}} to obtain it 
+	mode by default). If you don’t have one, click \href{https://nova.astrometry.net/api_help}{\textcolor{blue}{\underline{Astrometry ApiKey}}} to obtain it
```
10110111: You broke the correct phrase here.

plugins/NebulaTextures/src/gui/NebulaTexturesDialog.cpp
```diff
@@ -116,7 +116,7 @@ void NebulaTexturesDialog::createDialogContent()
 	ui->checkBoxShow->setChecked(getShowCustomTextures());
 	ui->checkBoxAvoid->setChecked(getAvoidAreaConflict());
 	ui->cancelButton->setVisible(false);
-	ui->label_apiKey->setText(QString("<a href=\"https://nova.astrometry.net/api_help\">") + q_("Astrometry API key:") + "</a>");
+	ui->label_apiKey->setText(QString("<a href=\"https://nova.astrometry.net/api_help\">Astrometry API key:</a>"));
```
10110111: Oh, this was good before. The text was simply inside a tag. Sorry for having listed this as problematic.

10110111:
Not sure why my comments don't appear here, I've re-posted them in the review format.

guide/guide.tex
```diff
@@ -242,6 +242,7 @@ \part{Extending Stellarium}
 \include{plg_scenery3d} % move to historical?
 \include{plg_atTheTelescope} % TelescopeControl, Oculars, various hardware projects with Raspberry, Arduino, ...
 \include{ch_scripting}
+\include{plg_nebulatextures}
```
alex-w: The file of paragraph plg_nebulatexture is not exist and should not be included here

alex-w:
@ultrapre thanks for the new plugin! Please assign issues #75 and #3170

PR #4031, opened 2024-12-29
Scale bottom and left panel buttons according to font size

10110111:
### Description

This is a continuation of the scalable GUI series. Currently the dialogs are scaled according to font size, but not the panels. This PR makes the panels, namely bottom and left one also scale. The Oculars plugin's panel (are there others?) remains unaffected, this will be fixed in a subsequent PR.

#### 13px font, old

stellarium-011

#### 13px font, new

stellarium-014

#### 19px font, old

stellarium-012

#### 19px font, new

stellarium-015

#### 40px font, old

stellarium-013

#### 40px font, new

stellarium-016

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/gui/SkyGui.cpp
+++ b/src/gui/SkyGui.cpp
@@ -386,6 +386,7 @@ void SkyGui::updateBarsPos()
 	progressBarMgr->setZValue(400);	
 
 	// Update position of the auto-hide buttons
+	btHorizAutoHide->setPos(1,btVertAutoHide->getButtonPixmapHeight()-btHorizAutoHide->getButtonPixmapHeight()+1);
 	autoHidebts->setPos(0, hh-autoHidebts->childrenBoundingRect().height()+1);
 	double opacity = qMax(animLeftBarTimeLine->currentValue(), animBottomBarTimeLine->currentValue());
 	autoHidebts->setOpacity(qMax(0.01, opacity));	// Work around a qt bug
--- a/src/gui/StelGuiItems.cpp
+++ b/src/gui/StelGuiItems.cpp
@@ -55,6 +55,16 @@
 #include <QSettings>
 #include <QGuiApplication>
 
+namespace
+{
+
+constexpr double DEFAULT_FONT_SIZE = 13;
+
+double fontSizeRatio()
+{
+	return StelApp::getInstance().getScreenFontSize() / DEFAULT_FONT_SIZE;
+}
+
 void brightenImage(QImage &img, float factor)
 {
 	for (int y=0; y<img.height(); y++)
@@ -78,6 +88,7 @@ void brightenImage(QImage &img, float factor)
 		}
 }
 
+}
 
 void StelButton::initCtor(const QPixmap& apixOn,
 						  const QPixmap& apixOff,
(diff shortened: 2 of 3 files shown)


alex-w:
Hmm… but you followed screen font size, not a GUI font size.

10110111:
Yes, it's consistent with the fact that panels follow it.

alex-w (approved):

gzotti:
The last example shows the danger, though - running out of screen space. Maybe thre should e an option to scale the buttons separate from screen font size. Yes, the bottom bar has a few issues still with placing the strings. IIRC these are currently placed on fixed locations, but those may depend on what's on their left side. (Long or short location name.) Only the time is right-adjusted. There may be sub-FullHD screens where users may accept smallish buttons mixed with their eyes' need for larger font. (Try on a 1280x720 window.) In this case maybe they may want to e.g. easily configure away FPS display to utilize the screenspace. 
It seems we need another tab "GUI" where we add yet a few more options, those now found in extras (activate various buttons) and tools (fonts, ...), and bottom menu options currently still only available by manual edits in config.ini.

alex-w:
> It seems we need another tab "GUI" where we add yet a few more options, those now found in extras (activate various buttons) and tools (fonts, ...), and bottom menu options currently still only available by manual edits in config.ini.

Good idea, but that’s for separate PR

10110111:
> The last example shows the danger, though - running out of screen space.

This danger already realizes on 800×600 screens with v24.4. And 1024×768 barely fits the bottom panel. Moreover, depending on the number of plugins enabled you can overflow it in 1024×768 case too.

> Maybe thre should e an option to scale the buttons separate from screen font size.

Maybe this could be added later, but currently the behavior is more or less the same as with high-DPI scaling, but with a saner handling of mouse clicks and positioning. (And this is the final goal: force Qt scale factor to 100% while scaling everything by the factor requested by the OS.)

> There may be sub-FullHD screens where users may accept smallish buttons mixed with their eyes' need for larger font.

If it's hard for someone to read text, it will be just as hard to discern the icons. Again, high-DPI scaling was introduced just because of this: one needs to enlarge everything, not just fonts.

Now, for those with very small screens there already exist many problems with the GUI of their DEs. I remember having such issues with some DE (I guess it was KDE?..) that had too high dialogs for my 1024×600 EEEPC 1015PN. But on Linux/X11 I could just move the dialog with <kbd>Alt</kbd>+leftMouseDrag and resize the window to fit contents with <kbd>Alt</kbd>+rightMouseDrag. This will work for Stellarium too.

I don't think anyone would use such screens with modern Windows though (and the old versions of it are gonna be obsoleted soon by our dropping of Qt5).

We could later add some support for, say, scrolling of the bottom panel, but it's not for this PR anyways.

gzotti:
Yes, on Windows 800x600 was no longer usable, therefore our min requirements (1024x768). On Ubuntu (18?) I found xrandr can scale down screens (SUG 2.1.1). And we introduced the configurable buttons to remedy the width problem. 
OK, I won't object this, but there remains room for improvement later.

PR #4046, opened 2025-01-01
Add Sh2-245 as Eridanus Loop

ultrapre:
A common name.

image

https://en.wikipedia.org/wiki/Orion%E2%80%93Eridanus_Superbubble

http://galaxymap.org/cat/view/sharpless/245


alex-w (approved):
Thanks!

PR #4054, opened 2025-01-05
AstroCalc: Add a filter by eclipse type

10110111:
### Description

This adds a way to select only particular types of eclipses to compute, e.g. only total, or only hybrid+annular. The UI might have been done better, but I'm not a UI designer :P

### Screenshots

stellarium-011

### Type of change

- [ ] Housekeeping

## Checklist:


alex-w:
Upper part of GUI in this tab is one for all subtypes and of course this filter is not applicable to lunar eclipses and to transits. Please move UI part of filter to bottom part of GUI.

10110111:
I'm retaining them at the same place, but now this block is used both for general and local solar eclipses (hiding "hybrid" checkbox for locals), and hidden for other pages.

alex-w:
What about “jumping” of time range block when subtabs are changing?

10110111:
> What about “jumping” of time range block when subtabs are changing?

Do you think it's critical? The tabs are mostly independent, aren't they?

10110111:
I've now extended it to lunar eclipses. So I think this placement in a common location is a good thing, and the jumpiness is not critical. Maybe just the aesthetics of the GUI are a bit suboptimal.

10110111:
Maybe this layout will be better:

stellarium-013

alex-w:
> Maybe this layout will be better:

What about placement of new block between name of table and time range block?

10110111:
> What about placement of new block between name of table and time range block?

IMO the filter by type is secondary to the time range, so the former shouldn't appear logically before the latter. Besides, I didn't like the two-checkbox-rows layout anyway, since it extends everything vertically in this HBox, leaving ugly space around. And putting the horizontal row in between will result in the name string being split into two lines (e.g. local eclipses title in Russian locale is quite long).
The new version seems coherent with both UI items being basically filters kept closely together.

alex-w (approved):
OK

gzotti:
I agree the new boxes should not go between table title and date range. But what about after? Then the first line reads "Table of xxx eclipses from year YYYY to the next NN years" (could also go in one continuous label, just keep bold as now) with selector boxes behind (after a stretch, right adjusted). Saves one vertical line. Or is the Russian name of local eclipses really too long for that?

alex-w:
> I agree the new boxes should not go between table title and date range. But what about after? Then the first line reads "Table of xxx eclipses from year YYYY to the next NN years" (could also go in one continuous label, just keep bold as now) with selector boxes behind (after a stretch, right adjusted). Saves one vertical line. Or is the Russian name of local eclipses really too long for that?

The current 2-line UI is good for translation to languages with "wide words"

alex-w:
Local eclipses:
stellarium-001

10110111:
> Local eclipses:

And to add to this, note that "Annular" is "Кольцеобразное", one more long word.

alex-w:
@10110111 please add ability to save/restore filter options in/from `config.ini` file

10110111:
> please add ability to save/restore filter options in/from config.ini file

If this is gonna be preserved, should it be per-page then? Or would a global state be sufficient?

gzotti:
Separate entries for partial/total lunar and solar please. And (as used elsewhere in AstroCalc) these settings should be stored immediately. On next call, the settings should appear as the last time user called the panel.

alex-w:
@10110111 did you have tested the saving/restoring feature? The config options look’s overcomplicated.

10110111:
> did you have tested the saving/restoring feature? The config options look’s overcomplicated.

Sure I have tested it. What exactly looks overcomplicated to you? We have 3 tabs, with 4,3,3 checkboxes. All have been requested to be kept independent, thus we get 10 independent options in the config file:
```ini
[astrocalc]
eclipse_filter/global_solar/annular_enabled = true
eclipse_filter/global_solar/hybrid_enabled  = true
eclipse_filter/global_solar/partial_enabled = true
eclipse_filter/global_solar/total_enabled   = true
eclipse_filter/local_solar/annular_enabled  = true
eclipse_filter/local_solar/partial_enabled  = true
eclipse_filter/local_solar/total_enabled    = true
eclipse_filter/lunar/partial_enabled        = true
eclipse_filter/lunar/penumbral_enabled      = true
eclipse_filter/lunar/total_enabled          = true
```

alex-w:
OK, thanks!

PR #4071, opened 2025-01-11
Scale lines according to screen font size

10110111:
### Description

Currently lines (constellations, DSO markers etc.) are scaled in thickness according to device pixel ratio. I'm continuing the work on basing this on font scale instead.

This PR uses a product of device pixel ratio and font size ratio to yield the final scaling coefficient of the lines (and dotted DSO markers). For this, a function called `screenScale` is introduced in both `ConstellationMgr.cpp` and `Nebula.cpp`. It's rather simple, but I'd rather not repeat its implementation everywhere, so I have a question: where is the best place for it? `StelApp`, or maybe `StelSkyDrawer`, or `StelPainter` or `StelProjector`? It seems `StelApp` is too general a place for an OpenGL-rendering-specific coefficient, but OTOH, after we get rid of the Qt scaling kludge, this function will be simplified to just `StelApp::screenFontSizeRatio()`. And `StelSkyDrawer` doesn't appear to have getters for such low-level factors. Not sure about `StelPainter` or `StelProjector` either.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/StelProjector.cpp
+++ b/src/core/StelProjector.cpp
@@ -21,6 +21,7 @@
 #include "StelTranslator.hpp"
 
 #include "StelProjector.hpp"
+#include "StelApp.hpp"
 
 #include <QDebug>
 #include <QString>
@@ -307,6 +308,13 @@ const SphericalCap& StelProjector::getBoundingCap() const
 float StelProjector::getPixelPerRadAtCenter() const
 {
 	return pixelPerRad;
+}
+
+float StelProjector::getScreenScale() const
+{
+	const float dppRatio = getDevicePixelsPerPixel();
+	const float fontRatio = StelApp::getInstance().screenFontSizeRatio();
+	return dppRatio * fontRatio;
 }
 
 //! Get the current FOV diameter in degrees
--- a/src/core/StelProjector.hpp
+++ b/src/core/StelProjector.hpp
@@ -222,6 +222,10 @@ public:
 	//! Get size of a radian in pixels at the center of the viewport disk
 	float getPixelPerRadAtCenter() const;
 
+	//! Get the scale factor for objects drawn on screen. This takes into account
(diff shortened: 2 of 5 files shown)


alex-w:
`StelProjector` of course - it has `getDevicePixelsPerPixel()` and `getPixelPerRadAtCenter()` already

10110111:
OK, this makes sense, since I already was passing an instance of `StelProjector` into the function. I just thought that it might be better for the function to not require construction of an instance of `StelProjector` and instead use a singleton like `StelApp`.

Anyway, I've updated the PR.

10110111:
With these changes, and a fix to make line strips & loops joined (in a miter joint style) I'm gonna push soon, I've been able to make the following version of the book cover. I'm sure Alex can make a better composition, and I forgot to enable constellation names, but notice how much crisper it looks compared to the current state at 100% scale. And when encoded in JPEG with 95% quality (via ImageMagick) it is even smaller in file size!

stellarium-011

gzotti:
The current link of screen font size and button size is a bit unbalanced IMO, buttons too large or bottom bar text too small. Is there a way to either fully decouple or add some relative scaling factor?

10110111:
> The current link of screen font size and button size is a bit unbalanced IMO, buttons too large or bottom bar text too small.

Do you mean you dislike the relation between the default sizes of 13px font and 32px buttons and want to use a different ratio? I suppose this is possible to implement by adding a factor into `screenFontSizeRatio()`.

gzotti:
Yes, please.

10110111:
OK, but this is unrelated to this PR.

alex-w (approved):
Thanks!

gzotti:
OK then, will be a new issue soon. Else no current objections.

PR #4092, opened 2025-01-21
Make all splash screen text scalable

10110111:
Currently the splash screen has rasterized branding text: `Stellarium` and `stellarium.org`. This text gets upscaled when scale factor or font size changes. It's better to have it rendered as text instead, which this PR does.

Later I intend to also provide a higher-resolution version of the background image.

Now, it looks as intended on Ubuntu 20.04, but I'm not completely sure how this will look on Windows, because `Sans` font will be different there. Please show me your screenshots from Windows before this could be mergeable.

This is how it looks on Ubuntu 20.04. To compare just try upscaling the current splash or run Stellarium with 26px GUI fonts or 200% scaling.

Screenshot_2025-01-21_13-28-15

Screenshot_2025-01-21_13-31-02


gzotti:
Will hopefully check during weekend. A few days ago there was another discussion involving the splash screen. When calling with --help, Stellarium dumps some info to stdout and exits. There is probably no need or use to show any graphics in this case. Should we suppress it or is there a benefit of a "flash screen"?

10110111:
> There is probably no need or use to show any graphics in this case. Should we suppress it or is there a benefit of a "flash screen"?

There's no longer a flash screen after 5564a42d8a56f7eaa010e0e67b51fb72c16743fe, because now the splash screen presented by `StelApp::init()`.

alex-w:
> What if you specify "Bitstream Vera Sans" explicitly?

stellarium-branch

10110111:
Please retry with the updated version.

gzotti:
Windows 11/Qt6.8:
image

alex-w (approved):

10110111:
> The look is much better, but why not using "DejaVu Sans" for text on splash?

Well, first of all, I didn't know that we distribute some fonts. Second, it appears that the alignment of the title and subtitle with DejaVu Sans is broken :\

BTW, there are several instances of DejaVu Sans in the tree, and all are different, is it intended?
```
-rw-rw-r-- 1 ruslan ruslan 757076 окт 24 18:36 ./data/DejaVuSans.ttf
-rw-rw-r-- 1 ruslan ruslan 720280 окт 24 18:36 ./plugins/RemoteControl/webroot/dejavusans-webfont.ttf
-rw-rw-r-- 1 ruslan ruslan 720012 окт 24 18:36 ./util/DSSToStellarium/DejaVuSans.ttf
```

10110111:
So, now it should be pixel-perfect in the sense that the rightmost pixels of all three text lines must be at the same column. This might be a bit suboptimal regarding very faint sides of vertical lines (where the better alignment would be by the neighboring bright column), but it's OK to me.

Here's how it looks on my Ubuntu 20.04:

Screenshot_2025-01-26_17-23-58

Screenshot_2025-01-26_17-23-50

If it looks similarly good on Windows and the Linux distro used by @alex-w, this can be merged.

alex-w:
This is may be strange, but what about some redesign the splash screen - version 25.1 will coming with serious changes and we can make a "Stellarium" text a bit less and add the version number in same string (e.g. "Stellarium 25.0" currently) + do a small shift of both lines (Stellarium & stellarium.org) to up.

@10110111 @gzotti @sushoff @henrysky @worachate001 your opinion?

alex-w:
> Well, first of all, I didn't know that we distribute some fonts. 

We doing it since version 10.1/10.2 (at least in 10.5 when I coming in the project the DejaVu fonts already was in distributive of Stellarium).

> Second, it appears that the alignment of the title and subtitle with DejaVu Sans is broken :\

This is sad, but probably our fonts is outdated now.

> BTW, there are several instances of DejaVu Sans in the tree, and all are different, is it intended?
> -rw-rw-r-- 1 ruslan ruslan 757076 окт 24 18:36 ./data/DejaVuSans.ttf

This is font which we distribute within Stellarium by default

> -rw-rw-r-- 1 ruslan ruslan 720280 окт 24 18:36 ./plugins/RemoteControl/webroot/dejavusans-webfont.ttf

We distribute it too, but for Remote Control plugin only (a font is using for compatibility of text rendering within WUI) - probably we should sync the font with main one.

> -rw-rw-r-- 1 ruslan ruslan 720012 окт 24 18:36 ./util/DSSToStellarium/DejaVuSans.ttf

This is font for utilite and he is outdated by the fact now and we do not distribute it within binary packages.

10110111:
> make a "Stellarium" text a bit less and add the version number in same string

I'm not sure it would look better. The current version placement looks nice to me, while simply having it follow the name would be somewhat boring, too standard. But anyway, this is something a designer should look at. From readability perspective we shouldn't lift the text too much up, otherwise it would be white on white.

alex-w:
> > make a "Stellarium" text a bit less and add the version number in same string
> 
> I'm not sure it would look better. The current version placement looks nice to me, while simply having it follow the name would be somewhat boring, too standard. But anyway, this is something a designer should look at. From readability perspective we shouldn't lift the text too much up, otherwise it would be white on white.

"Stellarium:25.0" :D ?

alex-w:
> If it looks similarly good on Windows and the Linux distro used by @alex-w, this can be merged.

The look of text is perfect for me

10110111:
>
> make a "Stellarium" text a bit less and add the version number in same
> string
>
> I'm not sure it would look better. The current version placement looks
> nice to me, while simply having it follow the name would be somewhat
> boring, too standard. But anyway, this is something a designer should look
> at. From readability perspective we shouldn't lift the text too much up,
> otherwise it would be white on white.
>
> "Stellarium:25.0" :-) ?
>
I don't like this, the colon doesn't make sense.

> @10110111 <https://github.com/10110111> is it possible to make a splash
> screen a grayscale in depending by config option? In this case we shouldn't
> be storing 2 images and we can avoid a cache issue.
>
What kind of cache? I wanted, to the contrary, to provide multiple images
optimally looking at different scales (dithered to avoid banding, and this
is lost on resampling). Otherwise we'd have to resize and dither a large
16-bpc source image at runtime, which may add a delay before the appearance
of the splash.

gzotti:
Win11/Qt6.8, 100% scaling:
image

Good for me.

alex-w:
So, the PR can be merge as is :)

10110111:
Oops, I updated the image to an experimental new version for this PR (accidentally committed after experimenting). It doesn't look too different, but it was intended as part of the next stage...

> Uh, I just see the commit number looks identical to above. I did of course checkout master, branch -D pr/4092, pull, checkout pr/4092. Cmake ran again, but apparently, sometimes commit numbers and even branch names are still not updated.

So well, does the actual new version look good for you?

gzotti:
TBH I did not see a difference or even reason to touch the ephemeral splash screen in the first place, so it's OK for me. 

The later comment about commit number not being updated by cmake is not related to the actual splashscreen topic. Maybe there is a git hook to execute after each checkout to recreate the relevant file(s)? That could be more reliable than cmake detecting little changes.

10110111:
> TBH I did not see a difference or even reason to touch the ephemeral splash screen in the first place

Well, I observe it very often while restarting Stellarium during development, so it does matter to me :)

> Maybe there is a git hook to execute after each checkout to recreate the relevant file(s)?

Ah, I see what you mean, it's only about the commit displayed. I also have been hit by this glitch. If we want to fix this, we could do similarly to the way I implemented in CalcMySky, [here](https://github.com/10110111/CalcMySky/blob/b66f57885db49f60183b96b025da4cf8a5dcfcec/CMakeLists.txt#L112) (see everything related to `version.cpp`). But beware of the drawback: you'll see "Linking CXX executable stellarium" on every change of a version, including commits, switches between dirty and clean state etc. Linking is fast for CalcMySky, but for Stellarium it takes about 11 seconds on my machine.

Issue #4043, opened 2024-12-30
The new star catalog should display more information


sunshuwei:
**Is your feature request related to a problem? Please describe.**
The new star catalog should display more information, such as Eta Carinae, which should show HR 4210 , HD 93308, SAO 238429，WDS 10451-5941 in addition to Gaia DR3 5350358584482202880.
Some important star names should also be added, such as α PsA B, α PsA C.
Similar examples include 12 Aur, HD 61563, HD 124953...
![微信图片_20241230125159](https://github.com/user-attachments/assets/9c662017-2b1e-4ba7-881a-31225f182491)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

alex-w:
Cross-id data is not updated yet

alex-w:
See also #633

sunshuwei:
> See also #633
HIP/Gaia should now be considered as a key reference, rather than just using HIP.

PR #4096, opened 2025-01-23
Added missing Bayer/Flamsteed designations for stars

alex-w:
### Description
This pull request added missing Bayer/Flamsteed designations for stars.

See #4043

### Type of change

- [X] Housekeeping


stars/hip_gaia3/name.fab
```diff
@@ -1,4943 +1,5108 @@
...
-5350358584482202880|η Car
+# HIP/Gaia3        |Bayer/Flamsteed designations
+# IAU constellation Andromeda (And)
+                677|α_And
+                841|22_And
+               5447|β_And
+               9640|γ1_And
+ 346231302441905920|γ2_And
+ 346231302441905920|γ_And_B
+ 346231302441905920|γ_And_C
```
gzotti, line 9: Is this really the same star, or can two stars share the same Gaia number? If component C is spatially unseparable from B, maybe a single BC entry is better.

alex-w: I think just a B will be enough here

gzotti:
I see many cases of binary component 2 or B listed as HIP and A/1 (expected to be the brighter or "main" component) as Gaia number, which is unexpected but may be OK, but I also have not checked the numbers :-). Sometimes a B component is now labeled without an explicit A component, cannot say if this is a flaw.  Else I see no other structural issues (like duplicate Gaia above). OK for me, occasional errors can be fixed later. Thanks for extending!

PR #4103, opened 2025-01-26
Make it possible to scale screen buttons independently

### LINKED ISSUE #4073: Relative scaling of buttons and screen text

Some recent GUI changes coupled button size and screen text size, but in a way which makes buttons needlessly large. 

I'd like to configure my screen with not too large buttons but still readable text. 

**Additional context**

Mentioned in PR #4071

10110111:
### Description

This adds an option in the Configuration dialog to apply a scale to screen (panel) buttons as requested in #4073.

Fixes #4073

### Type of change

- [ ] Housekeeping

## Checklist:

### DIFF
--- a/plugins/Oculars/src/Oculars.cpp
+++ b/plugins/Oculars/src/Oculars.cpp
@@ -189,6 +189,10 @@ Oculars::Oculars()
 	setObjectName("Oculars");
 	setFontSizeFromApp(StelApp::getInstance().getScreenFontSize());
 	connect(&StelApp::getInstance(), SIGNAL(screenFontSizeChanged(int)), this, SLOT(setFontSizeFromApp(int)));
+	connect(&StelApp::getInstance(), &StelApp::screenButtonScaleChanged, this,
+	        [this]{const int size = StelApp::getInstance().getScreenFontSize();
+	               setFontSizeFromApp(size); /* and repeat to apply all the geometry changes */
+	               setFontSizeFromApp(size);});
 
 	ccds = QList<CCD *>();
 	oculars = QList<Ocular *>();
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -450,6 +450,7 @@ void StelApp::init(QSettings* conf)
 
 	setScreenFontSize(confSettings->value("gui/screen_font_size", getDefaultGuiFontSize()).toInt());
 	setGuiFontSize(confSettings->value("gui/gui_font_size", getDefaultGuiFontSize()).toInt());
+	setScreenButtonScale(confSettings->value("gui/screen_button_scale", 100).toDouble());
 
 	SplashScreen::present(guiFontSizeRatio());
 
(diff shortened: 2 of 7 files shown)


gzotti:
Very good, thanks! I like 80%.
I think the GUI has enough space to fit the switch in the upper line: Font size: Screen [XX] GUI [YY] Buttons [ZZ%]. The (still optional?) Font switches could become a bit more narrow.
A question will turn up: What defines 100%? Is it just "designer's subjective preference" or anything that can be technically justified? (Like "width=X times size of 'M' letter" or such?)

10110111:
> I think the GUI has enough space to fit the switch in the upper line

I'm not sure it semantically fits the line of fonts, even though the fonts define the scale of all the screen objects (in addition to OS scaling factor).

> A question will turn up: What defines 100%? Is it just "designer's subjective preference"

I think the initial choice of 32px/56px buttons and 13px fonts that was made years ago was just the designer's (Fabien's?) subjective preference. The 100% scale preserves this ratio.

alex-w:
> What defines 100%

Standard size of UI elements

alex-w (approved):
Thanks

gzotti:
> > What defines 100%
> 
> Standard size of UI elements

Sure. But what makes it "standard"? "Original design choice (J.Meuris ~2006-08?) fitting to original choice of font size (13)" is OK-ish for me, understanding program history.

alex-w:
Oops... I think the new feature should have some limits - maybe 25%..250%

10110111:
> maybe 25%..250%

Currently it's 0..500%, with 500% being the limit of upscaling of the buttons. As for the lower limit, I guess it should be positive, but whether 25% is low enough I'm not sure. Maybe 10% would be OK...

alex-w:
> > maybe 25%..250%
> 
> Currently it's 0..500%, with 500% being the scale of the button images. As for the lower limit, I guess it should be positive, but whether 25% is low enough I'm not sure. Maybe 10% would be OK...

10% is acceptable, but definitely not 0%

gzotti:
Wouldn't it still be better understandable to have them e.g. 16...64 "pixels" wide? These may then be scaled pixels, but don't we use this scale for the fonts as well?

10110111:
> Wouldn't it still be better understandable to have them e.g. 16...64 "pixels" wide?

No, they are not at fixed pixel size.

> These may then be scaled pixels, but don't we use this scale for the fonts as well?

No, and no. Unless you're using the screen scaling kludge, that is. Without screen scaling the pixels of font size are real pixels, and the buttons are scaled according to the ratio between the current font size and the default one.

Moreover, even if we only ever had the OS-induced scaling and operated in fake pixels, this wouldn't be more understandable, as e.g. screen shooters, image editors, and I think also screen rulers would still operate in real pixels, so measuring size with these tools will only confuse the user.

The current solution is just "X% of the default", which IMO is quite sensible, as it would scale along with the UI without the confusion of the fake pixels.

alex-w:
The PR is OK for me

PR #4133, opened 2025-02-15
Update color of Neptune

10110111:
This patch makes the texture of Neptune look closer to that derived in the paper by P.G.J.Irwin et al. [1]. The texture is simply reweighted to get the same mean color as in the reference image while keeping the details similar in contrast.

The color in the .ini file is updated to the mean color of the reference image, scaled up to make one color channel reach 1.0.

Reference [1]: https://doi.org/10.1093/mnras/stad3761


alex-w (approved):

10110111:
Hmm, it seems I've made a mistake: I took Figure 2 as the reference, while I should've taken Figure 8 instead.

gzotti (approved):
OK for me. Let's not forget to add the reference to the SUG.

10110111:
I updated the color to match (on disc-average) the correct spectrum presented in figure 1.

Where should the reference go in the SUG? Should it me mentioned explicitly in some chapter, or just be present as an item in the bibliography list?

gzotti:
A bib entry goes into guide.bib, and reference should be mentioned at the end of H.4, with \newFeature{25.1} annotation.

10110111:
Is the current state OK to merge?

alex-w (approved):

PR #4140, opened 2025-02-17
Copy HTML version of the selection info

10110111:
In addition to current plain text being copied into clipboard on Ctrl+Shift+C, this patch also provides an HTML version, so that one could paste it e.g. into LibO Writer and get bold fonts, subscripts etc.

I'm not sure though that we want to keep the text color. It's very pale when pasted into LibO Writer with white background, but OTOH, I remember that copying from web browser also preserves the colors. We don't specify the font explicitly though, so it doesn't match after pasting. If we preserve the color, maybe we could wrap the HTML into one more `<font>` tag for font name.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/gui/SkyGui.cpp
+++ b/src/gui/SkyGui.cpp
@@ -99,15 +99,20 @@ void InfoPanel::setTextFromObjects(const QList<StelObjectP>& selected)
 		// just print details of the first item for now
 		// Must set lastRTS for currently selected object here...
 		StelCore *core=StelApp::getInstance().getCore();
-		QString s = selected[0]->getInfoString(core, infoTextFilters);
+		infoHTML = selected[0]->getInfoString(core, infoTextFilters);
 		selected[0]->removeExtraInfoStrings(StelObject::AllInfo);
-		setHtml(s);
+		setHtml(infoHTML);
 	}
 }
 
 const QString InfoPanel::getSelectedText(void) const
 {
 	return toPlainText();
+}
+
+QString InfoPanel::getSelectedHTML() const
+{
+	return infoHTML;
 }
 
 SkyGui::SkyGui(QGraphicsItem * parent)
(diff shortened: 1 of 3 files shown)


alex-w:
I’m not sure that this patch is necessary - why HTML formatted text is important for you?

10110111:
> why HTML formatted text is important for you?

Because there are subscripts that are lost on `toPlainText()`, which is not nice.

gzotti:
HTML sounds nice to keep titles&subscripts, yes. But I'm also not sure about the color necessity. For stars the color is derived from B-V, or a custom color can override the automatically derived ones, ... probably "current color and current font" is enough for the editor (sorry, is "LibO==LibreOffice?) Or do you fear glyphs may get lost if font name is not given?

alex-w:
I think the text should be black :)

alex-w (approved):

10110111:
> is "LibO==LibreOffice?

Yes.

> Or do you fear glyphs may get lost if font name is not given?

No, it's just a matter of completeness. Seems to make sense to preserve font if we do preserve colors. It's just a bit strange to see colored text that suddenly got serifs that weren't present in the Stellarium window.

> I think the text should be black :)

Ideally, the color shouldn't be defined if we decide not to preserve the colors, so that the editor's current color could take effect.

alex-w:
> Ideally, the color shouldn't be defined if we decide not to preserve the colors, so that the editor's current color could take effect.

Currently I see white text on white background - this is not very good combination for reading :)

10110111:
> Currently I see white text on white background - this is not very good combination for reading :)

Sure, because we are preserving the colors. If we decide not to, we shouldn't force it to be black either, because it's not our design choice then, we should simply omit the font color and let the editor use what it would if you type. In the current case we could just strip ` color="#......"` to achieve this.

alex-w:
Add stripping color data is good IMHO

PR #4145, opened 2025-02-20
Use commonNameMap instead of iterating over dsoArray

10110111:
The `search()` function is used when loading DSO outlines and discovery information. Its search for English names actually doesn't work, because it's called before any sky culture is loaded, which is when Nebula's English name is set. This doesn't appear to be a problem, because all the data we have are properly handled by a search by designation.

Now it's renamed to `searchForCommonName()`, and it uses `commonNameMap` instead of iterating over all the array of ~95k DSOs. This makes loading of outlines and discovery information almost instant, while before this patch it took about half a second to a second for each.

A small change in functionality is that now all the common names are searched, not only the first one per DSO, but it seems to be a useful change.

## Checklist:

### DIFF
--- a/src/core/modules/NebulaMgr.cpp
+++ b/src/core/modules/NebulaMgr.cpp
@@ -870,15 +870,12 @@ void NebulaMgr::drawPointer(const StelCore* core, StelPainter& sPainter)
 }
 
 // Search by name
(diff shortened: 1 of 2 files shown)


alex-w (approved):
Thanks!

PR #4154, opened 2025-02-26
Make SC converter deployable

10110111:
### Description

Since the SC converter appears to be useful for normal users, this PR makes it deployable. In Linux it's easy, since all the dependencies are available in the repos, but on Windows only the binaries of libgettextpo for x86_64 are easily available. So only this architecture (and thus only the Qt6 version) comes with the converter binary now.
On macOS I don't know how deployment works, I hope someone will advise me (@alex-w?). Currently the installation procedure is copied from the code for the `stellarium` target.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:


alex-w:
Is it really need to deploy as "inner" tool for Stellarium for each release?

10110111:
> Is it really need to deploy as "inner" tool for Stellarium for each release?

What do you mean by "inner"? The tool will evolve as I fix bugs, so each release will include an updated version.

gzotti:
Yes, that's important, at least for v25.1...25.4 or so. Some people have developed their private SCs which they will want to convert, and not all people are updating their installation with every version. This will have to be described for non-technical users (ethnologists, historians, astro enthusiasts) in the User Guide, section 9.3. If it's just a small/command-line tool, it can be packed with the distribution. If it's a larger thing, it could be offered as separate download, but this may make things needlessly complicated. Not many people will run it, but for some it will be essential.

alex-w (changes requested):
I don't see the patch for ISS scripts to distribute convertor within stellarium package for Windows

10110111:
> I don't see the patch for ISS scripts to distribute convertor within stellarium package for Windows

Shouldn't this line in `stellarium.iss.cmake:37` pick up all the DLLs?
```
Source: "@CMAKE_INSTALL_PREFIX@\bin\*.dll"; DestDir: "{app}"; Flags: ignoreversion
```

alex-w:
> > I don't see the patch for ISS scripts to distribute convertor within stellarium package for Windows
> 
> Shouldn't this line in `stellarium.iss.cmake:37` pick up all the DLLs?
> 
> ```
> Source: "@CMAKE_INSTALL_PREFIX@\bin\*.dll"; DestDir: "{app}"; Flags: ignoreversion
> ```

Yes, but where you see rule for `skyculture-convertor.exe`?

10110111:
> Yes, but where you see rule for skyculture-convertor.exe?

OK, this one I forgot. And actually, the DLLs also appear to not be present in the installed paackage...

alex-w:
> OK, this one I forgot. And actually, the DLLs also appear to not be present in the installed paackage...

The rule for DLLs is added for some specific cases of building the Stellarium

10110111:
It appears that I was copying the DLLs into the wrong directory. Regarding the EXE, is it OK if I just replace `stellarium.exe` with `*.exe`? Otherwise, if I have to specify `skyculture-converter.exe` explicitly, I'll need some kind of logic for the case then the converter is disabled.

alex-w:
I fear `ENABLE_SC_CONVERTER` should be equals `0` for AppImage and macOS packages due architecture specifics of these packages

10110111:
> architecture specifics

What specifics? How is `libgettextpo-0` different from, say, `libexiv2-27`?

alex-w:
> What specifics? How is `libgettextpo-0` different from, say, `libexiv2-27`?

No. _*.AppImage_ and _*.app_ are containers

10110111:
> No. *.AppImage and *.app are containers

So what? Why do we install `libexiv2-27` (see `stellarium-appimage-qt6.yml:57`) but can't do the same for `libgettextpo-0`?

alex-w:
> So what? Why do we install `libexiv2-27` (see `stellarium-appimage-qt6.yml:57`) but can't do the same for `libgettextpo-0`?

Well, you do not understand - you cannot call `skycultures-convertor` from `Stellarium.AppImage` package

alex-w:
I can propose move code into separate repo, add the GUI for tool and make a standalone package (with GUI it will be good for packages in macOS or for containers like an AppImage)

10110111:
> Well, you do not understand - you cannot call skycultures-converter from Stellarium.AppImage package

Gotcha. I'm always forgetting that just having dependencies in place is only part of the goal :)
So then, if @gzotti is OK with it, the converter will still remain a separate project, and will be built as a separate package.

> add the GUI for tool

Not sure it's worth the effort unless I create a very convenient cross-editor of the description and its translations. I'm currently not planning to do this :)

gzotti:
Yes, if that does not play nice with the architecture/concept of major distribution packages like AppImages, it's fine as separate download. My minimum idea was "available on Linux commandline", that would include WSL which also non-techs can probably extend their systems with. If you even build a GUI (conversion wizard?) around, I won't object. I expect just hundreds of users who will use it for a few runs of their previous work only, though (former SC authors), so yes, don't overdo it...

10110111:
OK, so the new repo for the converter is [stellarium-skyculture-converter](https://github.com/Stellarium/stellarium-skyculture-converter). Will try to devise something there.

sunshuwei:
> Actually, maybe we'll still want the Windows Qt6 x86_64 builds to contain the converter. I guess it should be easy enough to clone the repo and build it as part of the build process (maybe only when `[publish]` is given).
Maybe the converter could be available as a plug-in.

10110111:
How do I connect the new repo under Stellarium user to AppVeyor, so that it could be built in its own environment, rather than rely on Stellarium's one?
I've tried installing AppVeyor to this repo, but got a server error. The current status is like this:

Screenshot_2025-03-05_20-08-13

alex-w:
> So, I've pushed a working AppVeyor YAML to the `master` branch of this project, but to actually start building there, rather than in my fork, someone (@alex-w?) needs to enable AppVeyor integration there.

Hmm... https://ci.appveyor.com/project/alex-w/stellarium-skyculture-converter/builds/51643991

10110111:
So well, now this branch contains only the change to make the converter deployable on Windows. Let's go step by step.

10110111:
Hmm, apparently this PR isn't reopenable. Will create a new one...

10110111:
Could you @gzotti and @alex-w please check that the current `skyculture-converter.exe` supplied with the latest snapshot works as expected in Windows? I'm wondering whether the console output is actually visible.

10110111:
And generally, I'd like to have the converter tested by more than just me before we release Stellarium 25.1, to prevent problems when the users are suddenly required to switch to the new format.

alex-w:
> Could you @gzotti and @alex-w please check that the current `skyculture-converter.exe` supplied with the latest snapshot works as expected in Windows? I'm wondering whether the console output is actually visible.

Yes, the console output is visible

alex-w:
> And generally, I'd like to have the converter tested by more than just me before we release Stellarium 25.1, to prevent problems when the users are suddenly required to switch to the new format.

I don't have difficult skycultures for converting...

gzotti:
the SUG users needs section 9.3. Anything non-obvious to remark here? Will try later today.

PR #4156, opened 2025-02-27
Update star catalogs and cross-id with astroquery 0.4.9

### LINKED ISSUE #4124: Some stellar magnitudes wrong
### Expected Behaviour

All stellar magnitudes according to HIP/Gaia, or a better source if available.

### Actual Behaviour

This Issue is open to collect entries to be fixed before releases.

It seems these are off: (add your finds here!)

- &beta; Cen (Hadar) given as 1.96mag, but variable 0.61...0.65? Contradicting entries!

### System
* Stellarium version: master as of 2025-02-09
* Operating system: Win11
* Graphics Card: irrelevant
### Logfile 
not needed

henrysky:
This PR updates the star catalogs generated with astroquery 0.4.9 released last month (see henrysky/stellarium_star_catalogs#2). Astroquery 0.4.9 has a better method to query binary star systems so components for HIP star systems should be more complete. The rest of the updated star catalogs are here: https://www.astro.utoronto.ca/~hleung/shared/stellarium_new_catalog/2025_feb_26/

``cross-id.dat`` is generated as a part of the process to generate the star catalog now. Compared to the existing ``cross-id.dat``, about additional 79 HIP id & 689 SAO id & 62206 HD id get cross-identifiers.

Fixes #4124

Also fixes a handful of stars with exactly 7.50 mag (e.g., SAO 145064) are missing from the current star catalogs.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: MacOS 15
* Graphics Card: Apple M3

## Checklist:


alex-w (approved):
Thank you very much!

10110111:
> `cross-id.dat` is generated as a part of the process to generate the star catalog now.

Is there some script that does this based on the catalog files? Also, is it possible to switch to a binary format instead of the current textual one? It takes 550ms on my machine to load it, and most of the time is spent in parsing integers from string.

henrysky:
> Is there some script that does this based on the catalog files? Also, is it possible to switch to a binary format instead of the current textual one? It takes 550ms on my machine to load it, and most of the time is spent in parsing integers from string.

Not based on the catalog files (since the catalog files do not have SAO/HD/DR ids). Basically what happen is I query all possible HIP ids from SIMBAD, and then parse the result to build the catalog files, as well as the cross-id file. I can sure switch to a binary format.

alex-w:
> Also, is it possible to switch to a binary format instead of the current textual one? It takes 550ms on my machine to load it, and most of the time is spent in parsing integers from string.

I'll add patch to use packed version of cross-id data today

alex-w:
> Basically what happen is I query all possible HIP ids from SIMBAD, and then parse the result to build the catalog files, as well as the cross-id file. I can sure switch to a binary format.

Oops... I planned to packing data through command `gzip -nc cross-id.dat > cross-id.cat` and unpack it in the C++ code (similar to processing the `gcvs.cat` data at the moment). Do you want to use other way to create and use binary format for cross-id data?

10110111:
> Do you want to use other way to create and use binary format for cross-id data?

Of course, unpacking will only increase loading time :) My idea was to just have the numbers stored as normal little-endian 64-bit (or 32-bit when applicable) values and simply read them directly instead of parsing from string.

PR #4157, opened 2025-02-27
Using English terms for English names of constellations

alex-w:
### Description
This patch do switches to using English terms for English names of constellations for Modern and Modern (IAU) sky cultures

### Type of change

- [x] Housekeeping

### DIFF
--- a/skycultures/modern_iau/index.json
+++ b/skycultures/modern_iau/index.json
@@ -7,442 +7,442 @@
     {
       "id": "CON modern_iau And",
       "lines": [[677, 3092, 5447, 9640], [113726, 116631, 116805, 116584], [116631, 2912, 3092], [2912, 5447, 4436, 3881, 5434, 7607], [3092, 3031, 3693, 4463]],
-      "common_name": {"english": "Chained Maiden", "native": "Andromeda"}
+      "common_name": {"english": "Chained Maiden", "native": "Andromeda", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Ant",
       "lines": [[53502, 51172, 46515]],
-      "common_name": {"english": "Air Pump", "native": "Antlia"}
+      "common_name": {"english": "Air Pump", "native": "Antlia", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Aps",
       "lines": [[72370, 81065], [80047, 81852, 81065]],
-      "common_name": {"english": "Bird of Paradise", "native": "Apus"}
+      "common_name": {"english": "Bird of Paradise", "native": "Apus", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Aql",
       "lines": [[98036, 97649, 97278, 95501, 93805, 95501, 93747, 95501, 97804, 99473], [93244, 93747], [93805, 93429], [99473, 96468, 93805, 93747]],
-      "common_name": {"english": "Eagle", "native": "Aquila"}
+      "common_name": {"english": "Eagle", "native": "Aquila", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Aqr",
       "lines": [[102618, 106278, 109074, 110395, 110960, 111497, 110960, 110672, 109074], [109139, 106278, 109074, 110003, 112961, 114724, 115033, 115438, 115033, 114341, 115033, 113136, 112716, 112961]],
-      "common_name": {"english": "Water Bearer", "native": "Aquarius"}
+      "common_name": {"english": "Water Bearer", "native": "Aquarius", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Ara",
       "lines": [[85267, 85727, 82363, 83081, 83153, 85792, 88714, 85792, 85258]],
-      "common_name": {"english": "Altar", "native": "Ara"}
+      "common_name": {"english": "Altar", "native": "Ara", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Ari",
       "lines": [[8832, 8903, 9884, 13209]],
-      "common_name": {"english": "Ram", "native": "Aries"}
+      "common_name": {"english": "Ram", "native": "Aries", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Aur",
       "lines": [[25428, 23015, 23767, 24608, 28360, 28380, 25428], [23767, 23453, 23416, 24608, 28358, 28360]],
-      "common_name": {"english": "Charioteer", "native": "Auriga"}
+      "common_name": {"english": "Charioteer", "native": "Auriga", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Boo",
       "lines": [[69673, 72105, 74666, 73555, 71075, 71053, 69673, 67927, 67275], [69673, 71795], [71075, 69732, 70497, 69483, 69732]],
-      "common_name": {"english": "Herdsman", "native": "Boötes"}
+      "common_name": {"english": "Herdsman", "native": "Boötes", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau CMa",
       "lines": [[30324, 32349, 34444, 33579, 34444, 35904], [30324, 31592, 33152, 33579], [32349, 33347, 34045, 33160, 33347]],
-      "common_name": {"english": "Great Dog", "native": "Canis Major"}
+      "common_name": {"english": "Great Dog", "native": "Canis Major", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau CMi",
       "lines": [[37279, 36188]],
-      "common_name": {"english": "Lesser Dog", "native": "Canis Minor"}
+      "common_name": {"english": "Lesser Dog", "native": "Canis Minor", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau CVn",
       "lines": [[63125, 61317]],
-      "common_name": {"english": "Hunting Dogs", "native": "Canes Venatici"}
+      "common_name": {"english": "Hunting Dogs", "native": "Canes Venatici", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Cae",
       "lines": [[23595, 21861, 21770, 21060]],
-      "common_name": {"english": "Engraving Tool", "native": "Caelum"}
+      "common_name": {"english": "Engraving Tool", "native": "Caelum", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Cam",
       "lines": [[23040, 23522, 22783, 29997, 33694, 29997, 22783, 17959, 17884, 16228]],
-      "common_name": {"english": "Giraffe", "native": "Camelopardalis"}
+      "common_name": {"english": "Giraffe", "native": "Camelopardalis", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Cap",
       "lines": [[100064, 100345, 102485, 102978, 105881, 106723, 107556, 106985, 104139, 100064]],
-      "common_name": {"english": "Sea Goat", "native": "Capricornus"}
+      "common_name": {"english": "Sea Goat", "native": "Capricornus", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Car",
       "lines": [[30438, 45238, 50099, 52419, 51576, 50371, 45556, 42913], [51576, 53253], [52419, 54301, 54751, 54463, 53253], [45556, 41037, 38827, 39953]],
-      "common_name": {"english": "Keel", "native": "Carina"}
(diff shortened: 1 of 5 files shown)


10110111:
In the current state of the patch, won't we get these descriptive names on screen by default, confusing all the users at least in the Russian locale?

Also, if we later decide to translate the native names, we need to maintain some kind of compatibility with the current SCs in the SC repo used by Stellarium Mobile (ideally, with no need to rewrite SM to support our decisions). Particularly, there `"native"` contains actually _native_ names, not their romanized versions, e.g.:
```
    # Chinese SC in the SC repo
    "M 7": [{"english": "Fish", "native": "鱼", "pronounce": "Yú"}],
```

alex-w:
> In the current state of the patch, won't we get these descriptive names on screen by default, confusing all the users at least in the Russian locale?

Well, we need support for context for differentiation similar cases (example: `"context": "IAU constellation name"`)

> Also, if we later decide to translate the native names, we need to maintain some kind of compatibility with the current SCs in the SC repo used by Stellarium Mobile (ideally, with no need to rewrite SM to support our decisions). Particularly, there `"native"` contains actually _native_ names, not their romanized versions, e.g.:
> 
> ```
>     # Chinese SC in the SC repo
>     "M 7": [{"english": "Fish", "native": "鱼", "pronounce": "Yú"}],
> ```

This is next step - native names should be native (as in stellarium-skycultures) and `transliteration` field for romanized names

alex-w:
Note: the `stellarium-skycultures` repo has English terms for Western SC (a Modern in our terms), but S-M has separate POT file for each SC and of course we do not see troubles for this SC in S-M

alex-w:
@10110111 please see line 158 in  `ConstellationMgr.cpp` file - is it a joke? The next step, see `ConstellationMgr::updateI18n()` - by the fact context is used for trying translation of constellation names.

gzotti:
> This is next step - native names should be native (as in stellarium-skycultures) and `transliteration` field for romanized names

Yes, native is not translatable original spelling and may use non-Latin glyphs. `transliteration` (now called `pronunciation` but IMO as explained earlier this is sub-optimal) shall be initiated with Latin-based glyphs (with diacritics as needed) but be adaptable to user language, i.e. be translatable. For reasons of compatibility with S-M the tag name could still remain "pronunciation", but documented as somewhat inaccurate. Given that not all transliteration systems were developed as pronunciation aid for people only familiar with Latin characters, we should offer 'IPA' for pronunciation, and use it as optional feature.  

Once the tag names are known, I could develop the per-SC label building. (Was rather planned for 25.2, but with this going so fast now, first changes may come also here sooner than anticipated.)

10110111:
> please see line 158 in ConstellationMgr.cpp file - is it a joke?

If you mean the check for `"en"`, see `update_po.py` in the SC repo. The logic is a bit unintuitive indeed.

> The next step, see ConstellationMgr::updateI18n() - by the fact context is used for trying translation of constellation names.

Right, this is a mistake.

10110111:
> For reasons of compatibility with S-M the tag name could still remain "pronunciation", but documented as somewhat inaccurate.

I think the purpose of the tag is exactly pronunciation aid. Fabien's only wish is that "it should be easy to read for most people".

IMO, IPA is not easier than pinyin for people not familiar with any of these systems, so pinyin is just as good for Chinese terms. It could be an optional item though, if a creator wants to provide both a common romanization like pinyin and IPA.

In S-M `pronounce` appears in the "Also known as" list, see the screenshot with Chinese SC enabled:

Screenshot_2025-02-27_18-31-17

gzotti:
> I think the purpose of the tag is exactly pronunciation aid. Fabien's only wish is that "it should be easy to read for most people".
> 
> IMO, IPA is not easier than pinyin for people not familiar with any of these systems, so pinyin is just as good for Chinese terms. It could be an optional item though, if a creator wants to provide both a common romanization like pinyin and IPA.
> 
> In S-M `pronounce` appears in the "Also known as" list, see the screenshot with Chinese SC enabled:

I never proposed to replace pinyin by IPA. Pinyin is perfect as starting content of a "transliteration" for Chinese words (many people know and use it), but probably not for more remote languages with their unique character sets, like Sanskrit, for which other transliteration schemes surely exist. Most likely, any language with script in non-Roman glyphs meanwhile has one or more accepted transliteration systems to allow communication with the rest of the world. We can only start with one most compatible to use in our default program language, English.  IPA is an additional, standardized way of providing pronunciation aids, and, once learned, should be applicable to all languages without first learning all the other systems, which is why this would also be a non-translatable and optional item. We can hope that these IPA data can later (whenever) be converted to audio by voice synthesis, if needed.  On the other hand, transliteration _is_ dependent on user language and _should_ therefore be in principle translatable. Voice synthesis (VS) will then have to include "VS for pinyin, VS for Sanskrit-transliterated-to-their-favourite-system [sorry, I don't know the details], ..." The earlier example was (sorry, I can't type the glyphs now) Greek "Lyra" transliterated for Latin-char users into Latin-1 "Lyra", but transliterated for Russian-language users into Cyrillic "Lyra" and probably Indian, Chinese or Korean-character versions for those users.   But in fact, not all transliterations were even made for pronunciation, see the Wylie-Tibetan system which as I understand was made to be reversible, at cost of unintuitive character use that does not represent pronunciation. Therefore I expressed my concern about the accuracy of the tag name "pronunciation" when what we show is more accurately a "transliteration". But if S-M has this tag name already in use and it is hard to change, it can be kept. I would just not use the tag name as screen label for the user, and document this inaccuracy in the SC spec docs.

10110111:
But do we even want a transliteration that doesn't provide a pronunciation aid? The `pronounce` tag was intended to be such an aid, regardless of the system used to transcribe the names. If you're unhappy with some of our current transcriptions and consider them to just be transliterations, you should be welcome to make them better.

10110111:
> So, the pronounce field is used for transliteration/romanization by the fact in S-M and we should re-use it in S-D.

Fabien says it's OK to have IPA there as well, the name should just be easily enough readable. So it can be transcription as well.

gzotti:
If (real) researchers in a field where a non-pronunciative transliteration is widely used for international communication and publication (see example) describe a high-quality skyculture,  they will likely use the system best-established as Latin-letters style of writing in their field. To help "ordinary" people, other systems may appear better to them. I just propose to provide an optional IPA element for the sake of having one internationally, globally accepted standard, accept that for the sake of compatibility the current element "pronunciation" may have to keep its name, but still say that not all strings that will end up there help with actual pronunciation. But all strings which end up there provide a way of spelling in the character shapes best known to users of the currently set language. Often, but not always, reading by users of that language may produce something related to the original pronunciation of natives in the native language. And especially in these cases, the IPA content is certainly helpful.

10110111:
Do any of our current `pronounce` tags give a romanization that doesn't help pronunciation?

gzotti:
Ah, our thoughts crossed...  So, can we change the tag name to 'transliteration' then, for all software branches? And will it be translatable? The IPA is optional, but I'd like to have it to get an approximation of e.g. Vietnamese multi-stacked diacritical letters for which Europeans usually will not need a transliteration.

gzotti:
> Do any of our current pronounce tags give a romanization that doesn't help pronunciation?

Most, but not all, transliterations start as pronunciation aid for adherents of the "contact people" ("discoverers", missionaries, ...) and their own language. AFAICS, all currently in use look probably acceptable as pronunciation aid for people reading with English background. The German "translation" of some transliterations will in some points diverge (e.g., "cha" --> "tscha"), if not the German-speaking user has just negligently accepted to read the English version by now. And again, some languages/cultures have not been touched yet and may bring their own idiosyncrasies.  A hypothetical contribution of more Tibetan will most likely provide a common transliteration that was not developed to help pronunciation. The Rey shows the original mixed-capitalisation spelling from the book which invites applying such form in their translation to other languages where capitalisation is available.

10110111:
> Ah, our thoughts crossed... So, can we change the tag name to 'transliteration' then, for all software branches?

Not really. The field was meant to be a pronunciation aid. Particularly, it's given in extracted comments for translators as a pronunciation of a name, e.g.:
```
#. Chinese constellation, native: 参宿, pronounce: Shēnxiù
msgid "Three Stars"
msgstr "Три звезды"
```
I suppose we can have `transliteration` tag for the cases when the transliteration isn't helpful for reading, and `IPA` as a tag specially for IPA. I think the `IPA` data could be fetched by S-M as a fallback for the case when `pronounce` is not present—or simply be ignored, as 毕宿 is "known as" Bìxiù but isn't "known as" [pî.ɕjôʊ̯]. But if a transliteration is also useful for pronunciation, it should be in `pronounce` tag. Both `transliteration` and `pronounce` could be translatable, while `IPA` wouldn't be.

alex-w:
> In the current state of the patch, won't we get these descriptive names on screen by default, confusing all the users at least in the Russian locale?

OK, I've restored ability to use context for constellation names and I'll use official Russian constellation names as translation for English terms for constellation names. It should help us to avoid possible confusing the users (I hope translators will work good).

10110111:
> OK, I've restored ability to use context for constellation names

This is good in general, and I think this commit can be cherry-picked straight into `master`.

> I'll use official Russian constellation names as translation for English terms for constellation names.

So from your side it's OK. But will the other translators follow this?..

> I hope translators will work good

They are the weakest part of this situation, given that `msgid` says one thing, and `msgstr` should be quite unrelated.

alex-w:
> > I'll use official Russian constellation names as translation for English terms for constellation names.
> 
> So from your side it's OK. But will the other translators follow this?..
> 
> > I hope translators will work good
> 
> They are the weakest part of this situation, given that `msgid` says one thing, and `msgstr` should be quite unrelated.

This is a very good example of localization task :)

10110111:
> This is a very good example of localization task :)

Well, localization is more than just translation, and requires a higher expertise, which open-source projects can rarely afford (and many modern commercial projects couldn't care less about this). Remember that we have translators who translate `<a href="` as `<a href=„` — would you trust them such localization decisions?

alex-w:
> > This is a very good example of localization task :)
> 
> Well, localization is more than just translation, and requires a higher expertise, which open-source projects can rarely afford (and many modern commercial projects couldn't care less about this). Remember that we have translators who translate `<a href="` as `<a href=„` — would you trust them such localization decisions?

Let's see the real work

alex-w:
@10110111 @gzotti @sushoff is patch OK for you?

skycultures/modern_iau/index.json
```diff
@@ -7,442 +7,442 @@
...
       "id": "CON modern_iau Col",
       "lines": [[26634, 27628], [25859, 26634], [28328, 27628, 28199, 30277]],
-      "common_name": {"english": "Columba", "native": "Columba"}
+      "common_name": {"english": "Dove", "native": "Columba", "context": "IAU constellation name"}
     },
     {
       "id": "CON modern_iau Com",
       "lines": [[64241, 64241, 64394, 60742]],
-      "common_name": {"english": "Coma Berenices", "native": "Coma Berenices"}
+      "common_name": {"english": "Bernice's Hair", "native": "Coma Berenices", "context": "IAU constellation name"}
```
gzotti, line 130: Is it really "Bernice", not "Berenice" in English?

alex-w: Yeah, see https://www.iau.org/public/themes/constellations/

10110111:
I personally think it's not a good change. I'm pretty sure some translators
will mess it up. The mismatch between the extracted comments, context and
msgid is too big and confusing IMO.

gzotti (approved):
That means, the "context" is an untranslatable item, used to annotate .po files and get the correct translations later.
 
I Guess it's a start, to construct the per-SC multi-variant label selector. "native (translated)" for most of us probably.

10110111:
Oh, forgot to note: the `context` tag is your own creation, it's not supported anywhere yet :) Particularly, the newly-extracted lines don't have `msgctxt`.

alex-w:
> Oh, forgot to note: the `context` tag is your own creation, it's not supported anywhere yet :) Particularly, the newly-extracted lines don't have `msgctxt`.

Why? See [here](https://github.com/Stellarium/stellarium/commit/03bf0fd38bf31a943bb4e7d568264a876fe7f156#diff-f2dfc1caed51badc411d66f16d371016dbd926cd44b8bbd3ed8976850b9b446fR55916). I see some troubles for comments only at the moment.

10110111:
> Why? See [here](https://github.com/Stellarium/stellarium/commit/03bf0fd38bf31a943bb4e7d568264a876fe7f156#diff-f2dfc1caed51badc411d66f16d371016dbd926cd44b8bbd3ed8976850b9b446fR55916). I see some troubles for comments only at the moment.

Oops, my bad. Need to go to sleep. The comments are indeed strangely multiplied.

10110111:
> I see some troubles for comments only at the moment.

Well, I spent quite a bit of time to realize that they had been from the beginning, it's just that you had removed the duplicates before this PR. The duplicates are not a trouble in themselves: they represent many constellations in Modern-Rey SC that have no difference in names. Each of these constellations gets an extracted comment, so it's OK, in a way.

PR #4170, opened 2025-03-02
Use hips_frame and type to determine planet

10110111:
### Description

The current way of determining planet—by reading the filename of survey's URL—is quite hacky and doesn't let one find other survey components related to the same planet, e.g. the normal map. At the same time, all the planet HiPS currently available have properties that determine them as planetary, which planet they are for, and what type of map they represent.

So this PR uses these properties to determine planet.

## Checklist:

### DIFF
--- a/src/core/StelHips.hpp
+++ b/src/core/StelHips.hpp
@@ -60,8 +60,10 @@ public:
 							   const QVector<uint16_t>& indices)> DrawCallback;
 	//! Create a new HipsSurvey from its url.
 	//! @param url The location of the survey.
+	//! @param frame The reference frame from the survey's \c hips_frame property.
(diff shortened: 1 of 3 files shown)


src/core/StelHips.hpp
```diff
@@ -61,7 +61,7 @@ class HipsSurvey : public QObject
 	//! Create a new HipsSurvey from its url.
 	//! @param url The location of the survey.
 	//! @param releaseDate If known the UTC JD release date of the survey.  Used for cache busting.
```
gzotti, line 63: param type must be explained here.

10110111: Fixed.

alex-w (approved):
Thanks!

PR #4175, opened 2025-03-04
AstroCalc/Almanac

alex-w:
### Description
This pull request contains the new tool (Almanac) in Astronomical Calculations feature. The first subtool (Specific Time) in new tool is coming for compute seasons (as in typical Almanacs) and time of twilight/dawn/dusk, etc. with ability a quick set the time of these events in application. The next few subtools will be helpful for observers and some of them will similar to similar features from classical Almanacs.

The icon for new feature is re-used at the moment, and I have no idea for best graphical representation of new feature.

Another peculiarities of new feature is start splitting the huge tool into separate files/tools. The SUG is not reflected the changes yet.

stellarium-025

### Type of change

- [ ] Housekeeping

## Checklist:

### DIFF
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -1192,3 +1192,12 @@ QMessageBox {
 QColorDialog {
         background-color: rgb(143, 143, 143);
 }
+
+/*************************************************************************
+ * AstroCalcAlmanacWidget
+ *************************************************************************/
+
+AstroCalcAlmanacWidget QFrame[frameShape="4"],   /* QFrame::HLine == 0x0004 */
+AstroCalcAlmanacWidget QFrame[frameShape="5"] {  /* QFrame::VLine == 0x0005 */
+	border: 1px solid rgb(238, 238, 238);
+}
--- a/po/stellarium/POTFILES.in
+++ b/po/stellarium/POTFILES.in
@@ -58,6 +58,7 @@ src/gui/AstroCalcDialog.cpp
 src/gui/AstroCalcChart.cpp
 src/gui/AstroCalcExtraEphemerisDialog.cpp
 src/gui/AstroCalcCustomStepsDialog.cpp
+src/gui/AstroCalcAlmanacWidget.cpp
 src/gui/ObsListDialog.cpp
 src/gui/ConfigureDSOColorsDialog.cpp
 src/gui/ConfigureOrbitColorsDialog.cpp
@@ -86,6 +87,7 @@ src/ui_scriptConsole.h
 src/ui_astroCalcDialog.h
 src/ui_astroCalcExtraEphemerisDialog.h
 src/ui_astroCalcCustomStepsDialog.h
+src/ui_astroCalcAlmanacWidget.h
 src/ui_obsListDialog.h
 src/ui_dsoColorsDialog.h
 src/ui_orbitColorsDialog.h
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -356,12 +356,14 @@ IF(STELLARIUM_GUI_MODE STREQUAL "Standard")
           gui/ConfigureScreenshotsDialog.cpp
           gui/AstroCalcDialog.hpp
           gui/AstroCalcDialog.cpp
+          gui/AstroCalcChart.hpp
+          gui/AstroCalcChart.cpp
           gui/AstroCalcExtraEphemerisDialog.hpp
           gui/AstroCalcExtraEphemerisDialog.cpp
           gui/AstroCalcCustomStepsDialog.hpp
           gui/AstroCalcCustomStepsDialog.cpp
-          gui/AstroCalcChart.hpp
-          gui/AstroCalcChart.cpp
+          gui/AstroCalcAlmanacWidget.hpp
+          gui/AstroCalcAlmanacWidget.cpp
           gui/ObsListDialog.hpp
           gui/ObsListDialog.cpp
           gui/StelDialog.hpp
@@ -398,6 +400,7 @@ IF(STELLARIUM_GUI_MODE STREQUAL "Standard")
           gui/astroCalcDialog.ui
           gui/astroCalcExtraEphemerisDialog.ui
           gui/astroCalcCustomStepsDialog.ui
+          gui/astroCalcAlmanacWidget.ui
           gui/obsListDialog.ui
           gui/addRemoveLandscapesDialog.ui
           gui/dsoColorsDialog.ui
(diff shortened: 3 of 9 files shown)


10110111:
The behavior of "Current X Equinox/Solstice" is unintuitive. I expected that it would move to the event in the current real-time year, but instead it moves to the event in the current simulated year.

Also, the durations of polar nights/twilights in summer could be simply omitted instead of being 0h00m, similarly to the corresponding entry for the Moon. Otherwise we get
```
Sun  —  —  0h00m
Moon —  —
```

10110111:
It would be good if the data labels were selectable and copiable.

alex-w:
> It would be good if the data labels were selectable and copiable.

Any use case for this feature please...

alex-w:
> Also, the durations of polar nights/twilights in summer could be simply omitted instead of being 0h00m, similarly to the corresponding entry for the Moon. Otherwise we get
> 
> ```
> Sun  —  —  0h00m
> Moon —  —
> ```

This is correct - the duration of daytime is zero when polar night is coming :)

Atque:
Maybe the icon could be an actual calendar page?

alex-w:
> > Any use case for this feature please
> 
> Avoiding retyping the information. Particularly obvious with the Julian day.

But you can set the time immediately through respectively buttons - no need just a copy the data. But maybe you want to use these data outside the Stellarium?

> > This is correct - the duration of daytime is zero when polar night is coming
> 
> The Sun and the Moon are inconsistent here. Both don't rise, but only one has 0h00m.

The Moon can has set and rise time, but it doesn't have duration.

10110111:
> Maybe you want to use these data outside the Stellarium?

Of course, you don't just use an almanac to control the application that displays it :) I thought it was understood by default that the data are expected to be extractable.

alex-w:
> Of course, you don't just use an almanac to control the application that displays it :) 

If we can use an Almanac to control application then we should do it (without a part of copy/paste) :)

> I thought it was understood by default that the data are expected to be extractable.

For using inside the Stellarium we do not need extract data for copy/paste it by hands

10110111:
> If we can use an Almanac to control application then we should do it

I'm not saying we must not control it. I'm saying it's not the only use for the almanac.

> Extracting the data is have meanings when you want to use the data outside the application.

Do you mean the times of rises/sets/equinoxes/etc. are useless in the real life? What if I want to message my friend the time of sunrise tomorrow? Do I have to look at the Stellarium window and re-type the numbers from there? Or can Stellarium be so kind and allow me to copy and paste this information?

alex-w:
> > Extracting the data is have meanings when you want to use the data outside the application.
> 
> Do you mean the times of rises/sets/equinoxes/etc. are useless in the real life? What if I want to message my friend the time of sunrise tomorrow? Do I have to look at the Stellarium window and re-type the numbers from there? Or can Stellarium be so kind and allow me to copy and paste this information?

Okay, that makes sense.

alex-w:
> Maybe the icon could be an actual calendar page?

We using "calendar page" icon for WUT already...

gzotti:
Very nice addition indeed. Yes, some sort of data export would certainly be welcome. HTML table to clipboard, Excel, or a PDF report? (This could have been one application for the PrintSky plugin where we never found an actual application...) The table could add "created with Stellarium v25.1 for location LONG/LAT/TimeZone/..." for reference.

When Sun or Moon don't rise/set, `---` is good. "Duration 0h00m" is OK for me, but `---` would also signify "not applicable, given no sunrise". Matter of taste, I leave this decision to Alexander.

"Current year" etc is "current year in simulation" also elsewhere in the program, so OK for me. If you really want to express "current system time", the only icon that signifies this is the downward pointing triangle with small upward triangle below that says "now", this would have to be added. But the panel reacts to system time panel being set, so you could press the "now" button in the main tool bar. That's fine for me. 

I see a time zone issue, though. Season times reported apparently use timezones for the time of reporting. When I then click on the clock icon, I may jump to a time in the year when daylight time flips, and the correctly set time does not agree with the time displayed in the panel. When I e.g. then jump to "next same-season begin" and back to "previous same-season begin", the displayed time uses the now-flipped zone time. 

In effect, the given local time for seasons' beginnings should probably add "UT+XY", timezone name or/and be computed to fit the daylight time rules of the currently set timezone. Adding the time zone is IMHO never a bad idea.

alex-w:
> Very nice addition indeed. Yes, some sort of data export would certainly be welcome. HTML table to clipboard, Excel, or a PDF report? (This could have been one application for the PrintSky plugin where we never found an actual application...) The table could add "created with Stellarium v25.1 for location LONG/LAT/TimeZone/..." for reference.

Export as some formatted data maybe will added in future, but ability to select and copy dates and time is available right now.

> I see a time zone issue, though. Season times reported apparently use timezones for the time of reporting. When I then click on the clock icon, I may jump to a time in the year when daylight time flips, and the correctly set time does not agree with the time displayed in the panel. When I e.g. then jump to "next same-season begin" and back to "previous same-season begin", the displayed time uses the now-flipped zone time.

Oops... I've fixed it.

> In effect, the given local time for seasons' beginnings should probably add "UT+XY", timezone name or/and be computed to fit the daylight time rules of the currently set timezone. Adding the time zone is IMHO never a bad idea.

I'm not sure that showing the time zone is really needed here (we never use TZ info for local time in AstroCalc tool)

gzotti:
I'm afraid this did not solve it. See: I start with today in zone UT+1. I can see spring begin on March 20, click in the clock symbol and go there in time. As time zone stays UT+1, all is OK. 
stellarium-006

But then I'd like to go to Autumnal equinox. Display says Sept.22, 19h21. I click on the clock, and I am transferred to Sept.22, 20:20:42 in time zone UT+2. 
stellarium-007

So, the displayed time is based on the currently (simulation time) active daylight saving time rule when the table is filled. When I press now the "next autumn equinox", I see the date for 2026, then I press "previous autumn equinox", I see as time 20:21 for 2025, which is the correct answer with timezone UT+2, but differs from the result shown when I first opened the panel on March 4th. 

The casual user would expect results "as shown on their wall clock on that day", regardless of current state of DST, but could live with just that time offset string at least for clarity. Therefore my recommendation to add time zone information to it, or take active DST offset into consideration.

alex-w:
Ouch... done!

gzotti:
Yes, now it's fine for me!  That's a great tool for planning outdoor dusk sessions! 

Maybe a small fix: Currently, we read civil twilight, nautical twilight, astro twilight, and just below, sunrise. When you start with astro twilight, the sequence appears more logical to me.

alex-w:
> Maybe a small fix: Currently, we read civil twilight, nautical twilight, astro twilight, and just below, sunrise. When you start with astro twilight, the sequence appears more logical to me.

Okay, that makes sense.

PR #4179, opened 2025-03-06
Multiple label options for Skycultures

gzotti:
### Description

This is a step forward in the Skycultures expansion. Given various new naming components of which only a few tags may exist per skyculture, this should allow configuring user preferences, per skyculture, of how labels are to be configured from nativeName, pronunciation, translated name etc.

Fundamental is the introduction of StelObject::CulturalDisplayStyle, which allows cultural/linguistic awareness in all StelObjects. StelObject receives two new methods, getScreenLabel() and getInfoLabel(), which format the respective labels from the various components. 

### Done
 - [x] Label enum in StelObject, properties in StelSkyCultureMgr, GUI Switch, property storage
 - [x] Constellations
 - [x] Asterisms
 - [x] Star names: Do we need yet another different setting?
 - [x] Planet names: Do we need yet another different setting?
 - [x] DSO names: Do we need yet another different setting?

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win10/11
* Graphics Card: irrelevant

## Checklist:


gzotti:
When @10110111 was working on the big format switch, he asked about the value of separate translators for program GUI and skycultures. IIRC we tried to circumvent limitations of the old data files.  Please @alex-w, can you share some insights? We now have native names, transliteration/pronounciation etc., which should all make the native names understandable to the user in his own language. Do we still need two language settings?

The current state has the new data items available for constellations, asterisms, planets and DSO. Not polished yet, as the old solutions are still partially in use and could likely be simplified a bit. Class members which can have more than one cultural name own a list of CulturalNames (to never mix elements) from which the first is used as screen label, and all are visible in the Infostring label. Stars should come early next week, and I may have to extend the output options to include the "modern" name in a few more combinations, but I think the general direction is OK.  

In StarMgr, what does loadCultureSpecificNameForNamedObject()? The non-stars are handled by SolarSystem and NebulaMgr, so what is that for?

10110111:
> In StarMgr, what does loadCultureSpecificNameForNamedObject()? The non-stars are handled by SolarSystem and NebulaMgr, so what is that for?

It's for supplying names for stars specified by name, it uses the database from `skycultures/common_star_names.fab`.

gzotti:
Ah, OK. But I see it's never used, actually, and is a "just in case"? The only names following NAME so far are Planet names, at least for the SCs in master. DSO like M45 are labeled directly. Can we not just force SC authors to use HIP/Gaia designations directly (not a very high demand, right?) to simplify the code (i.e., remove that function)?

10110111:
> But I see it's never used, actually, and is a "just in case"?

This is in use in some SCs in the SC repo, e.g. Anutan.

> Can we not just force SC authors to use HIP/Gaia designations directly (not a very high demand, right?) to simplify the code (i.e., remove that function)?

I'm afraid not. I'd be glad to do this (and in my view we should do it), but it seems Fabien wants to extend this functionality even further, to allow not just names, but also things like `"lines": [["* gam Cru", "* del Cru"]],`, which will require some additional processing.

gzotti:
> > But I see it's never used, actually, and is a "just in case"?
> 
> This is in use in some SCs in the SC repo, e.g. Anutan.

Ah, like that. Hmm, OK. Still, I regard this as unneeded burden for software maintenance. But if it's used now, I'll comply.

> > Can we not just force SC authors to use HIP/Gaia designations directly (not a very high demand, right?) to simplify the code (i.e., remove that function)?
> 
> I'm afraid not. I'd be glad to do this (and in my view we should do it), but it seems Fabien wants to extend this functionality even further, to allow not just names, but also things like `"lines": [["* gam Cru", "* del Cru"]],`, which will require some additional processing.

I honestly don't see the necessity for that. All stars are meanwhile addressable by HIP or Gaia numbers.

gzotti:
To get a bit of clarity around the not or wrongly documented entries:

StarsMgr::commonNamesMap: This is the list from skycultures/common_star_names.fab, but can be modified when loading a skyculture. However, SC names now have several more components, I think this should not be mixed.

Is the StarsMgr::additionalNamesMap and additionalNamesMapI18n exclusively reserved for the Cultural names? If yes, I will replace by a QMap<starId, QList<CulturalName>>

10110111:
Actually, I was a bit wrong about this. The names from `skycultures/common_star_names.fab` only get into `commonNamesMap` when `skyCulture.fallbackToInternationalNames` is set. When this happens, these names are added first. After this the logic with `commonNamesMap`, `commonNamesMapI18n` etc. is the same as before the switch to the new format.

The names used for finding stars specified in `index.json` by name are assigned to `commonNamesIndexToSearchWhileLoading` local variable in `StarMgr::updateSkyCulture` and this is the only index used for searching.

gzotti:
More questions on code quality in StarMgr, @alex-w : The lists doubleHipStars, variableHipStars, algolTypeStars, classicalCepheidsTypeStars, hipStarsHighPM are lists of 1-element QMaps. Is that really intended? Why not just QMaps? Or lists of QPairs? They are used in Astrocalc, but in a very simple way. You can iterate over QMaps as well.

alex-w:
These lists using in AstroCalc and Search tools. I didn’t use QMap to not mixed iteration QList/QMap elements (though it was added before refactoring AstroCalc).

alex-w:
> Please @alex-w, can you share some insights? We now have native names, transliteration/pronounciation etc., which should all make the native names understandable to the user in his own language. Do we still need two language settings?

Probably we can remove second dropdown list for languages (skycultures) after introduction the support of extra options for labels. At least I don’t see critical necessary for second list.

alex-w:
Using starid:QStringList for storing and retrieving data for creating a screen or info label is very good, but it may be a bit painful for creating a suggestions in search tools.

10110111:
> it may be a bit painful for creating a suggestions in search tools

Why? Shouldn't search tools do a reverse lookup, by name?

gzotti:
I think we can assume the commonNames are a 1:1 relationship?
Scientific and scientific extra designations: if there is one scientific designation that has a preferred role, we could collate those into one QMap or QHash <StarId,QStringlist>sciDesignations  where the first element is the primary sci designation. Or even a QMultiMap or multihash, but here it seems the last-entered name is the fastest/first retrieved, so a stringlist and if needed first(), would be simpler. 
Finally, why are the sci, double and var star lists named ...I18n? I see nothing translated here. @alex-w , @henrysky ?

alex-w:
> > it may be a bit painful for creating a suggestions in search tools
> 
> Why? Shouldn't search tools do a reverse lookup, by name?

Yes, but please see the code `search*` and `listMatchingObjects` methods

alex-w:
> I think we can assume the commonNames are a 1:1 relationship? (each star only 1 name here, and each name only one star)

Yes, but without guarantee that situation will not changed in future

> Scientific and scientific extra designations: if there is one scientific designation that has a preferred role, we could collate those into one QMap or QHash <StarId,QStringlist>sciDesignations where the first element is the primary sci designation. Or even a QMultiMap or multihash, but here it seems the last-entered name is the fastest/first retrieved, so a stringlist and if needed first(), would be simpler.

QMap or QHash <StarId,QStringlist>sciDesignations should enough here. Currently `sciDesignations*` contains Bayer/Flamsteed designations of stars and `sciExtraDesignations*` has other scientific designations for stars (from the various catalogs of double stars). Plus we have separate lists for designations of variable stars (from GCVS) and designations for double stars (from WDS). `sciDesignations` and `sciExtraDesignations` was splitted to added ability for users do not show too many designations for stars on the screen.

> Finally, why are the sci, double and var star lists named ...I18n? I see nothing translated here. 

It was added by me. Currently you can remove part _I18n_ from names of maps/hashes (the data will never have translatable strings) for designations/cross index data.

gzotti:
> > I think we can assume the commonNames are a 1:1 relationship? (each star only 1 name here, and each name only one star)
> 
> Yes, but without guarantee that situation will not changed in future

Given the IAU star names, this should hold from now, and I will add a warning if a second name is read from here. 

> 
> > Scientific and scientific extra designations: if there is one scientific designation that has a preferred role, we could collate those into one QMap or QHash <StarId,QStringlist>sciDesignations where the first element is the primary sci designation. Or even a QMultiMap or multihash, but here it seems the last-entered name is the fastest/first retrieved, so a stringlist and if needed first(), would be simpler.
> 
> QMap or QHash <StarId,QStringlist>sciDesignations should enough here. Currently `sciDesignations*` contains Bayer/Flamsteed designations of stars and `sciExtraDesignations*` has other scientific designations for stars (from the various catalogs of double stars). Plus we have separate lists for designations of variable stars (from GCVS) and designations for double stars (from WDS). `sciDesignations` and `sciExtraDesignations` was splitted to added ability for users do not show too many designations for stars on the screen (this is important!).

Yes, short labels is surely important. With the sciDesignations only Bayer/Flamsteed indeed, we should keep those separate (will rename for clarity). Maybe it is enough to have one common index for the reverse lookup then, will see. When a sci name is needed, test BayerFlamsteed first, and add sciExtra.first() if needed, to show just one name. 


> 
> > Finally, why are the sci, double and var star lists named ...I18n? I see nothing translated here.
> 
> It was added by me. Currently you can remove part _I18n_ from names of maps/hashes (the data will never have translatable strings) for designations/cross index data.

OK. It just confuses, will rename.

alex-w:
> Given the IAU star names, this should hold from now, and I will add a warning if a second name is read from here.

Wait! `commonNames` is used for all common names currently (at least in master) and one star can have many names

gzotti:
The common_star_names.fab has unique names. The star names from the skycultures go into the new CulturalNames structure.

10110111:
> Ouch... Probably it should be `iau_star_names.fab` for clarify

No it shouldn't. It's the common names that are used for searching for the objects specified in `index.json`. They could generally include not only IAU-defined names. I just took IAU ones as the most overlapping with the `common_names.tab` from the SC repo.

gzotti:
Ah, therefore ref 2 everywhere. IMHO it would be good to have the unique "official" names in this file, now that such official register exists. A comment header can be added stating that these are Stellarium's default names, coming from not 30 sources, but from the one and only relevant body for contemporary observers, IAU. Personally, I'd like to avoid loading names coming from the IAU Exoplanet list (official or not, many are just silly), so I'd like so see them marked differently. I would then develop something similar to the DSO names list now. 
For varieties of names and alternative names, we have now the skycultures lists. (And again, here an exclusion scheme will come >;-)

gzotti:
Ah, I finally see another data issue: The modern* SCs have "native" and "english" constellation names, but star names are coded as "english" (i.e., translations) only. I think these entries should become "native", and probably the existence of "native" or at least "pronounce" should be checked, mandatory with at least warning, if not skipping entries when both native and pronounce are missing. Not sure what "english" (the actual translation) should then contain, either the same name or preferrably a best-knowledge english translation of wherever the name comes from? E g., native=Capella, english=Goat kid; native=Beteigeuze, english=Left hand of giant, ... 

In addition, at least some SCs (seen in Boorong and Inuit) have additional tags "description". Are these for translators only (then call them translator_notes) or shall we display them also (then they need their place in the CulturalNames, including translation and display options). Or shall the description moved to the actual file description.md? 

I am afraid there are several SCs where the naming is messed up. E.g. kamilaroi has their native names as "english" only, without further details or translations. Should be "native" only IMHO. Should I convert those as I see fit, or are there objections from S-mobile side?

@sushoff @10110111 @alex-w @xalioth  opinions?

10110111:
> In addition, at least some SCs (seen in Boorong and Inuit) have additional tags "description".

Do you mean in the SC repo? Because in desktop SCs they aren't present.
These tags are used for short descriptions in Stellarium Mobile, e.g.:

Screenshot_2025-04-03_19-13-32

gzotti:
Ah yes. But if the long-time goal is to share the format, these should be combined. OK, that means this tag shall be ignored in the desktop version, at least for now. 
More important is the interpretation of english/native tags. and the current (25.1) interminglings of commonNames and additionalNames. I have replaced the latter with culturalNames which have all relevant fields from the JSON.  Just playing with setting up better star labels from the selected options.

alex-w:
> I think we had translation context in some skycultures. I try to allow that in all cultural names, but would that mean Transifexing everything anew?

Checking and confirmation - yes

sushoff:
Sorry, I am a bit lost (many projects simultaneously) and I apologize if I again misunderstand something. I'll respond to what I understand. 

 1) modern star names are "english" only because the IAU-names are for research papers in astrophysics where only English is used as language. Yes, I agree, it is save to assume that for this list "english" equals "native". However, the IAU-Style Guide allows only ascii, no diacritics (and hyphens?), which is why WGSN sometimes gives two options: one option with "simple" diacritics like the French accents or German umlaut, and one "cleaned" without them. In this case, in Stellarium "english" and "native" would differ.
 
 2) I don't really understand the 1:1-debate above: I thought, we would have one list of official (IAU) names, and all the alternatives will be displayed in the "online query"-window. From my part of data management, the data for this part has yet to be prepaired, but I envisioned this as the solution: not hardcoded in the software, but in the transfer from the encyclopaedia project... Do I again misunderstand the question? 
 
 3) FYI the "one reference = IAU" is a great idea. If you take the website that I host (https://exopla.net/star-names/modern-iau-star-names/), it will work - the IAU website is now redesigned, but unfortunately all but complete. I'm still negociating... will take a while but hopefully doable within this year

gzotti:
OK, thanks. 

re 1. If there are technical requirements to do otherwise (@10110111 , @xalioth ?), I propose to change the star name entries in the modern* SCs to be tagged "native". Here, the "accented" versions would come in.  The "english" term usually provides a _translation of the meaning_ if available, which is further translated to user language. I proposed that we might add here the translation of the meaning (see above, Capella etc.)   It is debatable whether an accent-free version could go into the pronounce tag, at least in the "pure" modern (default) SC. If anyone wants this (IMO terrible) pronunciation guide from S&T, this may be applicable to the modern_st SC. Non-US users will always be free to switch it off, or translate away... (OMG, this is a "single" SC. Should translations even be allowed then?)

re 2. No, the online query window is a separate thing. We have the cultural names for the currently active skyculture on-screen, with a user selectable mix of Native, scientific transliteration, Pronounce, IPA, and Translated names. Just that there are too many lists and ideas how these are presented or which names fall back to which other names etc. 

There is a common_star_names.fab file (in the old format) which (now) contains (or should contain) the IAU star names, and could always be available as background knowledge. This list is and should be (but @10110111  had warned it may develop differently) 1:1, each star has one name, and each name belongs to a single star. Then there are the culture-specific names, which are many-to-many. A star can have several names, and the same name may apply to several stars. These cases must be mapped differently (with different data structures) in the program. 

Now, I see the online query window mostly as a cross-cultural thing, or providing additional information that exceeds what is stored in the SC files. Depending on what your website will deliver, I envision the following scenarios:
- you select a star, planet or DSO and ask the ASE via OnlineQuery. Result is a webpage that lists information about the object in all available cultures, with extra images, stories, legends, ... that culture historians have compiled in ASE. 
- you select a constellation and ask via OnlineQuery. Result is a webpage either going deeper about the constellation for the current skyculture, or provides information about all stars the selected constellation consists of, as found in other cultures (for now, the stars listed in the stick figure lines). This could be something like "The seven stars of the dipper are a wagon in Germanic and Belarus tradition, a dipper in China and the US, Seven Buddhas in Mongolia, three mourning daughters behind the bier in Arab tradition, ..." but also partial matches, like "Orion was part of the early Arab constellation "djauza" which may have consisted of ... and was a mythological figure that made...." or "the three belt stars formed constellation XY in Z culture"... Here results around overlaps are increasingly difficult and may require filtering (when becoming too large. Users may then want to exclude results for Oceanian SCs when dealing with Africa, etc. But this is a topic for ASE to solve.)
- I assume ASE to become a website that is useful in its own right, not just in interaction with Stellarium. Like Peakfinder is for landscapes. 
- ASE may/should be able to extract, for a culture selected on the website, a skyculture directory in Stellarium's new format for download as "current snapshot". This can be installed (optimally into the running program; files will be copied to the user data directory and will also be available on next program launch). The easy download is made in your system webbrowser, just retrieving a ZIP-file, and installing is similar to the custom Landscape installation. Installation in the user directory means older definitions for that skyculture that may have been included in a regular release will be hidden. 
- The OnlineQuery window is just an embedded Chromium-based web browser. It can display whatever a webbrowser should be able to do, but I have no fancy interaction in mind. I may have to have another look regarding the behaviour of its separate window, esp. when it comes to using on a second monitor. 

re 3. Yes, that list should be (made) available in a format similar to the DSO nebula file, with references listed in a commented section at the beginning of the file, and the old/still-current `HIP|name [ref]` format. It would be good to have Gaia DR3 numbers for the stars without HIP number. CSV download is good, we could start from here.

sushoff:
AH! ... please do not maintain yet another IAU-star name-file! There are too many (outdated ones) anyway! This list (the so called IAU-CSN = Catalog of Star Names) is updated monthly: *you* don't want to do this, because you are software developers and your skills are needed in other fields

gzotti:
Well, we will need to maintain our local version in any case, prior to each release then. (I.e., add one more step to the release building checklist: download CSV and process.) It just should not be used as master reference by other people (as Stellarium is increasingly done...). A simple CSV->fab converter can do the work, or another download option on your site.

gzotti:
As said before, desktop Stellarium works primarily offline, so files should be installed locally. Who curates the actual content of skycultures is a different topic. But the technical details of the exchange format are important, and here we have to converge to a useful solution.

gzotti:
Question on asterisms. Currently we still have telescopic asterisms defined as those given via coordinate lists, and then a search for a dim star is performed. If it fails, asterism is skipped. (Or a wrong star was found, see last bug report on the "stargate".) Now that we can address Gaia stars, I think this category with star search is no longer necessary. Coordinate lists (without resolving into stars!) would be useful (and shall be added) for polygons outlining dark constellations, deliberately without stars. Not sure if this is needed for telescopic asterisms, here "dark asterisms" would be equivalent to dark clouds, and these are best presented as DSO polygons already. A classification as telescopic asterism (auto-detectable by use of any Gaia StarId?) may still be useful to present them only when useful, with small enough fov or even a manual switch, to avoid screen clutter. @alex-w @10110111 ?

alex-w:
Yes, after introducing Gaia catalog using coordinates for asterisms is not needed anymore and probably we may remove type "TelescopicAsterism" (after updating data in JSON files). Using polygons for dark constellations is good idea, but we really having  dark asterisms?

10110111:
> Now that we can address Gaia stars, I think this category with star search is no longer necessary.

Yes, I think we can remove this separation.

alex-w:
> Hmm, a 3-way switch is complicated to explain for the users. Do they care whether the author wants no well-known star names? The names are only displayed with a setting that includes "modern". I can make a boolean "ignore/obey" though.

Why not use QSlider (auto; enable; disable)?

gzotti:
I mean you will need a paragraph in the User Guide stating "Auto means following the author's intention that is mapped in the index.json. To see whether the author intended to show well-known names or not, open a text editor and load skyculture/xy/index.json from the Stellarium installation folder... and then decide whether yo want to show, ignore, or let the author decide. Decide this for every entry in the skyculture list."
If I want to show well-known (IAU) star names while studying something like Boorong constellations to support what I am doing, I don't really care whether the author has chosen to not use them.  It is clear to me as user that the Boorong don't by default use these names. By just not even offering a switch and ignoring this entry (which may have its right in s-m) we save screen space and yet another enum, and modern names would in any case only ever be displayed when using one of the name styles that provides it.

10110111:
> Why not use QSlider (auto; enable; disable)?

What? These are three discrete non-numeric options, why `QSlider`? `QComboBox` is natural for this.

> open a text editor and load skyculture/xy/index.json from the Stellarium installation folder... and then decide whether yo want to show, ignore, or let the author decide. Decide this for every entry in the skyculture list."

No, this is not how defaults work :) What I'd expect is that this option is Auto by default, and then the user can force a particular choice. Of course, if you in principle disagree with the decision made by an author, you could maybe change the JSON file (having consulted with the author etc.).

gzotti:
In combining all kinds of labels, would it be possible to show markup like bold or italics? Simple HTML tags like `<i>` don't work.

10110111:
> would it be possible to show markup like bold or italics?

Given that they are rendered with `StelPainter::drawText`, you need to change `StelPainter::getTextTexture` to use some rich-text-enabled way of rendering, rather than `QPainter::drawText` (and make sure this parsing doesn't conflict with the other callers of `StelPainter::drawText`).

gzotti:
Ah, thanks, yes. That would put some QGraphicsTextItem in the process I guess. Hmm, maybe later. I will use angle brackets as in "<modernNameForReference"> for now where applicable.

gzotti:
Again about the skyCulture.fallbackToInternationalNames: If it means "this skyculture does not come with own star names, but uses the commonNames in its culturalNames", or at least, "this SC uses commonNames and a few of its own", this would make sense. It would usually be applied to modern variants which just don't care too much for star names. 
I would still read the commonNames list in any case and make it available as referential system when user selects one of the "..._Modern" display styles.

10110111:
In the SC repo this flag is only set for these:
```
$ grep fallback_ stellarium-skycultures/*/index.json
stellarium-skycultures/chinese_contemporary/index.json:  "fallback_to_international_names": true,
stellarium-skycultures/romanian/index.json:  "fallback_to_international_names": true,
stellarium-skycultures/western_hlad/index.json:  "fallback_to_international_names": true,
stellarium-skycultures/western/index.json:  "fallback_to_international_names": true,
stellarium-skycultures/western_rey/index.json:  "fallback_to_international_names": true,
stellarium-skycultures/western_SnT/index.json:  "fallback_to_international_names": true,
```
So basically yes, the "modern" ones, more or less.

alex-w:
@gzotti `QMultiHashIterator` is not existing in Qt5

src/core/modules/StarMgr.cpp
```diff
@@ -751,7 +751,7 @@ auto StarMgr::loadCommonNames(const QString& commonNameFile) const -> CommonName
 #if QT_VERSION >= QT_VERSION_CHECK(6,0,0)
 			StarId hip = recMatch.captured(1).toULongLong(&ok);
 #else
-			StarId hip = recMatch.captured(1).toInt(&ok);
+			StarId hip = recMatch.captured(1).toULongLong(&ok);
```
alex-w, line 754: Ifdef for Qt is not needed here

gzotti: I wondered why @10110111 used the fallback to Int here and suspected a problem with Qt5.12 which I no longer have. If CI is happy, this of course goes away.

gzotti:
> @gzotti `QMultiHashIterator` is not existing in Qt5

Yes, seen and fixed already.

Will we ever upgrade to Qt5.15? E.g. after Qt5 will be EOS in late May?

src/core/StelTranslator.cpp
```diff
@@ -146,18 +146,18 @@ QString StelTranslator::tryTranslateChineseStar(const QString& s, const QString&
...
 		}
 	}
-	const auto extra = match.captured(3);
+	const QString extra = match.captured(3);
 	if (!addedPresent) return translatedConstellation + number + extra;
 
-	const auto& translatedAdded = qtranslate(" Added", "chinese skycultures");
-	return QString("%1%2%3%4").arg(translatedConstellation, translatedAdded, number, extra);
+	const QString& translatedAdded = qtranslate("Added", "chinese skycultures");
+	return translatedConstellation + " " + translatedAdded + number + extra;
```
10110111, line 160: It wasn't "missing", this is intended. Some locales (e.g. German) want a space here, others (e.g. Chinese, or even Russian) don't.

src/core/modules/StarMgr.cpp
10110111: I think it was just a change by inertia when I was changing the conversion of a `QJsonValue`. Of course, `QString` doesn't have this problem.

alex-w:
> Will we ever upgrade to Qt5.15? E.g. after Qt5 will be EOS in late May?

AppVeyor has Qt 5.15.2 with serious regression on lconvert/lupdate tools. Of course, we may upgrade to Qt5.15, but cost of upgrading is discarding of localization support.

src/core/StelTranslator.cpp
gzotti: Ah, OK. Then this must be annotated in the typical // TRANSLATORS:... comment and be visible in Transifex.

gzotti:
5.15.2 works again, at least on a 8GB PC I think. 5.15.0 and 1 were unusable as we have reported... I will test again on weekend. Some things changed around 5.14 which might be advantageous.

alex-w:
> 5.15.2 works again, at least on a 8GB PC I think. 

Well, just try to rebase the branch - I've added instance for Qt5.15 in master now

src/core/StelTranslator.cpp
gzotti: What to do that translators will see the note that the space is important?

alex-w: Please see https://github.com/Stellarium/stellarium/commit/ebc48cb876ab69a8329ae091d924a1c7f3a57d63

gzotti:
The screen result is taking shape. 
Now, the two comboboxes with the incomplete choice of labeling options (actually there should be 2^6 combinations plus abbreviated constellations!) have become quite long, challenging small screens and maybe even intimidating some users. Maybe these should be handled by a flag/bitfield with 2x6 checkboxes? (plus extra for abbrev. constellation labels)

With DSO names, when a SC loads modern names, I have an InfoText line with cultural name, then the common names. Of course duplicate strings are possible. The optimal combination of names depends on existing data, demand for pronounce/transliteration fields (when native is Latin-based in the first place),  and user knowledge and expectations, so the settings are stored/reloaded immediately per culture.

alex-w:
Please fix codefactor suggestions for empty lines

alex-w:
Hmm… why English names for stars and asterisms is native now in modern SC?

gzotti:
Codefactor and other cleanups of course come before merge. 

The "native" constellation names in modern* should be Latin. "english" are the secondary translatable explanatory bynames ("Chained Woman" etc.). These should be translated. Most users familiar with the standard constellations will configure "native" and be happy. 

Again, note that the name mix selection is stored per-skyculture, so when you re-activate some SC which you had configured for your taste before, you will find the previously used configuration. 

Given that many SCs have incomplete datasets, I configured a few fallback strings, e.g., then transliteration is asked for, but only pronounce is available, the latter is provided.

gzotti:
What would be required soon, before or rather after merge: As next step, let's get rid of the second translator to simplify things (?), and see that all translations work correctly with and without context. (Context may be recommended as json element for each name, and should be extracted when preparing .po files for Transifex, right?) In this PR I have also updated a few SCs to follow the new system, but probably new strings for (re-?)translation will come.

alex-w:
IMHO `i18n` string shouldn't be in screen labels

alex-w:
I'm not sure that saving settings for labels per SC is good idea

gzotti:
> IMHO `i18n` string shouldn't be in screen labels of DSO

Why not? What else? The user-language translated name used to be there.

gzotti:
> I'm not sure that saving settings for labels per SC is good idea

I am very sure about this, as described several times. Switch between e.g. Modern, Chinese, Belarus, Siberian, and Arab multiple times, and you will like that the system remembers your per-culture preferences. When we have more data (more complete pronunciation or IPA entries), it may become even more important.

alex-w:
> > IMHO `i18n` string shouldn't be in screen labels of DSO
> 
> Why not? What else? The user-language translated name used to be there.

Let’s see the screen label for M45 - `i18nПлеяды` - it’s ugly.

gzotti:
> > > IMHO `i18n` string shouldn't be in screen labels of DSO
> > 
> > 
> > Why not? What else? The user-language translated name used to be there.
> 
> Let’s see the screen label for M45 - `i18nПлеяды` - it’s ugly.

Ah, yes, sorry. the "i18" in the string was for debugging.

gzotti:
> I’m not checked, but is it possible to switch settings for labels via scripts now?

Yes. The settings are not numbers, but string values converted via Qt MetaObject, See strings in StelObject.hpp, enum class CulturalDisplayStyle and in StelSkyCultureMgr.hpp:

```
        //! Returns the screen labeling setting for the currently active skyculture
	StelObject::CulturalDisplayStyle getScreenLabelStyle() const;
	//! Sets the screen labeling setting for the currently active skyculture
	void setScreenLabelStyle(const StelObject::CulturalDisplayStyle style);
	//! scripting version
	void setScreenLabelStyle(const QString &style);

	//! Returns the InfoString labeling setting for the currently active skyculture
	StelObject::CulturalDisplayStyle getInfoLabelStyle() const;
	//! sets the InfoString labeling setting for the currently active skyculture
	void setInfoLabelStyle(const StelObject::CulturalDisplayStyle style);
	//! scripting version
	void setInfoLabelStyle(const QString &style);

```

Not sure if we need the string value accessible via scripting. This can be added if needed via `QVariant::fromValue(style).toString()`

gzotti:
> Displaying discovery circumstances for minor planets and comets in infobox are broken now :(

Uh yes, still more duplicates to remove...

alex-w:
Oops… @gzotti did you checked working scripts like zodiac or constellations_tour?

gzotti:
@alex-w @10110111 @sushoff any further comments? Do you see obstacles preventing acceptance and merge?

10110111:
I'll try testing it tonight, haven't tried yet.

gzotti:
Can you please add which SC you tried, and whether you are talking of stars or constellations? Of course planet names can only be given if they exist in the respective SC, else the standard names are used. 

Regarding the mix of IPA, Translit etc.: when these fields are empty, I tried to provide at least "something useful". If you only select IPA, but IPA is empty (as is is currently most of the cases, with lots of future work for SC authors), there should be "something" that helps reading, be it pronounce, translation or native. The current lacunae (missing data) and mixed requirements for non-Latin native and only optional pronounce fields also necessitate the option store per-SC.

10110111:
> Can you please add which SC you tried

Modern, the default one.

gzotti:
> * When I turn off translation for Screen label (the only checkbox that was on for it initially), the labels get translated, while with this option on they are not. When this checkbox is checked, and I check "sci. translit." or "IPA", the labels also get translated to the same names. Not very intuitive.

What is the best default label for constellations when all options are switched off? For Modern SC, most users will likely select "native" to have the well-known Latin constellation names, and adding "translated" will add the newly-added explanatory additions. IMO the translated name is the best default option. Indeed IPA alone shows currently nothing. Switch off constellation label for no label. 


> 
>     * By default (with only "translation" enabled) a star is called Navi, while making all names translated by any of the above mentioned ways turns it into "γ Cas", which is weird.

If you switch off any cultural naming options, you fall back to the scientific designation. Not too bad IMO. The more interesting effects can be seen in e.g. Chinese SC.
 
>     * Somehow the planets aren't affected by these settings, they retain translated names regardless.
In Modern SCs we have no native alternatives. Please try Egyptian or any other SC that actually provides own planet names.

gzotti:
> > What is the best default label for constellations when all options are switched off? For Modern SC, most users will likely select "native" to have the well-known Latin constellation names
> 
> I'd say most Russian-speaking users will want to get the same standard Russian names (those adapted from Latin and used in the literature) by default, not the "native" ones: e.g. Водолей instead of Aquarius.

When I deselect everything, I see "Water Bearer" as label for Aquarius, which is what is given in the "english" tag. In Russian, I expect the translation of that to be Водолей. This is to have "some meaningful label". IMO the translation is better than the native in this case, given potential dangers around unknown glyph systems. (But indeed I did not provide fallbacks for switching on only IPA when IPA is not providing data. This results in empty labels.) Iy you intend to switch off label, switch off labeling directly, and don't exploit missing data, though. When I switch on "translated", it looks identical to having no selection about label style. When I switch on "Native+Translated", I have "Aquarius (Water Bearer)", 


> > If you switch off any cultural naming options, you fall back to the scientific designation.
> 
> Well the problem is that switching them OFF is the only way I found to turn ON translation. And Modern is the most visible SC, so this matters. And, to clarify, I was looking at the star names, not constellations (which I didn't even check).

One distinction is about star names. When switching off all cultural options, we still see the names from common_star_names.fab. Those are regarded as not endemically belonging to this sky culture, but to be internationally approved proper scientific names. Those which then have cultural names just show their sci. designations. 

Ah, I just see a quirk here. Both native and translation show "Cih" (You will likely see Navi, I have excluded reference 11 for me...), while adding "modern" switches this star to "&gamma; Cas", instead of adding the "modern name" (which should add "<&gamma; Cas>". TBD...

10110111:
> When switching off all cultural options, we still see the names from common_star_names.fab.

OK, so I get the names translated from `common_star_names.fab` when everything is off. This makes sense. But how do I get translated names (from whatever source) when something is on, e.g. "translation"? The constellations indeed seem to follow the choice, but not the stars, see these screenshots:

Screenshot_2025-05-04_14-21-15

Screenshot_2025-05-04_14-21-25

10110111:
I guess this is because after ca0ec35b7a99cffb979ec76dc917509ee2cf1184 the `modern/index.json` entries for the stars lack the `english` name, which then prevents translation I suppose. Are `native` names supposed to be translated now?

gzotti:
Star names in the modern SCs are supposed to show the canonical 
international (usually English, but at least Latin-char) spelling. With 
an english Tag, we could allow translated versions. I don't know 
practice e.g.in Russia, if Star names transliterated to cyrillic, we may 
need Star names both in native and English fields.

Am 04.05.2025 08:33, schrieb Ruslan Kabatsayev:

> 10110111 left a comment (Stellarium/stellarium#4179) [1]
> 
> I guess this is because after ca0ec35 [2] the modern/index.json entries 
> for the stars lack the english name, which then prevents translation I 
> suppose. Are native names supposed to be translated now?
> 
> --
> Reply to this email directly, view it on GitHub [1], or unsubscribe 
> [3].
> You are receiving this because you were mentioned.Message ID: 
> ***@***.***>

-- 


Links:
------
[1] 
https://github.com/Stellarium/stellarium/pull/4179#issuecomment-2849037770
[2] 
https://github.com/Stellarium/stellarium/commit/ca0ec35b7a99cffb979ec76dc917509ee2cf1184
[3]

10110111:
> I don't know practice e.g.in Russia, if Star names transliterated to cyrillic, we may need Star names both in native and English fields.

They are adapted, not always transliterated. E.g. Proxima Centauri becomes Проксима Центавр**а**, with the Russian genitive of Centaur and a transliteration of "Proxima". Another example: Polaris becomes Полярная, which is a translation (while the transliteration "Поларис" [refers to a missile](https://ru.wikipedia.org/wiki/%D0%9F%D0%BE%D0%BB%D0%B0%D1%80%D0%B8%D1%81) instead). Most of the others are indeed transliterated.

So we need an English entry, so that the users can see the normal localized star map by default.

sushoff:
sorry being late again - semester start (busy with new student projects and lectures... the beginning is always much work for the teachers in order to enable them to, then, learn & develop themselves: looking forward to great contributions now)

I suggest (again) to leave the selection as a dropdown menu. In the info label, we just present a chain of designators, but in the screen label, my intuition would be to leave the choice of orders to the user. I would design it somewhat like this: 

SCmenu-lbl-sel_menu

You could also make 4 dropdowns if you like, but I guess, nobody would want to display all four simultaneously ... but how knows: probably, we can't predict all use cases of all users ...

sushoff:
note for later

In the modern sky culture:  We will have to rework the description in in near future (but not now, because other people are not ready yet): the IAU webteam indeed managed to break all links! Namely, links to scientific bodies (e.g. working groups) don't work anymore and links to topics such as the definitions of constellations simply don't exist: in their first release, these reference pages don't exist. 
I will create an issue for myself (in my private issue board).

sushoff:
@gzotti where did we discuss the simbad-issue: I want to look up what you wanted me to give you an example for ... 

Can we please create a "project" in gitHub (so that I can sort issues in board layout) for the SC-project and the star catalogue-project: because in the huge list of open issues, I start loosing the overview

gzotti:
> I suggest to leave the selection as a dropdown menu. In the info label, we just present a chain of designators, but in the screen label, my intuition would be to leave the choice of orders to the user.

I think it is complicated enough to have two identically working selections of options. If we make one in the newer proposed checkbox style and the other in a 4/5-dropdown style with permutations, IMHO it gets too wild for both users and developers saving options. The sequence of "`native (reading aids, translation) <modern>"` looks for me to be the most logical, and yes, also I expect most users to select only 1-3 parts here, but it's very language/culture/expertise dependent. 

> where did we discuss the simbad-issue: I want to look up what you wanted me to give you an example for ...

Uh, I don't even remember what point we disscussed around Simbad, sorry. 

> Can we please create a "project" in gitHub (so that I can sort issues in board layout) for the SC-project and the star catalogue-project: because in the huge list of open issues, I start loosing the overview

Must have a look how this works. I think we used projects some time ago, but I cannot find the projects feature now (?)

sushoff:
> > Can we please create a "project" in gitHub (so that I can sort issues in board layout) for the SC-project and the star catalogue-project: because in the huge list of open issues, I start loosing the overview
> 
> Must have a look how this works. I think we used projects some time ago, but I cannot find the projects feature now (?)

click on your GitHub icon (opens a menu) and then on "projects", then you can create a new one with a big fat green button

sushoff:
> > I suggest to leave the selection as a dropdown menu. In the info label, we just present a chain of designators, but in the screen label, my intuition would be to leave the choice of orders to the user.
> 
> I think it is complicated enough to have two identically working selections of options. If we make one in the newer proposed checkbox style and the other in a 4/5-dropdown style with permutations,
> 
> I think it is complicated enough to have two identically working selections of options. If we make one in the newer proposed checkbox style and the other in a 4/5-dropdown style with permutations, 


au contraire! 

I think, four dropdowns, where the user needs to read and understand the terms once and then they only repeat (plus it really looks as all users are used to for 20 years now!), is much simpler than six(!) radio buttons!

>IMHO it gets too wild for both users and developers saving options. The sequence of "native (reading aids, translation) " looks for me to be the most logical, and yes, also I expect most users to select only 1-3 parts here, but it's very language/culture/expertise dependent.
>

gzotti:
> You could also make 4 dropdowns if you like, but I guess, nobody would want to display all four simultaneously ... but how knows: probably, we can't predict all use cases of all users ...

This is another point I am trying to make. The best way to let the user assemble his/her label from a choice of 6 entries is actually allowing all the choices, with the only restriction being the sequence of items (first "native", then "reading aids", then "translation", then "modern name" where required. Changes in this respect are also possible at a later date, when a few more SCs actually come with all those new extra data! If we want to go over the top, someone could disable various checkboxes if a SC has none of the respective data fields, but IMO it's really not worth the effort.

gzotti:
> OK, acknowledged. Star names in modern* SCs need to be in native and english fields.

Does anyone know a simple tool to replicate the native entries into english fields? But just in the star names. Not sure if I should write a Python tool that parses the JSON, or if sed/awk is all that's required. @10110111 ?

10110111:
> Not sure if I should write a Python tool that parses the JSON, or if sed/awk is all that's required.

I think `sed` should be sufficient, you just need to `s@"native\(":\s*"[^"]\+"\)@"english\1, "native\1@` or something similar. Just don't forget that this should only apply to `common_names`, not anything before it. Example (assuming `common_names` is the last entry):
```
sed -i '/"common_names"/,$ s@"native\(":\s*"[^"]\+"\)@"english\1, "native\1@' skycultures/modern/index.json
```

gzotti:
Alright, @10110111 please check again if this works for you now.

10110111:
Ah, I guess it's because with the previous commit the default SC was broken, and `getCurrentSkyCultureID()` returned an empty string.

gzotti:
Yes, I hope that was the case... :-o

gzotti:
@10110111 I hope you can have a reviewing look at this ASAP.

10110111:
The way the transliteration and IPA are merged in square brackets as e.g.

> Лось [Los’, ɫ̪os̪ʲ] (Elch)

is misleading, since square brackets are normally used for proper phonetic transcription (and slashes for phonemic transcription). It would be better to decorate the transliteration in some other way (braces?).

gzotti:
The square brackets are now used to collect all kinds of reading aids, while brackets enclose translation. You prefer {braces, like, these}? Or which ones? I was not even aware of slashes, always thought they are just replacements for angle brackets on some typewriters...

10110111:
> You prefer {braces, like, these}? Or which ones?

It's not that I prefer, they just seem to not be in use currently, so it's just a suggestion. Any way you choose, I'd like it to be clearly distinct and separate from the transcription. 

> I was not even aware of slashes, always thought they are just replacements for angle brackets on some typewriters...

Well, outside of phonetics they are sometimes used as substitutes for missing bracket characters, but when we are in the context of phonetics it's rather well-established to use square brackets for phonetic transcription and slashes for phonemic one.

gzotti:
> > You prefer {braces, like, these}? Or which ones?
> 
> It's not that I prefer, they just seem to not be in use currently, so it's just a suggestion. Any way you choose, I'd like it to be clearly distinct and separate from the transcription.

I have no formal education in linguistics. For me, all things inside the angle brackets are there to help reading/pronouncing (as seen in many books; some show transliterations, others show phonetics, others both), so I just collected everything there. But I have not seen curly braces in print use anywhere outside mathematical set notation (or programming). I also don't really expect users to always show all optional pronunciation/reading aids. Do you think an appearance of "Лось {Los’, [ɫ̪os̪ʲ]} (Elch)" is better? IMO that's too many brackets.

10110111:
> Do you think an appearance of "Лось {Los’, [ɫ̪os̪ʲ]} (Elch)" is better?

Absolutely not. I was thinking of something like "Лось {Los’} [ɫ̪os̪ʲ] (Elch)" instead. Otherwise, if we just leave it at the current state where two variants are mixed inside square brackets, the comma may be mistaken for an IPA secondary stress (despite having a space after it), and the whole expression taken as a single transcription.

gzotti:
"IPA secondary stress"... wow. I was not aware of that, OK. I still don't like the curlies, what about "Tortoise shell brackets" (U+3014/3015) for the pronounce/transliteration name bits?

10110111:
> I still don't like the curlies, what about "Tortoise shell brackets" (U+3014/3015) for the pronounce/transliteration name bits?

I guess it would work too. My only wish here is that nothing gets confused with/within the transcription.

10110111:
A side effect of this PR is that now abbreviated names are translated, and Ara becomes "Жертвенник" in the Russian locale, i.e. the full name of the constellation. So the abbreviation `.po` entries need a context.

gzotti:
Yes, I hope these translation details can be worked out ASAP after merge of this: where do we need context, can/should context be provided automatically, and esp. should we get rid of the second (skyculture) translator.
Actually, the short constellation/asterism labels should be used if you click "Abbreviated labels" below the name part selectors.

10110111:
> Actually, the short constellation/asterism labels should be used if you click "Abbreviated labels" below the name part selectors.

This is exactly how I discovered the problem.

alex-w:
IMHO abbreviations shouldn’t be translated

10110111:
> For abbreviations? Of course abbreviation for constellation may use other-glyph systems, but we shouldn’t use Chinese hieroglyphs or Cyrillic text as English source for translation.

No one said they'd be used as sources. We just need to disambiguate the English strings like "Ara", which can be both "Altar" and "Ara" in different contexts.

alex-w:
> Who said that? The abbreviations must be Latin characters to allow general developers' work with them, IMO not even accents/diacriticals allowed. But users may want to see them translated into their own language. And I discourage use of numerical labels unless you have a good reason (like enumeration from a book or so).

OK, but why abbreviations are translated now?

gzotti:
To let non-Latin users see their own characters on screen. What @10110111 noted is that the abbreviation for Ara is Ara itself, but the e.g. Cyrillic translation of (abbreviation) Ara should be "Apa", not "Жертвенник". This can be solved by giving context "abbreviated label", not "constellation name".

gzotti:
Probably, for some automation, constellation/asterisms labels and names can both use id as context:

name: "Ara" context: "id:CON modern Ara"
abbreviation: "Ara" context "abbrev:CON modern Ara" 

And actually, just like in the code, the abbreviations must be extracted from the id first.

alex-w:
Seriously, you’re don’t understand me. Why abbreviated text is translatable in the code?

10110111:
> Why abbreviated text is translatable in the code?

Hadn't you agreed to this when we discussed it months ago? I think it was in the main SC format PR, but that thread is too long to search (GitHub doesn't work well there).

alex-w:
> > Why abbreviated text is translatable in the code?
> 
> Hadn't you agreed to this when we discussed it months ago? I think it was in the main SC format PR, but that thread is too long to search (GitHub doesn't work well there).

Abbreviations are not extracted into pot file, so, I don’t expect translation the text, which is not translatable.

gzotti:
Abbreviations now undergo translations, as we have discussed we want that. I expect that this currently mostly fails (as e.g. Boo or Psc is not yet translated), but for the abbreviation Ara, the translation for the full-name Ara is found.

10110111:
> ```
> [SCInfoLabelStyle]
> modern                                          = Translated
> russian_siberian                                = Native_Pronounce_Translated_IPA
> ```
If these are lists of enabled items, why not separate them by commas rather than underscores?

10110111:
> Can we place fixing translation issues (adding context, and removing the second [skyculture language] translator) into a future branch to be started ASAP after merging this?

OK. This branch already has some unrelated changes like edits to irrelevant parts of SUG or some CMakeLists change that I don't get why it has to be a part of this PR.

10110111:
Why were constellation titles in `kamilaroi/description.md` demoted from the format-mandated level 5 to level 3?

gzotti:
> > ```
> > [SCInfoLabelStyle]
> > modern                                          = Translated
> > russian_siberian                                = Native_Pronounce_Translated_IPA
> > ```
> 
> If these are lists of enabled items, why not separate them by commas rather than underscores?

Ah, it never occurred to me to see them as separable lists. They are actually enum constants, the names of which are retrieved by the Qt MetaObject. When using commas, people may try to swap strings which would then fail.

gzotti:
> Why were constellation titles in `kamilaroi/description.md` demoted from the format-mandated level 5 to level 3? Same for `tupi`.

Uh, format-mandated? Sorry. The point was that the level five headings are smaller than the regular paragraph text, this looks very bad. Where are the style sizes configured?

10110111:
> Where are the style sizes configured?

HTML styles are in `data/gui/normalHtml.css`.

gzotti:
I know that file. But there are no styles for h4 and h5 (or any hX, actually...), so I hoped this was defined elsewhere. Do you find the tiny h5 style OK, or should we do something about it?

10110111:
> Do you find the tiny h5 style OK, or should we do something about it?

Well it looked OK to me (I didn't even notice the size mismatch) until you paid my attention to it :) I think the CSS can be fixed separately from this PR, and in this one the change to heading levels should be reverted.

10110111:
> They are actually enum constants, the names of which are retrieved by the Qt MetaObject. When using commas, people may try to swap strings which would then fail.

From the changes to `.ssc` files I see that these names are visible more than just in the config file. I think it would be more user-friendly to turn these lists into real lists, with arbitrary order, so that script writers could simply add whatever item is needed at the end, rather than look up the right name in the table of allowed names.

gzotti:
You mean string lists on .SSC API level, and the current enum internally, or really dealing with string lists everywhere? TBH I see no added value here. I chose the enum as it can be handled as bitfield with fast operations. I don't know many scripting applications where switching would be performed in the first place. The existing examples are brute-force overrides which I much dislike, and settings are restored after execution.

gzotti:
This means replacing underscores by commas for storing, and manipulating whatever string list is read to assemble whatever valid enum string can be assembled from the parts. OK, that's possible, it's just needlessly complicating the elegant 1-line solution provided by `QVariant(enum).value<enumName>()` .

10110111:
So it looks like the default if the string fails to be parsed is no checkboxes at all. Wouldn't it be better to use the global default in this case?
Also, for some reason I get no transliteration if 'native' is unchecked and IPA+transliteration+translation are checked (tested on Russian Siberian SC). If I enable native, both native and transliteration appear.

gzotti:
There is actually no global default. If no checkboxes are active, the old fallback was Translated, to have just "something" visible. Must check tomorrow what it is now. The other issue sounds weird, must also check that...

gzotti:
The label fallback for "nothing checked" should actually be the translated international (modern) name, to have something, and for those who don't actually care for names. While parsing the label settings, I check and combine all bits. If they don't parse, all bits will be off. This should now happen once, while going from enum-named config values to CSV-style values. 

The other issue should now be OK.

10110111:
OK, now the selections seem to work well. But for the transliteration, do you also get some extraneous spacing around the tortoise brackets? I guess you should, because they are square-letter (CJK) characters. I think the normal spaces around them should be removed.

10110111:
Oh, and you need to fix the scripts now too.

gzotti:
Ah yes, the scripts, which are totally ignoring user preferences for constellation tours... I probably won't run these default scripts again.

10110111:
> Any other brackets you prefer (but no curlies please)?

All Unicode brackets are listed [here](https://stackoverflow.com/a/33357311/673852). I suggest the tortoise shell brackets but the "black" ones: `⦗⦘`, U+2997 and U+2998. They seem to extend the same way as the parentheses (at least on my Stellarium screen).

gzotti:
Any further issues?

alex-w:
> Any further issues?

I think other issues should be fixed in separate branches

src/core/StelSkyCultureMgr.cpp
```diff
@@ -564,14 +578,14 @@ bool StelSkyCultureMgr::setCurrentSkyCultureNameI18(const QString& cultureName)
 }
 
 //! returns newline delimited list of human readable culture names in english
```
10110111, line 580: The comment is outdated.

10110111:
In many places the indentation of continuation lines is wrong, e.g. in the definition of `StelSkyCultureMgr::createCulturalLabel`.

src/core/StelSkyCultureMgr.cpp
```diff
@@ -756,3 +770,224 @@ QString StelSkyCultureMgr::skyCultureI18ToDirectory(const QString& cultureName)
...
+		default:
+			break;
+	}
+	// simple cases done. Now build-up label of form "primary [common transliteration aka pronounce, scientific transliteration, IPA] (translation) <modern>"
+	// "primary" is usually either native or one of the reading aids, or translation or even modern when other options are switched off.
+	// Rules:
+	// Styles with Native_* start with just native, but we must fallback to other strings if native is empty.
+	// Styles with Pronounce_* start with Pronounce or transliteration, but Pronounce_Translit_... must show Translit in braces when both exist.
+	// Styles with Translit_* start with Transliteration or fallback to Pronounce
+	// Styles with ...IPA... must add IPA (when exists) in braces, conditionally comma-separated in the same braces after Transliteration
```
10110111, line 931: IPA is displayed in square brackets, not braces (same for some other lines of this pseudo-multiline comment). Also, you've updated it to not include transliteration in the same brackets.

src/core/StelSkyCultureMgr.cpp
```diff
@@ -756,3 +770,224 @@ QString StelSkyCultureMgr::skyCultureI18ToDirectory(const QString& cultureName)
...
+	if (styleInt & int(StelObject::CulturalDisplayStyle::Native))
+	{
+		label=nativeOrPronounce;
+		// Add pronounciation and Translit in braces
+		if (styleInt & int(StelObject::CulturalDisplayStyle::Pronounce))
+			braced.append(pronounceStr);
+		if (styleInt & int(StelObject::CulturalDisplayStyle::Translit))
+			braced.append(cName.transliteration);
+	}
+	else // not including native // if (styleInt ^ int(StelObject::CulturalDisplayStyle::Native))
```
10110111, line 945: This commented out `if(A^B)` is not equivalent to `if(!(A&B))`. What is it supposed to be — an alternative implementation or a piece of documentation?

src/core/modules/Asterism.cpp
```diff
@@ -127,16 +207,18 @@ bool Asterism::read(const QJsonObject& data, StarMgr *starMgr)
 			}
 			case Type::TelescopicAsterism:
 			{
+					QMessageBox::information(nullptr, "Bad asterism", "Asterism with coord list found: " + id);
```
10110111, line 210: Is the issue so serious to warrant a message box while other SC issues are simply logged? This may be a dozen times that the user will have to click _OK_ if multiple asterisms have coordinate lists.
Also, this `nullptr` parent can make the message box hidden behind the Stellarium window, making the latter seem frozen.

gzotti: All asterisms in the default SCs are fixed. I wanted to make sure to signal, very loudly and directed to the SC author, that the coordinate lists are no longer supported. Should we hide this notice and just log an "ignore"?

gzotti:
> In many places the indentation of continuation lines is wrong, e.g. in the definition of `StelSkyCultureMgr::createCulturalLabel`.

I start to get the impression that indentation settings on my 3 computers change with either switching Qt targets or on every Qt update, but maybe only at full moon or when my neighbour's dog is hungry. I re-loaded the .clang-format file and re-ran indentation over the whole file. Is this all one has to do, or are there additional settings? 

Ah, there are apparently two settings. Our old xml file that worked for ages, but it may be overruled by the newer .clang-format file. Do we need both, are they fighting each other or do they work together?

src/core/modules/Asterism.cpp
10110111: We don't do this for any other issue, however serious it may be. And most external SCs are very imperfect format-wise anyway, so their authors will get other errors in the log anyway and should look for them.

src/core/StelSkyCultureMgr.cpp
```diff
@@ -48,117 +48,117 @@ constexpr auto SkipEmptyParts = QString::SkipEmptyParts;
...
-
-				string.replace(substr, newSubstr);
-
-				replaced = true;
-				break;
-			}
-		}
-	} while(replaced);
+    static const auto pPattern = []{
+        // Locates any paragraph. This will make sure we get the smallest paragraphs. If
```
10110111, line 52: This is worse than before this commit. Tabs have been replaced with spaces...

alex-w:
@gzotti please disable clandformat plugin in the QtCreator

10110111:
@alex-w It seems the importing and exporting feature is completely broken, at least in Qt Creator 15. First of all, it doesn't appear to want any `.xml` (judging by its file open filter), and whatever settings I export, they appear to be identical to our `.clang-format` file (whose contents also appear to be wrong, e.g. what is `IndentWidth: 3`?). So they simply don't reflect the checkboxes I choose.

gzotti:
The Qt settings dialog reports "there is a .clang-format file that defines the indentation and formatting settings". So, the xml is probably ignored or at least overruled. I don't see this "broken", but by current QtCreator design. Otherwise, the clang plugin provides important analysis, so I prefer to use it. If we decide to not use the .clang-format file (which existed since 2017 and was last touched in 2021, so probably after that time it gained preference over XML-based loadable styles), we should probably just delete this file from the git and reactivate the xml (plus the manual settings that influence tab widths or whatever, there are just too many confusing settings here), but not disable the plugin. Or we could also make good use of the .clang-format file and retire the xml file.

alex-w:
Removing the .clang-format file is a quick solution. As far as I remember we did not use clang formatter as a primary tool for format the source code (previously we’ve used astyle formatter). I’m not sure that we can fully switch to use .clang-format without changes in our coding style. :(

gzotti:
It must have had affected formatting, when indent=3 etc became active. However, there are these two files, and then there are countless settings in the GUI. Do they override the .clang-format file, or is that file the "live config" of these various checkboxes? I have no idea how the old xml file and clang-format interact. But when that config file is doing unwanted things and is otherwise not wanted, we better delete it. What does the xml file cover that the newer file doesn't?

I have disabled .clang-format (by renaming) on my "daytime system" before restarting QtCreator and reformatting.  Are these settings the same on your systems (Sorry for German labels, I hope your systems are structurally the same):

image

image

image

image

image

image

In the top area you see "Formatierungsart". options are "Integrated code model"/"just indentation"/"complete formatting". What is the required setting?

If this is so relevant, the path to consistent settings must be foolproof on all platforms. It should be OK to require "QtCreator 15 or later" (if available on all platforms). What annoys me as well are what may be caused by differences in tab width and tabbed formatting "later" in the lines. Tabs always mess up what should be eye-friendly visual alignments.

10110111:
Interesting, I've just tried switching between the _Formatting mode_ checkbox values, and the "classic" UI for the configuration only appears with the "Use built-in indenter". But the _Export_ button only ever exports ClangFormat files. So the old-style XML is simply not supported by Qt Creator 15 and ignored when imported. I assume that we shouldn't use the built-in indenter anymore in the new Qt Creator versions.

Now, using "Indenting only" mode we instead get a textual representation of the clang-format file, and editing it directly affects the preview next to the input field. So far I've modified it to this (our in-tree file simply represents the defaults, and doesn't follow our style):
```
---
Language:        Cpp
AccessModifierOffset: -4
AlignAfterOpenBracket: Align
AlignArrayOfStructures: None
AlignConsecutiveAssignments:
  Enabled:         false
  AcrossEmptyLines: false
  AcrossComments:  false
  AlignCompound:   false
  PadOperators:    false
AlignConsecutiveBitFields:
  Enabled:         false
  AcrossEmptyLines: false
  AcrossComments:  false
  AlignCompound:   false
  PadOperators:    false
AlignConsecutiveDeclarations:
  Enabled:         false
  AcrossEmptyLines: false
  AcrossComments:  false
  AlignCompound:   false
  PadOperators:    false
AlignConsecutiveMacros:
  Enabled:         false
  AcrossEmptyLines: false
  AcrossComments:  false
  AlignCompound:   false
  PadOperators:    false
AlignEscapedNewlines: DontAlign
AlignOperands:   Align
AlignTrailingComments:
  Kind:            Always
  OverEmptyLines:  0
AllowAllArgumentsOnNextLine: true
AllowAllParametersOfDeclarationOnNextLine: true
AllowShortBlocksOnASingleLine: Never
AllowShortCaseLabelsOnASingleLine: false
AllowShortEnumsOnASingleLine: true
AllowShortFunctionsOnASingleLine: Inline
AllowShortIfStatementsOnASingleLine: Never
AllowShortLambdasOnASingleLine: All
AllowShortLoopsOnASingleLine: false
AlwaysBreakAfterDefinitionReturnType: None
AlwaysBreakAfterReturnType: None
AlwaysBreakBeforeMultilineStrings: false
AlwaysBreakTemplateDeclarations: Yes
AttributeMacros:
  - __capability
BinPackArguments: false
BinPackParameters: false
BitFieldColonSpacing: Both
BraceWrapping:
  AfterCaseLabel:  false
  AfterClass:      true
  AfterControlStatement: Never
  AfterEnum:       false
  AfterExternBlock: false
  AfterFunction:   true
  AfterNamespace:  false
  AfterObjCDeclaration: false
  AfterStruct:     true
  AfterUnion:      false
  BeforeCatch:     false
  BeforeElse:      false
  BeforeLambdaBody: false
  BeforeWhile:     false
  IndentBraces:    false
  SplitEmptyFunction: false
  SplitEmptyRecord: false
  SplitEmptyNamespace: false
BreakAfterAttributes: Never
BreakAfterJavaFieldAnnotations: false
BreakArrays:     true
BreakBeforeBinaryOperators: All
BreakBeforeConceptDeclarations: Always
BreakBeforeBraces: Custom
BreakBeforeInlineASMColon: OnlyMultiline
BreakBeforeTernaryOperators: true
BreakConstructorInitializers: BeforeComma
BreakInheritanceList: BeforeColon
BreakStringLiterals: true
ColumnLimit:     100
CommentPragmas:  '^ IWYU pragma:'
CompactNamespaces: false
ConstructorInitializerIndentWidth: 4
ContinuationIndentWidth: 8
Cpp11BracedListStyle: true
DerivePointerAlignment: false
DisableFormat:   false
EmptyLineAfterAccessModifier: Never
EmptyLineBeforeAccessModifier: LogicalBlock
ExperimentalAutoDetectBinPacking: false
FixNamespaceComments: true
ForEachMacros:
  - forever
  - foreach
  - Q_FOREACH
  - BOOST_FOREACH
IfMacros:
  - KJ_IF_MAYBE
IncludeBlocks:   Preserve
IncludeCategories:
  - Regex:           '^<Q.*'
    Priority:        200
    SortPriority:    200
    CaseSensitive:   true
IncludeIsMainRegex: '(Test)?$'
IncludeIsMainSourceRegex: ''
IndentAccessModifiers: false
IndentCaseBlocks: false
IndentCaseLabels: false
IndentExternBlock: AfterExternBlock
IndentGotoLabels: true
IndentPPDirectives: None
IndentRequiresClause: true
IndentWidth:     8
IndentWrappedFunctionNames: false
InsertBraces:    false
InsertNewlineAtEOF: false
InsertTrailingCommas: None
IntegerLiteralSeparator:
  Binary:          0
  Decimal:         0
  Hex:             0
JavaScriptQuotes: Leave
JavaScriptWrapImports: true
KeepEmptyLinesAtTheStartOfBlocks: false
LambdaBodyIndentation: Signature
LineEnding:      DeriveLF
MacroBlockBegin: ''
MacroBlockEnd:   ''
MaxEmptyLinesToKeep: 1
NamespaceIndentation: None
ObjCBinPackProtocolList: Auto
ObjCBlockIndentWidth: 4
ObjCBreakBeforeNestedBlockParam: true
ObjCSpaceAfterProperty: false
ObjCSpaceBeforeProtocolList: true
PackConstructorInitializers: BinPack
PenaltyBreakAssignment: 150
PenaltyBreakBeforeFirstCallParameter: 300
PenaltyBreakComment: 500
PenaltyBreakFirstLessLess: 400
PenaltyBreakOpenParenthesis: 0
PenaltyBreakString: 600
PenaltyBreakTemplateDeclaration: 10
PenaltyExcessCharacter: 50
PenaltyIndentedWhitespace: 0
PenaltyReturnTypeOnItsOwnLine: 300
PointerAlignment: Right
PPIndentWidth:   -1
QualifierAlignment: Leave
ReferenceAlignment: Pointer
ReflowComments:  false
RemoveBracesLLVM: false
RemoveSemicolon: false
RequiresClausePosition: OwnLine
RequiresExpressionIndentation: OuterScope
SeparateDefinitionBlocks: Leave
ShortNamespaceLines: 1
SortIncludes:    CaseSensitive
SortJavaStaticImport: Before
SortUsingDeclarations: Lexicographic
SpaceAfterCStyleCast: true
SpaceAfterLogicalNot: false
SpaceAfterTemplateKeyword: false
SpaceAroundPointerQualifiers: Default
SpaceBeforeAssignmentOperators: true
SpaceBeforeCaseColon: false
SpaceBeforeCpp11BracedList: false
SpaceBeforeCtorInitializerColon: true
SpaceBeforeInheritanceColon: true
SpaceBeforeParens: ControlStatements
SpaceBeforeParensOptions:
  AfterControlStatements: true
  AfterForeachMacros: true
  AfterFunctionDefinitionName: false
  AfterFunctionDeclarationName: false
  AfterIfMacros:   true
  AfterOverloadedOperator: false
  AfterRequiresInClause: false
  AfterRequiresInExpression: false
  BeforeNonEmptyParentheses: false
SpaceBeforeRangeBasedForLoopColon: true
SpaceBeforeSquareBrackets: false
SpaceInEmptyBlock: false
SpaceInEmptyParentheses: false
SpacesBeforeTrailingComments: 1
SpacesInAngles:  Never
SpacesInConditionalStatement: false
SpacesInContainerLiterals: false
SpacesInCStyleCastParentheses: false
SpacesInLineCommentPrefix:
  Minimum:         1
  Maximum:         -1
SpacesInParentheses: false
SpacesInSquareBrackets: false
Standard:        Auto
StatementAttributeLikeMacros:
  - Q_EMIT
StatementMacros:
  - Q_UNUSED
  - QT_REQUIRE_VERSION
  - Q_CLASSINFO
  - Q_ENUM
  - Q_ENUM_NS
  - Q_FLAG
  - Q_FLAG_NS
  - Q_GADGET
  - Q_GADGET_EXPORT
  - Q_INTERFACES
  - Q_MOC_INCLUDE
  - Q_NAMESPACE
  - Q_NAMESPACE_EXPORT
  - Q_OBJECT
  - Q_PROPERTY
  - Q_REVISION
  - Q_DISABLE_COPY
  - Q_SET_OBJECT_NAME
  - QT_BEGIN_NAMESPACE
  - QT_END_NAMESPACE
  - QML_ADDED_IN_MINOR_VERSION
  - QML_ANONYMOUS
  - QML_ATTACHED
  - QML_DECLARE_TYPE
  - QML_DECLARE_TYPEINFO
  - QML_ELEMENT
  - QML_EXTENDED
  - QML_EXTENDED_NAMESPACE
  - QML_EXTRA_VERSION
  - QML_FOREIGN
  - QML_FOREIGN_NAMESPACE
  - QML_IMPLEMENTS_INTERFACES
  - QML_INTERFACE
  - QML_NAMED_ELEMENT
  - QML_REMOVED_IN_MINOR_VERSION
  - QML_SINGLETON
  - QML_UNAVAILABLE
  - QML_UNCREATABLE
  - QML_VALUE_TYPE
TabWidth:        8
UseTab:          ForIndentation
WhitespaceSensitiveMacros:
  - BOOST_PP_STRINGIZE
  - CF_SWIFT_NAME
  - NS_SWIFT_NAME
  - PP_STRINGIZE
  - STRINGIZE
...

```

gzotti:
I have disabled the clang-formatter plugin (Help->Plugins), restarted, and the xml support (R/W) is back. So, whichever way we decide, with the plugin the xml may be ignored and clang-format rules are observed as long as that file exists and the plugin is active (which it is by default). 

UseTab options: My interpretation:
AlignWithSpaces spaces only
ForIndentation Tabs at line start, spaces later (?). 

The actual decision about this is not part of this PR, though. Is the rest OK now?

10110111:
> Is the rest OK now?

There are several unresolved comments I left that are still relevant.

src/core/StelSkyCultureMgr.cpp
gzotti: remains of older idea, removed.

10110111:
Oh, indeed, GitHub mislead me to think these comments were still in their old form. They are OK now.

Unfortunately, the alignment of the continuation lines starting with `<<` is now broken. Same for `const QMap<QString, QString>options` in `getLicenseDescription` and other places where function argument lists are split into several lines. Just try looking at `git diff dcaec055d781fb0e497a821d1085621a9a5367df` and you'll notice this immediately.

gzotti:
I "fixed" the non-standard indentation in the license QMap. I have not reverted the 1-character indentation changes (also QString(...)/.arg continuations) which I actually find providing a more consistent view. Both git diff and my QCreator show the << well-aligned now, don't know what you see.

alex-w:
@10110111 would you update the `.clang-format` file?

10110111:
> would you update the .clang-format file?

I'll try doing it in a couple days.

gzotti:
I installed the git tools and did not reconfigure them. Especially, I don't use vim. For me apart from the one group near top the << sections look better than before. Yes, the 4/8 tabwidth difference causes permanent problems. As you have changed tab width to 4 on your system, I assume 8 is the default. Also compare appearance in github. And I agree, generally not using tabs and deciding for one indentation width using spaces would solve many such frustrating issues and delays. 

image

10110111:
> As you have changed tab width to 4 on your system, I assume 8 is the default.

Yes, for git-diff this can be achieved by setting git config's variable `pager` to `less -x1,5`, e.g. (if it doesn't exist):
```
git config --global --add core.pager 'less -x1,5'
```
> Also compare appearance in github.

GitHub can also show 4-char tabs, so there I also notice the issue.

Anyway, I've now fixed the remaining parts of the file, so I suppose this PR can go in.

gzotti:
Thanks, that (you running re-indent with your settings) was my hope for today. I will rebase to fix conflicts, then merge later today.

sushoff:
> > > But I see it's never used, actually, and is a "just in case"?
> > 
> > 
> > This is in use in some SCs in the SC repo, e.g. Anutan.
> 
> Ah, like that. Hmm, OK. Still, I regard this as unneeded burden for software maintenance. But if it's used now, I'll comply.
> 
> > > Can we not just force SC authors to use HIP/Gaia designations directly (not a very high demand, right?) to simplify the code (i.e., remove that function)?
> > 
> > 
> > I'm afraid not. I'd be glad to do this (and in my view we should do it), but it seems Fabien wants to extend this functionality even further, to allow not just names, but also things like `"lines": [["* gam Cru", "* del Cru"]],`, which will require some additional processing.
> 
> I honestly don't see the necessity for that. All stars are meanwhile addressable by HIP or Gaia numbers.

you may or may not have seen: 
There is a new sup-project by a team of computer science students who are actually great software developer, who develop a "sky culture maker plugin". Yesterday, we finished the second of five 14days-sprints. I trust they are really good: the function of drawing (poly)lines and snapping them to stars exists already. With that, SC-contributors can visually define their constellations and the entry (HIP or Gaia or any catalog number) will be added automatically to the JSON file - then, I think, names like *gam Cru are not necessary. 

Also, the function of anchoring at coordinates (instead of stars) allows for definition of dark constellation ... we "only" need the option to store that in JSON and display it. I can imagine that the average person who knows a SC does not like to sit down and type the telephon numbers from any modern star catalog into a (manually written) JSON file

10110111:
@gzotti So I've pushed the new `.clang-format`, you can try using it in "Indenting only" mode. If it works well, we can document it in the Coding Style document (leaving the old file for old Qt Creators too, because Ubuntu 22.04LTS ships QtC 6.0.2).

PR #4180, opened 2025-03-07
Use a glTF-compatible normal map for the Moon

10110111:
This loses a bit of precision per normal component, but lets us load normal maps from externally-supplied HiPS, which will most likely use this standard format.

This solves the problem that the `moon` HiPS looks too shallow compared to the built-in Moon.


alex-w (approved):
Thanks!

PR #4182, opened 2025-03-07
Add SC converter on AppVeyor/Windows x86_64

10110111:
The converter is now git-cloned on AppVeyor, and for `[publish]` builds also is built and installed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -19,26 +19,36 @@ skip_commits:
 max_jobs: 5
 # environment variables
 environment:
+  gettextSrcURL: https://ftp.gnu.org/gnu/gettext/gettext-0.23.1.tar.xz
+  gettextSrcBaseName: gettext-0.23.1
+  gettextURL: https://github.com/vslavik/gettext-tools-windows/releases/download/v0.23.1/gettext-tools-windows-0.23.1.zip
+  gettextBaseName: gettext-tools-windows-0.23.1
   matrix:
   - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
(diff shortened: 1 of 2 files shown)


alex-w:
Just an interesting: it can be more effectively when package for converter will be download from release page at GH and incorporated into stellarium installer?

10110111:
Yes, it can and likely will be. The question is whether we want it this way.

10110111:
> Making some official releases for converter may be important for maintainers in UNIX-like OS.

Yes, I'm just hesitant to make a release right now, just for this AppVeyor script to work.

.appveyor.yml
```diff
@@ -63,7 +73,25 @@ before_build:
   - if [%PUBLISH_BINARY%]==[true] set PATH=C:\stellarium-OpenSSL;%PATH%
   - if [%PUBLISH_BINARY%]==[true] appveyor DownloadFile https://github.com/Stellarium/stellarium-data/releases/download/qt-5.6/%SSL%Win%BITS%.zip -FileName c:\stellarium-OpenSSL.zip
   - if [%PUBLISH_BINARY%]==[true] 7z e c:\stellarium-OpenSSL.zip -aoa -oc:\stellarium-OpenSSL
-  - mkdir stellarium-%qtver%-%qtbin%
```
alex-w, line 66: Why this line is removed?

alex-w (changes requested):
Binary package cannot be installed into c:\stellarium-%qtver%-%qtbin% because the dircetory is not created

.appveyor.yml
10110111: Because it creates a directory in the current directory, which has changed by some `cd` commands above. And editing it to an absolute path didn't make sense to me, because it will be created by CMake's `install` target anyway.

alex-w (approved):
OK, thanks!

Issue #4178, opened 2025-03-05
How to use Gaia ID in Sky Culture


sunshuwei:
For example:

    {
      "id": "CON chinese 090",
      "lines": [[51232, 5350358584482202880, 56606, 56561, 57363]],
      "common_name": {"english": "Sea and Mountain", "native": "Hai Shan"}
    },

    "Gaia DR3 5350358584482202880": [{"english": "Sea and Mountain II"},
                 {"english": "Hǎishān II", "translators_comments": "Pinyin transliteration"}],      

It won't work

10110111:
Long ids should be passed as JSON strings rather than integers:

```
"lines": [[51232, "5350358584482202880", 56606, 56561, 57363]],
```

sunshuwei:
> Long ids should be passed as JSON strings rather than integers:
> 
> ```
> "lines": [[51232, "5350358584482202880", 56606, 56561, 57363]],
> ```

It still doesn't seem to work.
The constellation lines still does not use "5350358584482202880". 
And "Gaia DR3 5350358584482202880" also does not display the name "Sea and Mountain II"


I have uploaded [index.json](https://github.com/user-attachments/files/19122001/index.json)

The following are the changes made:

    {
      "id": "CON chinese 090",
      "lines": [[51232, "5350358584482202880", 56606, 56561, 57363]],
      "common_name": {"english": "Sea and Mountain", "native": "Hai Shan"}
    },
...

    "Gaia DR3 66715105696289024": [{"english": "Hairy Head Added X"},
                 {"english": "Mǎoxiù zēng X", "translators_comments": "Pinyin transliteration"}],  
    "Gaia DR3 211805079552264832": [{"english": "Five Chariots Added XVIII"},
                 {"english": "Wǔchē zēng XVIII", "translators_comments": "Pinyin transliteration"}],   
    "Gaia DR3 1233902704963092608": [{"english": "Great Horn Added II"},
                 {"english": "Dàjiǎo zēng II", "translators_comments": "Pinyin transliteration"}],     
    "Gaia DR3 3138519829402291328": [{"english": "South River Added XI"},
                 {"english": "Nánhé zēng XI", "translators_comments": "Pinyin transliteration"}],       
    "Gaia DR3 3626803636103730048": [{"english": "Recommending Virtuous Men Added VI"},
                 {"english": "Jìnxián zēng VI", "translators_comments": "Pinyin transliteration"}],        
[...]

PR #4195, opened 2025-03-10
Improvement fix for issue #4178

alex-w:
Added improvement to fix issue #4178

### DIFF
--- a/src/core/modules/StarMgr.cpp
+++ b/src/core/modules/StarMgr.cpp
@@ -919,7 +919,7 @@ void StarMgr::loadCultureSpecificNames(const QJsonObject& data, const QMap<QStri
 		// Let's allow Hipparcos and Gaia designations only
 		if (key.startsWith("HIP "))
 		{
-			loadCultureSpecificNameForStar(it.value().toArray(), key.mid(4).toULongLong());
+			loadCultureSpecificNameForStar(it.value().toArray(), key.mid(4).toUInt());
 		}
 		else if (key.startsWith("Gaia DR3 "))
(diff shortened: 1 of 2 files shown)


src/core/modules/StarMgr.cpp
```diff
@@ -916,9 +916,14 @@ void StarMgr::loadCultureSpecificNames(const QJsonObject& data, const QMap<QStri
 	for (auto it = data.begin(); it != data.end(); ++it)
 	{
 		const auto key = it.key();
+		// Let's allow Hipparcos and Gaia designations only
 		if (key.startsWith("HIP "))
 		{
-			loadCultureSpecificNameForHIP(it.value().toArray(), key.mid(4).toULongLong());
+			loadCultureSpecificNameForStar(it.value().toArray(), key.mid(4).toULongLong());
```
10110111, line 922: Now it should be `toUInt()` here.

alex-w: Why? The second parameter for method `loadCultureSpecificNameForStar` has _StarId_ type

10110111: Because we are parsing a HIP, which cannot exceed a 32-bit int range.

alex-w: I still see no reasons to twice type conversion

10110111: It's not done twice. It parses the string once (also validating the range), then zero-extends once.

10110111 (approved):
LGTM

PR #4201, opened 2025-03-12
Add support for loading external SCs

10110111:
Some email discussions indicated the desire to be able to load an external SC. This PR aims at implementing this. I'm not sure if it will be ready for the nearest release though.

### DIFF
--- a/src/core/StelSkyCultureMgr.cpp
+++ b/src/core/StelSkyCultureMgr.cpp
@@ -181,14 +181,36 @@ QString StelSkyCultureMgr::getSkyCultureEnglishName(const QString& dir, const QS
 StelSkyCultureMgr::StelSkyCultureMgr()
 {
 	setObjectName("StelSkyCultureMgr");
-	makeCulturesList();
 }
 
 StelSkyCultureMgr::~StelSkyCultureMgr()
 {
 }
 
-void StelSkyCultureMgr::addNewCulture(const QString& dir, const bool reportErrorsToGUI)
+void StelSkyCultureMgr::addCustomCulture(const QString& dir, const bool reportErrorsToGUI)
+{
+	if (!addNewCulture(dir, reportErrorsToGUI)) return;
+
+	additionalCultureDirs << dir;
+
+	QSettings*const conf = StelApp::getInstance().getSettings();
+	conf->beginGroup("skycultures");
+	conf->beginWriteArray("additional_cultures");
+	int i = 0;
+	for (const auto& dir : additionalCultureDirs)
+	{
+		conf->setArrayIndex(i);
+		conf->setValue("dir", dir);
+		++i;
+	}
+	conf->endArray();
+	conf->endGroup();
+
+	auto& culture = dirToNameEnglish[dir];
+	qInfo() << "Added a new sky culture:" << culture.englishName;
+}
+
+bool StelSkyCultureMgr::addNewCulture(const QString& dir, const bool reportErrorsToGUI)
 {
 	constexpr char indexFileName[] = "/index.json";
 	const QString filePath = QFileInfo(dir).isAbsolute() ? dir + indexFileName : StelFileMgr::findFile("skycultures/" + dir + indexFileName);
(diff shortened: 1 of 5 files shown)


gzotti:
In principle, yes, we should have such thing in 25.2. However, I am not sure about this additional_cultures entry. We have the user data dir, and in there is the user's private SC dir. To install a new one, just copy files there. So I see two useful modes: Copy over SC files from another open directory into `$USERDATA/skycultures/<newSC>`, or unpack a ZIPfile much like for downloaded landscapes. After copying/unpacking, make the new SC known to the system. 

Or what's the application for the additional_cultures?

10110111:
> Or what's the application for the additional_cultures?

Well, I thought the external SC would just be pointed at, and there'd be no need to copy it around.

gzotti:
I don't consider this necessary or beneficial. Reminds me of a 3dsmax OBJ export where the MTL file points anywhere else for textures instead of packing everything in one directory. A nightmare when just wanting to copy USERDIR to another system, suddenly you have missing tentacles.... Or is there an actual use case? I would edit my private SCs in my USERDIR.   @alex-w , @sushoff opinions?

10110111:
> Or is there an actual use case? I would edit my private SCs in my USERDIR.

Well, my expectation was that a creator's SC would reside somewhere on a USB stick, in a Dropbox directory or elsewhere, but not in Stellarium's special directory. I guess the user could add a symlink in Stellarium's directory in this case, but AFAIK there's no GUI for this in Windows, so many will not even know of this option.

gzotti:
Users capable of creating a SC should know where their USERDIR is, even if it's hidden by default on all platforms. (You can use another dir on any drive, pointed at by an ENVVAR, to keep your system drive clean.) There one author would work on it. When finished, author can pack it to a ZIP and distribute. Receivers would download ZIP anywhere and install to their USERDIR with your add button. This is the process for adding landscapes for simple users, and IMHO should be applied also here.

sushoff:
I agree seeing the necessity for a function to load external. I think, the initial request does not target the "users capable to create SCs" but all the others (the majority!) 

- Some years ago, we had some South African SCs in our repository which we had not included in the standard download because something essential was missing (I think, it was the licence, but doesn't matter) until I had pestered them long enough that they provided the necessary detail. However, until we included it in the download, nobody except they themselves (and we) knew about their work. It would have been great for everybody to at least see the options. 
- Then, my next idea was, that it would actually be great to have all SCs downloadable on request from an external directory. This would have serveral advantages for us: (1) if the contributor changes a detail (as research goes on, sometimes researchers change their mind), then the downloaded version will always have the state of the art. (2) more important for us legally: if we have to withdraw a dataset, we can actually withdraw it from our external repository - in contrast, we have no control about the private directories on our millions of users' personal devices. This will make us both, scientifically & legally up to date by obeying the [CARE principles](https://en.wikipedia.org/wiki/CARE_Principles_for_Indigenous_Data_Governance#:~:text=The%20CARE%20Principles%20are%20'people,accessible%2C%20interoperable%2C%20reusable).) of research data management, and technically more to date: see, for instance, the "stickers" in WhatsApp: the user does not have to download them all, but can choose to download specific series as they like -> that seems to be state of the art in user friendliness. 
- Then, Georg, you (independent from my idea and due to different lines of thought) also came up with the idea to exclude all the SCs from the download. Your motivation simply was to make the download package slimmer (although, I think, the SCs contribute only a small share, compared to - e.g. - the huge modern catalogues of all sorts of surveys). 
- Therefore, I support this inital request takes us to the right path that we want to develope towards anyway.

10110111:
@sushoff
First, the question wasn't about making Stellarium download something. It was about whether we want to keep the SC data where it resides in the system, or force the user to copy it into Stellarium's config directory.

As for making the SCs basically a service rather than just data, you seem to be thinking that everyone has perpetual stable Internet access. This is not so in reality.

sushoff:
1. no, we don't want to have the files distributed over the whole system - but it may helpful for some people to maintain their test versions outside the local Stellarium installation files
2. I don't think that everybody has perpetual internet access (I know, that's not the case). but typically, we know in advance what we need. For instance, during a planetarium show or in a school lesson, the presenter may not have stable internet connection but knows in advance what they want to use and will prepare. Likewise, when an amateur astronomer travels to a remote place to get the greatest astrophotos, they will prepare for the travel and download in advance (like packing the suitcase) what will be needed. 

When do we need the SC? I know two types of main application: one is research (falls in the category of travel preparation), the other is presentation for the public/in education (falls in cat. 1).

alex-w (changes requested):
Please fix unit test

alex-w:
We should use stardard directories in USERDIR for storing skycultures IMHO

PR #4216, opened 2025-03-17
Set correct light pollution level when location=auto

### LINKED ISSUE #2762: Location=auto fails to set correct light pollution level
### Expected Behaviour

Launching newly-installed Stellarium should set location according to some ways of self-location, and corresponding light pollution level should be set. At the very least this should work for major cities.

### Actual Behaviour

Launching Stellarium in Saint Petersburg, Russia, with `[init_location]/location=auto` sets correct geographical coordinates and a name of the location, "St Petersburg". Light pollution level is set to 23.3 mag/arcsec² (Bortle class 2), which is way too low for Saint Petersburg.

OTOH, selecting location from the _Location_ dialog as a particular entry in the list, "Saint Petersburg, Eastern Europe", results in a much more realistic light pollution level of 18 mag/arcsec² (Bortle class 9).

### System
* Stellarium version: v1.0-241-g1f44dedae3
* Operating system: Windows, Ubuntu

### Logfile 

[log.txt](https://github.com/Stellarium/stellarium/files/9842028/log.txt)

alex-w:
### Description
This patch should obtaining a realistic light pollution level for auto discovered locations (at least for big cities)

Fixes #2762 (issue)

### Type of change

### DIFF
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -1165,7 +1165,7 @@ void StelLocationMgr::changeLocationFromNetworkLookup()
 			if (!ipCity.isEmpty())
 				ipLoc = locationForString(QString("%1, %2").arg(ipCity, regionName));
 			else
-				ipLoc = pickLocationsNearby("Earth", longitude, latitude, 5.f).first();
+				ipLoc = pickLocationsNearby("Earth", longitude, latitude, 1.f).first();
 
 			if (ipLoc.isValid())
 				luminance = ipLoc.lightPollutionLuminance.toFloat();
--- a/src/core/StelLocationMgr.cpp
+++ b/src/core/StelLocationMgr.cpp
@@ -1160,13 +1160,15 @@ void StelLocationMgr::changeLocationFromNetworkLookup()
 
 			QString regionName = pickRegionFromCountryCode(ipCountryCode.toLower());
 			float luminance = StelLocation::DEFAULT_LIGHT_POLLUTION_LUMINANCE;
+			StelLocation ipLoc;
+			// Check location in our database and fetch light pollution luminance if it possible
 			if (!ipCity.isEmpty())
-			{
-				// Check location in our database and fetch light pollution luminance if it possible
-				StelLocation loctmp = locationForString(QString("%1, %2").arg(ipCity, regionName));
-				if (loctmp.isValid())
-					luminance = loctmp.lightPollutionLuminance.toFloat();
-			}
+				ipLoc = locationForString(QString("%1, %2").arg(ipCity, regionName));
+			else
+				ipLoc = pickLocationsNearby("Earth", longitude, latitude, 5.f).first();
+
+			if (ipLoc.isValid())
+				luminance = ipLoc.lightPollutionLuminance.toFloat();
 
 			StelLocation loc;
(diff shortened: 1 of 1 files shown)


alex-w:
> Where will it ultimately come from?

From out database if city name is known

alex-w:
@10110111 could you check how works getting the location (and light pollution level) from IP for St. Petersbourg?

10110111:
> could you check how works getting the location (and light pollution level) from IP for St. Petersbourg?

Not sure about SPb, since I don't have access to my machine there, but for Kuta, Indonesia I get a very wrong location:
```
Got location ", , Indonesia (-6.175, 106.829; Asia/Jakarta)" for IP "106.0.50.27"
```
So there's no city, but the coordinates indicate Jakarta.
 
And when I choose light pollution setting "From location database", the value gets set to Bortle class 2, which is very wrong for Kuta that's the brightest part of Bali, and just as well wrong for Jakarta that's the brightest part of Java.

I think it would be better to use some kind of a light pollution map instead of a city-based table. E.g. [The World Atlas of the Artificial Night Sky Brightness](http://www.lightpollution.it/worldatlas/pages/fig1.htm).

gzotti:
The IP-based location lookup has set me to places 18km (regularly) or even 160km away from my place. It seems OK to set the location close enough to get a roughly correct sky, but for anything else we'd need more. Looking up the IP-based coordinates in a LP database will therefore also not really help. 

The various OSes have introduced their own location APIs which _may_ work better, but many users dislike being "tracked". Should we try to make use of that?

10110111:
> The various OSes have introduced their own location APIs which may work better, but many users dislike being "tracked". Should we try to make use of that?

Isn't it just an API to the GPS sensors, whose input we already support?

gzotti:
They must do something around network traffic. My desktop PC and laptop don't have sensors, but in the system control panel there are settings that work considerably better than our IP-based lookup.

10110111:
Well, if it's indeed something more than the GPS, then I suppose we could add support for this, maybe as a button near the "Get from GPS" one.

alex-w:
Added code for processing locations, which presents by coordinates only

src/core/StelLocationMgr.cpp
```diff
@@ -1158,16 +1158,28 @@ void StelLocationMgr::changeLocationFromNetworkLookup()
...
 			qDebug() << "Got location" << QString("%1, %2, %3 (%4, %5; %6)").arg(ipCity, ipRegion, ipCountry).arg(latitude).arg(longitude).arg(ipTimeZone) << "for IP" << locMap.value("ip").toString();
 
+			QString regionName = pickRegionFromCountryCode(ipCountryCode.toLower());
+			float luminance = StelLocation::DEFAULT_LIGHT_POLLUTION_LUMINANCE;
+			StelLocation ipLoc;
+			// Check location in our database and fetch light pollution luminance if it possible
+			if (!ipCity.isEmpty())
+				ipLoc = locationForString(QString("%1, %2").arg(ipCity, regionName));
+			else
+				ipLoc = pickLocationsNearby("Earth", longitude, latitude, 5.f).first();
```
gzotti, line 1168: This will select a random location within 5° (~550km) of coordinates, which IMO does not make much sense. You would have to iterate over the map to find the entry closest to coordinates.

alex-w:
At least it works now (not 100% good but better than nothing)

PR #4217, opened 2025-03-17
Update StelFader.hpp

pkrawczun:
Explicitly inheriting the operator = from LinearFader. Not an error but produces a lot of warnings and might potentially cause an issue in the future.


src/core/StelFader.hpp
```diff
@@ -140,6 +140,7 @@ class ParabolicFader : public LinearFader
 {
 public:
 	// Create and initialise to default
+	using LinearFader::operator=;
```
gzotti, line 143: While this may fix the warnings which I also have seen in CI builds, writing it between a comment and the function which is commented is not a good move, as it makes the comment which is now obviously commenting your line hard to understand in the future. Basically you hijacked the comment...

Just move it down a bit, and maybe even explain in its own comment why it is needed.

pkrawczun: I see what you mean.

gzotti (approved):
Looks OK for me, thanks!

PR #4221, opened 2025-03-17
Use QtPositioning for location lookup

gzotti:
### Description

Discussed in #4216 

Our IP-based location lookup was a nice feature in 2014, but Qt has introduced some while ago the QtPositioning library. 
This can utilize the OS services for location lookup and may deliver more accurate results.

Operating systems may limit access to these services and require permissions that must be granted or can be denied. 

The current implementation first tries to get permission for a high-precision lookup, if that is not granted, a low-precision lookup. If this is not granted, the previous IP-based lookup is executed.

Note that the permission system was only introduced in Qt6.5 or 6.6, so the first versions here will likely not compile in our CI.

For discussion: 
- [ ] Shall we remove the previous IP-based lookup fallback? 
- [ ] We have a contribution that uses its own QGeoPositionInfoSource for recurring position updates as alternative to GPS, but is #ifdeffed to Windows only. It may be also useful on Linux.
- [ ] If this is difficult on Qt5, we could just exclude it there.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11 23H2
* Graphics Card: irrelevant
* Qt:
* [x] 6.8.2
* [x] 6.5
* [x] 5.15.2  Attn, fallback lookup fails with "TLS initialization failed"! So, our old IP lookup fails with Qt5.

## Checklist:


gzotti:
Given that this did not need new buttons or labels, if it works on your platforms, it could maybe even still go into 25.1.

src/core/StelLocationMgr.cpp
```diff
@@ -502,6 +505,12 @@ StelLocationMgr::StelLocationMgr()
 	planetName="Earth";
 	planetSurfaceMap=QImage(":/graphicGui/miscWorldMap.jpg");
 	connect(StelApp::getInstance().getCore(), SIGNAL(locationChanged(StelLocation)), this, SLOT(changePlanetMapForLocation(StelLocation)));
+
+	// configure the QGeoPositionInfoSource which can be queried from OS
+	qGeoPositionInfoSource = QGeoPositionInfoSource::createDefaultSource(this);
+	if (qGeoPositionInfoSource)
+		connect(qGeoPositionInfoSource, SIGNAL(positionUpdated(QGeoPositionInfo)),
```
10110111, line 512: Why do you still use this antiquated string-based connection mechanism?

gzotti: Directly taken from Qt6.8 docs.

gzotti:
> I do not getting coordinates/location in linux and he switches to use custom time zone when feature is disabled

The logfile should contain a line "Position received from OS query:...". What does it say? If the location is invalid, this should then fallback to the old version. (TODO, I assumed the reply is meaningful.)

Not sure what you mean by custom timezone. 
This lookup does not provide place name or timezone. I initialize the location from the current location and change a few data entries. I have left TODO notes where the timezone and light pollution could be put in. The timezone could come from the system clock (Assuming you travel somewhere far away, adjust PC to local time zone and then call this).

gzotti:
Maybe the logic should be changed: 
Retrieve both, QtPosition (long/lat/altitude?) and old IP lookup (long/lat/0, timezone, placename, country, ...).
Process data like with old solution, just replace long/lat/alt with the new data, This would retrieve at least timezone from IP lookup.
The place name retrieved from IP lookup does not make sense when the actual position is tens of km away, so placename from coordinates seems OK to me.

If you then know where to get LP values for coordinates, this can be inserted where indicated by TODO.

alex-w:
> The logfile should contain a line "Position received from OS query:...". What does it say? If the location is invalid, this should then fallback to the old version. (TODO, I assumed the reply is meaningful.)

Nothing...
 
> Not sure what you mean by custom timezone. This lookup does not provide place name or timezone. I initialize the location from the current location and change a few data entries. I have left TODO notes where the timezone and light pollution could be put in. The timezone could come from the system clock (Assuming you travel somewhere far away, adjust PC to local time zone and then call this).

Stellarium after startup:
stellarium-008

OK, click "Get location from network":
stellarium-009

In the log I see:
```
[   156.295][DBG ] Doing OS service lookup for location...
[   156.296][DBG ] Doing OS service lookup for location... postRequest 
```
nothing else :(

OK, click "Get location from network" again, to disable using this feature:
stellarium-010

WTF???

alex-w:
Effect switching to using custom time zone can be reproduced in master too :(

gzotti:
This may be caused by `LocationDialog::ipQueryLocation(bool state)`. Not sure why it has 
`updateTimeZoneControls(!state);` -- Since 2019.

gzotti:
I see a problem in that there are two asynchronous methods called. We could call the old IP method if QtPositioning result is bad. But when the QtPositioning message reply does not even appear, should we add some wait timer that fires in this case?

src/core/modules/Landscape.cpp
```diff
@@ -156,32 +156,23 @@ void Landscape::loadCommon(const QSettings& landscapeIni, const QString& landsca
...
+		location.planetName = landscapeIni.value("location/planet", "Earth").toString();
 		// Tolerate decimal values in .ini file, but round to nearest integer
-		if (landscapeIni.contains("location/altitude"))
-			location.altitude = qRound(landscapeIni.value("location/altitude").toDouble());
-		if (landscapeIni.contains("location/latitude"))
-			location.setLatitude(static_cast<float>(StelUtils::getDecAngle(landscapeIni.value("location/latitude").toString())*M_180_PI));
-		if (landscapeIni.contains("location/longitude"))
-			location.setLongitude(static_cast<float>(StelUtils::getDecAngle(landscapeIni.value("location/longitude").toString())*M_180_PI));
+		location.altitude = qRound(landscapeIni.value("location/altitude", 0.).toDouble());
+		location.setLatitude(static_cast<float>(StelUtils::getDecAngle(landscapeIni.value("location/latitude", 0.).toString())*M_180_PI));
```
alex-w, line 162: Latitude can’t be outside the range -90..90 degrees, so, maybe using 100 degree as default values for wrong latitude?

10110111: How will it behave though if applied?

gzotti: An invalid StelLocation should be just marked invalid by the dedicated flag, role='!', which I retry here and now. For now I added checks hopefully wherever StelLocations are generated, to set them valid only if useful. It may be useful to load other roles [CRNOL] from the landscape.ini. However, are these roles used anywhere in the program? Maybe to guesstimate light pollution?

The code change above changes nothing in behaviour BTW. long/lat was and still is 0 for a fresh StelLocation.  It seems in 2013 (!) we did not trust the default values and were super careful when parsing an ini file. If you think it's needed, we now can introduce qBound clamping in StelLocation::setLongitude() and StelLocation::setLatitude(). Or we just keep making the user responsible for own errors.

alex-w:
Debug messages should be more informative for users IMHO

alex-w:
The big deal: how it need test for all cases of potential troubles?

gzotti:
I have worked earlier this week on a desktop PC and experienced failures in ` StelLocationMgr::positionUpdatedFromOS(const QGeoPositionInfo &info)`, when a serial port was found, but no NMEA device. Now back on my notebook, pressing "Get location from GPS or system device" works again as the NMEA stuff quickly fails with port not found, then QtPositioning takes over. For debugging, we had activated extended debugging with CLI option `--verbose`. 

The QtPositioning addon was contributed on Windows only but should actually work on all platforms where QtPositioning is working properly. I have read there may be separate plugins that may have to be activated in code first on the U*X platforms. I am not there yet.

In the last days I have modified StelLocation in this branch so that it is default-invalid (role='!'), and solved an ultimative lastResortLocation (Paris, unconditional). This alone should help avoiding #4266, you could test that on a fresh install on Linux and Mac please.

gzotti:
Earlier in March I tried to extend the "location from IP" button. You said it does nothing on Linux. 
I now tried to remove the Windows-only limit of uwes-ufo QtPositioning extension.

A few more diagnostics show:

```
[    16.556][DBG ] Neither GPSD nor NMEAhelper. Try OS positionSource, one of
[    16.556][DBG ] QList("nmea", "winrt")
[    16.556][DBG ] Our QGeoPositionInfoSource is : "winrt"
[    16.556][DBG ] Setting up new positionSource...
[    16.572][DBG ] QFlags(0x1|0x2|0x4|0x8|0x10|0x20|0x40|0x80|0x100|0x200|0x400|0x800|0x1000|0x2000|0x4000|0x8000|0x10000|0x20000|0x40000|0x80000|0x100000|0x200000|0x400000|0x800000|0x1000000|0x2000000|0x4000000|0x8000000|0x10000000|0x20000000|0x40000000|0x80000000)
[    16.611][DBG ] Setting up new positionSource...done
[    16.899][DBG ] Location in progress: Long= 16.2599  Lat= 48.1814  Alt 0
[    16.899][DBG ] New location named  "GPS N48 E16"
[    16.899][DBG ] queryOK, resetting GUI
```

Building on WSL, I get:

```
[    25.627][DBG ] Neither GPSD nor NMEAhelper. Try OS positionSource, one of
[    25.627][DBG ] QList("geoclue2", "nmea")
[    25.627][DBG ] Our QGeoPositionInfoSource is: "geoclue2"
[    25.627][DBG ] Setting up new positionSource...
[    25.627][DBG ] QFlags()
[    25.627][DBG ] Setting up new positionSource...done

```

So, it seems the geoclue2 plugin on Ubuntu (WSL) does not behave like the winrt plugin on Windows, which may have been why the location from IP extension did not work on Linux. 

(Note that the nmea is Qt's own version, not the NMEALookupHelper Florian had made with me in 2017.)

According to https://doc.qt.io/qt-6/position-plugin-geoclue2.html, the geoclue2 plugin requires installation of 
`sudo apt install libgeoclue-2-0`
and required dbus communication. I installed this and libgeoclue-2-dev but it changed nothing.
I am not familiar enough with dbus. How to check if if dbus is even running on WSL? 

The situation is quite stupid, admittedly. The minimal solution would probably be to fix, in master, just the fallback to lastResortLocation to hit Paris, just to find ourselves in a useful location if all else fails. The "get location from GPS or OS" button may not work on all systems. Or an ifdef OS_WIN orgy in this branch tomorrow. :-(

alex-w:
Hmm... what about `geoclue-2-demo` and `geoclue-2.0` packages to check on WSL?

gzotti:
Can you please share any information what to do after installing those? I see /etc/geoclue/geoclue.conf and added an entry for stellarium similar to the firefox entry. Changed nothing.

gzotti:
Hmpf...

```
sudo systemctl start geoclue.service
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```

And according to https://wiki.ubuntuusers.de/geoclue/, Mozilla has shut down their WiFi location service. But what to do to start dbus on WSL?

alex-w:
> Can you please share any information what to do after installing those? I see /etc/geoclue/geoclue.conf and added an entry for stellarium similar to the firefox entry. Changed nothing.

I'm not familiar with geoclue and I don't have hardware with incorporated GPS/GLONASS/GALILEO/BAIDU device. I'll try use GPS/GLONASS dongle.

gzotti:
Ah, then probably the nmea plugin will kick in. It may work, but only for those with GPS/GLONASS hardware. But we already support operation with GPS hardware since 2017...  

Alright, sorry, I think I better today deactivate again all new things around QtPositioning for non-Windows for 25.2 and open up further experiments in August, maybe on a fresh Kubuntu installed on a then-ex-Win10 workhorse. My idea was simply to provide a more accurate location when asking for a network location,  But for an all-platform solution around QtPositioning, we need a workflow that incorporates the geoclue2 plugin, or that properly detects if something does not work.

alex-w:
Is it working on Windows?

P.S. Technically I can test it in Linux, macOS and Haiku platforms currently

gzotti:
Qt on Windows uses the "winrt" QGeoPositionInfoSource and just works. And of course whatever I did in February/March just worked. I have not tested recently what happens if I actively disallow geolocation (no tracking in system settings). But of course it is wise to test capabilities first before querying. Actually, I think the GPS button is now safe for Linux, it just doesn't provide data when geoclue isn't running.  I still need to secure the "location from network" checkbox action.

10110111:
Why is all this magic with WSL needed? If you need to test on Linux, you can just use a real VM booting real Linux (e.g. a Live CD).

gzotti:
I just have no time to set up everything again today, sorry. Stellarium builds and runs perfectly on WSL, just an optional system service is not available. And there may be other reasons (privacy, firewalls, ...) for QGeoPositionInfoSource not delivering data.  So, I will just test whether the QGeoPositionInfoSource delivers data (without test we had the fallback to not even Paris but zero), and catch and fall back to the old path if not.

alex-w:
Is it fixed issue #4266 in Windows?

gzotti:
I never had the problem... But I am just restricting the IP lookup to valid- tested sources, and will test with disabled network later to see if I land in Paris or in null island.

gzotti:
OK, in system panel I can disallow location access to Stellarium. However, I still get a permission. Then the service doesn't deliver. Weird, but I can catch this. So be it...

gzotti:
When I rip out the cable and want to set location from IP in Linux, 

```
[    40.537][DBG ] permission not granted or no QGeoPositionInfoSource, doing old IP service lookup for location
[    60.585][DBG ] Failure getting IP-based location:
         "Host freegeoip.stellarium.org not found"
```

When I then press "return to default", I get 

`[   126.064][CRIT] StelCore::returnToDefaultLocation: Location  "auto|||!|0|0.000000N|0.000000E|0|5.17088e-05||Earth|" is invalid. Store an entry from the locations list as default location.
`

So this still has to be fixed...

gzotti:
A pristine installation should now retrieve location from OS service (if that is ready to deliver something meaningful) or IP lookup. If not on a network, user fallback to Paris is OK again. I could not trigger a zero-island location now, but the worst possible is a recommendation to temporarily rename config.ini and restart to configure location, then never touch "location from network" in a network-restricted installation. 

I might want to rebase -i a few things, but I hope it at least solves #4266. Testing geoclue support from my side has to wait for whenever I get my hands on a Linux system with working geoclue2 on it. (It may even work already, though...)

10110111:
I get different behavior of the "get from network" checkbox on two machines in the same network (see below). What could be the reason for the difference?

* Ubuntu 20.04 + Qt 6.4.1
```
[    16.424][DBG ] Doing OS service lookup for location...
[    16.424][DBG ] Doing OS service lookup for location... postRequest 
[    17.285][DBG ] Position received from OS query: QGeoPositionInfo(QDateTime(2025-06-22 01:22:06.000 WITA Qt::LocalTime), QGeoCoordinate(0.93469856012, 104.4583), HorizontalAccuracy=23000)
```
* Kubuntu 25.04 (Qt 6.8.3)
```
[    13.245][DBG ] Got location ", , Indonesia (-6.175, 106.829; Asia/Jakarta)" for IP "106.0.50.27"
[    13.250][DBG ] Close location:  "Bekasi"  --  19 km
[    13.250][DBG ] -- TAKEN!
[    13.250][DBG ] Close location:  "Bogor"  --  46 km
[    13.250][DBG ] Close location:  "Ciampea"  --  44 km
[    13.250][DBG ] Close location:  "Cibinong"  --  34 km
[    13.250][DBG ] Close location:  "Cikarang"  --  37 km
[    13.250][DBG ] Close location:  "Cikupa"  --  36 km
[    13.251][DBG ] Close location:  "Cileungsir"  --  28 km
[    13.251][DBG ] Close location:  "Ciputat"  --  16 km
[    13.251][DBG ] -- TAKEN!
[    13.251][DBG ] Close location:  "Citeureup"  --  34 km
[    13.251][DBG ] Close location:  "Curug"  --  31 km
[    13.251][DBG ] Close location:  "Depok (West Java)"  --  24 km
[    13.251][DBG ] Close location:  "Jakarta Planetarium and Observatory"  --  2 km
[    13.251][DBG ] -- TAKEN!
[    13.251][DBG ] Close location:  "Jakarta"  --  4 km
[    13.251][DBG ] Close location:  "Kresek"  --  49 km
[    13.251][DBG ] Close location:  "Pamulang"  --  21 km
[    13.251][DBG ] Close location:  "Parung"  --  29 km
[    13.251][DBG ] Close location:  "Pasarkemis"  --  33 km
[    13.251][DBG ] Close location:  "Rengasdengklok"  --  51 km
[    13.251][DBG ] Close location:  "Sawangan"  --  25 km
[    13.251][DBG ] Close location:  "Sepatan"  --  28 km
[    13.251][DBG ] Close location:  "Serpong"  --  24 km
[    13.251][DBG ] Close location:  "South Tangerang"  --  17 km
[    13.251][DBG ] Close location:  "Tangerang"  --  21 km
[    13.252][DBG ] Close location:  "Teluknaga"  --  22 km
[    13.252][DBG ] Close location:  "Utan"  --  13 km
[    13.252][DBG ] Closest known place: "Jakarta Planetarium and Observatory" at 2.02359 km
```

alex-w:
QtPositioning (Geoclue?) is working on Ubuntu 20.04 machine and not working on Kubuntu 25.04 machine

gzotti:
Looks like the first finds a QGeoPositionInfoSource with Qt<6.6 (so geoclue2 works here? Hooray!), the second has retrieved an old-style IP answer with enough data to retrieve a closer place to retrieve light pollution data. I think I have never seen this output, actually... Or maybe in February/March before I tried to get a better location by QtPositioning.

Can you run with CLI option --verbose ? It gives more outputs.

gzotti:
And yes, can you press the "Get location from GPS" button on the U20.4 device? It should start emitting results, likely the button will be yellow indicating a 2D position (no altitude). Press a second time to stop.

gzotti:
Just reactivated my NUC (Ubuntu 24.4) for a quick Linux test... My geoclue2 may run, but I am not allowed to run it (where-am-i demo). How to configure that? With this half-working state the location lookup doesn't retrieve location. Grmbl...

gzotti:
Sorry, I will have to re-disable this part for Linux for now. Enabling location service seems non-standard, and I don't see how I can diagnose this unknown plugin. The net effect now is that lookup changes to better accuracy for Windows, but all changes in IP lookup on Linux will have to wait for when s.o. finds out how to retrieve location data via geoclue. That I read Mozilla has apparently stopped supporting some underlying service does not help here. I finally found the location switch. Now the dbus error has vanished, but the source still is quiet. I would like to know what /usr/libexec/geoclue-2.0/demos/where-am-i should deliver. Here it waits for a few seconds, then returns without any output.

alex-w:
Is agent running? 

P.S. /usr/libexec/geoclue-2.0/demos/agent

gzotti:
This program answers "nicht verwendeter Ortungsdienst" ("unused location service"). Not sure about the meaning, as I would definitely like to use it! BTW, it does not return. Does it expect messages, status updates etc via dbus? 

In any case, it means, on this Ubuntu 24.4 (current LTS) it does not work out of the box, and I see it's just too late for 25.2 to start digging in vague instructions. I need a 1/2 page howto, no treatises about dbus. If this is an advanced feature that needs system changes, we must write instructions for our users, and the expected behaviour is failure. The instructions must be complete for Ubuntu/Debian and maybe 2 other distribution families. This needs participation from a few people here. When someone can write a half-page of setup instructions for geoclue2 for my Ubuntu NUC, it would help a lot. this will be another PR. It must instruct users to install the right apt packages, start whatever new services are required, modify config files (AND WHAT CHANGES EXACTLY!). When it works, we can flag those packages as recommended and write instructions, but Stellarium still must work without these optional packages. 

We can probably merge this now to avoid zero-island issues, and make a new branch for geoclue2 integration in 25.3 (hopefully).

gzotti:
Ah, agent gets updated. When I run where-am-i in another terminal, a new enlightening message appears: 
"verwendeter Ortungsdienst" ("used location service"). Maybe translation should be "location service in use". After running where-am-i, the first message is printed again. And a little geotag symbol appears in the upper right corner. Following into the system settings, I am again pointed to mozilla's location service website. It appears to deliver minimal information (country and time). But neither agent not where-am-i seem to do anything about it. Clearly, something is still missing.

10110111:
I've successfully gotten it to work on Kubuntu 25.04 by starting the agent. But something is still strange with permissions or something.

Here I check "get location from network":
```
[    24.879][DBG ] permission not granted or no QGeoPositionInfoSource, doing old IP service lookup for location
[    26.053][DBG ] Got location ", , Indonesia (-6.175, 106.829; Asia/Jakarta)" for IP "106.0.50.27"
[    26.057][DBG ] Close location:  "Bekasi"  --  19 km
[    26.057][DBG ] -- TAKEN!
[    26.057][DBG ] Close location:  "Bogor"  --  46 km
[    26.058][DBG ] Close location:  "Ciampea"  --  44 km
[    26.058][DBG ] Close location:  "Cibinong"  --  34 km
[    26.058][DBG ] Close location:  "Cikarang"  --  37 km
[    26.058][DBG ] Close location:  "Cikupa"  --  36 km
[    26.058][DBG ] Close location:  "Cileungsir"  --  28 km
[    26.058][DBG ] Close location:  "Ciputat"  --  16 km
[    26.058][DBG ] -- TAKEN!
[    26.058][DBG ] Close location:  "Citeureup"  --  34 km
[    26.058][DBG ] Close location:  "Curug"  --  31 km
[    26.058][DBG ] Close location:  "Depok (West Java)"  --  24 km
[    26.058][DBG ] Close location:  "Jakarta Planetarium and Observatory"  --  2 km
[    26.058][DBG ] -- TAKEN!
[    26.058][DBG ] Close location:  "Jakarta"  --  4 km
[    26.058][DBG ] Close location:  "Kresek"  --  49 km
[    26.058][DBG ] Close location:  "Pamulang"  --  21 km
[    26.058][DBG ] Close location:  "Parung"  --  29 km
[    26.058][DBG ] Close location:  "Pasarkemis"  --  33 km
[    26.058][DBG ] Close location:  "Rengasdengklok"  --  51 km
[    26.058][DBG ] Close location:  "Sawangan"  --  25 km
[    26.059][DBG ] Close location:  "Sepatan"  --  28 km
[    26.059][DBG ] Close location:  "Serpong"  --  24 km
[    26.059][DBG ] Close location:  "South Tangerang"  --  17 km
[    26.059][DBG ] Close location:  "Tangerang"  --  21 km
[    26.059][DBG ] Close location:  "Teluknaga"  --  22 km
[    26.059][DBG ] Close location:  "Utan"  --  13 km
[    26.059][DBG ] Closest known place: "Jakarta Planetarium and Observatory" at 2.02359 km
```
Here I uncheck it:
```
[    52.705][DBG ] permission granted, doing OS service lookup for location...
[    52.705][DBG ] Location provider: "geoclue2"
[    52.706][DBG ] Location provider supported caps: QFlags(0x100|0x200|0x400|0x800|0x1000|0x2000|0x4000|0x8000|0x10000|0x20000|0x40000|0x80000|0x100000|0x200000|0x400000|0x800000|0x1000000|0x2000000|0x4000000|0x8000000|0x10000000|0x20000000|0x40000000|0x80000000)
[    52.706][DBG ] permission granted, doing OS service lookup for location... postRequest 
[    53.433][DBG ] Position received from OS query: QGeoPositionInfo(QDateTime(2025-06-22 10:14:29.000 +08 Qt::LocalTime), QGeoCoordinate(0.91679732204, 104.4583), HorizontalAccuracy=21000)
```

10110111:
> And yes, can you press the "Get location from GPS" button on the U20.4 device? It should start emitting results, likely the button will be yellow indicating a 2D position (no altitude). Press a second time to stop.

Yes, the button gets pressed, then red and unpressed. The logs contain the following, although I can't find where the `ttyS4` is configured. It's not mentioned anywhere in `~/.stellarium`.
```
[    25.526][DBG ] Only one port found at  "ttyS4"
[    25.535][DBG ] Query GPS NMEA device at port  "ttyS4"
[    31.627][WARN] "NMEA general error: UpdateTimeoutError"
[    31.627][DBG ] GPS queryError, resetting GUI
[    33.658][DBG ] setFieldsFromLocation(): Empty loc.ianaTimeZone!
```

gzotti:
Red button indicates no data, i.e. geoclue does not provide results. This button was originally (2017) developed to access GPS mice (USB or serial) sending NMEA strings (SUG 5.5.4) or connect to gpsd where available. A NMEA device is configured in the config.ini with a serial port name, but if there is only one in the system (as is frequent with USB<->serial converters plugged into field laptops), the config entry is not required. It then detects no communication on ttyS4 and gives up. This should be the same behaviour as in e.g. 0.20.  The button would be green while receiving 3D fixes (with altitude) or yellow (only long/lat). Later we received the first access of OS location query (Windows only). Given that OS location comes without altitude, only yellow results can be expected.  If geoclue would run as expected, it should behave similar, which was the idea in dropping the limitation to Windows.  However, it seems I need concise instructions how to make geoclue work, not just formal, but actually delivering coordinates. 

As a funny side discovery, it turned out that pressing the "get location from network" checkbox actually triggers an IP-based query when switching on and again when switching off. This came unexpected, but I did not touch it (as touching some unexplained constructs often turn out to trigger landmines...). IMO it should only trigger one lookup on activating. (The activation also sets an entry to config.ini to retrieve location from network on program launch.) I am not happy with resolution as I am often located 160km off.

The search for close locations (now silenced unless you run --verbose) is an attempt to set light pollution from nearby locations in the location DB. I am not so sure whether it is meaningful to accept a data value for a closest location of 25km, as even 5km can make a huge difference. Anyhow, the algorithm works for now.

I saw no other way for 25.2 to again limit OS-based queries to Windows. If someone instructs me (and users) to properly setup geoclue2 to actually provide location data (something that seems to work on one of your systems!?), we can continue. In any case, checking unavailability of data must become more robust as well.

10110111:
> something that seems to work on one of your systems!?

The `where-am-i` demo now works on both. I had to start the agent, which was already running on the system that worked at my first attempt. With this setup I get a 25 km estimated accuracy, with the point being 400 m away from the real location. If I add the URL for https://positon.xyz API to `/etc/geoclue/geoclue.conf`, then the estimated accuracy is 24 m, while the point is about 30 m away. In both cases this is much better than geoIP, which locates me hundreds of kilometers away from the real place.

I wonder though, why the agent is a part of the "demo" package. This name sounds as something that's only intended to demonstrate the correct operation, rather than be the main modus operandi.

gzotti:
OK, so you actively changed the config after installation. It means it doesn't work out of the box but the configuration process can (and must) be described for laypeople in the User Guide, chapter 5.5.4. And it means the old geoIP path must definitely remain working default, at least on Linux. Better 200km off at first launch than nothing. OK, let's continue this after 25.2 is out.

I can again only guess, but these are command-line demos for features that may be available and used in mapping/navigation programs.

10110111:
No,I didn't have to change config to make it work, just start the agent.
The config change just improves precision, but even without it geoclue is
orders of magnitude better than the old freeGeoIP method.

gzotti:
Then still, we have systems (currently: one of yours) where it works out of the box, and systems where it doesn't (one other of yours (meanwhile working), and 2/2 of mine). I hope the OP of #4334 can enlighten us.

PR #4224, opened 2025-03-18
SUG: Document sky culture converter

10110111:
This adds the missing chapter 9.3. There were some issues, like U+02BB being not acceptable by default, and HTML snippets protruding too far to the right when put into `\texttt{}`, so I put them into separate lines formatted as file chunks. The `dso_names.fab` still protrudes, but not as seriously, so I decided to keep it as is.


guide/ch_skycultures.tex
```diff
@@ -712,9 +712,31 @@ \subsection{Deep-Sky Objects Names}
...
+\item[\texttt{-\/-native-locale LOCALE}] Some sky cultures have, in addition to \texttt{star\_names.fab} or \texttt{dso\_names.fab}, localized versions of them, like e.g. \texttt{star\_names.zh\_CN.fab}, that were never actually used but do contain useful information. If you pass the locale (\texttt{zh\_CN} in this example) as the \texttt{LOCALE} parameter, these names will be read and put into the \texttt{native} section of the corresponding JSON entry. Note that the order of stars/DSOs in the normal and localized files much be the same, otherwise there's no way to match the names.
+\item[\texttt{-\/-translated-md}] To check the look of the translated description texts you can use this option. The output directory will contain, in addition to \texttt{description.md}, files named like \texttt{description.es\_419.DO\_NOT\_COMMIT.md}, with the "DO NOT COMMIT" part reminding you that they are not a part of the sky culture.
```
alex-w, line 738: Please change quotes for _DO NOT COMMIT_ string or use slashes to escape it

10110111: What exactly is wrong with these? It gets typeset as expected.

alex-w: It should be ```DO NOT COMMIT''`

alex-w:
OK for me now

alex-w (approved):

gzotti:
For whatever reason I cannot install the RC2. Installer just stops doing anything after a few seconds. Will try to fix some of the overshoots...

gzotti:
Finally calling the installer from the command line worked. Very strange. 
I added a few things. Should be OK for the release.

But I also saw some SCs (started with Anutan) have both mative and translated names tagged "english". I fixed that for Anutan, shall I go through the others?

10110111:
> I fixed that for Anutan, shall I go through the others?

Yes please. These are just artifacts of the old format of the SCs.

PR #4226, opened 2025-03-19
Fix accents in a JSON snippet

10110111:
I think it's better to include the accents than to apologize. Additionally, the JSON data have been updated to match the latest version of the SC.


alex-w (approved):

gzotti (approved):
Thank you for finding this solution, yes!

PR #4237, opened 2025-03-20
Add pronounce and IPA tags in Siberian SC

10110111:
So the recently added SC TODO mentions a lack of pronounce tags in the Siberian SC. This commit adds them. They might be not 100% correct, but reflect most of what I know about the Russian phonetics (however strange this might sound from a native speaker).

Notice how ugly IPA looks in monospaced font. Even in LibreOffice Writer (with proportional fonts) it looks wrong. It's OK in the browser's address bar though... I suppose most translators will use Transifex or poedit, so this might be OK for them.

Also I didn't try rendering it using Qt, but I suppose this may be doubly ugly in our texture-based rendering on the scene.

Thus, we may have to think twice before showing pronunciations of names on the screen.

### DIFF
--- a/skycultures/russian_siberian/index.json
+++ b/skycultures/russian_siberian/index.json
@@ -7,21 +7,21 @@
     {
       "id": "CON russian_siberian Elk",
       "lines": [[67301, 65378, 62956, 59774, 58001, 53910, 54061, 59774]],
-      "common_name": {"english": "Elk", "native": "Лось", "IPA": "ɫ̪os̪ʲ"}
+      "common_name": {"english": "Elk", "native": "Лось", "pronounce": "Los’", "IPA": "ɫ̪os̪ʲ"}
     },
     {
       "id": "CON russian_siberian Kic",
       "lines": [[26241, 26268, 26727, 26311, 25930]],
-      "common_name": {"english": "Threshers", "native": "Кичиги", "IPA": "k̟ʲɪˈt̠̪͡ʲɕ̪iɡ̟ʲɪ"}
+      "common_name": {"english": "Threshers", "native": "Кичиги", "pronounce": "Kichigi", "IPA": "k̟ʲɪˈt̠̪͡ʲɕ̪iɡ̟ʲɪ"}
     },
     {
       "id": "CON russian_siberian DNe",
       "lines": [[17499, 17489, 17531, 17579, 17664, 17851, 17847, 17608, 17499]],
-      "common_name": {"english": "Duck Nest", "native": "Утиное гнездо", "IPA": "uˈt̪ʲin̪ə.jə ɡn̪ʲɪˈz̪d̪o"}
+      "common_name": {"english": "Duck Nest", "native": "Утиное гнездо", "pronounce": "Utinoye gnyezdo", "IPA": "uˈt̪ʲin̪ə.jə ɡn̪ʲɪˈz̪d̪o"}
     }
   ],
   "common_names": {
     "HIP 11767": [{"english": "Golden stake"}],
-    "NAME Venus": [{"english": "Night Star", "native": "Ночная звёздочка", "IPA": "n̪əˈt̠̪͡ʲɕ̪n̪a.jə ˈz̪vʲo.zd̪ə.t̠̪͡ʲɕ̪kə", "translators_comments": "Native name of Venus in Siberian culture"}]
+    "NAME Venus": [{"english": "Night Star", "native": "Ночная звёздочка", "pronounce": "Nochnaya zvyozdochka", "IPA": "n̪əˈt̠̪͡ʲɕ̪n̪a.jə ˈz̪vʲo.zd̪ə.t̠̪͡ʲɕ̪kə", "translators_comments": "Native name of Venus in Siberian culture"}]
   }
 }
--- a/skycultures/russian_siberian/index.json
+++ b/skycultures/russian_siberian/index.json
@@ -21,7 +21,7 @@
     }
   ],
   "common_names": {
-    "HIP 11767": [{"english": "Golden stake", "native": "Золотой кол", "pronounce": "Zolotoy kol"}],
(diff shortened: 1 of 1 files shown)


gzotti:
Ah, I am surprised now. IPA should go into (future) IPA tags. The point here is, European-trained readers cannot read cyrillic. We decided that "pronounce" should provide the basis for english-language readers. You can just rename this to IPA (as first sample) and please add english pronunciation aids.

alex-w:
Please don’t mix `pronounce` and `IPA` fields

skycultures/russian_siberian/index.json
```diff
@@ -7,21 +7,21 @@
...
       "id": "CON russian_siberian Elk",
       "lines": [[67301, 65378, 62956, 59774, 58001, 53910, 54061, 59774]],
-      "common_name": {"english": "Elk", "native": "Лось"}
+      "common_name": {"english": "Elk", "native": "Лось", "pronounce": "ɫ̪os̪ʲ"}
     },
     {
       "id": "CON russian_siberian Kic",
       "lines": [[26241, 26268, 26727, 26311, 25930]],
-      "common_name": {"english": "Threshers", "native": "Кичиги"}
+      "common_name": {"english": "Threshers", "native": "Кичиги", "pronounce": "k̟ʲɪˈt̠̪͡ʲɕ̪iɡ̟ʲɪ"}
```
alex-w, line 15: Obvious fix: Kichigi

skycultures/russian_siberian/index.json
```diff
@@ -7,21 +7,21 @@
...
     },
     {
       "id": "CON russian_siberian DNe",
       "lines": [[17499, 17489, 17531, 17579, 17664, 17851, 17847, 17608, 17499]],
-      "common_name": {"english": "Duck Nest", "native": "Утиное гнездо"}
+      "common_name": {"english": "Duck Nest", "native": "Утиное гнездо", "pronounce": "uˈt̪ʲin̪ə.jə ɡn̪ʲɪˈz̪d̪o"}
```
alex-w, line 20: Obvious fix: Utinoje gnezdo

skycultures/russian_siberian/index.json
```diff
@@ -7,21 +7,21 @@
...
     }
   ],
   "common_names": {
     "HIP 11767": [{"english": "Golden stake"}],
-    "NAME Venus": [{"english": "Night Star", "native": "Ночная звёздочка", "translators_comments": "Native name of Venus in Siberian culture"}]
+    "NAME Venus": [{"english": "Night Star", "native": "Ночная звёздочка", "pronounce": "n̪əˈt̠̪͡ʲɕ̪n̪a.jə ˈz̪vʲo.zd̪ə.t̠̪͡ʲɕ̪kə", "translators_comments": "Native name of Venus in Siberian culture"}]
```
alex-w, line 25: Obvious fix: Nochnaja zvjezdochka

skycultures/russian_siberian/index.json
10110111: Not obvious at all. The "j" will be read differently by German, French, English, and Spanish people. And it definitely isn't [j] in English.

skycultures/russian_siberian/index.json
10110111: And you get the ugliness of the "deyofication": it's zvyozdachka, if we are going by pronunciation and not (vulgarized) spelling.

skycultures/russian_siberian/index.json
alex-w: Yeah, I’ve mixed styles, sorry. It should be: Utinoye gnyezdo

skycultures/russian_siberian/index.json
alex-w: Nochnaya zvyozdochka

10110111:
So OK, let this PR be about IPA then so that we don't lose this. It's tedious to compile the symbols for the words...

Are there any rules we should be using when writing "pronounce" entries? E.g. do we denote stressed syllables, how?

alex-w:
Hmm… I think we should use one of various existing styles (or schemes) for transliteration Russian text onto English in future.

alex-w:
Golden take - Золотой кол - Zolotoy kol

gzotti:
Yes, for languages using Cyrillic, just write how you would find it in English literature or newspapers, so generally without accents. This is why I argued that the pronounce tag should be translatable to adjust for what would be found in transcriptions in French, German, Italian, Spanish, ... . The IPA will be fixed. I am fully aware that IPA is a lot of work (yes, also finding the glyphs...), needs dedicated language experts and a lot of time. But we should be open for it, in the long run.

gzotti (approved):
IPA for Golden Stake is still missing, and you must know better how to transcribe Russian. For me it looks OK for now. (Squash it.)

alex-w:
> Thus, we may have to think twice before showing pronunciations of names on the screen.

I think we may have some special block in description (an automatic generated table?) for names/native names/pronounces/IPAs/etc.

alex-w (approved):
@10110111 please add IPA for star and it can be merged

gzotti:
> I’m not sure that it should be stored per-SC

I am afraid I have thought too long about this to say one fits all. It depends on languages a user knows or wants to know more about. If e.g. Chinese characters are completely illegible to you, you will likely not show them but show Pinyin, and maybe translation. But you may be able to read Cyrillic and want to improve your language knowledge here. Or maybe you are just beginning and need Cyrillic plus transcription (unfortunately called pronunciation now, but this cannot be changed as discussed before), but you don't need the translation any longer. Same goes for any mix of Arab, Korean, Sanskrit or even Cuneiform (when added). Decorative squiggles uselessly obscuring the screen, or meaning? So, here, I don't see how a single setting can be sufficient. With the limitations of the ini format, maybe it's even a new json file in the user's skycultures directory. Or actually, two dedicated sections in config.ini may be enough. And I recommend immediate-store mode for all: easily changed, remembered last setting for next time you activate this SC, and an inadvertent flip does not break anything.

alex-w (approved):

PR #4289, opened 2025-04-16
Add Manual Dependency Download Instructions to README

ultrapre:
#4288


alex-w:
But you can use proxy or prepare builded packages of dependencies also

alex-w (approved):

Issue #3095, opened 2023-03-11
Mollweide Projection


lf0k0nle:
The Mollweide projection is used extensively by professional astronomers to show the celestial sphere when referring to surveys of solar system objects as well as deep sky objects , but is missing from Stellarium 1.2. The Hammer-Aitoff projection comes close, but departs from the Mollweide projection as you approach the celestial poles. Therefore I would like to humbly request that the Mollweide projection be added to Stellarium. Thanks.

CurlyQuokka:
Hi, I'd like to try to implement that, if that's OK with the maintainers. :)

CurlyQuokka:
Hi all! Sorry for a long time wothput info - sadly I had some other thins to attend to. :(
I've tried to implement the feature starting with the code for Hammer-Aitoff projection, but I am still having problem  [with the assertion in AtmospherePreetham.cpp](https://github.com/Stellarium/stellarium/blob/master/src/core/modules/AtmospherePreetham.cpp#L343) . Assertion is not passed when value of x coordinate is greater than 100, I think. If I'll comment out the assertion I can get some display, but no ground is being displayed and there are issues with displaying the atmosphere.

I am afraid that I'll be not able to figure this out (at least not on my own). So if anyone would like to take a look at this issue to give me some feedback or just take it from here my current code is available [here](https://github.com/CurlyQuokka/stellarium/tree/mollweide-projection).

I am really, really sorry for the trouble here. :(

10110111:
> no ground is being displayed and there are issues with displaying the atmosphere

These (and a few other modules) use the shader version of the projector, and your shaders appear to contain C++ code instead of GLSL.

gzotti:
@CurlyQuokka any news, or did you give up on this?

gzotti:
@CurlyQuokka any news, or did you give up on this?

alex-w:
Possible implementation: https://github.com/dusan-prascevic/stellarium/tree/mollweide-projection

dusan-prascevic:
Put this issue down for a while, but I believe I have a working implementation now. I have prepared a PR draft (https://github.com/Stellarium/stellarium/pull/4330).

Feedback is very welcome, let me know if there's anything missing or that needs adjustment

lf0k0nle:
This is great news! Thanks Dusan!

Issue #4328, opened 2025-05-16
Inner planet orbits extend too far when viewed perfectly from the side


Atque:
When Earth passes through the orbital planes of Mercury and Venus (May/November, June/December respectively), their orbits extend far beyond their true extent. See the screenshot below. It does not seem to matter what projection is used.

![Image](https://github.com/user-attachments/assets/b900fd68-a7c5-4c57-a5a1-e7be124fd052)

Current master
Windows 11 24H2
Qt 6.9.0
RTX 4070 Ti Super

10110111:
On the opposite side of the orbit we can see some extraneous lines, which move with respect to the scene during panning and disappear on the next day of simulation:

![Image](https://github.com/user-attachments/assets/ab2077ce-88f9-49ff-848f-ab581eabe1ae)

![Image](https://github.com/user-attachments/assets/01aa9a06-e542-45ce-8893-3c2d4f4e39b7)

10110111:
Might be related: #4321. And might also be related to the change in #4013.

10110111:
> But if you draw on-screen (x, y) from (100, 100) to (1000, 1000) to (351, 349), then the line extend beyond

So the easy reproducer is this:
```diff
diff --git a/src/core/modules/Planet.cpp b/src/core/modules/Planet.cpp
index 82c597f050..795c7ffd65 100644
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -5042,6 +5042,7 @@ void Planet::drawOrbit(const StelCore* core)
 
     sPainter.setLineSmooth(true);
 
+#if 0
     for (int n=0; n<=nbIter; ++n)
     {
         if (prj->project(orbit[n],onscreen) && (vertexArray.size()==0 || !prj->intersectViewportDiscontinuity(orbit[n-1], orbit[n])))
@@ -5056,6 +5057,13 @@ void Planet::drawOrbit(const StelCore* core)
             vertexArray.clear();
         }
     }
+#endif
+    vertexArray.append(100);
+    vertexArray.append(100);
+    vertexArray.append(1000);
+    vertexArray.append(1000);
+    vertexArray.append(351);
+    vertexArray.append(349);
     orbit[ORBIT_SEGMENTS/2]=savePos;
     if (!vertexArray.isEmpty())
     {
```
Looks like there's a bug in `StelPainter` then.

PR #4330, opened 2025-05-18
Implemented Mollweide projection

dusan-prascevic:
Implementation of Mollweide projection (requested in https://github.com/Stellarium/stellarium/issues/3095)

### Description

Implemented Mollweide projection following the standard mathematical formulation (solving for auxiliary angle via Newton-Raphson iteration). I tried to stick to the same approach as for other projections (e.g. handling out of bounds points etc.)

Fixes # (issue) https://github.com/Stellarium/stellarium/issues/3095

image

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Tested the projection visually and via added test functions in `testStelProjector.cpp` (kept the same approach as for testing the other projections)

**Test Configuration**:
* Operating system: Ubuntu 24.04.2 LTS
* Graphics Card: Intel Iris Xe

## Checklist:


.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
```
gzotti, line 1: This file is not required.

Atque:
Hmm, the angular distortions normally expected only far from the center of projection do not seem to change with FOV. Is this intentional?

gzotti:
For me it looks (mostly) good!

Two things: 
- You need to include StelUtils.hpp in StelProjectorClasses.hpp to reach M_PIf. 
- Sometimes there are border lines escaping from the map. (See north pole.)
image

Maybe this relates to #4328, so it would add to "small/harmless gitches" if there is no trivial solution.

alex-w:
Please check artefacts, like this:
stellarium-002

10110111:
This artifact is also present in other discontinuous projections, so can be
left here as well.

alex-w:
> @alex-w so if I understand you correctly, on MacOS the projection ellipse is getting clipped (even at full FOV)? That's strange, I'll look into it!

Yes, but don't worry - this is not a bug in the projection

alex-w:
Why the maximum FOV is 185 degrees?

gzotti:
Compare with Hammer-Aitoff and Sinusoidal. Both have 185°, meaning it just exceeds the from-pole-to-pole vertical fov. The three behave very similar in this respect.

Atque:
Wouldn't 180 degrees be more intuitive? Or is it intentionally limited to just above 180, to get both the poles visible at max FOV?

10110111:
> Wouldn't 180 degrees be more intuitive? Or is it intentionally limited to just above 180, to get both the poles visible at max FOV?

I think it was a tradeoff that enabled a bit more of the sides to be included at some typical screen/window aspect ratios used by the person who decided on these limits. Ideally the limit should be based on horizontal FoV, or maybe on some kind of minimum or maximum of the two.

gzotti:
I think the point of 185 for Hammer/Sinusoidal was having the bottom not interact with the bottom status/button bar. But it's another task to rebalance all projections for whatever new demands arise.

dusan-prascevic:
Thanks to everyone for taking the time to review the PR and provide feedback

PR #4350, opened 2025-06-03
Skycultures 2.0: Zodiac and Lunar Systems

### LINKED ISSUE #437: Indian vedic sky culture: Show longitude for the 27 asterisms
_Original report by vishvAs vAsuki:_ https://bugs.launchpad.net/stellarium/+bug/1710443

Indian vedic sky culture divides the sky into 27 strips (full list, with western equivalents here: https://goo.gl/2PVi28 ), each approximately 360/27 ~ 13 degrees. This division is critical for several purposes including timing rituals and determining the birth lunar mansion/ nakShatra. As it is, users are forced to enable the ecliptic grid and then imagine parallel grid lines running through asterisms in the set of 27. Example pic: https://i.imgur.com/gRcwqcD.jpg . This is clumsy and inaccurate. So, the vedic indian sky culture should show the required 27 grid lines by default.

PS: Provided guidance, I will be glad to assist if needed - I am an active in the sanskrit programming community ( groups.google.com/forum/#!forum/sanskrit-programmers ) .

### LINKED ISSUE #1627: zodiacal starlore

I often want to know what the sign is of the selected date.

A starlore that shows the abstract signs in a band along the ecliptic.

The band would 
- start from the vernal equinox 
-  extend 23.43 (declination of tropics) north and south of the ecliptic.
- It would be divided in 12 squares of 360/12 wide by 2 x 23.43 high 
- and would contain white (or some other colour) symbols in a transparent box.

image

It would also help in better recognising the astronomical symbols of the zodiacal constellations.

I think that this could be a good starter issue for someone who wants to create starlores.

(As a side issue, there exist official astronomical unicode symbols for events such as conjuntion ☌ , opposition ☍, lunar nodes ☊ ☋, planets ☉ ☿♁♂♃♄♅♆ : some of these could be very useful in the Astro calc module; tooltips could then show the normal textual meaning, and pixel space could be won)

A separate calendar, such as http://robinfo.oma.be/nl/efemeriden/jaar-2021/

**Additional context**

https://en.wikipedia.org/wiki/Astronomical_symbols
https://en.wikipedia.org/wiki/Astrological_symbols#Signs_of_the_zodiac

_And  no: I don't believe in astrology - just to make things clear._

gzotti:
### Description
This branch should introduce optional cultural elements: zodiac and various flavours of Lunar stations or mansions. The latter are important in many Asian cultures, while the Zodiac is best known from Mesopotamia to Europe. There are different definitions and variants to be found in the literature which we have to learn and may have to adjust and extend even after the PR will be regarded good for a first release.

Fixes #437, fixes #1627 

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: irrelevant

## Checklist:


sushoff:
@gzotti as we talked about the invention of nakshatras, zodiac etc. yesterday: 

I know, it makes it more complicated but: shouldn't we add a quality like "visibility" for these mathematical concepts? For instance, if the zodiac was invented in 400 BCE, then it must not be displayed before 410 BCE because otherwise people would derive crazy things.

gzotti:
This won't sound very difficult. However, is it possible that someone in 400BC would have tried to compute backward? First things first, this has to work, but we can add some "introduced" tag later.
EDIT: I am sure most users are interested in today's application of these things anyhow.

skycultures/modern/index.json
```diff
@@ -1675,6 +1675,53 @@
...
+              "partitions": [ 12, 30],
+              "extent": 8,
+	      "names": [ { "symbol": "\u2648", "native": "Aries", "english": "Ram"},
+			 { "symbol": "\u2649", "native": "Taurus", "english": "Bull"},
+			 { "symbol": "\u264A", "native": "Gemini", "english": "Twins"},
+			 { "symbol": "\u264B", "native": "Cancer", "english": "Crab"},
+			 { "symbol": "\u264C", "native": "Leo", "english": "Lion"},
+			 { "symbol": "\u264D", "native": "Virgo", "english": "Virgin"},
+			 { "symbol": "\u264E", "native": "Libra", "english": "Scales"},
+			 { "symbol": "\u264F", "native": "Scorpio", "english": "Scorpion"},
```
sushoff: the name is "Scorpius" (Latin, not Italian!)

skycultures/modern/index.json
```diff
@@ -1675,6 +1675,53 @@
...
+	      "link": { "star": 65474, "offset": 180},
+	      "names": [ { "symbol": "LS01", "native": "अश्विनी",             "pronounce": "Ashvini",    "english": "Physician to the Gods"},
+			 { "symbol": "LS02", "native": "भरणी",             "pronounce": "Bharani",    "english": "Bearer"},
+			 { "symbol": "LS03", "native": "कृत्तिका",           "pronounce": "Krittika",    "english": "Pleiades"},
+			 { "symbol": "LS04", "native": "रोहिणी",            "pronounce": "Rohini",     "english": "Red One"},
+			 { "symbol": "LS05", "native": "मृगशिर",            "pronounce": "Mrigashira", "english": "Deer's Head"},
+			 { "symbol": "LS06", "native": "आर्द्रा",             "pronounce": "Ardra",      "english": "Storm God"},
+			 { "symbol": "LS07", "native": "पुनर्वसु ",            "pronounce": "Punarvasu",  "english": "Two Restorers of Goods"},
+			 { "symbol": "LS08", "native": "पुष्य",              "pronounce": "Pushya",    "english": "Nourisher"},
+			 { "symbol": "LS09", "native": "आश्ळेषा/आश्लेषा",       "pronounce": "Āshleshā",   "english": "Embrace"},
```
sushoff: why does "Āshleshā" have diacritica but the other not?

skycultures/modern/index.json
```diff
@@ -1675,6 +1675,53 @@
...
+			 { "symbol": "LS10", "native": "मघा",              "pronounce": "Maghā",     "english": "Bountiful"},
```
sushoff: please make it consistently (either all with or all without diacritics)

skycultures/modern/index.json
```diff
@@ -1675,6 +1675,53 @@
...
+			 { "symbol": "LS20", "native": "पूर्व आषाढा",          "pronounce": "Purva Ashadha",  "english": "First Invincible"},
+			 { "symbol": "LS21", "native": "उत्तर आषाढा",         "pronounce": "Uttara Ashadha", "english": "Later Invincible"},
+			 { "symbol": "LS22", "native": "श्रवण",             "pronounce": "Shravana",       "english": ""},
+			 { "symbol": "LS23", "native": "श्रविष्ठा/धनिष्ठा",         "pronounce": "Dhanishta/Shravishthā", "english": "Most Famous/Swiftest"},
+			 { "symbol": "LS24", "native": "शतभिष/शततारका",      "pronounce": "Shatabhisha Nakshatra", "english": "Requiring a hundred Physicians"},
+			 { "symbol": "LS25", "native": "पूर्व भाद्रपदा/पूर्व प्रोष्ठपदा",   "pronounce": "Purva Bhadrapada",  "english": "First of the Blessed Feet"},
+			 { "symbol": "LS26", "native": "उत्तर भाद्रपदा/उत्तर प्रोष्ठपदा", "pronounce": "Uttara Bhādrapadā", "english": "Second of the Blessed Feet"},
+			 { "symbol": "LS27", "native": "रेवती",            "pronounce": "Revati",            "english": "Prosperous"}, 
+			 { "symbol": "LS27", "native": "अभिजित",          "pronounce": "Abhijit",           "english": "Invincible"}]},			 
   "edges_type": "iau",
```
sushoff, line 1693: really? 

In my view, as the nakshatras represent an own, ~4000 years old culture, they should not be mixed with modern boundaries

src/core/StelSkyCultureSkyPartition.hpp
```diff
@@ -0,0 +1,156 @@
...
+//! "partitions": [ 12, 30],
+//! "extent": 8,
+//! "names": [ { "symbol": "\u2648", "native": "Aries",       "english": "Ram"},
+//! 	       { "symbol": "\u2649", "native": "Taurus",      "english": "Bull"},
+//! 	       { "symbol": "\u264A", "native": "Gemini",      "english": "Twins"},
+//! 	       { "symbol": "\u264B", "native": "Cancer",      "english": "Crab"},
+//! 	       { "symbol": "\u264C", "native": "Leo",         "english": "Lion"},
+//! 	       { "symbol": "\u264D", "native": "Virgo",       "english": "Virgin"},
+//! 	       { "symbol": "\u264E", "native": "Libra",       "english": "Scales"},
+//! 	       { "symbol": "\u264F", "native": "Scorpio",     "english": "Scorpion"},
```
sushoff: dito: Scorpius

skycultures/modern/index.json
gzotti: I quickly copy/pasted the strings from Wikipedia. I don't pretend to read/write native Indian. At this point, in this branch while in draft, I need strings, and will gladly take corrections later. :-)

skycultures/modern/index.json
gzotti: I wanted to look a bit medieval here, whatever is commonly found on old prints. Can be reversed of course.

skycultures/modern/index.json
gzotti: 30 minutes vs. 6 hours. Yes, later.

skycultures/modern/index.json
gzotti: As said, this is a branch in draft. The Nakshatras will not stay in the modern SC, this is just to have the display immediately after launch. Obvious candidate for integration is Vedic, but you can create your isolated version as well. But let me get the implementation right before moving and fixing the json.

src/core/StelSkyCultureSkyPartition.hpp
gzotti: -"-

skycultures/modern/index.json
gzotti: And I need a 100% correct certified source in UTF-8 that allows copy/paste.

gzotti:
OK, we are getting close. GUI is there, but not totally working (fader setup? - and I don't get why Zodiac fade time is frozen at 10s.) The Chinese LM (currently only in Chinese SC) are taken from WP again, it seems they agree with the old Mansion "Borders/Edges" except for LM5. Given that probably all "defining stars" bear the number one in their respective constellation, this line (through a star II) may be wrong in WP, but it needs confirmation by SC experts. 
Is there any historical source explaining the vastly different mansion sizes? I see LM4 very narrow, LM8 quite wide. Does that have to do with Chin. numerology and lucky/unlucky numbers? What about LM20? Super-unlucky?

Currently the mansion lines are linked to stars in current epoch/equinox, lines drawn in equinox of date. Defining stars are marked with a little circle, that was a good idea!

Now the question remains, should the mansions really be set up on epochs, and then lose their defining stars, letting stars slip by precession etc? How would that have been observed? Or were they defined in a way that "from now on these stars define the LM", using celestial poles and equator at time of observation?

Atque:
The labels for Sagittarius and Capricorn are reversed:
image

Atque:
Symbols need to be reversed too; Sagittarius = ♐︎, Capricornus = ♑︎. ;)

alex-w:
> OK, we are getting close. GUI is there, but not totally working (fader setup? - and I don't get why Zodiac fade time is frozen at 10s.) 

The result of quick look of the code of new features - the both faders run incorrectly. Please see using faders in grids and constellations classes for examples.

Atque:
Mirroring the view strangely prints the labels upside-down.
image

gzotti:
> > OK, we are getting close. GUI is there, but not totally working (fader setup? - and I don't get why Zodiac fade time is frozen at 10s.)
> 
> The result of quick look of the code of new features - the both faders run incorrectly. Please see using faders in grids and constellations classes for examples.

Yes, just found the mess. Late-night work... :-)

skycultures/modern/index.json
sushoff: ASE

sushoff:
> OK, we are getting close. GUI is there, but not totally working (fader setup? - and I don't get why Zodiac fade time is frozen at 10s.) The Chinese LM (currently only in Chinese SC) are taken from WP again, it seems they agree with the old Mansion "Borders/Edges" except for LM5. Given that probably all "defining stars" bear the number one in their respective constellation,

by definition (this is the only star that was given with coordinates in hist. Chin. star cat.s 

> this line (through a star II) may be wrong in WP,

not necessarily: the counting has not been canonical prior to the Jesuits

What WP page did you use? You can cross check with our "Chinese Song Dyn." (the one which has a map in description) and "Chinese Chenzhuo" star catalog data. There is almost a millennium between these two historical documents - but both our SCs have the LM as "boundaries". Any potential difference between the two SCs is definitely in the historical source. 

> but it needs confirmation by SC experts. Is there any historical source explaining the vastly different mansion sizes? I see LM4 very narrow, LM8 quite wide.

is correct - see hist. map in our Chin. SC description. 

> Does that have to do with Chin. numerology and lucky/unlucky numbers? What about LM20? Super-unlucky?
> 
> Currently the mansion lines are linked to stars in current epoch/equinox, lines drawn in equinox of date. Defining stars are marked with a little circle, that was a good idea!

Thanks!
> 
> Now the question remains, should the mansions really be set up on epochs, and then lose their defining stars, letting stars slip by precession etc? 

no, I would still anchor the LM at stars (and move with precession and pm).  

> How would that have been observed?

armillary sphere and similar instruments ... there were many, but I would declare their adjustment a source of error and the definition at stars as "correct" / original/ meant to be ... 

> Or were they defined in a way that "from now on these stars define the LM", using celestial poles and equator at time of observation?

yes

sushoff:
### something really IMPORTANT

- Scorpi**O** is Italian 
- Scorpi**US** is Latin 

The Latin term is used for the sign (ecliptic section) and for the constellation in astronomy. 

The bad habit of using the Italian term in popculture-astrology must not be distributed further in Stellarium!!!

skycultures/modern/index.json
sushoff: https://xing.fmi.uni-jena.de/mediawiki/index.php/Indian_Names_(All_Terms)

gzotti:
> ### something really IMPORTANT
> 
>     * Scorpi**O** is Italian
> 
>     * Scorpi**US** is Latin
> 
> 
> The Latin term is used for the sign (ecliptic section) and for the constellation in astronomy.
> 
> The bad habit of using the Italian term in popculture-astrology must not be distributed further in Stellarium!!!

Actually, consulting my Latin dictionary, Scorpio, -onis is the primary Latin form (closer to Greek), and Scorpius, -ii is listed as "poetic variant". It has nothing to do with "Italian". In my (probably incomplete) searches I have the impression that the *zodiacal sign* is more often called "Scorpio" and described as "associated with the constellation Scorpius". 

EDIT: Consulting maps in D. Warner, The Sky Explored.
- Allard ca 1706: Scorpio
- J. Blaeu 1648 Scorpius
- Brentel 1573 (Bavaria): Scorpio
- Brunacci 1687: Scorpione (and this is the Italian form!)
- Andreas Cellarius has Scorpio for the figure and Scorpius in the 30° gradation in the limbus!
- Chr. Cellarius: Scorpius
- Coronelli: Scorpione (Italian!)
- Doppelmayer: Scorpius
- Dürer 1515: Scorpio
- Eimmart 1705: Seems to have sneaked out by not giving a name to the figure (I just cannot see it.)
- Habrecht 1628 Map 2B: Scorpio
- Halley 1678: Scorpius
- Hevelius: Scorpius
- de la Hire (all French names, Le Scorpion)
- Lamb ca 1673 Scorpius
- Plancius, Globe 9 (1649): Scorpio
Note that these are all maps, planispheres or globes. 

Further:
- Gallucci 1588: Table has Scorpius, figure has Scorpio.
- Rost, Atlas Portatilis 1723: Scorpius (Finally, also in the description of the Zodiacal signs!)

Checking some astrolabes: Here it is common that the ecliptic is graduated and labeled with the 12 names of the signs. (Some makers just used the symbol to not get involved in such discussions centuries later ;-)
In Decker, 
- Buot ca 1675: Scorpius

Webster, Western Astrolabes:
- Stöffler 1513: Scorpio
- 1: Scorpius
- 2: Scor9 (read Scor...us)
- 3: Scorpi9 (-ius; same maker)
- 4: Scorpio along the ecliptic, Scorpius in the dorsal limbus
- 5, 6 (Hartmann): Scorpius
- 8, 9, 10 (Arsenius): Scorpius
- 11 (1559): Scorpio
- 12 (1567) Scorpi. (indicates Scorpius)
- 13 (before 1582) Scorpio
- 16 (c.1620) S'orpio (engraver's error fix?)
- 17, 18, 19, 21, 28, 29, 31 Scorpius
- 24 Scorpio on the ecliptic, Scorpius in the dorsal limbus

and finally, 
- Zotti 2003: Scorpius (Phew! :-)

OK, with Arsenius and Hartmann having used Scorpius, I agree to change that.

gzotti:
>>The Chinese LM (currently only in Chinese SC) are taken from WP again, it seems they agree with the old Mansion "Borders/Edges" except for LM5. Given that probably all "defining stars" bear the number one in their respective constellation,
> 
> by definition (this is the only star that was given with coordinates in hist. Chin. star cat.s
> 
> > this line (through a star II) may be wrong in WP,
> 
> not necessarily: the counting has not been canonical prior to the Jesuits
>
> 
> What WP page did you use? 

https://en.wikipedia.org/wiki/Twenty-Eight_Mansions

They list &alpha; Sco, you had defined the line with &sigma; Sco as found also elsewhere. I changed that now to &sigma;

>You can cross check with our "Chinese Song Dyn." (the one which has a map in description) and "Chinese Chenzhuo" star catalog data. There is almost a millennium between these two historical documents - but both our SCs have the LM as "boundaries". Any potential difference between the two SCs is definitely in the historical source.

I have added (copy/pasted) the (currently only single) description to the other Chinese SCs. A few more borders are different, (and currently the new definition goes through the star and equatorial poles of date, not of epoch, while the "abused borders" are frozen in epoch).


> > but it needs confirmation by SC experts. Is there any historical source explaining the vastly different mansion sizes? I see LM4 very narrow, LM8 quite wide.
> 
> is correct - see hist. map in our Chin. SC description.

Yes, but presumably there was a reason for this. There would be enough candidate stars for a more equal distribution. Proper motion or even only precession seems not to be sufficient as "width-changing" factor.  

> > Does that have to do with Chin. numerology and lucky/unlucky numbers? What about LM20? Super-unlucky?
> > Currently the mansion lines are linked to stars in current epoch/equinox, lines drawn in equinox of date. Defining stars are marked with a little circle, that was a good idea!
> 
> Thanks!
> 
> > Now the question remains, should the mansions really be set up on epochs, and then lose their defining stars, letting stars slip by precession etc?
> 
> no, I would still anchor the LM at stars (and move with precession and pm).
> 
> > How would that have been observed?
> 
> armillary sphere and similar instruments ... there were many, but I would declare their adjustment a source of error and the definition at stars as "correct" / original/ meant to be ...

Yes, I expect the same. With this you can easily observe equatorial coords of date, but not some precession-shifted lines after 300 years.

 
> > Or were they defined in a way that "from now on these stars define the LM", using celestial poles and equator at time of observation?
> 
> yes

That means, the current implementation of the new lines is OK (apart from declaring a few different defining stars), and the epoch-frozen old borders are no longer relevant?

gzotti:
Observations: 

Comparing the Song SC's old "Borders" solution to the new system, it seems the "Borders" were computed rather for 100AD, not 1052? Did I miss something? Should it look like that? Then, all stars mismatch.

Comparing the Chenzhuo version for 300AD, there are two different lines: 
- the right (&delta;) instead of left (&zeta;) belt star of Orion, 
- LM14 defined by &zeta; And (?) instead of &eta; And

If the Chinese mansions are defined by the 28 stars, IMHO the lines should be allowed to change their position in the sphere: Always go from current celestial pole to pole and through the star. This seems most plausible to me. If a "freeze the lines in 100AD" is really required/intended, I can add this later.

sushoff:
I'll get back later ...

skycultures/modern/index.json
gzotti: To make it consistent, should I use the column with or without diacritics? Look in LS2, and a few others: You list totally different names in the columns. How do I know what's the correct transliteration of the Sanskrit? You know what: I let you fix the strings. :-) I moved the block into Indian SC now.

gzotti:
> @gzotti as we talked about the invention of nakshatras, zodiac etc. yesterday:
> 
> I know, it makes it more complicated but: shouldn't we add a quality like "visibility" for these mathematical concepts? For instance, if the zodiac was invented in 400 BCE, then it must not be displayed before 410 BCE because otherwise people would derive crazy things.

I am still not sure if it's a good idea, as going back at least a few centuries can help understanding the formation of the systems. If you really want that, please provide a list of "introduced" tags (years with decimal fraction) for:

- "tropical zodiac" (modern SCs)
- Indian lunar stations (nakshatras. What about the earlier 28-station system? [this may be defined later])
- Indian "sidereal" Zodiac (same period? later?)
- Chinese mansions. Probably a very concrete date can be given here, i.e. the maps that we try to replicate. 

I don't know about the Arab stations: are they just the asterisms, or also ecliptic sections that can be defined?

src/gui/ViewDialog.cpp
```diff
@@ -619,10 +619,10 @@ void ViewDialog::createDialogContent()
 	}
 	else
 	{
-		ui->constellationHullsCheckBox->hide();
-		ui->constellationHullsColorButton->hide();
-		ui->constellationHullsThicknessSpinBox->hide();
-		ui->constellationHullsFadeDurationDoubleSpinBox->hide();
+                ui->constellationHullsCheckBox->setEnabled(false);
```
gzotti: My intention really was to not even show these elements to beginning users. Those 20 (?) researchers interested in the convex hulls should enable the setting in config.ini. Showing but disabling make zero sense IMHO.

alex-w: After adding UI elements for Zodiac and Lunar Systems the "Hulls" elements is looks like an "empty' place in UI - this is not very good (I mean the standard look)

alex-w:
@gzotti what about English names for Zodiac? Currently not all Zodiac elements using English names as a synonyms of English names of IAU constellations. I think it should be synced (English names and context)

src/gui/ViewDialog.cpp
gzotti: But we have other "holes" elsewhere in the GUI. Look into the SSO tab. I consider this "hole" here not so critical, and certainly better than filling it with visual but disabled clutter. I would like to just skip your commit on next rebase -i. 

We could think of moving const. hulls below const. boundaries, and zodiac and lunar stations to the right side. This would then have a "hole" in the lower left corner, maybe less disturbing for you. 

What should still come (and was on my task list of further refinements) is to check after switching a SC and (visually) disable zodiac or LS elements if a SC does not have it. Plus, exchanging the label by the actual culture-relevant name. But these are minor details, and can even wait for 25.3, when maybe all terms have been translated.

Speaking of translation, using just one user language setting instead of two is still open, and apparently some issues with missing context (Men, Leo, Ara, ...), so maybe some context rules should be established?

gzotti:
I had the similar discussion around the "native" spelling of the Scorpius/Scorpio above. I am very unsure whether any english-speaking person uses "Sea Goat" as name of the zodiacal sign commonly written as "Capricorn". The same is true for German "Steinbock". Nobody says "Ziegenfisch", unless you are about to explain the Babylonian origin of the figure in a planetarium show. We had user complaints already.

src/gui/ViewDialog.cpp
alex-w: > But we have other "holes" elsewhere in the GUI. Look into the SSO tab. I consider this "hole" here not so critical, and certainly better than filling it with visual but disabled clutter. 

I see no "holes" in the SSO tab (we have no hiding elements here).

> I would like to just skip your commit on next rebase -i.

OK

> We could think of moving const. hulls below const. boundaries, and zodiac and lunar stations to the right side. This would then have a "hole" in the lower left corner, maybe less disturbing for you.

What about visual separation - adding a horizontal line before zodiac and lunar stations?

> What should still come (and was on my task list of further refinements) is to check after switching a SC and (visually) disable zodiac or LS elements if a SC does not have it. 

+1. Good addition
 
> Speaking of translation, using just one user language setting instead of two is still open, and apparently some issues with missing context (Men, Leo, Ara, ...), so maybe some context rules should be established?

The context should be support by Python (management of the `*.pot` files) and C++ code. Right now context is exist in C++ code (at least it expected context from JSON data), but Python code is not extracted elmenets with context (at least abbreviations and names of newly added features are not extracted).

alex-w:
> I had the similar discussion around the "native" spelling of the Scorpius/Scorpio above. I am very unsure whether any english-speaking person uses "Sea Goat" as name of the zodiacal sign commonly written as "Capricorn". The same is true for German "Steinbock". Nobody says "Ziegenfisch", unless you are about to explain the Babylonian origin of the figure in a planetarium show. We had user complaints already.

Supporting of the context is expected here in other words...

gzotti:
Yes, indeed, it seems necessary. Now, can context be formed from "skyculture + zodiac" and "skyculture + lunar_system" automatically, or do we need a "context" tag in the json ? This question may appear in several places now. Even, it appears, down to "skyculture+constellation abbreviation", "skyculture+constellation translated name", "skyculture+constellation pronunciation".

alex-w:
> Yes, indeed, it seems necessary. Now, can context be formed from "skyculture + zodiac" and "skyculture + lunar_system" automatically, or do we need a "context" tag in the json ? This question may appear in several places now. Even, it appears, down to "skyculture+constellation abbreviation", "skyculture+constellation translated name", "skyculture+constellation pronunciation".

It can be formed as will be coded in python code for extraction the translatable data. Right now all translatable data are extracted without context if "context" element is not defined in JSON. Of course we may add one context for whole SC in python and C++ code (Skyculture ID for example) and removed support special JSON elements "context". Or adding JSON elements "context" for all places where context is needed.

@10110111 @sushoff  please your opinion

sushoff:
> Observations:
> 
> Comparing the Song SC's old "Borders" solution to the new system, it seems the "Borders" were computed rather for 100AD, not 1052? Did I miss something? Should it look like that? Then, all stars mismatch.
> 
> Comparing the Chenzhuo version for 300AD, there are two different lines:
> 
> * the right (δ) instead of left (ζ) belt star of Orion,
> * LM14 defined by ζ And (?) instead of η And
> 
> If the Chinese mansions are defined by the 28 stars, IMHO the lines should be allowed to change their position in the sphere: Always go from current celestial pole to pole and through the star. This seems most plausible to me. If a "freeze the lines in 100AD" is really required/intended, I can add this later.

my Chinese colleague confirmed:

"Please use this as principle: Always go from current celestial pole to pole and through the determinative stars."

sushoff:
> > @gzotti as we talked about the invention of nakshatras, zodiac etc. yesterday:
> > I know, it makes it more complicated but: shouldn't we add a quality like "visibility" for these mathematical concepts? For instance, if the zodiac was invented in 400 BCE, then it must not be displayed before 410 BCE because otherwise people would derive crazy things.
> 
> I am still not sure if it's a good idea, as going back at least a few centuries can help understanding the formation of the systems. If you really want that, please provide a list of "introduced" tags (years with decimal fraction) for:
> 
it was a question ... perhaps not

> * "tropical zodiac" (modern SCs)
> * Indian lunar stations (nakshatras. What about the earlier 28-station system? [this may be defined later])
> * Indian "sidereal" Zodiac (same period? later?)
> * Chinese mansions. Probably a very concrete date can be given here, i.e. the maps that we try to replicate.
> 
> I don't know about the Arab stations: are they just the asterisms, or also ecliptic sections that can be defined?

our wizard (or djinn), Khalid, defined sections: see "Arabic lunar stations" SC

gzotti:
> my Chinese colleague confirmed:
> 
> "Please use this as principle: Always go from current celestial pole to pole and through the determinative stars."

And does it mean we can then delete the old edge definition, setting it to "none"?

10110111:
> please your opinion

If context is required in >50% of entries, then I'd go with the fully automated solution. If it's just needed in <25% of entries, I'd prefer manual specification. The cases in between are a judgement call.

gzotti:
> our wizard (or djinn), Khalid, defined sections: see "Arabic lunar stations" SC

Ah, indeed. But I see several shortcomings after 25.1, maybe they can be improved (by name experts, you or Khalid?): 
- constellation names are Arab and translated (english) only, pronounce is missing. 
- numerical constellation abbreviations. Unless there is a good reason like referencing some external index, I recommend short character-based abbrevs.
- With 13/14 degrees wide lunar stations, this must again be coded as new variation. Is it just a list of ecliptical longitudes of J2000, and can this list be provided please (or provide an accessible reference)? Are there sub-divisions? Would it be OK to limit the zone along the ecliptic to a common width (e.g. &pm; 10 or 15°?)
- I recommend repeating the anwa-defining stick figures as asterisms to clearer highlight them.

gzotti:
> > please your opinion
> 
> If context is required in >50% of entries, then I'd go with the fully automated solution. If it's just needed in <25% of entries, I'd prefer manual specification. The cases in between are a judgement call.

I am afraid it's not easy to foresee ambiguous words among all languages, so requiring context everywhere seems the safe way out. It may lead to having a few more duplicates (like repetition in all "modern_*" SCs. Sometimes it appears translators don't read the small-print context. But this is the danger of such community project.

Atque:
Zodiac + lunar mansions should probably only be visible from Earth (Zodiac maybe also from the Moon). I get crazy views from other planets.

skycultures/modern/index.json
gzotti: @vvasuki provided the strings now.

Atque:
Would it be possible for the translated signs of the Zodiac to use new strings? It messes up the Swedish translation, where I want to keep definite forms for Western constellations and zodiacal signs, and indefinite forms for cultures with no real consensus on translations.

image

No one uses the indefinite form "Lejon" about the sign Leo, it's always the definite "Lejonet". And "Krabba" is not even the correct word ("Kräftan" would be), but using that messes up other cultures were crab can't be translated to kräfta. (Strictly speaking, kräfta means crayfish.)

gzotti:
This is what we mean with context above. So, a Crab in context of zodiacal sign might be different from crab as constellation, and see my (resolved) discussion around the two Latin forms of Scorpio/-ius and the (as it turned out) seemingly random usage preferences in historical maps and instruments.  I even wonder if, for reasons of space inside the sign box, we should add a third layer of cultural labelling just for the signs/stations, or a combo where user can select just one of the string parts. This may also come in 25.3, though, when we have more cultures more complete and when the waves around "one translator is enough" and "context everywhere" have calmed. 

If some Swedish terms are entirely wrong, please become active on Transifex.

Atque:
Hmm, what about using the translation of the Modern sky culture as Zodiacal sign? Because now it seems to use the same string as used in e.g. `greek_almagest`. Or could that lead to contradictions anywhere?

And yes, I am active on Transifex.

gzotti:
The names of the zodiacal signs and Lunar stations come per skyculture. But you may want to have native+transcription elsewhere on screen, but with not enough space only one word as sign label. I don't recommend deliberately mixing names from various SCs, but it can happen when no context is given. E.g., there is a star "War" in an Australian SC. No connection to a conflict here, but someone translated it like that! We should just provide systematic and easy to follow rules for the "context" in SCs.

gzotti:
@alex-w , @10110111 please your comments/review? After completion, we should go for the QtPositioning/locationIP issues to fix that before 25.2.

alex-w (approved):
The new feature is OK for me, but it has problem for localization support (though it's not this PR issue)

gzotti:
Yes, I agree that after this is merged, the various shortcomings in translations that appeared with the new features in the last months should be solved, at the latest in 25.3. Some issues like the multiple-scripts-one-language issue (#4365) can hopefully be clarified in translation guidelines.

10110111:
Why is the Zodiac marking enabled by default? No other lines are, why should this one be?

gzotti:
Argh, right. I started with this before I had the GUI and forgot to revert. Will fix ASAP.

alex-w:
> Argh, right. I started with this before I had the GUI and forgot to revert. Will fix ASAP.

I fixed it already

PR #4383, opened 2025-06-21
fix i18n to #4003

ultrapre:


ultrapre:
@alex-w 
I missed some translatable string extraction, can it be merged into the upcoming v25.2 as soon?

alex-w (approved):
Thanks

PR #4385, opened 2025-06-21
Sky culture converter20062025

MherMnatsakanyan03:


MherMnatsakanyan03:
- Added SkyCultureConverter via "convert" button in start dialog
- option to disable it via "-DSCM_SHOULD_ENABLE_CONVERTER=Off"
- automatic disabling when Qt Version 5

- convertion happens in parrallel thread, so that ui-thread is not stuck possibly

PR #4402, opened 2025-07-01
Fix HIP number for Menkar in NavStars

decretist:
I have never created a pull request before, so apologies in advance if I'm not following the right procedure. I've submitted an extremely minor fix, correcting the HIP number for Menkar in the Apollo Navigation Stars list. I also updated the URL for the relevant reference on the Apollo Lunar Surface Journal site.


alex-w (approved):
Thanks for fix!

PR #4459, opened 2025-08-17
Make it possible to choose between multiple planetary HiPS per planet

10110111:
### Description

Currently we can only have one HiPS pack per planet (albedo, normals, horizons). This PR enables multiple HiPS per planet to choose from.

The surveys are arranged into groups, with the groups defined either by `group` property of a set of surveys, or by the hipslist's URL.

To test this, you can use my new Moon survey pack: http://86.106.181.97/hips/hipslist. Additional set of surveys can be obtained from [this URL](https://alasky.cds.unistra.fr/MocServer/query?expr=hips_frame!%3Dequatorial%2Cgalactic%2Cecliptic+%26%26+hips_frame%3D*+%26%26+dataproduct_type!%3Dcatalog%2Ccube+%26%26+hips_service_url%3D*&get=record) (with a nuance that their coordinate system is shifted by 180° of longitude with respect to ours, which should be taken care of in a later change).

If you are going to merge this, please keep the commits separate.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Ubuntu 25.04
* Graphics Card: AMD Ryzen AI 9 HX 370 w/ Radeon 890M

## Checklist:

### DIFF
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -55,6 +55,7 @@ QSizeGrip {
 QTextEdit,
 QPlainTextEdit,
 QListWidget,
+QTreeWidget,
 QLineEdit {
 	border: 1px solid rgb(0, 0, 0);
 	border-radius: 0;
@@ -67,6 +68,7 @@ QLineEdit {
 
 QTextEdit:disabled,
 QListWidget:disabled,
+QTreeWidget:disabled,
 QLineEdit:disabled {
 	background: rgb(90, 90, 90);
 }
@@ -424,6 +426,7 @@ QGroupBox::indicator {
  * indicators, but there's a problem with spacing discussed at
  * https://stackoverflow.com/q/77422984 */
 QListWidget::indicator,
+QTreeWidget::indicator,
 QListView::indicator {
 	margin: 0;
 	background: none;
@@ -431,6 +434,7 @@ QListView::indicator {
 
 QCheckBox::indicator:checked,
 QListWidget::indicator:checked,
+QTreeWidget::indicator:checked,
 QListView::indicator:checked,
 QGroupBox::indicator:checked {
 	image: url(:/graphicGui/uieCheckbox-checked.png);
@@ -438,6 +442,7 @@ QGroupBox::indicator:checked {
 
 QCheckBox::indicator:indeterminate,
 QListWidget::indicator:indeterminate,
+QTreeWidget::indicator:indeterminate,
 QListView::indicator:indeterminate,
 QGroupBox::indicator:indeterminate {
 	image: url(:/graphicGui/uieCheckbox-partial.png);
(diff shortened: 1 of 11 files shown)


gzotti:
Nice addition! I like the various moon maps. Any clue why they are reversed? (Or why ours are reversed if the other ones fulfill some standard?)
I have not found the logic, but sometimes (no, actually often) when activating nomenclature (crater symbols&names) it seems one survey is stuck and does not disable. Then after clicking around some more, disabling nomenclature, it suddenly works again. Not sure if and how nomenclature could have anything to do with that, though. Maybe switch off nomenclature (how/why?) and move time by a few minutes also helped?

Also, frequently a message pops up "This group of surveys doesn't have an albedo map. Can't display it". I then have these trouble to disable displaying the survey.

10110111:
> I have not found the logic, but sometimes (no, actually often) when activating nomenclature (crater symbols&names) it seems one survey is stuck and does not disable.

I don't reproduce this. Could you make a screencast or something to show this?

> Any clue why they are reversed? (Or why ours are reversed if the other ones fulfill some standard?)

I guess theirs should be right, after all they are more actively interacting with the authors of the HiPS standard. As to why ours is different, I suppose it's because Guillaume didn't know better when he was implementing it, and now it's a matter of backwards compatibility. Now I also wonder if I should follow the reversal in my Moon survey too, before it's too late.

> Also, frequently a message pops up "This group of surveys doesn't have an albedo map. Can't display it". I then have these trouble to disable displaying the survey.

OK, the failure to disable the survey after the message is a bug, I'll fix it soon.

As for the frequent popup, it's because you're supposed to disable the group if you don't want any survey to be shown, or just enable another survey from the group if you want to switch to it (the previously selected one will be unchecked automatically). Maybe there's a way to make UX better, any ideas?

Atque:
How does one add your Moon survey? It doesn't appear in the list.

10110111:
> How does one add your Moon survey?

SUG chapter 10.2. 

> It doesn't appear in the list.

It should appear by the group name: `Moon horizons + normals + albedo for Stellarium`.

10110111:
After a few more commits it's even more important to keep them separate when merging. But at least now I've addressed all the comments here as well as a few issues I've found. Are there any objections to merging this?

10110111:
> Any clue why they are reversed? (Or why ours are reversed if the other ones fulfill some standard?)

Just stumbled upon this code in `Planet::drawSurvey`:
```cpp
// Apply a rotation otherwise the hips surveys don't get rendered at the
// proper position.  Not sure why...
painter->getProjector()->getModelViewTransform()->combine(Mat4d::zrotation(M_PI * 0.5));

...

// Undo the rotation we applied for the survey fix.
v = Mat4d::zrotation(M_PI * 0.5) * v;
```
And this comes from the very beginning, #110, namely [this commit](https://github.com/Stellarium/stellarium/pull/110/commits/c7678ce206112fcb9b664919b69e8f17f099ad6c). I'm still not sure what exactly would go wrong if this kludge weren't included. Maybe Guillaume thought that his HiPS were correctly generated while they weren't, and introduced this workaround into the C++ code, breaking the actual HipsGen-generated surveys.

In any case, this is working now, but the code contains two rotations: one in `Planet::drawSurvey`, and another in `StelHips.cpp!shiftPix180deg`, which may be worth trying to simplify in the future.

PR #4496, opened 2025-09-05
Fixed  translated constellation names in ukrainian language

zayats1:
Fixed  translated constellation names in ukrainian language

### DIFF
--- a/po/stellarium-skycultures/uk.po
+++ b/po/stellarium-skycultures/uk.po
@@ -37563,7 +37563,7 @@ msgstr "Орел"
 #. Modern (IAU) constellation, native: Andromeda
 msgctxt "IAU constellation name"
 msgid "Chained Maiden"
-msgstr "Прикута Діва"
+msgstr "Андромеда"
 
 #. Modern constellation, native: Sculptor
 #. Modern (IAU) constellation, native: Sculptor
@@ -37586,7 +37586,7 @@ msgstr "Scl"
 #. Modern (IAU) constellation, native: Ara
 msgctxt "IAU constellation name"
 msgid "Altar"
-msgstr "Вівтар"
+msgstr "Жертовник"
 
 #. Modern constellation, native: Libra
 #. Modern (IAU) constellation, native: Libra
@@ -37598,7 +37598,7 @@ msgstr "Терези"
 #. Modern (IAU) constellation, native: Cetus
 msgctxt "IAU constellation name"
 msgid "Sea Monster"
-msgstr "Морське Чудовисько"
+msgstr "Кит"
 
 #. Modern constellation, native: Aries
 #. Modern (IAU) constellation, native: Aries
@@ -37627,7 +37627,7 @@ msgstr "Sct"
 #. Modern (IAU) constellation, native: Pyxis
 msgctxt "IAU constellation name"
 msgid "Mariner Compass"
-msgstr "Компас Моряка"
+msgstr "Компас"
 
 #. Abbreviation of constellation in Modern sky culture
 #. Abbreviation of constellation in Modern (Chinese) sky culture
@@ -37650,7 +37650,7 @@ msgstr "Волопас"
 #. Modern (IAU) constellation, native: Caelum
 msgctxt "IAU constellation name"
 msgid "Engraving Tool"
-msgstr "Гравер"
+msgstr "Різець"
 
 #. Abbreviation of constellation in Modern sky culture
 #. Abbreviation of constellation in Modern (Chinese) sky culture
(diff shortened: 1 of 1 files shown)


alex-w:
@yurchor please check it

yurchor:
> @yurchor please check it

Actually, that's replacing translations with IAU names of constellations. Should be rejected.

10110111:
@gzotti Why do we have `msgctxt "IAU constellation name"` for a name that it isn't supposed to be translated as an IAU constellation? Also, is it intended that Modern (IAU) has `english` for the names (e.g. Sea Monster) that are `byname`s in Modern?

gzotti:
The (native Latin) IAU constellation names can be translated to national spelling variants of the canonical IAU name. English took over the original term, but we had the case of Lithuanian which finally triggered the byname separation.

OTOH, yes, the Sea Monster, Hero,  etc should go into byname everywhere.

zayats1:
> > @yurchor please check it
> 
> Actually, that's replacing translations with IAU names of constellations. Should be rejected.

to be fair: IAU names are the only names I learn in the astronomy literature and videos.  
And I can`t search the exact constellations with current translations 

image
image

image


ps: russian translation have original names: Perseus -> "Персей", not "Герой"

zayats1:
> OTOH, yes, the Sea Monster, Hero, etc should go into byname everywhere.

100% agree as an amateur astronomer/stargazer

gzotti:
As mentioned numerous times, the descriptive names (tradition since al-Sufi) were welcome by some researchers and non-Europeans. I had assumed all Europeans would just happily use the original Latin names, but given the dismay expressed by some Europeans with this solution, we have now added the byname add-on to allow national translation of the English "actual" name (Latin IAU spelling now used in English). This is visible in current betas. 

If Russian translation now does "Hero->Perseus", it should be adapted for 25.3!

@zayats1 usually translations are processed via Transifex.

10110111:
Well it's Ukrainian, not Russian. But anyway, the point that beta should be
used still stands.

As for Transifex, just imagine what we'd get here if the translations were
just silently updated to the incorrect ones.

gzotti:
@zayats1 had written (as reference) that the Russian translation also does not follow the actual wording. 

On Transifex, we have responsible translators per-language who have the power to approve or fix the correct translation. We must rely on those and their responsibility. Once a translation has been approved, users can still raise a question there to suggest a better translation.

yurchor:
From my point of view, the current state of the art is unbearable. I cannot predict what the Stellarium developers mean if the instruction is "Do not translate as it is, translate otherwise". The recent updates are rather trashy with the rarest sky cultures on Earth and their widest explanation with the contents of mininovel, confusing rewriting of the names, and bunches of typos and bad English.

gzotti:
We are all only humans, make errors and learn to improve. The SC development is topic of the year, so of course brings lots of extra work. Typos should be reported (or even better, PR'ed). I explained the misexpectation of development and SC researchers around the Sufi names and hope it has been rectified via bynames. Just that now byname "Hero" should not be back-translated to "Perseus" (etc.) The grief applies to about measly 5-7 strings and may take 10 minutes per language to repair.

gzotti:
> @gzotti Why do we have `msgctxt "IAU constellation name"` for a name that it isn't supposed to be translated as an IAU constellation? Also, is it intended that Modern (IAU) has `english` for the names (e.g. Sea Monster) that are `byname`s in Modern?

"IAU constellation name" indicates that the usual user-language translation should use whatever a language uses canonically for that constellation. But indeed with the latest developments there is a stupid divergence. 

Our main source for the "Modern (IAU)" SC is no longer available, but may have been archived in https://iauarchive.eso.org/public/themes/constellations/ (Should we change the reference link 1?)

- Most users who reacted did so to state they want to see "Orion", not "Hunter", etc. For some languages, apparently, just displaying the canonical Latin spelling is not an option. This is why 'byname' has been introduced and can be used in our "Modern" SC, which is a melting pot of tens of sources and attempts to serve most users (the general observing amateur astronomer community without special interest in other cultures' constellations). 
- If we want to authentically provide what IAU has to offer (and this is the idea behind a "single"-classified SC), we actually should use the names that shocked some users in the "english" names, which should then be translated as regular name to user language, and not even use bynames as the used IAU source list does not flag such extras, they just use them as english names. But as of now it's washed out, maybe by some oversight in its previous commit. Andromeda is OK for "common users" but breaks our rule of "represent what's in the source". Orion, Pegasus, Perseus are still what IAU prescribes (in-line with "single" representation) with what we now wrote into byname in our default Modern SC. 

OTOH, also they refer to S&T as source of the pronunciations which look horrible IMHO. I have added these only to the Modern (S&T) SC so far, as US-Americans seem to need them most. Of course, any other language must write their own version for this! (And they don't have to imitate US-American pronunciation...)


I tend to prefer representing the "single"-classified SCs as close as possible to the source, and to actually revert the reversion, i.e., use "Chained Maiden" in the 'Modern (IAU)' SC because it is written like that in the source (and indeed remove the byname in this SC, as it's just not there!). Use default 'Modern' to see what you expect to see from your decades of tradition. If anything, also here (when this website is given as master source) the US-American pronunciations might be added, but then I wonder what's the remaining difference between IAU and S&T SCs.  When we introduced it, I think the IAU SC did not have lines (as the IAU has not ratified lines.)

yurchor:
Can this be closed? I promise to provide appropriate translations where it's needed.

PR #4507, opened 2025-09-10
CLI: Try to inhibit screensaver in fullscreen (Windows only)

gzotti:
### Description

When running Stellarium in fullscreen mode for a presentation, it may annoy when the screensaver kicks in after a minute.
A new command-line option should inhibit this. 

Idea taken from https://forum.qt.io/topic/38504/solved-qdialog-in-fullscreen-disable-os-screensaver/20

Enabled for Windows for now. I don't know how to do this in Mac/Linux. Some old Mac solution is on this website. 

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -62,6 +62,8 @@
 #include <QStorageInfo>
 #ifdef Q_OS_WIN
 	#include <QPinchGesture>
+	#include <Windows.h>
+	#include <WinUser.h>
 #endif
 #include <QOpenGLShader>
 #include <QOpenGLShaderProgram>
@@ -1475,10 +1477,31 @@ void StelMainView::initTitleI18n()
 void StelMainView::setFullScreen(bool b)
 {
 	if (b)
+	{
 		showFullScreen();
+#ifdef Q_OS_WIN
+		if (qApp->property("onetime_inhibit_screensaver").toBool())
+		{
+			qDebug() << "Disabling screensaver while in fullscreen";
+			SystemParametersInfo(SPI_SETSCREENSAVEACTIVE, FALSE , NULL, SPIF_SENDWININICHANGE);
+		}
+		else
+			qDebug() << "Not touching screensaver business";
+#endif
+	}
 	else
 	{
 		showNormal();
+#ifdef Q_OS_WIN
+		if (qApp->property("onetime_inhibit_screensaver").toBool())
+		{
+			qDebug() << "Re-enabling screensaver when leaving fullscreen";
+			SystemParametersInfo(SPI_SETSCREENSAVEACTIVE, TRUE , NULL, SPIF_SENDWININICHANGE);
+		}
+		else
+			qDebug() << "Not touching screensaver business";
+#endif
+
 		// Not enough. If we had started in fullscreen, the inner part of the window is at 0/0, with the frame extending to top/left off screen.
 		// Therefore moving is not possible. We must move to the stored position or at least defaults.
 		if ( (x()<0)  && (y()<0))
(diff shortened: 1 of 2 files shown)


10110111:
It would be better to put this code into a separate function, taking as a parameter whether the window is fullscreen, than to litter the two branches with platform-specific code.

On the Linux side you'll have to check what windowing system is in use. In Wayland there's the `zwp_idle_inhibit_manager_v1`, which *seems* to be controllable via Qt's `QWaylandIdleInhibitManagerV1`. In X11 there appears to be an [Idle Inhibition Service Draft](https://people.freedesktop.org/~hadess/idle-inhibition-spec/index.html) spec implemented by [xssproxy](https://github.com/vincentbernat/xssproxy) that we could try using (not by installing the daemon, but simply by supporting the protocol it implements).

gzotti:
Seems it does not even work on the laptop where I need it. Maybe there are still other mechanisms.

maybe this works: https://gitlab.com/qt-goodies/qscreensaver

alex-w:
VLC has code for suppressing screensaver on all platforms.

gzotti:
https://github.com/Simbiat/SnoozeGuard works for now. An integrated solution may come later...

gzotti:
> but simply by supporting the protocol it implements).

Wouldn't that finally require implementing D-Bus support? (It was a long-standing wish, but without the "killer app" to show off its capabilities. The RC plugin API is used instead...)

10110111:
> Wouldn't that finally require implementing D-Bus support?

Yes it would. But given how tiny the use case is, the dependency on `libdbus` should be optional.

gzotti:
This works as originally intended, on Windows for now, and is ready to merge. It's a limited use case, if users on the other OSes are interested, somebody else working on those platforms might be able to do that later. (I am still waiting for somebody on Mac fulfilling the Syphon request that was answered on Windows with Spout...)

10110111:
I think the screensaver inhibition code should be moved to its own class, like `ScreenSaverInhibitor`, together with the timer and the methods.

gzotti:
This is just a timer and two short methods, one of them for the timer. Isn't that overengineering?

10110111:
> Isn't that overengineering?

Well, this is a good question. Given that this is highly platform-specific and has a potential to get more platform-specific code for other platforms, I don't like having all this bunch of `#ifdefs` polluting `StelMainView`. This is why I wanted to isolate this a bit.

But I'm not insisting. If you feel it's not worth it, OK.

gzotti:
If it becomes clearer what has to be done on the other platforms (and maybe the methods need to expand?), it may be good to then move all these methods to another class later. Whoever does it: I can imagine it's something along similar OS specific calls on Mac, and two different paths for X11 and Wayland. I have no hardware for the first, and no time now (and need) for the latter.  Are there then even more variations? For now I'd just squash and merge this little change.

10110111:
> For now I'd just squash and merge this little change.

OK. But please change your new `connect()` calls to use function pointers instead of strings. This relic of the past is too problematic to maintain, since it breaks silently when functions get changed or removed.

src/StelMainView.cpp
```diff
@@ -1996,3 +2003,46 @@ QRectF StelMainView::setWindowSize(int width, int height)
...
+// take action to inhibit screensaver when display is running in fullscreen mode
+// private slot. Connect to fullScreenChanged(b)
+void StelMainView::disableScreensaver(bool fullscreen)
+{
+    if (fullscreen)
+    {
+        qInfo() << "Disabling screensaver while in fullscreen";
+        screensaverInhibitorTimer = new QTimer(this);
+        screensaverInhibitorTimer->setTimerType(Qt::VeryCoarseTimer);
+        connect(screensaverInhibitorTimer, SIGNAL(timeout()), this, SLOT(bumpScreensaver()));
```
10110111: This one too.

gzotti: There are some other older-style connects. Shall I update those as well or do you want to isolate them into another later commit?

10110111: Better isolate. Since we're not moving the code to a separate file, it's better for the future implementors of the code for other platforms to know what to change, and for this a smaller diff is better.

gzotti:
@alex-w any comments?

Issue #4493, opened 2025-09-03
Arabic Al-Sufi lunar mansions' names are shifted with respect to their sections


10110111:
### Expected Behaviour

The labels of the lunar mansions should be centered in the mansions' sections. Ideally, the number should be fixed to the name.

### Actual Behaviour

In Arabic Al-Sufi, when Native and Translated names are on, the Arabic name is centered, while the translated one is attached to its left and, depending on the zoom level, overlaps with the mansion number.

<img width="1920" height="1044" alt="Image" src="https://github.com/user-attachments/assets/ea0ce50c-f74f-414a-8e5d-9c9ec94a682b" />

### Steps to reproduce

1. Switch to Arabic (Al-Sufi) sky culture
2. Make sure lunar mansions are enabled
3. Enable Native and Translated names (I did so in both rows)
4. Zoom out to see several mansions at the same time
5. Zoom in gradually, notice how the names are shifting with respect to mansion numbers, and how decentered they are.

### System
* Stellarium version: v25.2-189-g7093812068
* Operating system: Ubuntu 20.04
* Graphics Card: Intel UHD Graphics 620
* Screen type (if applicable): 1920×1080, no scaling

### Logfile 

[log.txt](https://github.com/user-attachments/files/22119550/log.txt)

gzotti:
Yes, with the Arab names now completed, I also have seen the issue around RTL languages. 
- The canonical layout of labels should have been to have the native (original glyph) part always on the left side. 
- As mentioned in #4339, apparently there are font metrics issues which cut off strings under some unknown circumstances (is that the string->texture conversion with too small texture?) Maybe the cuneiform fonts don't provide bounding box values? Or whatever font mixing happens behind the screens, maybe this fails?
- Probably we need another particle setting for the Zodiac/LunarStation strings to avoid them to become too long. Presumably users would here only use one part for brevity, so a simple 5-value combo could be enough. And separate font sizes.

Not sure how to solve this. Should the string->texture conversion work with the name parts, and assemble larger textures from the name part textures?

10110111:
A wild guess is that maybe the bounding box is computed correctly, but has large negative offsets (because the Arabic text does look well centered). I haven't checked the code though, maybe this is already taken into account.

gzotti:
I just see the ConstellationMgr and AsterismMgr have their own fonts, modify sizes, but never set the actual font. This looks more than weird to me.

alex-w:
[...]

Issue #4339, opened 2025-05-25
Bab. SCs: add cuneiform in "Native" tags


sushoff:

sushoff:
done for Bab.MUL.APIN (almost complete)

- [x] let colleagues check 
- [x] copy for Bab.Seleucid

alex-w:
@sushoff any progress here?

sushoff:
yes, most of them are added ... 

I was about to close the issue but there are some uncertainties which will be worked on for the Sept. release - anyway no need to delete the native-tags

gzotti:
@sushoff will we see this in 25.3?

sushoff:
oh! yes, I just realized that I had added them only for Bab-MUL.APIN and not Bab-Seleucid. 

Scheduled it next.

gzotti:
Uh, I just see we will need to solve the font issue. @alex-w, @10110111  I downloaded and installed the Noto Sans Cuneiform font and set it as application font. But I still see only the font rectangles. Any ideas? Or @sushoff do you see the characters?

sushoff:
yes, I see cuneiform (not tofu) - that used to be the case since May or so ... but Assyriologists may blame us (me) for using such an old font (Old Babylonian instead of NeoAssyrian). 

Here is a screenshot that I had made some months ago for my work report:

![Image](https://github.com/user-attachments/assets/26d89a19-81ce-4574-97aa-f7cc65b4d6fc)

sushoff:
we had agreed some months ago that it is not worth to include all fonts that someone might want in the Stellarium download package (@10110111 had found the size and it was obviously unacceptable)

Is there any option to "load fonts on demand" for SCs?

10110111:
> I downloaded and installed the Noto Sans Cuneiform font and set it as application font. But I still see only the font rectangles. Any ideas?

Which OS are you on? I have NotoSansCuneiform-Regular on Ubuntu 20.04 and the cuneiforms are shown correctly in Stellarium (with DejaVu Sans being the app font).

> @10110111 had found the size and it was obviously unacceptable

This was for the whole pack of Noto fonts. Noto Sans Cuneiform Regular is 802K, which may be OK.

> Is there any option to "load fonts on demand" for SCs?

First we need to make sure that we know how to make them work on all platforms...

gzotti:
[...]

PR #4510, opened 2025-09-11
GUI: Fix handling of app font change

gzotti:
### Description

StelApp is the only element which ever accepts a font change.
Currently, several StelModules have their own fonts which may get initialized at startup, but then only size was changed.

When switching font, the font change should become effective immediately, not only after relaunch.

This is a first step to hopefully fix the font issues which appeared with Skycultures demands.

This branch should be used to
- [x] make font changes effective on screen labels immediately
- [x] fix issues with missing glyphs (#4339) and 
- [x] fix issues with cropped strings (#4339) and 
- [x] fix bounding box issues (#4493).

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/StelSkyCultureMgr.cpp
+++ b/src/core/StelSkyCultureMgr.cpp
@@ -974,12 +974,41 @@ QString StelSkyCultureMgr::createCulturalLabel(const StelObject::CulturalName &c
 					       const QString &commonNameI18n,
 					       const QString &abbrevI18n) const
 {
+	// Each element may be in an RTL language (e.g. Arab). However, we want a canonical order of left to right elements.
+	// This requires building Unicode isolation cells.
+	// Unicode constants from Unicode Standard Annex #9, section 2.
+	//static const QString LRE{"\u202a"}; // Left-to-right embedding: Treat following text as embedded left-to-right
+	//static const QString RLE{"\u202b"}; // Right-to-left embedding: Treat following text as embedded right-to-left
+	//static const QString LRO{"\u202d"}; // Left-to-right override: Force following characters to be treated as strong left-to-right chars.
+	//static const QString RLO{"\u202e"}; // Right-to-left override: Force following characters to be treated as strong right-to-left chars.
+	//static const QString PDF{"\u202c"}; // Pop directional formatting: terminate scope of last LRE/RLE/LRO/RLO
+	static const QString LRI{"\u2066"}; // left-to-right isolate: Treat following text as isolated and left-to-right
+	static const QString RLI{"\u2067"}; // right-to-left isolate: Treat following text as isolated and right-to-left
+	static const QString FSI{"\u2068"}; // First strong isolate: Treat following text as isolated and in the direction of its first strong directional character.
+					    // ASSUMPTION: Can be used as autodetect feature? Mark all parts inside embeddings?
+	static const QString PDI{"\u2069"}; // Pop directional isolate: terminate scope of last LRI/RLI/FSI
+	//static const QString LRM{"\u200e"}; // left-to-right mark: zero-width char
+	//static const QString RLM{"\u200f"}; // right-to-left mark: right to left zero-width non-Arabic char
+	//static const QString ALM{"\u061c"}; // right-to-left mark: right to left zero-width Arabic char
+
+	StelObject::CulturalName lName;
+	// copy over filled elements from cName, but enclose each with Unicode isolation markers.
+	if (!cName.native         .isEmpty()) lName.native          = FSI+cName.native+PDI;
+	if (!cName.pronounce      .isEmpty()) lName.pronounce       = FSI+cName.pronounce+PDI;
+	if (!cName.pronounceI18n  .isEmpty()) lName.pronounceI18n   = FSI+cName.pronounceI18n+PDI;
+	if (!cName.transliteration.isEmpty()) lName.transliteration = FSI+cName.transliteration+PDI;
+	if (!cName.translated     .isEmpty()) lName.translated      = FSI+cName.translated+PDI;
+	if (!cName.translatedI18n .isEmpty()) lName.translatedI18n  = FSI+cName.translatedI18n+PDI;
+	if (!cName.IPA            .isEmpty()) lName.IPA             = FSI+cName.IPA+PDI;
+	if (!cName.byname         .isEmpty()) lName.byname          = FSI+cName.byname+PDI;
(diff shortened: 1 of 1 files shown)


gzotti:
It seems at least the full strings are now again visible, and font change is immediately visible.

gzotti:
@10110111 @alex-w do you see Cuneiform even if you have set Noto Sans as app font? I must switch to Noto Sans Cuneiform, and then vertical spacing is weird.

The bounding box issues may depend on QString's append and handling of mixed LTR/RTL strings. Are you familiar with the problem? I would like to have the native string always on the left side.

10110111:
> do you see Cuneiform even if you have set Noto Sans as app font?

Yes, I set Noto Sans in the rightmost combobox, and the cuneiforms are still visible. Not sure what the one on the left does, but it's set to Latin by default, and doesn't contain any cuneiform option.

> The bounding box issues may depend on QString's append and handling of mixed LTR/RTL strings. Are you familiar with the problem? I would like to have the native string always on the left side.

I'm not familiar with the issue, but what I'd try is render the separate strings to separate surfaces, and then join them in the desired order physically (rather than typographically), separated by the width of a space character.

gzotti:
> > do you see Cuneiform even if you have set Noto Sans as app font?
> 
> Yes, I set Noto Sans in the rightmost combobox, and the cuneiforms are still visible. 

Argh, this is different on Windows then. I was looking for optional OS features like "international support", but did not find anything relevant (aside from eastern Asian fonts of course)

> Not sure what the one on the left does, but it's set to Latin by default, and doesn't contain any cuneiform option.

The left combo allows setting a font family, and then the right then provides the list which fonts contain the respective glyphs, with samples where applicable. Indeed, no Cuneiform family here, and the Noto Cuneiform also has no glyph samples. It seems that some font features are optional and missing.

> 
> > The bounding box issues may depend on QString's append and handling of mixed LTR/RTL strings. Are you familiar with the problem? I would like to have the native string always on the left side.
> 
> I'm not familiar with the issue, but what I'd try is render the separate strings to separate surfaces, and then join them in the desired order physically (rather than typographically), separated by the width of a space character.

That was an option I see as last-resort, but sure, it's a way to go. Unless we want to keep "gravity" labels.

... Grmpf - Gravity labels don't work with Cuneiform! Just showing question mark roadsigns.  OTOH, Arab strings are placed to the left where I want them. ... BUT character-reversed! OMG, new features, new nightmares...

EDIT:

Maybe this helps for label collation:
- QChar::Direction (just test first character of each sub-string?), and if needed, insert
- https://www.unicode.org/reports/tr9/#Implicit_Directional_Marks

src/main.cpp
```diff
@@ -484,6 +491,7 @@ int main(int argc, char **argv)
 	qDebug() << "Max thread count (Global Pool): " << QThreadPool::globalInstance()->maxThreadCount();
 	// Share available cores with the TextureLoader and other jobs
 	//QThreadPool::globalInstance()->setMaxThreadCount(qMax(1,QThread::idealThreadCount()/2-1));
+	app.dumpObjectInfo();
```
alex-w: Seems `if` block is missing here

alex-w:
_Qt5_: error: cannot call member function ‘QStringList QFontDatabase::families(WritingSystem) const’ without object

src/main.cpp
gzotti: No, the line is bogus. No idea how it came here.

gzotti:
> _Qt5_: error: cannot call member function ‘QStringList QFontDatabase::families(WritingSystem) const’ without object

I am developing this for diagnostics. Not sure we will ever need it in the end, or it can just be encircled with ifdefs.

10110111:
Just checked on Windows 11 Home installation that has never been on the Internet, and Notepad shows cuneiform characters regardless of the font configured in its settings, even Times New Roman. OTOH, Stellarium shows empty rectangles instead of cuneiforms in all labels, info strings and input fields.

So the conclusion is that some cuneiform-enabled font is already present in the system (apparently, it's _Segoe UI Historic_), but Qt isn't able to pick it up. Maybe it needs some special configuration for this, I don't know.

gzotti:
Thanks for this. Seems we need to understand a few things around Qt's Unicode and font handling much better.  :-(

gzotti:
This creates good labels and IMHO solves #4493. However, it may be better to do that FSI/PDI enclosing already when reading the culturalName from file, and change that from struct to object, with getters and setters that guarantees such embeddings. @alex-w ,  @10110111 opinions?

I have not started changing code around gravityLabels. Probably the complete label will have to be split on FSI/PDI regexps, and where applicable, the angular rendering handled in some yet-to-invent ways.

gzotti:
@10110111 's hint on QFontDatabase::addApplicationFallbackFontFamily() nailed it. No need for extra fonts, just set one as fallback. My previous question about when to place the Unicode isolation characters remains open. And again, GravityLabels are not fixed in this PR.

10110111:
Note that `QFontDatabase::addApplicationFallbackFontFamily` is only available since Qt 6.8. If we want to support older Qt versions, we need to add a version check. Of course, our official Qt6-based package then must use Qt 6.8+.

gzotti:
Uh, again thanks for that hint. I hope it is no problem to set the Windows CI to Qt6.8 also for Arm ( @alex-w ?) I will add the version check, and the feature will just not work nicely on Qt5 builds.

alex-w:
It's interesting - I see hieroglyphs and cuneiform on the mac when DejaVue Sans font selected :)

alex-w:
> I hope it is no problem to set the Windows CI to Qt6.8 also for Arm ( @alex-w ?) 

I tried use Qt6.8 for arm, but without success

alex-w (approved):

gzotti:
> It's interesting - I see hieroglyphs and cuneiform on the mac when DejaVue Sans font selected :)

Apparently just what Linuxians experience. Either your version of DejaVue Sans contains the glyphs, or a fallback handler becomes active. This is what the latest change did to Windows: declare which font to use to retrieve missing characters of a particular character class. Finally I can again set any font (decorative handwriting, blackletter, ... however ugly I want my UI to be), and the Cuneiform letters come from Segoe UI Historical.

gzotti:
I will rebase a bit before merging. But happy to have reached at least this stage in 25.3.

PR #4515, opened 2025-09-13
Add support for Windows virtual keyboard

### LINKED ISSUE #4478: No virtual keyboard on stellarium with tactile PC
Dear all
I use stellarium software with my microsoft surface go2.
No problem when I use the physical keyboard.
But I want also want to use stellarium without physical keyboard, the "tablet mode" of my surface go.
When I unplug the physical keyboard, the virtual keyboard never appears on Stellarium unlike the classical softwares (word, excel etc).
So I can't type anything... For example I can't type an object number/name to search it.
Does anyone knows how to solve this ?
Note that this thread has been discussed shortly today here : https://github.com/Stellarium/stellarium/discussions/4477 it has been reproduced with other equipment than surface go, and it seems there was an old patch to solve the issue. Does anyone have it ?
Regards
Raphaël

10110111:
### Description

This adds support for calling `osk.exe` to get an on-screen keyboard whenever a `QLineEdit`, `QComboBox` or `QDoubleComboBox` are in focus, and closing the keyboard window when they are out of focus. This is not the normal keyboard provided by `TabTip.exe`, the _Touch Keyboard and Handwriting Panel_, but instead the _Accessibility On-Screen Keyboard_.

While the normal keyboard may be possible to use, [the method](https://superuser.com/a/1742459) is undocumented and may break in newer Windows versions, including by crashing the app implementing it (because of relying on a particular undocumented ABI). So this PR doesn't do this.

Fixes #4478

### Screenshots

Screenshot 2025-09-13 155420

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11 Home

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -57,11 +57,16 @@
 #include <QTimer>
 #include <QWidget>
 #include <QWindow>
+#include <QSpinBox>
+#include <QLineEdit>
 #include <QMessageBox>
+#include <QDoubleSpinBox>
 #include <QStandardPaths>
 #include <QStorageInfo>
 #ifdef Q_OS_WIN
 	#include <QPinchGesture>
+	#include <windows.h>
+	#include <shellapi.h>
 #endif
 #include <QOpenGLShader>
 #include <QOpenGLShaderProgram>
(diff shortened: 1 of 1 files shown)


gzotti:
Just tried. 
- With the F3 search (lineEdit), it seems reasonable. After entering, keyboard does not vanish though.
- With time panel, clicking the day up-arrow calls the keyboard and then repeats the up-arrow click action, i.e., day continues to increase. One could argue to call the keyboard (limit to numbers?) when the click is exactly in the number field. I also dislike the obtrusiveness. The time panel can be left on screen, but preferably without keyboard. 

Does anybody really need it? Or is it technical debt that just has to be maintained for the sake of it? As found in #4478 the operating system's virtual keyboard does work after all.

10110111:
> After entering, keyboard does not vanish though

As long as the focus is in the input field, the keyboard stays. All as expected.

> Does anybody really need it? Or is it technical debt that just has to be maintained for the sake of it?

I'm not really sure. Since even the OP of the issue says that the solution without this PR is sufficient, I don't think we need this anymore. (And a downside of this accessibility keyboard is that it's slower than the normal one.)

gzotti:
> As long as the focus is in the input field, the keyboard stays. All as expected.

The panel is configured to vanish after search. So, the dialog is hidden, just the keyboard covers 1/3 of tablet screen. Testing this on my big screen, I have not checked how this appears on the small notebook. Maybe the kbd even covers the dialog? 

If you also are in doubt, I agree to close it.

PR #4517, opened 2025-09-15
Use a more realistic texture of the Sun

10110111:
### Description

Current texture of the Sun looks cartoonish and doesn't reflect the real look of the Sun. It not only has wrong colors, but also wrong granule sizes and sunspot sizes, locations, and contrast.

The new texture is made from the SDO HMI observations, which automatically fixes these issues. The texture is a result of stitching the maps obtained from multiple HMI images, with the equatorial region containing the sunspots taken from observations from 2025-08-07 to 2025-08-31, while the rest is filled with spotless data of 2019-12-15. All the source images were considered valid only in the central 60% of the solar disk, because outside that region the faculae become more prominent, which is undesired for such a generic texture expected to be viewed from different angles.

The faculae are not displayed, since they'd require special processing to make them more visible near the rim and less visible in the center.

### Screenshots

stellarium-018

stellarium-019

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Ubuntu 20.04 LTS


alex-w:
Hmm - _8k_ x _4k_ and `wepb` format - what about minimal VRAM for GPU? 

What about license for texture of the Sun (editing CREDITS.md)?

P.S. Almost 13 MB for file - could be file size reduced without losing quality?

gzotti:
Looks impressive! However, IIRC ANGLE systems (and OpenGL ES SBCs) are limited to 2k textures. Should we keep the old one as fallback?

10110111:
> Hmm - 8k x 4k and wepb format - what about minimal VRAM for GPU?

There's a decimation option for low-resources machines (although it may need to be made more flexible, to apply a higher coefficient to larger textures).

> What about license for texture of the Sun (editing CREDITS.md)?

SDO data are public domain, so we can license the texture as any other image we created ourselves. So GPLv2+ I suppose. But yes, will edit the CREDITS file in a bit.

> Almost 13 MB for file - could be file size reduced without losing quality?

The original PNG is 25.4 MiB. I've already tried to make as small as possible while preserving the granulation in almost the same quality.

10110111:
> However, IIRC ANGLE systems (and OpenGL ES SBCs) are limited to 2k textures

We resize the textures when they are out of limits for the current OpenGL implementation.

alex-w (approved):
The new texture is OK for me, but please add notes for new texture and remove notes for old texture in CREDITS.md file.

10110111:
So the CREDITS file is updated. Is the formulation OK?

CREDITS.md
```diff
@@ -277,7 +277,7 @@
 	4.3a Jupiter map created by James Hastings-Trew from Cassini data. "The 
 	     maps are free to download and use as source material or resource
 	     in artwork or rendering (CGI or real time)."
-	4.3b The Sun and Iapetus maps and rings of Uranus 
+	4.3b The Iapetus maps and rings of Uranus
```
alex-w: Probably typo: 'The Iapetus maps' -> 'The Iapetus map'

gzotti (approved):
OK for me, thanks!

gzotti:
Tried on Raspberries 4 and 3. All OK, but you must install qt5-image-formats-plugins (or the respective qt6 package) for webp. I did not check on other Linux systems, but I assume it's similar. This must be stated in install dependencies.

10110111:
> This must be stated in install dependencies.

Where exactly should this be stated? In the AppImage package (I think it's supposed to include all the dependencies)? Is the Windows installer affected?

alex-w:
> Where exactly should this be stated? In the AppImage package (I think it's supposed to include all the dependencies)? 

I've added `qt6-image-formats-plugins` as dependency for packages in our PPA (+ calcmysky package) + `qt5-image-formats-plugins` as dependency for Qt5-based AppImage (Qt6-based AppImage not affected).

> Is the Windows installer affected?

No

10110111:
It seems GPL is not a great choice for images, because we don't only fail to provide any script or other info on stitching the source images, but even the final output is lossily compressed.
Do you think CC-BY-SA would be more appropriate?

gzotti:
I think for images yes, CC-BY-SA or FAL.

PR #4559, opened 2025-10-02
Split translations of the Constellations section

10110111:
It appears that the format of the *Constellations* section of an SC description mandates more than just the sequence of H5 headers with descriptions. There are a few more requirements:

 * Each header must coincide with an `english` name of a constellation;
 * Each description is a separate translatable string.

This in particular is used in Stellarium Mobile where the descriptions appear in the object info panel for a constellation currently selected.

This PR makes Stellarium desktop follow the second rule (relying on the first one), with a backwards compatibility fallback to support badly formatted *Constellations* sections of the preexisting SCs. (These SCs should be fixed in the future too, and maybe the fallback could be removed then.)

New SCs should be checked for following this section format.

This PR also adds a tool that can split the existing translations, so that the `.po` files processed by it could be uploaded to Transifex to avoid losing translations.

### DIFF
--- a/src/core/StelSkyCultureMgr.cpp
+++ b/src/core/StelSkyCultureMgr.cpp
@@ -25,6 +25,7 @@
 #include "StelObjectMgr.hpp"
 #include "StelTranslator.hpp"
 #include "StelLocaleMgr.hpp"
+#include "ConstellationMgr.hpp"
 #include "StelApp.hpp"
 
 #include <md4c-html.h>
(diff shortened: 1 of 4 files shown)


alex-w:
Partially related question to this PR: is it possible to create a validator or linter to verify structure of `description.md` file?

10110111:
> is it possible to create a validator or linter to verify structure of description.md file?

Yes, I was going to do it too.

PR #4560, opened 2025-10-02
Oculars: Try a drop shadow to increase GUI panel clarity

### LINKED ISSUE #4555: Oculars plugin control panel transparency
When a HIPS source where background is not very dark is loaded it is practically impossible to see some UI elements in the Oculars control panel. More specifically some arrows and rotation text are almost invisible.

This could be solved by ability to decrease the transparency of the control panel. 

Altenatively being able to define a different color scheme for mentioned UI elements would also help.

**Additional context**
Screenshot of the problematic situation with a bright HIPS loaded:

Image

gzotti:
### Description

As noted in #4555 a noisy background like a dense HiPS makes the GUI arrows in the Oculars panel almost disappear. 
Adding a drop shadow like in the main InfoPanel helps. However, it is a bit problematic. Adding the effect to the main panel seems to make the buttons disappear. 

Also, a problem also in 25.3, the arrow keys and rotation text buttons, dark in the beginning, brighten up on mouse-over, but then don't turn dark again. They should not be so dark to begin with.

The menu is pretty old code (2011) and I may not have investigated deeply enough. There is certainly still room for improvement!

Fixes #4555 (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant
* Qt 6.8.2

## Checklist:

### DIFF
--- a/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
+++ b/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
@@ -46,9 +46,6 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
 	StelApp& stelApp = StelApp::getInstance();
 	QSettings* conf = stelApp.getSettings();
 	Q_ASSERT(conf);
-	//qDebug() <<  "Oculars: effectiveOpacity:" << effectiveOpacity() << "Setting opacity 0.75";
-	//setOpacity(0.75);
-	//qDebug() <<  "Oculars: now effectiveOpacity:" << effectiveOpacity();
 
 	setContentsMargins(0, 0, 0, 0);
 
@@ -108,7 +105,7 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
 	setLayout(mainLayout);
 
 	//Widgets with control and information fields
-	// TENTATIVE FIX:
+
 	// We can use a DropShadowEffect for enhanced visibility. Note that we cannot re-use it, so each panel needs to get its own copy.
 	// The blur radius is different to test varying radii.
 	// It would be fine to blur the background, but Qt has no way for this. Even an underlying QGraphicsWidget with QGraphicsBlurEffect does not blur the background.


gzotti:
To be still decided: Radius of drop shadow. In the main program (Infopanel) I had set 6 (if configured, SUG 6.3.3), now we can compare visual impact of a few sizes. And maybe it should be on by default. 

One thing is not solved here: At least when configuring brighter buttons (SUG 6.3.2), mouse-over will brighten up the buttons, but they will not fade back. I don't find the responsible code, though. Is CSS involved here?

gzotti:
Any opinions?

10110111:
Somehow I don't get any observable effect from this change (having set `gui/flag_info_shadow` to `true`). Could you show a screenshot before and after, that shows the effect you achieve on your system?

10110111:
Actually I do see a shadow around the "Left"/"Right" scroller buttons, but this doesn't look like a real improvement, since the GUI text still becomes unreadable when stars or labels are behind it, while the buttons don't get any more discernible—only dirtier.

gzotti:
The drop shadow improves readability when there are large bright patches behind the text. Indeed, a noisy star field has too much variance to be solved by this, just like for the main InfoPanel. I tried to apply a blur filter (with an auxiliary panel underneath), but this does not blur the background, just the panel content. Given that I had added the drop shadow for the main GUI InfoPanel, I propose adding the same to the Oculars panel for the same reason, improving readability. (Either with the same shadow size or one of those implemented here for testing.) The effect has to be activated following SUG instructions. Another (additional) step could be to make the panel less transparent, but this is not intended here.

alex-w:
Maybe just disable opacity for Oculars panel?

10110111:
I think the best solution would be to blur the scene using all the windows
and panels as a stencil. This would automatically solve allso the
readability of the windows atop the star field. The only thing not solved
by this single-pass aproach would be the readability of overlapping windows.

gzotti:
As said, this is another, more brutal step to mitigate readability at cost of more visible GUI background. Configuring panel opacity may probably even go into the Oculars plugin settings. But the same principal issue applies to all on-screen texts when they overlay small structures (also e.g. menu bar over landscape.).

gzotti:
With shadow set to 6 pixels radius, this is how the text readability improves. (Sorry, I cannot easily show the same without effect.)
grafik

It is just "do the same as in main GUI now also in Oculars." If you have a better solution, I close this.

10110111:
I might try to implement the blurring when I have some time.

gzotti:
OK

10110111:
Ah, I completely missed that this PR was supposed to solve a very different problem that what I thought! It's about the background that's close in color to the text, not about the high-frequency details interfering with the readability.

gzotti:
Well, similar colors between background pixels and foreground text is what impedes readability. Here the drop shadow helps a bit to make a tiny dark area around each character of the foreground text.  It has improved InfoPanel readability on daylight sky or against the Milky Way. When background features are of size similar to or even smaller than characters (e.g. dense star field, globular cluster, that HiPS from issue report), it still disturbs, and here large-scale blurring would be of greater benefit.

10110111:
I guess we might need a combination of these two effects. Could you give a screenshot of the real use case of the drop shadow (presumably with some particular gray HiPS?)?

gzotti:
In #4555 I give another screnshot against gray background. However, I don't know which HiPS the OP used.

gzotti:
This is from a private HiPS (yes, unfortunately still misaligned a bit, and therefore not publishable). Left: 25.3. Right (this PR): The text stands out from the background. (Also not same sky view and language, and font engine is different, therefore somewhat different sizes.) Also not adressed here: font color for the sensor rotation buttons, they really vanish in the gray, but in the new view are somewhat better.

grafik


grafik

gzotti:
Current PR only replicates on the Oculars panel what has been done for the main InfoText panel. @10110111 do you want to add blurring in this branch and in 25.4, or at any later time? In the latter case I'd like to bring this branch/PR to completion to at least have a small improvement. Blurring is a separate task. Apart from removing some comments and maybe clarifying (@alex-w ?) the question around parenting buttonBar (OcularsPanel.cpp L59), this is done.

10110111:
> do you want to add blurring in this branch and in 25.4, or at any later time?

I think I won't have time to do this for 25.4, so later.

gzotti:
Any objections then against a merge?

10110111 (approved):
Given that it's optional, LGTM.

PR #4563, opened 2025-10-03
SCM Plugin: Option to merge constellation lines into polylines

xLPMG:
- added function `mergeLinesIntoPolylines` that merges individual constellation lines into polylines
- added checkbox to export dialog to enable/disable this function

### Description
Currently, constellation lines drawn using the SCM plugin are exported to json as single line arrays, e.g.
`"lines": [[53910,67301],[67301,72607],[72607,11767]]`
As mentioned in the original PR of this plugin,  this should be merged into a polyline, e.g.
`"lines": [[53910,67301,72607,11767]]`

Possible solutions to this issue include
1. Generation of polylines whenever the user draws a continuous line
2. Merging individual lines on export

I considered solution 1 and tried prototyping, but the hurdles here are
1. Major changes to the drawing code (ScmDraw) are needed
2. A lot of extra processing is done. The eraser has to split polylines and when two existing lines are connected, they also have to be merged again(?). This would have to be done while the user is editing

As for solution 2, I created a working function `mergeLinesIntoPolylines` which merges some lines on export. The issue with such a function are that it is
1. not perfect (meaning that it does not explore all possible combinations for merges)
2. might look complicated at first, thus not easily maintainable

In my tests, it worked quite well at least. The highest priority is of course to keep the same information as the user intended. That is to not remove, add or alter any lines drawn by the user.

I would be happy to hear about any feedback or test results. Thank you!

### Type of change
- [ ] Housekeeping

### How Has This Been Tested?
While a unit test would be ideal, I do not know if this is possible for plugins. Therefore, I simply drew a bunch of different constellations, verified that they looked the same after export and compared the index.json files with and without the feature enabled. However, I ask you to please test this for yourself! My knowledge on Stellarium and sky cultures is still very limited, so the resulting sky culture json may or may not comply with your guidelines. I hope you can give me some feedback on this.

**Test Configuration**:
* Operating system: MacOS 26.0.1
* Graphics Card: Apple M3 Pro

## Checklist:

### DIFF
--- a/plugins/SkyCultureMaker/src/ScmConstellation.cpp
+++ b/plugins/SkyCultureMaker/src/ScmConstellation.cpp
@@ -299,68 +299,70 @@ void scm::ScmConstellation::mergeLinesIntoPolylines(QJsonArray &lines) const
 	}
 
 	// Step 1: merge line ends with other line starts
-	for (int i = 0; i < lines.size(); ++i)
+	for (int growableLineIdx = 0; growableLineIdx < lines.size(); ++growableLineIdx)
 	{
-		QJsonArray currentLine = lines.at(i).toArray();
-		QJsonValue currentEnd  = currentLine.last();
+		QJsonArray growableLine = lines.at(growableLineIdx).toArray();
 
-		// Look for a line that starts where the current line ends
-		for (int j = i + 1; j < lines.size(); ++j)
+		// Look for a line that starts where the growableLine ends
+		for (int attachableLineIdx = growableLineIdx + 1; attachableLineIdx < lines.size(); ++attachableLineIdx)
 		{
-			QJsonArray nextLine  = lines.at(j).toArray();
-			QJsonValue nextStart = nextLine.first();
+			QJsonArray attachableLine = lines.at(attachableLineIdx).toArray();
 
-			// Merge nextLine into currentLine
-			if (currentEnd == nextStart)
+			// Merge attachableLine into growableLine
+			if (growableLine.last() == attachableLine.first())
 			{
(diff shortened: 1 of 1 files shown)


10110111:
From my testing it seems to work indeed, but I'd like to have better index names than `i`, `j`, e.g. `growableLineIdx`, `attachableLineIdx` or similar. Similarly, "current" and "next" would be better called functionally, like "growable" and "attachable" or something.

~~The `k` could be eliminated by using a range-based `for` loop, like~~ Sorry, I missed that `k` starts with 1.

> `lines.replace(i, currentLine);`

Wouldn't `lines[i] = currentLine;` be more readable? Or am I missing something?

xLPMG:
> From my testing it seems to work indeed, but I'd like to have better index names than `i`, `j`, e.g. `growableLineIdx`, `attachableLineIdx` or similar. Similarly, "current" and "next" would be better called functionally, like "growable" and "attachable" or something.
> 
> ~The `k` could be eliminated by using a range-based `for` loop, like~ Sorry, I missed that `k` starts with 1.
> 
> > `lines.replace(i, currentLine);`
> 
> Wouldn't `lines[i] = currentLine;` be more readable? Or am I missing something?

I simply went with the wording you suggested, as I couldn't really find a better alternative. I also replaced the k-loops with range based loops and simply removed the first or last element. In my opinion, that makes the code much more readable (with slightly less performance, but I doubt that makes a big difference here)

10110111 (approved):

10110111:
This PR LGTM, but please note #4593 that I discovered during testing of this PR.

PR #4594, opened 2025-10-19
SCM: Move constellation descriptions to the Constellations dialog

10110111:
The _Constellations_ section has a fixed structure, which isn't currently enforced. And this structure makes it more sensible to place descriptions for each constellation in its editor's UI instead of having a single editable section prone to mismatches between constellation names and subsection names, misformatting etc.

This PR adds a text edit field for a constellation in the Constellation Editor dialog and removes the Constellations section in the SC description editor.

### DIFF
--- a/guide/ch_skycultures.tex
+++ b/guide/ch_skycultures.tex
@@ -87,16 +87,18 @@ some cultural/ethnographical background of the users of this sky culture,
 history of sky culture research that provided this work, tables of names/translations,
 links to external resources, whatever seems suitable.  The length of the description texts is not limited.
 
-The skyculture description should include the following Level 2 sections. Any other section must be a subsection (i.e. Level 3 or deeper) of one of these sections. Each of these sections will have its own entry in the translations files.
+The skyculture description should include the following Level 2 sections. Any other section must be a subsection (i.e. Level 3 or deeper) of one of these sections. Each of these sections except \texttt{Constellations} will have its own entry in the translations files. The \texttt{Constellations} section will be split, so that each constellation's description will be translated separately.
 
 \begin{description}
 \item \texttt{Introduction} -- a summary that could ideally be shown on a small screen entirely.
 \item \texttt{Description} -- the actual text of the description.
-\item \texttt{Constellations} -- a sequence of Level 5 sections, where the header names a constellation, and the contents gives some information on it.
+\item \texttt{Constellations} -- a sequence of Level 5 sections, where the header is the English name of a constellation (as given in the \file{index.json} file), and the body gives some information on this constellation.
 \item \texttt{References} -- a list of references, see \ref{SC:references} for details.
 \item \texttt{Authors} -- all authors and acknowledgments go here.
 \item \texttt{License} -- either a license id (see \ref{SC:license}), or a sequence of lines separated by empty lines (i.e. Markdown paragraphs) describing which part of the culture has which license, in the format "Sky culture item: License-ID".
 \end{description}
+
+Note that the \texttt{Constellations} section has the strict format described above. In particular, there's no place for an introduction for all the constellations or a similar deviation from the format of this section. If this kind of text is desired, it can be placed in a subsection of the \texttt{Description} section.
 
 To allow the correct display of special characters like ÄÖÜßáé you must provide the file in UTF-8 encoding. 
 If you write only English/ASCII, this may not be relevant.


xLPMG:
Is the constellation description optional or required? Before, it was required and this PR makes it partially optional.

1. Create Constellation without filling description, because it says "optional" -> OK
2. Try to export SC -> "Error: The sky culture description is not complete. Please fill in all required fields correctly."

I think we should either make it required for each constellation or make it fully optional. Fully optional would mean to remove `!constellations.trimmed().isEmpty()` and `constellations.contains("##### ")` from `Description.hpp: isComplete`

10110111:
Ah, here is the check... Yeah, I've missed it. Will remove it completely. Constellation descriptions aren't really required, although they are very welcome.

gzotti:
Did not look into the code for now: Can there be a section introduction before the list of constellations? Or is the Constellations section really just a list of h5 sections for technical reasons?

10110111:
> Can there be a section introduction before the list of constellations?

I don't think it's expected in Stellarium Mobile or in the `update_po.py` script of the `stellarium-skycultures` repo. So for now we should consider it strictly a list of descriptions.

10110111:
> Can there be a section introduction before the list of constellations?

I think the authors can add a level-3 Constellations subsection in the Description section if needed.

gzotti:
OK. This would be to explain some general things. The strictness of the (h2 Constellations)-->{h5 Const , ...) breakdown must be clarified in SUG 25.4. Also how to structure any extended naming entries in the descriptive text.

10110111:
> Also how to structure any extended naming entries in the descriptive text.

What do you mean? Extended naming entries are given in `index.json`. What the descriptive text could do about them is explain why/how/when these names are as they are. Not repeat the names themselves.

gzotti:
Exactly such details. Authors most likely expect MD files to be immutable, while you earlier suggested you could inject the names from index.json into the h5 constellation descriptions and so SC authors should not bother repeating those names. (Which of the cultural names should then be used in the h5 title?)

Which SC in master currently fulfills your formatting requirements?

10110111:
> Which SC in master currently fulfills your formatting requirements?

Arabic (Arabian Peninsula) appears to be the closest (although it still has some flaws). The version in stellarium-skycultures repo is the best (because it's the version that I ported to Stellarium Mobile, so it had to be fixed). I hope it will soon be updated in desktop Stellarium.

> Which of the cultural names should then be used in the h5 title?

English, of course, as the base that should exist everywhere. The `.pot` generating script will warn you if it finds a name that isn't present in any `english` tag of the constellations in `index.json` (or if the Constellations section is misformatted).

> Authors most likely expect MD files to be immutable

Well the files *are* immutable. But the contents are used in different ways: e.g. the *References* section is modified to be pointable to by the numbered references in the text body, the *License* section is expanded with an explanation of what a license means, etc. Similarly the Constellations section is generally not even supposed to be shown as a section: its main use (in Stellarium Mobile) is to have the descriptions appear in the object info windows, like this:

Screenshot_20251019-195158

Similarly we can do whatever we consider best with these descriptions. Even move them to object info (which is currently rather empty for constellations).

guide/ch_skycultures.tex
```diff
@@ -87,17 +87,19 @@ \section{Text description}
...
 
-The skyculture description should include the following Level 2 sections. Any other section must be a subsection (i.e. Level 3 or deeper) of one of these sections. Each of these sections will have its own entry in the translations files.
+The skyculture description should include the following Level 2 sections. Any other section must be a subsection (i.e. Level 3 or deeper) of one of these sections. Each of these sections except \texttt{Constellations} will have its own entry in the translations files. The \texttt{Constellations} section will be split, so that each constellation's description will be translated separately.
 
 \begin{description}
 \item \texttt{Introduction} -- a summary that could ideally be shown on a small screen entirely.
 \item \texttt{Description} -- the actual text of the description.
-\item \texttt{Constellations} -- a sequence of Level 5 sections, where the header names a constellation, and the contents gives some information on it.
+\item \texttt{Constellations} -- a sequence of Level 5 sections, where the header is the English name a constellation (as given in the \file{index.json} file), and the body gives some information on this constellation.
```
xLPMG:
```suggestion
\item \texttt{Constellations} -- a sequence of Level 5 sections, where the header is the English name of a constellation (as given in the \file{index.json} file), and the body gives some information on this constellation.
```

alex-w (approved):

PR #4595, opened 2025-10-19
SCM: Fix: Deleted lines during creation of a constellation end up in the output (#4593) 

### LINKED ISSUE #4593: SCM: Deleted lines during creation of a constellation end up in the output
To reproduce:

1. _Create_ a new SC via the SCM plugin
2. _Add Constellation_
3. Click _Pen_, create a line between two stars
4. Click _Eraser_, erase the line
5. Click _Pen_, create another sequence of lines
6. Name the constellation, _Save_
7. Fill in the required fields in the description
8. Export the SC
9. Check the resulting `index.json`, notice that the line that was deleted in the SCM UI is still there.

xLPMG:
This PR is a rework of ScmDraw that I wanted to do for a while. It merges CoordinateLines (a constellation line with coordinate data) and StarLines (a constellation line with optional star ids/names) into a ConstellationLine object with the same properties. In my opinion there is no reason to keep two vectors of the same lines. It only introduces unnecessary complexity. This PR simplifies the code and fixes #4593.

Fixes #4593

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Manual test.

## Checklist:


xLPMG:
You may already review this, but I'd like to wait until the two other SCM PRs have been merged, to check for any merge conflicts and if more needs to be adjusted.

plugins/SkyCultureMaker/src/types/SkyPoint.hpp
```diff
@@ -34,10 +34,17 @@ namespace scm
 struct SkyPoint
 {
 	//! The coordinate of the point.
-	Vec3d coordinate;
+	Vec3d coordinate = Vec3d(0, 0, 0);
 
 	//! The optional star at that coordinate.
-	std::optional<QString> star;
+	std::optional<QString> star = std::nullopt;
```
10110111: Why does it have to be an optional? Isn't `QString::isEmpty()` sufficient for checking that a string is present?

xLPMG: Hm I guess I wasn't sure how exactly null safety works for QStrings. Apparently a QString can be null and in that case .isEmpty will still work and return true.

Anyway, I changed it and removed the optionals

gzotti: See https://doc.qt.io/qt-6/qstring.html#QString:

> QString::QString()
> 
> Constructs a null string. Null strings are also considered empty.
> 
> See also [isEmpty](https://doc.qt.io/qt-6/qstring.html#isEmpty)(), [isNull](https://doc.qt.io/qt-6/qstring.html#isNull)(), and [Distinction Between Null and Empty Strings](https://doc.qt.io/qt-6/qstring.html#distinction-between-null-and-empty-strings).
> 

QString() is a null string. QString s("") is not null but empty. No need for the std:: parts.

plugins/SkyCultureMaker/src/types/SkyPoint.hpp
```diff
@@ -37,13 +37,13 @@ struct SkyPoint
 	Vec3d coordinate = Vec3d(0, 0, 0);
 
 	//! The optional star at that coordinate.
-	std::optional<QString> star = std::nullopt;
+	QString star = QString("");
```
10110111: This can just be left default-initialized as `QString star;`, no need to pass an empty string to it.

plugins/SkyCultureMaker/src/types/SkyPoint.hpp
```diff
@@ -37,13 +37,13 @@ struct SkyPoint
...
 
 	//! Resets the SkyPoint to default values.
 	void reset()
 	{
 		coordinate.set(0, 0, 0);
-		star = std::nullopt;
+		star = QString("");
```
10110111: `star = {};` will suffice (or `star = "";`, which will make it non-null but is also not as verbose as your variant).

plugins/SkyCultureMaker/src/types/SkyPoint.hpp
xLPMG: Alright, I changed it to null strings

plugins/SkyCultureMaker/src/types/SkyPoint.hpp
xLPMG: Hm saw your comment too late. What about .clear()? It sets the QString to null

10110111:
BTW, why do you have to maintain a change log file? No plugin has a `CHANGELOG.md`, and those who do have a change log, use the GNU format for it and name it `ChangeLog`.

I don't think a change log is any useful though, since all the changes to the plugin can be found by
```
git log --oneline -- plugins/SkyCultureMaker
```
while having a separate change log file can result in collisions between pull requests.

xLPMG:
> requests.

I added a changelog because in the future there might be a new student group of 3-5 people, who will work on this plugin. I thought it would then be nice to have a log of bugs that have been fixed or what has been added. Especially for presentations, one could say "this group did x and we did y".

alex-w:
@10110111 is it OK now?

10110111:
> is it OK now?

Yes, my approval is still valid.

Issue #4554, opened 2025-10-01
How do you build the User Guide?


10110111:
I'm trying to build the Guide on Ubuntu 25.04, so for this I installed all the packages listed in the README. But when running `make pdf` (after having CMake generated `version.tex`) I get the following output:
```
(/usr/share/texlive/texmf-dist/tex/latex/tools/longtable.sty
(/usr/share/texlive/texmf-dist/tex/latex/tools/longtable-2020-01-07.sty))
(/usr/share/texlive/texmf-dist/tex/latex/tools/tabularx.sty

LaTeX Warning: You have requested release `2024/06/01' of LaTeX,
               but only release `2020-02-02' is available.

(/usr/share/texlive/texmf-dist/tex/latex/tools/array.sty
(/usr/share/texlive/texmf-dist/tex/latex/tools/array-2016-10-06.sty)))
(/usr/share/texlive/texmf-dist/tex/latex/gensymb/gensymb.sty)
(/usr/share/texlive/texmf-dist/tex/generic/babel/babel.sty
(/usr/share/texlive/texmf-dist/tex/generic/babel/txtbabel.def)
! Missing \endcsname inserted.
<to be read again> 
                   \@raw@classoptionslist 
l.4132   \fi}
             
? 
```
I found that I can work past this error by commenting out `\RequirePackage[2020-02-02]{latexrelease}`, but then I get the other error that this command was supposed to work around:
```
...
(/usr/share/texlive/texmf-dist/tex/latex/hyperref/puenc.def)
(/usr/share/texlive/texmf-dist/tex/generic/bitset/bitset.sty
(/usr/share/texlive/texmf-dist/tex/generic/bigintcalc/bigintcalc.sty))
(/usr/share/texlive/texmf-dist/tex/latex/base/atbegshi-ltx.sty

LaTeX Warning: You have requested package `atbegshi',
               but the package provides `atbegshi-ltx'.

)) (/usr/share/texlive/texmf-dist/tex/latex/hyperref/hpdftex.def
(/usr/share/texlive/texmf-dist/tex/latex/base/atveryend-ltx.sty

LaTeX Warning: You have requested package `atveryend',
               but the package provides `atveryend-ltx'.

) (/usr/share/texlive/texmf-dist/tex/latex/rerunfilecheck/rerunfilecheck.sty
(/usr/share/texlive/texmf-dist/tex/generic/uniquecounter/uniquecounter.sty)))
[...]

PR #4629, opened 2025-11-01
Object Narration

gzotti:
### Description
This branch introduces object narration (text to speech output).

REQUIRES Qt6.4 OR LATER. Should emit debug output for lower versions.

StelObject received a "getNarration()" method, derived classes should tell users something of interest.
IMO reading the coordinate numbers is not interesting. But think of what your local observatory guide would tell you.

Some coordinates are often useful on screen but still not if read aloud. Another line of info flags is added to compose the narration string where users can select what to tell. Also, the narration may form sentences which combine the data bits in another sequence from the InfoString tables, it's not just a sequential read. 

The action is managed by StelObjectMgr. After selecting, press Shift-R and listen.

The default implementation just names the object.
These object types have working narration:
- [x] Nebula
- [x] Planet
- [ ] MinorPlanet
- [ ] Comet
- [x] Nomenclature (planet features)
- [x] Star
- [x] Constellation
- [ ] Asterism (26.1?)
- [ ] Quasar (26.1?)
- [ ] Pulsar (26.1?)
- [ ] Exoplanet (26.1?)
- [ ] Satellite (26.1?)
- [ ] ... (please add) (26.1?)

In addition, landscape and skyculture descriptions can be read out.

This should always select available voices for the current program language. 
- [ ] If no language found, there is no narration, but no crash either.

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Irrelevant

## Checklist:


src/core/StelApp.cpp
```diff
@@ -327,22 +329,23 @@ StelApp::~StelApp()
...
-	delete videoMgr; videoMgr=Q_NULLPTR;
-	delete stelObjectMgr; stelObjectMgr=Q_NULLPTR; // Delete the module by hand afterward
-	delete textureMgr; textureMgr=Q_NULLPTR;
-	delete planetLocationMgr; planetLocationMgr=Q_NULLPTR;
-	delete moduleMgr; moduleMgr=Q_NULLPTR; // Delete the secondary instance
-	delete actionMgr; actionMgr = Q_NULLPTR;
-	delete propMgr; propMgr = Q_NULLPTR;
-	delete renderBuffer; renderBuffer = Q_NULLPTR;
-	delete randomGenerator; randomGenerator=Q_NULLPTR;
+	delete tmp; tmp=nullptr;
```
10110111, line 332: Can these maintenance things happen separately from feature PRs?

src/core/StelSpeechMgr.cpp
```diff
@@ -0,0 +1,183 @@
...
+#include <QTextToSpeech>
+#include <QVoice>
+#include <QMediaPlayer>
+
+StelSpeechMgr::StelSpeechMgr()
+{
+#if (QT_VERSION>=QT_VERSION_CHECK(6,6,0))
+	m_speech=new QTextToSpeech(static_cast<QObject*>(this));
+#else
+	qWarning() << "Text to Speech requires Qt6.6 or higher";
```
10110111: I think it would make sense to move this warning to the CMake stage.

gzotti: There should be some technical-explanatory output IMO. When conditions are not met, some dummy behaviour is necessary. But I have not tested with incapable configs so far.

src/core/StelApp.cpp
gzotti: I tend to just fix these as I see them in a WIP file.

src/core/StelSpeechMgr.cpp
10110111: Well, maybe it's OK to leave this here, but still a `message(WARNING "blabla")` would be good in the CMakeLists, so that packagers know what they are missing.

src/core/StelApp.cpp
```diff
@@ -601,6 +604,14 @@ void StelApp::init(QSettings* conf)
 	videoMgr->init();
 	getModuleMgr().registerModule(videoMgr);
 
+	// Init speech manager
+#ifdef ENABLE_MEDIA
+	SplashScreen::showMessage(q_("Initializing speech output..."));
+#endif
```
alex-w, line 609: Why only notification inside ifdefs?

gzotti: The feature only becomes active when building with media support. The initialisation else is trivial and message can well be skipped. I can still add the Qt6.6 stuff here.

alex-w: I fear it should be under ENABLE_SPEECH flag and code under this flag shouldn't be enabled for Qt prior version 6.6

gzotti: Do you really need a separate flag, and not run under MEDIA?

alex-w: Qt5 and Qt6 prior 6.6 has media, but not speech ;)

gzotti: You can decide to build with MEDIA support and have audio output (including speech), or not. If there are technical obstacles (Qt<6.6), your MEDIA installation lacks speech. But I don't see a reason to have speech and no Media support.

gzotti:
The first object narrations work, at least in principle, and does indeed provide not too bad read-out. A lot can obviously be derived by following the `getInfoString()` methods without table layout, but reformulating the terse labels into sentences. But when I see how long and complex the getInfoString() for some object classes are, we should discuss at this point which parts of the object data should be narrated. Maybe some parts can also be split into standardized sections reused by StelObject subclasses. 

I think, generally no coordinates readout should be given, and also not more than the first few names and designations. For cultural names, it would be great to read the original wording, but this has to be detected. Educated humans can easily switch to "english transcription mode" when seeing the pronounce tag, but the computer voice must be told. And I see no language-agnostic voice setting for "read the IPA strings".  Many other numbers need to be re-worked to hear "thousand" or "point" (translated to "comma" in German) instead of a silent dot, "arcseconds" instead of `"`, or to read "H hours, M minutes and S seconds".  And I'd appreciate hints on how to trigger a short pause. 

The InfoString is influenced by the InfoFlags and by useDecimalDegrees. Should the narration follow the same user selection (minus coordinates), or should we provide another set of narration info selections? (I will add a new dialog to be called by a button in Config/Language where users can configure speech engine, speaker voice, volume etc. given what's available on their systems, pretty much following the Qt example. But on this extra dialog, another box of Info selections would be possible.)

The `Constellation::getNarration()`  should obviously include what is written in the `description.md/##Constellations` level5 section (respectively its translation). For this to work, these sections must exist in the format intended by @10110111, and there must be a method to extract this. @10110111 any advise here?  `Asterism::getNarration()` should work in a similar way, but for this the description spec has to be extended. 

Should we even think of a multi-voice (male/female) readout to break the monotony? Like, have two tick boxes in the info selection who reads what? (May not be possible if there are not enough voices.) This would require `getNarration()` to return a different data structure.

The voice language must follow the GUI language/locale of course. It is pretty hilarious to run the program (built without NLS) in English, but use the voice for the default German locale. The result sounds like an 80-years old German with all the stereotype "German accent" patterns. Now, I just hope that many, also smaller, languages have good voice support.

10110111:
> For this to work, these sections must exist in the format intended by @10110111, and there must be a method to extract this.

You can put some code that would use the output of `StelSkyCultureMgr::getConstellationsDescriptions` as called from `StelSkyCultureMgr::convertMarkdownLevel2Section`, saving it to either a local map field, or to each constellation (or both, filling the constellations from this map when they are actually read). Just make sure that `StelSkyCultureMgr::getCurrentSkyCultureHtmlDescription` has been called when you need to fill in the constellations' descriptions.

src/core/StelApp.cpp
alex-w: Ok

10110111:
> Should we even think of a multi-voice (male/female) readout to break the monotony?

This idea sounds weird to me. Does this ever happen during a real lecture?

> It is pretty hilarious to run the program (built without NLS) in English, but use the voice for the default German locale. The result sounds like an 80-years old German with all the stereotype "German accent" patterns.

I vaguely recall a similar experience in about 2020 when my teammates were trying to use a Qt5-based text-to-speech engine to read Russian text with an English voice :) Later they did find a Russian voice, so the feature was implemented successfully.

This was the time of Qt5, and given this, I wonder why we have to require Qt 6.6+ for this.

gzotti:
> This was the time of Qt5, and given this, I wonder why we have to require Qt 6.6+ for this.

Qt5 is about as dead as Windows 10 (indeed, meaning not really dead, for still some years to come...), so I decided when using a new feature I skip Qt5.  selectEngine() was introduced in 6.4, and other changes that may be relevant were in 6.6. Current LTS is 6.8. I just don't want to dig into outdated stuff at this moment. If the rest works nicely, we can still try to backport, but I don't think it's necessary.  Maybe 6.4 is enough, so that systems on 6.5LTS will be satisfied. Will see that later.

gzotti:
> > Should we even think of a multi-voice (male/female) readout to break the monotony?
> 
> This idea sounds weird to me. Does this ever happen during a real lecture?

Think of a radio talkshow conversation between moderator and guest vs. frontal single-narrator presentation. A clever mix of voices may sound better. But TBH I'd have to test more of the voices first to get a feeling whether it's worth it.

gzotti:
@alex-w @xalioth Why is the default StelTranslator::globalTranslator initialized to system locale and remains so when NLS is disabled? (It is so since 2006...) Wouldn't it be better "en",  given our default program language?

src/core/StelApp.cpp
gzotti: Now I remember why I saw MEDIA as necessity - most Speech output plugins needs QAudioDevice from the Multimedia package.

alex-w:
> Why is the default StelTranslator::globalTranslator initialized to system locale and remains so when NLS is disabled?

By historical reasons and default behavior of operating systems from these times.

> Wouldn't it be better "en", given our default program language?

If locale is unknown - yes, of course, but what if locale is known?

gzotti:
When you build without NLS, the program shows all-english behaviour (untranslated english strings, GUI etc.). When building without NLS, I don't expect a German voice reading english text (and it sounds stupid). The speech engine is launched with system locale and has to be set to program locale during initialisation. If locale without building with NLS is inherited from the system, what is its purpose?

alex-w:
> When you build without NLS, the program shows all-english behaviour (untranslated english strings, GUI etc.). When building without NLS, I don't expect a German voice reading english text (and it sounds stupid). The speech engine is launched with system locale and has to be set to program locale during initialisation. If locale without building with NLS is inherited from the system, what is its purpose?

Oops… my answer was for situation when NLS is enabled. Of course when NLS disabled we should use English locale only.

gzotti:
> Just make sure that `StelSkyCultureMgr::getCurrentSkyCultureHtmlDescription` has been called when you need to fill in the constellations' descriptions.

It currently is called only when the dialog is open. But what's worse is getConstellationDescriptions seems to suffer from the WIndows CRLF convention. The captured sectionname ends with \r and never matches. Can the l5SectionNamePat be formulated safer?

10110111:
> seems to suffer from the WIndows CRLF convention

Hmm, I never expected that the description would have CRLF line endings. I guess we should strip CR characters (or maybe replace CRLF→LF) on loading the file and never bother in the further processing.

gzotti:
OK, I added `markdown.replace("\r", "");` to getCurrentSkyCultureHtmlDescription(). Now a few bits here finally start to work

gzotti:
For the GUI to configure engine and voices, should I create a new dialog or just a new "Speech" tab in the Config dialog?

gzotti:
@alex-w , @10110111 any opinion on tab in config dialog vs. extra dialog called from button next to language setting, and on using same vs. own info bits selection vs. using our own idea of what's interesting?

alex-w:
> For the GUI to configure engine and voices, should I create a new dialog or just a new "Speech" tab in the Config dialog?

For configuration the engine and voice adding block “Speech” near block “Language” should be enough IMHO. For extra options adding special dialog via button (like in other places).

10110111:
I think speech settings should be in a separate dialog, because it seems to be quite a rarely-usable feature whose settings shouldn't clutter the main dialogs.

10110111:
Actually no, since the narration is activatable by a simple hotkey (I initially thought it would be a pure scripted thing), then Speech should indeed be a tab in the Config dialog, next to Tools.

alex-w:
> Actually no, since the narration is activatable by a simple hotkey (I initially thought it would be a pure scripted thing), then Speech should indeed be a tab in the Config dialog, next to Tools.

Separate tab for 2-3 options is a strange strategy. Maybe splits tools tab onto 2 tabs (the speech options can be not exists - Qt5-based package of Qt6 prion version 6.6) and one of them should contain optional speech options?

10110111:
> Separate tab for 2-3 options is a strange strategy

Honestly, it seems there will be more. Not just a choice of a voice, speaking speed etc., but maybe also some details on how different objects get pronounced. But without knowing what is supposed to be there, I can't make a real decision.

> Maybe splits tools tab onto 2 tabs (the speech options can be not exists - Qt5-based package of Qt6 prion version 6.6) and one of them should contain optional speech options?

Then I suppose we can just put a button to a Speech config dialog in the Tools tab, grayed out when it's not available.

gzotti:
QtCreator comes with a demo app for QTextToSpeech. I think of a similar layout with a simple text entry for testing pitch and volume. This may now look too empty when compared to other tabs, but we can replicate the Info option selections for speech. Or those info options for speech go onto the other Info selection tab,  so that we'd have two tick boxes per info option. 

Looking at the dialog again, if I rearrange the Ephemeris settings a bit, I could maybe add the Voice options to the Language area directly.

Constructing a natural-sounding reading/narration of the object data will be some work, and may take to 26.1 for all object types to not sound terrible. I start with getInfoString, but remove all flag queries and compose more narration-like sentences. But before butchering away all flag-dependent stuff, we should decide whether we want flag selections for configuring the speech. I would definitely deselect coordinate output, although I like to see it on the screen. OTOH on a small screen I might switch off physical details for space reasons but want the voice to tell me about the physical properties of stars or nebulae, or how long the comet takes around the sun. (Also reading spectral classes may have to be converted to better sounding pronunciation...) 

As I see it currently, the QtTextToSpeech just reads ordinary text with a given voice.  Voice engine is part of the OS, or extra voice engines can be installed, but I am not familiar with flite or speechd on Linux.  Available voices depend on locale/language. IIRC you can install extra voices in the OS config settings.  So, some UI combos are necessary. 

Regarding demonstrating the right pronunciation, this would be interesting, but we could target this only with an "authoritative IPA voice" which I don't see available. I don't intend to go beyond what the OS speech engines provide. A short search indicates it is difficult, Microsoft has its own service on azure... I leave it to linguist experts to add this.

Another place to read longer text would be the info texts in Landscapes (when it says more than "my backyard observatory"), Skycultures and Scenery3D. This might just need setting a button next to the text field. 

Another task, more geared towards visually impaired users, is holding the mouse over a GUI element and reading its label. Not sure what it takes to allow this by default. (And of course, allow disabling this!)

All GUI elements, also tabs, can be hidden when Stellarium is built without speech support. 

And sure, this must also become a scripting command. core.say(...).

gzotti:
It seems I will need a CMake config flag after all. QtTextToSpeech may not be available on many Linux systems. 

Ubuntu seems to provide installable packages here: https://launchpad.net/ubuntu/+source/qt6-speech

But they give no instructions for add-apt-repository.  And how would I configure the CI?

alex-w:
`qt6-speech-dev` + `libqt6texttospeech6`?

10110111:
> But they give no instructions for add-apt-repository.

Of course, it's available in `universe` since Noble: https://packages.ubuntu.com/noble/qt6-speech-dev.

10110111:
> `qt6-speech-dev` + `libqt6texttospeech6`?

No need to explicitly list the second, it's already a dependency of the first.

10110111:
> And how would I configure the CI?

Similarly to `ENABLE_MEDIA` I suppose.

gzotti:
> qt6-speech-dev + libqt6texttospeech6?

Of course I just tried this. (WSL, U22.04.05 LTS)

> sudo apt install libqt6texttospeech6
> Reading package lists... Done
> Building dependency tree... Done
> Reading state information... Done
> E: Unable to locate package libqt6texttospeech6
 

Ah, 24.04. Time to upgrade then... And fix the TeX problem.  #4554 :-/ 
I meant, the CI installations will have to include these libs in the yaml files (if I understand correctly; how to declare support optional? If apt doesn't find a package, just continue and deactivate speech in CMake later.), and if they are not found, I need to deactivate this in the CMakeFile, so yes, another CMake variable (which I wanted to avoid, but OK, so be it).

gzotti:
Config GUI is hopefully done. Landscape and Skyculture textbrowsers can play narrations. However, for the text contents, not everything should likely be read out aloud: The narration should likely stop before References or License information. For the Skyculture panel the new structure of description.md may help, even with a few checkboxes to be added to select sections to be read. For the Landscapes, the format is still general HTML without prescribed structure. Probably the HTML tags can be annotated similar to "notr", with "nospeech", and a preprocessor step could remove these elements? (Same goes to the Scenery3D plugin.) Now, is the HTML ever preprocessed into a format where "nospeech" elements could be easily stripped?

gzotti:
Trying to get this to run on Ubuntu 24.4 now. This is a 2015 NUC on which I never had necessity for sound. Finally after a BIOS update I can hear system test sounds. However, it seems neither installing flite nor speech-dispatcher helps in Qt::TextToSpeech ever detecting a speech engine. Any idea how to configure this?

alex-w:
macOS 26.1 + Qt 6.9.3 = everything OK

gzotti:
Very good. Meanwhile I found what to do on Ubuntu 24.04 (packages were still missing), described in BUILDING.md. I hope I can continue to a merge-ready point by Monday, so that a first version can still go into 25.4. I think it is not required that all objects tell us everything in polished language at this stage.

10110111:
For some reason the _Voice_ combobox opens a very long list that I can't even fit in my screen, and there is no scrollbar in this list.

Also, somehow there are only variants of Russian in this list, and even switching program language to English doesn't change it. The result of pressing _Say_ on the default test text results in a gurgling voice reminding me of Windows XP's Microsoft Sam. Besides, it doesn't even pronounce Russian text that I type in, instead it just names the letters one after another. E.g. "проверка голоса" sounds like "pe er o ve e er ka ge o el o es a".

(I'm trying it on Ubuntu 24.04, with the speech engine available being speechd.)

Here's the log from this session: [log.txt](https://github.com/user-attachments/files/24013551/log.txt).

10110111:
OK, installing `flite` as per the instruction does result in a much better voice in English, but then there are no Russian voices available. And an attempt to voice a Russian text with the default `slt` voice of the `flite` engine results in a segfault originating in `libqtexttospeech_flite.so`.

10110111:
Here's the log file from the flite session: [log.txt](https://github.com/user-attachments/files/24014379/log.txt). And the stack trace:
```cpp
#0  QAudioSink::processedUSecs (this=0x0) at /usr/src/qt6-multimedia-6.8.3-0ubuntu1/src/multimedia/audio/qaudiosink.cpp:282
#1  0x00007fff4ed500e2 in QTextToSpeechProcessorFlite::startTokenTimer (this=0x55555b4d05f0)
    at /usr/src/qt6-speech-6.8.3-0ubuntu1/src/plugins/tts/flite/qtexttospeech_flite_processor.cpp:40
#2  0x00007fff4ed52008 in QTextToSpeechProcessorFlite::audioOutputCb (w=0x7ffe9c019590, start=<optimised out>, size=320, last=0, asi=0x7ffe9c0017d0)
    at /usr/src/qt6-speech-6.8.3-0ubuntu1/src/plugins/tts/flite/qtexttospeech_flite_processor.cpp:71
#3  0x00007fff4ca9163a in synthesis_body
    (params=<optimised out>, str=0x7ffe9c00bd70, fs=<optimised out>, framem=<optimised out>, cg_db=0x7ffdfd9f5c60 <cmu_us_slt_cg_db>, asi=0x7ffe9c0017d0, mlsa_speed_param=<optimised out>) at ../../src/cg/cst_mlsa.c:163
#4  0x00007fff4ca864e6 in cg_resynth (utt=0x7ffe9c001900) at ../../src/cg/cst_cg.c:715
#5  cg_synth (utt=0x7ffe9c001900) at ../../src/cg/cst_cg.c:174
#6  0x00007fff4ca91ac2 in apply_synth_method (u=<optimised out>, meth=0x7fff4caae8a0 <synth_method_text+160>) at ../../src/synth/cst_synth.c:162
#7  0x00007fff4caa0e5c in flite_do_synth (u=0x7ffe9c001900, voice=<optimised out>, synth=0x7fff4ca91af0 <utt_synth>) at ../../src/synth/flite.c:209
#8  0x00007fff4caa0ec8 in flite_synth_text (text=<optimised out>, voice=<optimised out>) at ../../src/synth/flite.c:224
#9  0x00007fff4caa14b7 in flite_text_to_speech (text=<optimised out>, voice=<optimised out>, outtype=0x7fff4ed5b1de "none") at ../../src/synth/flite.c:369
#10 0x00007fff4ed53932 in QTextToSpeechProcessorFlite::processText
    (this=0x55555b4d05f0, text=..., voiceId=<optimised out>, pitch=0, rate=0, outputHandler=0x7fff4ed517a0 <QTextToSpeechProcessorFlite::audioOutputCb(cst_wave_struct const*, int, int, int, cst_audio_streaming_info_struct*)>) at /usr/src/qt6-speech-6.8.3-0ubuntu1/src/plugins/tts/flite/qtexttospeech_flite_processor.cpp:183
#11 QTextToSpeechProcessorFlite::processText
    (this=0x55555b4d05f0, text=..., voiceId=<optimised out>, pitch=0, rate=0, outputHandler=0x7fff4ed517a0 <QTextToSpeechProcessorFlite::audioOutputCb(cst_wave_struct const*, int, int, int, cst_audio_streaming_info_struct*)>) at /usr/src/qt6-speech-6.8.3-0ubuntu1/src/plugins/tts/flite/qtexttospeech_flite_processor.cpp:164
#12 0x00007ffff4fca39a in QObject::event (this=0x55555b4d05f0, e=0x7fffd4001a90) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qobject.cpp:1418
#13 0x00007ffff6d980c8 in QApplicationPrivate::notify_helper (this=<optimised out>, receiver=0x55555b4d05f0, e=0x7fffd4001a90)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/widgets/kernel/qapplication.cpp:3296
#14 0x00007ffff4f813f0 in QCoreApplication::notifyInternal2 (receiver=0x55555b4d05f0, event=0x7fffd4001a90)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qcoreapplication.cpp:1177
#15 0x00007ffff4f8143d in QCoreApplication::sendEvent (receiver=<optimised out>, event=<optimised out>)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qcoreapplication.cpp:1617
#16 0x00007ffff4f817df in QCoreApplicationPrivate::sendPostedEvents (receiver=0x0, event_type=0, data=0x55556d18de90)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qcoreapplication.cpp:1951
#17 0x00007ffff51e92cf in postEventSourceDispatch (s=0x7ffe9c0012a0) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qeventdispatcher_glib.cpp:246
#18 0x00007ffff48f7de2 in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#19 0x00007ffff49691f8 in ??? () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#20 0x00007ffff48f8223 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#21 0x00007ffff51e6c63 in QEventDispatcherGlib::processEvents (this=0x7ffe9c000ef0, flags=...)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/kernel/qeventdispatcher_glib.cpp:399
#22 0x00007ffff4f8bd5b in QEventLoop::exec (this=0x7fff18bfcad0, flags=...) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/global/qflags.h:34
#23 0x00007ffff508356d in QThread::exec (this=<optimised out>) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/global/qflags.h:74
#24 0x00007ffff50fc0b0 in operator() (__closure=<optimised out>) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/thread/qthread_unix.cpp:375
#25 (anonymous namespace)::terminate_on_exception<QThreadPrivate::start(void*)::<lambda()> > (t=<optimised out>)
    at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/thread/qthread_unix.cpp:311
#26 QThreadPrivate::start (arg=0x55557eb8c710) at /usr/src/qt6-base-6.8.3+dfsg-0ubuntu2/src/corelib/thread/qthread_unix.cpp:339
#27 0x00007ffff46a27f1 in start_thread (arg=<optimised out>) at ./nptl/pthread_create.c:448
#28 0x00007ffff4733b5c in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

gzotti:
I have seen the incredibly long list of voices on Linux, yes. Any idea how to show a slider in the ComboBox? (Other GUI element?)  On Windows I have 3 voices to choose from. And it seems the language/locale is currently only reconfigured after a program language change AND then changing the engine. This can hopefully be fixed.  A "normal" user is expected to have selected some program language and keep it. Then only one voices from the available engine/languages combination should be selected. 

I cannot change quality of the voices themselves. Some sound bad, some are better. And I also expect less usable voices in some minor languages, This is a problem, but not ours to solve. I welcome configuration/optimisation hints for speechd for the SUG. 

The crash; It seems as no voice for Russian locale was found, the first in the list was taken regardless of locale, and then this English voice crashed in its attempt to speak Russian. So the case of "default" or "any voice" must be changed, and better find no voice than take "whatever is here".

10110111:
For the long list, the following patch helps me:
```diff
diff --git a/data/gui/normalStyle.css b/data/gui/normalStyle.css
index f8607835b3..f1609f9480 100644
--- a/data/gui/normalStyle.css
+++ b/data/gui/normalStyle.css
@@ -675,6 +675,7 @@ QComboBox QAbstractItemView {
 
 QComboBox {
        padding-left: 3px;
+       combobox-popup: 0;
 }
 
 /*
diff --git a/src/gui/ConfigurationDialog.cpp b/src/gui/ConfigurationDialog.cpp
index 620bb1b8f2..46382dcdc8 100644
--- a/src/gui/ConfigurationDialog.cpp
+++ b/src/gui/ConfigurationDialog.cpp
@@ -2584,6 +2584,7 @@ void ConfigurationDialog::populateVoiceCombo()
 
                int idx=ui->comboBox_Voice->findData(currentVoice);
                ui->comboBox_Voice->setCurrentIndex(qMax(idx, 0));
+               ui->comboBox_Voice->view()->setVerticalScrollBarPolicy(Qt::ScrollBarAlwaysOn);
        }
 }
 
diff --git a/src/gui/configurationDialog.ui b/src/gui/configurationDialog.ui
index b2cabcd6d2..5daa6a892a 100644
--- a/src/gui/configurationDialog.ui
+++ b/src/gui/configurationDialog.ui
@@ -3071,6 +3071,9 @@
                <property name="toolTip">
                 <string>Voice in program language.</string>
                </property>
+               <property name="maxVisibleItems">
+                <number>7</number>
+               </property>
               </widget>
              </item>
             
```
Note that the CSS change affects all the comboboxes, particularly non-editable ones. This will make the dropdown actually a dropdown, rather than a list that overlaps the combobox button. In fact, I'd prefer it to be applied regardless of this PR.

As for the other changes, the scrollbar is necessary to actually notice that there are many items, otherwise it's not obvious, and to be able to scroll when the mouse wheel/touchpad scrolling aren't available. `maxVisibleItems` property might need some tweaking, I chose 7 because it's the maximum on my screen+font size combo.

Unfortunately the dropdown doesn't become a drop-up; this seems to be a Qt issue with the `QGraphicsView` viewport not being taken into account. In fact, this is the real reason why this list is so problematic.

10110111:
With only two weeks remaining before the release, I don't think the changes to the SC descriptions should really go in now. Otherwise they won't get translated to most of the languages except the few most often updated. And thus, this PR may still be better off postponed to the next release.

gzotti:
My hope was to cherry-pick at least the SC descriptions into master later today so there are still ample 2 weeks for translation while fiddling with the technical implementation. The GUI strings will take a few minutes only and can come next weekend. I don't know how many percent of users just use English, but I guess "many". Their feedback in functionality would be valuable in Jan-Mar.

gzotti:
This branch now includes explanations for each constellation in modern SC. I rebased and squashed a bit to have all SC changes in one commit which could be cherry-picked into master, so that translators could start to work. This is mostly independent from speech output, but should happen ASAP if translated constellation descriptions should be available in 25.4.  For the branch/PR, there would be only a few GUI strings, I think this could still be manageable when finished merged next Sunday. Comments? Or just postpone to 26.1 with 3 curious beta testers starting in January? @alex-w , @10110111 ?

alex-w:
I vote to risk - merge it ASAP

gzotti:
OK, description.md upgraded in master.

10110111:
For the GUI strings, if you expect that they are mostly done, they could be temporarily pushed via `QT_TR_NOOP` in some file under `src/gui` to be picked up now, rather than in a week.

gzotti:
There is a file src/translations.h already, this looks like a temporary place to fill, maybe tonight (Indeed there are a few more string series...). Not sure if QT_TR_NOOP would have been the recommended way (?)

gzotti:
I wondered why I was missing many constellation narrations on Transifex. Is seems to have to do with the names which were in Latin :-( @10110111 are only those extracted which match a name in index.json? You explained some testing tool to Youla. Does it also test matches between index.json and description.md? This could be integrated to CI testing or the GH PR checklist notes. 

If this was the reason, I could cherry-pick the updated file to master again.

10110111:
> You explained some testing tool to Youla.

This is our good old `generate-pot.py` :) You can just run it yourself, passing the single SC you want to check to the script using the `-s` flag.

> are only those extracted which match a name in index.json?

Yes, and those that exist in the description but not in the index result in a warning.

> This could be integrated to CI testing or the GH PR checklist notes.

It could be integrated if we manage to fix all our current SCs, which currently emit loads of warnings.

gzotti:
Thanks. That's picky with visually empty 1-space lines... but OK. We have no names in ray helper asterisms, for now I leave the warnings. And I see weeks of fixing ahead :-o

gzotti:
@10110111 can you still reproduce the previous crash? 
I updated the descriptions in master. How to trigger (1) extraction to Transifex and (2) fetch from Transifex even though German translation is not finished yet? (And I won't manage to finish all by myself.)

10110111:
> can you still reproduce the previous crash?

Yes, I just opened the Speech tab, replaced the English text with "тест голоса", pressed "Say" and got the crash.

gzotti:
Did you actually set a voice (did GUI list a Russian voice?) the last part (~20 lines?) of logfile would be helpful.

10110111:
> Did you actually set a voice (did GUI list a Russian voice?) the last part (~20 lines?) of logfile would be helpful.

I used the flite engine with the default `slt` voice. There is no language-specific voice in the list. Here's the log from this session:

[log.txt](https://github.com/user-attachments/files/24149438/log.txt)

With speechd + Russian voice it works without crashing.

gzotti:
I have just added some more diagnostics in say() which may identify what happens. Can you please repeat?
It seems that when no voice for current language/locale can be found, some default English voice is taken. Maybe this then is unable to cope with a Russian string, or there is really no proper voice configured. On Windows the fallback English voice just does what it can do reading other Latin-based text, but it silently jumps over the Russian. I have no Russian voices installed.
Should we exclude flite on Linux?

10110111:
> I have just added some more diagnostics in say() which may identify what happens. Can you please repeat?

Nothing changes. The issue seems to be inside `QTextToSpeechProcessorFlite`. Namely, in Qt 6.8.3 (Ubuntu 25.10) `QTextToSpeechProcessorFlite::audioOutputCb` calls `QTextToSpeechProcessorFlite::startTokenTimer`, which depends on `QAudioSink` already existing. But it's only created after this call, in `QTextToSpeechProcessorFlite::audioOutput` that calls `QTextToSpeechProcessorFlite::initAudio`.

What does avoid crash is `Say`ing a word in English, and then continuing with the Russian input. Also `Say`ing a mixture where the first fragment (even a single letter) is English, and the rest is in Russian. Of course, in both these cases the Russian text remains silent, but at least there's no crash.

So the solution could be to `Say` a short string of English text (e.g. `a`) with volume 0 when `flite` is selected (so that audio sink gets created), including when it's the default engine, and then continue normally with the volume set in GUI.

gzotti:
Phew, that's weird. Or maybe I missed another check if engine is ready after language change, and flite is just slower than the other.

10110111:
Just noticed that, if I select kal16 in the voice list, play a Russian string (having already played English to avoid the crash), then the voice gets stuck: switching from it to other voices doesn't work, and playing English text too. Resetting the engine by selecting speechd, playing an English string, and going back to flite helps.

gzotti:
I also discovered that stop() causes a crash when sapi is used. I can fix that by interrupting with another say(".  ").  

I prepend the same inaudible ".  " if flite is used. Does that help? Else I will put in the "ee " at zero volume. I think else it works reasonably well.

10110111:
> I prepend the same inaudible ". " if flite is used. Does that help?

No, ". " itself crashes Stellarium just as the Russian text does. Apparently, the sound sink doesn't get created when there are no sound samples to emit.

gzotti:
Like this?

10110111:
Yes, now it doesn't crash. The comment should be corrected though: the question is not in no valid voice existing, but in no sound samples emitted. E.g., as I said above, ". " also crashed despite containing no Cyrillic characters.

gzotti:
SUG additions still missing. It also just occurred to me I probably have not cared for translating constellation narration before voice output. This needs a few translated strings to test...

gzotti:
The Speech tab in the configuration panel has the same icon as "Tools", which I regard sufficient for now, but am happy about volunteers to design a new icon for 26.1.

gzotti:
Is it just my system? Transifex gobbles linefeeds! This breaks structuring :-(((

10110111:
> Transifex gobbles linefeeds!

Where do you see this?

gzotti:
When I translate in Firefox, any "return" characters (little arrows that bend down/left) are visible while editing and new lines are shown in the edit box, but as soon as I press "save" the arrows are gone, and the whole string comes without newlines.

10110111:
> any "return" characters (little arrows that bend down/left) are visible while editing and new lines are shown in the edit box, but as soon as I press "save" the arrows are gone, and the whole string comes without newlines.

Yeah, I see this in your recently changed strings. This works fine in Chromium 137.0.7151.103 for me, as I've checked by fixing some of these strings.

gzotti:
It's stupid that once changed I could not revert/abort after storing. Must switch to Chrome then.

10110111:
> It's stupid that once changed I could not revert/abort after storing.

Indeed, I have seen some strange stripping of my translations by Transifex (in some older versions of Chromium), so I normally copy the final string to clipboard before storing.

gzotti:
I try to reduce SC narration in the GUI to the sections Introduction and Description. Somehow the MD processing works a bit weird, and I just don't find the right way to throw away everything starting with "## Constellations". I could suppress the sections one-by-one, but the text (not title) from the last section is still visible. Sorry, I am probably tired and not fully operational... Ideas?

gzotti:
Wow, that was stupid...

gzotti:
IMO it is OK for now. I cannot test how translations work before merge, though. Any objections? @alex-w @10110111

gzotti:
@alex-w I have translated the modern constellation lore to German in Transifex, but dozens of strings for other cultures remain untranslated in German, so this 100%-finished trigger does not fire. How can I still trigger that whatever is there will be copied back to our repo?  I expected the translated strings to be at least visible in the SC GUI screen, but also here I see German constellation names with English text, either by not being translated or by technical error of not translating before display.

alex-w:
See settings/integration/github “manual sync” at Transifex

10110111:
> I have translated the modern constellation lore to German in Transifex, but dozens of strings for other cultures remain untranslated in German, so this 100%-finished trigger does not fire.

There's a timeout of 30 minutes IIRC, so that we don't get too many Transifex commits for some few string changes while translators are still working.

gzotti:
I see my constellation descriptions in the SC panel long text now (so they are in my master checkout), but I have to change something to actually read them translated (Constellation InfoString and narration) But I don't see the Info/narration output  related new strings, which I had offered in advance for Transifex translation in master, and the exchange after the narration merge also did not retrieve them. Is there still some manual step involved, or is src/translation.h no longer used?

10110111:
> Is there still some manual step involved

Yes. There is the `generate-pot` target (not the `util/skycultures/generate-pot.py` script, but an actual build target). E.g. I can run it like so:
```
cmake --build ~/package-builds/stellarium-build-qt6/ --target generate-pot
```

gzotti:
I am afraid i need more guidance stepping out of my comfort zone. I am usually just within QtCreator and building exe's. Is there a target in QtCreator to call in the menu? Alternatively, I have a WSL build directory build_wsl next to my stellarium checkout where I called cmake ../stellarium, that writes the CMakeCache (to be edited, toggle plugin build etc). and make -j20 does the rest. I launch the program from within the stellarium checkout. Would I call that command from my build_wsl or in a fresh build directory? And, would that change files in the source dir that then need to be pushed to Github? Your command suggests you have a directory package-builds where all versions are built. From which dir do you call that command then?

10110111:
> Is there a target in QtCreator to call in the menu?

Not in the menu, but in the project config (appears on pressing <kbd>Ctrl</kbd>+<kbd>5</kbd>). There in the Build Steps you can click _Details_, and in the opening controls check the target you're interested in (unchecking the default one). Then after you "build" the project this target should be executed instead of the one you unchecked. `generate-pot` changes the `.pot` (and a `.jst`) files in the source directory, regardless of where the build directory is. These files must be committed.

> Alternatively, I have a WSL build directory build_wsl next to my stellarium checkout where I called cmake ../stellarium, that writes the CMakeCache (to be edited, toggle plugin build etc). and make -j20 does the rest.

Then you can just `make generate-pot` and get the same result as described above.

> Would I call that command from my build_wsl or in a fresh build directory?

If you go the `cmake` route as I said in my previous comment, you already specify the build directory on the command line, and the source directory is stored in `CMakeCache.txt`. So you can run `cmake` from any directory.

src/core/modules/NomenclatureItem.cpp
```diff
@@ -296,6 +298,102 @@ QString NomenclatureItem::getInfoString(const StelCore* core, const InfoStringGr
...
+		{
+			const QString orStr = qc_("or", "object narration");
+			const QString latin = getNomenclatureTypeLatinString(nType); // not always latin!
+			oss << QString(", %1 %2, ").arg(orStr, latin);
+		}
+
+		const QString description = getNomenclatureTypeDescription(nType, planet->getEnglishName()); // TODO Needs articles etc?
+		if (nType!=NomenclatureItem::niUNDEFINED && nType<NomenclatureItem::niSpecialPointPole && !description.isEmpty())
+		{
+			const QString verb = qc_("that is", "object narration)");
```
10110111, line 332: This fragment is untranslatable. This kind of splitting of sentences into English-centric units goes against the main principle of translatable software: _preserve context and make grammar flexible_. Particularly, this "that is" could be "i.e.", or "[something] that is [called foobar]", etc., and of course this could be implemented as a very different grammatical structure at all.

Qt provides the `%N` placeholders exactly for this purpose: to make translatable strings with only a few terms substitutable, but with the general grammatical structure remaining flexible, and semantics being obvious.

gzotti: So, these sentences should be longer, and translatable strings shown with % signs? I haven't seem them as translator in a while now, but I agree I had mostly English/German in mind. Probably several structures need re-formulation in more complete sentences, at the latest in 26.1, but now that I see this in action, it's easier to work with it. Also, finally the strings have appeared on Transifex, and I just write them AQAP while my cmake-build is still under way (single-threaded? Not sure, is the --target  generate-pot supposed to also build the whole program?

10110111: > So, these sentences should be longer, and translatable strings shown with % signs?

Of course, they've always had to be as full as possible. English and German are both Germanic languages, which makes their grammar very similar. So the experience only with them is not exactly helpful to realize how different other languages' grammars can be.

> while my cmake-build is still under way (single-threaded? Not sure, is the --target generate-pot supposed to also build the whole program?

@alex-w has already generated the pot files. But yes, apparently `generate-pot` has the main target as a dependency.

Why it's single threaded I don't know, but you can always supply `-jN` in the _CMake arguments_ field.

gzotti: OK, 100% German done. Does that trigger Transifex immediately, or shall I sync again to speed up? Taking a short break now, but will continue with a test and then, well, reformulating StelObject ans StarWrapper first.

gzotti: Meh, for reasons unclear, line StelUtils::1649: qc_("point", "decimal separator") was not processed.

10110111: Hmm, that's interesting. For some reason `StelUtils.cpp` is not listed in `po/stellarium/POTFILES.in`. And many files that are found by `find src -name '*.cpp' -o -name '*.hpp' | grep -v /external/` aren't listed there too.

@alex-w Is there any reason for this? Or should these files be added?

gzotti: Oh, StelUtils was not in POTFILES? Didn't expect this. Are there no other q_ string in it than my new ones? On that matter, can/should all files with at least one "q_(" or "qc_(" be listed (automatically/automatable) there or would that break something?

10110111: BTW, while in English (and presumably German) it's normal to say "12.345 hours" as "twelve point three four five hours", in other languages the rules are different: e.g. in Russian it's an equivalent of "twelve wholes, three hundred fourty-five thousandths of an hour". See e.g. [Google Translate's pronunciation of both English and Russian](https://translate.google.ru/?sl=en&tl=ru&text=12.345%20hours&op=translate) (the Russian emitted is "двенадцать целых, триста сорок пять тысячных часов", which you can type in and hear to be exactly the same (this has the "of an hour" replaced with "hours", but we're interested in the number here)).

While it's not easy to make code for narration in every possible language, it's not a good idea to manually carve this structure. How does your system pronounce "12.345" in English or "12,345" in German (possibly with a unit like "hours" appended)? Isn't it doing a good job already?

gzotti: Unfortunately, with 12.345 (using C locale's dot) it said "Punkt" and continued with dreihundertfünfundvierzig, which is "point" and is simply not the way these numbers should be spoken. This was my motivation behind narrateDecimal(). I could also just make a locale-dependent "if 'de' do as now else just do whatever is set in your language engine when delivered a dot-based decimal number", and leave other languages to individual contributing developers.

10110111: And what if you type a comma? (and a unit?) I think we could force a locale in this function to format the number (e.g. with `std::ostream::imbue` and `std::locale` if not with `QLocale`).

gzotti: I just added a "German only" exception and would spend the next hours with StarWrapper. As said, I expect refinements for 26.1, so, only the worst issues should be solved immediately. The feature works for English, and "we aim to improve" for others.

gzotti: Working with correct locales would certainly make sense. Typing a comma-based number in the test text box works well.

10110111:
Why are there some instances of "minutes", "seconds", "degrees", "and", "called", "is", "minus", "plus" without translation context? Even if we don't fix these being single words taken out of phrase for 25.4, they at least should have the context for the translators to be able to look into the sources and figure out how to translate them best.

10110111:
> Its Separation for %1 is %2 arc-seconds (%3).

Was it supposed to have "its component separation", rather than "its separation"? Otherwise the question arises, "its separation from what?"

> Its type is %1 with variable type %2

Shouldn't it be "variability type"? At least this is the name of the function that's called to get the relevant variable: `StarMgr::getGcvsVariabilityType`.

gzotti:
> Why are there some instances of "minutes", "seconds", "degrees", "and", "called", "is", "minus", "plus" without translation context? Even if we don't fix these being single words taken out of phrase for 25.4, they at least should have the context for the translators to be able to look into the sources and figure out how to translate them best.

I had expected these really very common words to be already translated or actually not needing context. But most instances of "is" are now inside %'ed sentences. A problem are optional supplementary data that - if present - best fit within one spoken sentence require long alternatives (see NomenclatureItem). I hope the grammatical concept of combining two main clauses with  "and" can be found in most languages. 

Component separation: yes, that's clearer, please replace (all instances in StarWrapper! See #4698). (I probably should stop working at 2:30 am..) Variable type:: Unsure about the 100% best word in English and am open to corrections. In German it's "Veränderlichentyp", not "Veränderlichkeitstyp". If you are sure about the best English word, please update.

10110111:
> If you are sure about the best English word, please update.

This is the original term from GCVS: https://heasarc.gsfc.nasa.gov/w3browse/all/gcvs.html . And it makes more sense than "variable type", which could mean a type that changes, or a type of a variable, both of which aren't really applicable to variable stars.

> I hope the grammatical concept of combining two main clauses with "and" can be found in most languages.

Possibly so. But "plus"/"minus" as a sign can be translated differently than "plus"/"minus" as an operator — even in English it's sometimes said "negative one" to mean -1. I guess in other languages such construction might be mandatory.

Same for "degrees", "minutes", "seconds": while English may not distinguish between cases of these words, in other languages they may be in nominative case when discussed (e.g. "these minutes"), and in other cases when counted (e.g. "seven minutes"). This of course is aggravated by the need to distinguish "one minute", "three minutes", and "eight minutes" in e.g. Russian, where all these phrases have "minute" in different cases and pluralizations, but at least it shouldn't be nominative everywhere.

And the "called": it's also a fragment that can be translated very differently depending on context.

---

I've fixed the component separation and the variability type.

10110111:
There's also a problem with the "Rise on this day is at", which is used in narration like "Rise on this day is at: 10 hours, and 5 minutes". This results in the preposition that is broken off its continuation word, e.g. "Восход в этот день происходит в: 5 часов, и 29 минут.". Not only is the "and" useless here (this is another reason why times and dates should be given to the translator as a whole), but also RHVoice fails to pronounce "в:" smoothly, instead it pronounces it as a letter name followed by a pause.

IMO the colon should be removed. Then the sentence gets its natural structure, and RHVoice appears happy with it.

gzotti:
the separate plus and minus are IIRC from the various StelUtil::*narrate. (Side note: Do we really need so many number manglers? History shows...) Funny enough, the plusminus sign is pronounced well in English and German.

minutes and seconds: Yes, we'd also need differentiating between singular and plural forms etc. While German voice on Windows correctly reads "um 9:23:45" mit "um neun Uhr 23 Minuten und 45 Sekunden" (The particle "um" indicates a point in time like "at"), "um 9:23.45,6" will become "um neun [pause] dreiundzwanzig [pause] fünfundvierzig [pause] sechs" (as, apparently, seconds with a decimal are regarded as so unlikely that "this cannot be a time"). And I don't know what voices on Mac or Linux do here. Explicitly separating out the numbers and units and building the strings "manually" seemed best unless you find a better solution. (I'd rather avoid making an if (lang=="de") test. What about Spanish, Portuguese, Arab, ... Better "stupid computer voice doesn't get the singular right" than "voice doesn't give units", IMO.)  Still not done: 0°0'0.4'' should not become "zero degrees, zero arcminutes and zero point 4 arcseconds", but just "zero point four arcseconds".

Colon, point etc. What sounds good in English may sound worse in Russian. Sure. I hear a little "pause of announcement" after colon which I want. Translators may omit it where it sounds better. And I need to attach points and spaces and even " . . " to audibly separate sentences and add slightly longer pauses. There are also some words like "Doppelstern" which sound stupid (sounds like "Doopelstern" with a weird long o) in German voice on Windows. Again, no experience with the other OSes. Separating written and spoken labels is an option, writing "Doppel Stern" (wrong grammar) sounds better.

As said, I did not expect to provide perfect strings in 25.4, but the coming weeks provide the chance to react to user experience. I did also not complete Planet narration (I did not intend to tell everything at first, but now with the selectable flags it would be possible and should come in 26.1, together with narration for more object types). We may also collect experience with integrating better-sounding AI-based voice output on Linux (OONX).

I had another issue with "rising time". Context free, it was translated by someone as "Aufgangszeit", but the context was variable star light curve, where "Anstiegszeit" (growing time) seems more appropriate.

10110111:
> "um 9:23.45,6" will become "um neun [pause] dreiundzwanzig [pause] fünfundvierzig [pause] sechs"

You have a period after 23 here, which may have influenced the readout. Google Translate pronounces "um 9:23:45,6" better, but the final "sechs" is pronounced as a separate completely unrelated number.

> Colon, point etc. What sounds good in English may sound worse in Russian. Sure. I hear a little "pause of announcement" after colon which I want. Translators may omit it where it sounds better.

I'd happily omit it, but it's not included in the translatable string, instead it's added in `%1: %2`.

> There are also some words like "Doppelstern" which sound stupid (sounds like "Doopelstern" with a weird long o) in German voice on Windows. Again, no experience with the other OSes. Separating written and spoken labels is an option, writing "Doppel Stern" (wrong grammar) sounds better.

I guess this is an OK tweak to be done by the translators. But the source English shouldn't be littered with this kind of kludges. (We have `en_US`, `en_GB` etc. where they may be implemented if English appears to be pronounced badly.)

And of course, if you tweak something for Windows, you should check that it doesn't break on Linux (the latter is free to check, unlike the former).

> I had another issue with "rising time". Context free, it was translated by someone as "Aufgangszeit", but the context was variable star light curve, where "Anstiegszeit" (growing time) seems more appropriate.

Oh, I may have made a similar mistake in the Russian translation then. There should be a `TRANSLATORS:` comment that explains this.

gzotti:
> > "um 9:23.45,6" will become "um neun [pause] dreiundzwanzig [pause] fünfundvierzig [pause] sechs"
> 
> You have a period after 23 here, which may have influenced the readout. 

Nah, I had used a colon in the text field, just mistyped here.... 

> Google Translate pronounces "um 9:23:45,6" better, but the final "sechs" is pronounced as a separate completely unrelated number.

That's what I mean. It does not expect seconds to have further fractions. Therefore, we must find our own solution.

 
> > Colon, point etc. What sounds good in English may sound worse in Russian. Sure. I hear a little "pause of announcement" after colon which I want. Translators may omit it where it sounds better.
> 
> I'd happily omit it, but it's not included in the translatable string, instead it's added in `%1: %2`.

Ah, a leftover after the latest round... That means, all narration strings must take care to not copy the InfoString too closely. 
 
> > There are also some words like "Doppelstern" which sound stupid (sounds like "Doopelstern" with a weird long o) in German voice on Windows. Again, no experience with the other OSes. Separating written and spoken labels is an option, writing "Doppel Stern" (wrong grammar) sounds better.
> 
> I guess this is an OK tweak to be done by the translators. But the source English shouldn't be littered with this kind of kludges. (We have `en_US`, `en_GB` etc. where they may be implemented if English appears to be pronounced badly.)

The problem is that I need two source strings, one for display, the other for narration. This can be solved with translation context and indeed more "//TRANSLATORS" comments. . I can solve it for English and German (mainly on Windows.), but not for others.  And yet other voices/speech plugins may have other surprises, like the crash you found.
 
> And of course, if you tweak something for Windows, you should check that it doesn't break on Linux (the latter is free to check, unlike the former).

Sure. I run Linux on my Raspbies and various tools on the indispensable WSL. But still need W for daily income, so my time exposure is like 92% W, and "testing on Linux"  requires a moderately capable system (U24, not 22 like my old "battleship" which cannot be upgraded thanks to NVidia dropping drivers, or dual boot) with sound. My 10-year old NUC is not really fast but OK for a quick function test with ear plug. The RasPi4 is ARM, so probably a special case. I will not finetune Linux sound output there. From the download numbers, I see Windows has >50% downloads from Github/stellarium.org, but of course Linucians use their package manager which has other channels. Do you have an estimate of Linux users?

I probably won't be active here for a few hours now. Merry Christmas!

PR #4664, opened 2025-11-21
SCM: Fix: Thickness of constellation lines in SCM is not tied to Stellarium's settings (#4659)

### LINKED ISSUE #4659: SCM: Allow line thickness to be changed, or matched with constellation line setting
In my planetarium, I set the constellation line thickness to 5 (the maximum) to make it easier for the audience to see. Lower than this can be difficult, especially with a 1080p projector.

The current Sky Culture Maker lines are very thin and difficult to see in a planetarium dome. This is also an accessibility issue for people that need thicker lines.

Either:

- match the constellation line thickness preference
- add a setting in SCM to change line thickness

**Additional context**

Screenshot shows default constellation line thickness vs CSM line thickness. Note, this is a screenshot while spherical mirror distortion was enabled, so it looks a bit warped.

Image

xLPMG:
First of all, I apologize for making so many code changes again, even though this could have been a quick fix. In my opinion, there is a lot of unnecessary complexity in the code which reduces maintainability. I am slowly trying to clean this up step by step.

In its core, I copied the code for the `constellationLineThickness` and `scale` properties from the `ConstellationMgr`, which can be used to set the constellation line thickness.

Furthermore, some parts of the code just seemed confusing: `drawConstellation` was drawing lines, but also calling separate draw functions for the name and artwork. The doc in the header did not mention at all what was being drawn exactly. Each drawing function had its own isHidden check, which imo is also the wrong place.

I created separate drawing functions (`drawLines`, `drawName`, `drawArtwork` and `draw`, which draws everything). The variable isHidden can now be used to check whether `draw` should be called or not. Each drawing function now also only takes in the parameters that it really needs. I think the code is more flexible now :)

Fixes #4659

### Type of change

- [ ] Housekeeping

## Checklist:

### DIFF
--- a/plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
+++ b/plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
@@ -38,8 +38,8 @@ void scm::ScmSkyCulture::setFallbackToInternationalNames(bool fallback)
 }
 
 scm::ScmConstellation &scm::ScmSkyCulture::addConstellation(const QString &id,
-															const std::vector<ConstellationLine> &lines,
-															const bool isDarkConstellation)
+                                                            const std::vector<ConstellationLine> &lines,
+                                                            const bool isDarkConstellation)
 {
 	constellations.push_back(std::make_unique<scm::ScmConstellation>(id, lines, isDarkConstellation));
 	return *constellations.back();
--- a/plugins/SkyCultureMaker/src/ScmDraw.cpp
+++ b/plugins/SkyCultureMaker/src/ScmDraw.cpp
@@ -25,9 +25,9 @@
 #include "ConstellationMgr.hpp"
 #include "StelActionMgr.hpp"
 #include "StelModule.hpp"
+#include "StelModuleMgr.hpp"
 #include "StelMovementMgr.hpp"
 #include "StelProjector.hpp"
-#include <StelModuleMgr.hpp>
 #include <QApplication>
 #include <QDebug>
 #include <QKeyEvent>


xLPMG:
On a separate note: is there a way to change the global constellation line thickness in the Stellarium GUI? I couldn't find this information in the SUG. There seems to be only the option to change it for each sky culture separately, but that was not saved to my `config.ini`. To test this feature, i manually added `constellation_line_thickness` to the config file.

gzotti:
This is not saved per-skyculture. You can just query 

`GETSTELMODULE(ConstellationMgr)->getConstellationLineThickness()
`

xLPMG:
> This is not saved per-skyculture. You can just query
> 
> `GETSTELMODULE(ConstellationMgr)->getConstellationLineThickness()`

Interesting, I suppose I should also replace
```
QSettings *conf = StelApp::getInstance().getSettings();
constellationNameFont.setPixelSize(conf->value("viewing/constellation_font_size", 15).toInt());
```
with
```
constellationNameFont.setPixelSize(GETSTELMODULE(ConstellationMgr)->getFontSize());
```

10110111:
> Interesting, I suppose I should also replace

And also subscribe to `ConstellationMgr`'s `fontSizeChanged` signal and update font size accordingly.

gzotti:
Either this if you have your own variable, or just retrieve ConstellationMgr's values whenever you need them like you proposed. If used frequently, retrieve GETSTELMODULE into a proper variable to avoid needless lookups.

xLPMG:
I turned ScmConstellation into a QObject in order to properly connect to the config values. I did not do this for ScmDraw, since it was more work than expected... a compromise could be to reload the config in the `resetDrawing` function in ScmDraw. So whenever the user wants to draw a new constellation, the current config would be loaded

plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
```diff
@@ -25,6 +25,12 @@
 #include "types/Classification.hpp"
 #include <utility>
 #include <QFile>
+#include <QTextStream>
```
xLPMG, line 28: I added this include because QTextStream is used in the file but was never included before.

alex-w:
@10110111 @gzotti any comments?

10110111:
What was the point of switching from a vector of constellations to a vector of pointers to constellations? This makes the code harder to comprehend, but doesn't seem to add any benefits.

xLPMG:
> What was the point of switching from a vector of constellations to a vector of pointers to constellations? This makes the code harder to comprehend, but doesn't seem to add any benefits.

I didnt want to do this, but had to in order to make Constellations QObjects. And this was done so that i can use connect()

10110111:
> I didnt want to do this, but had to in order to make Constellations QObjects.

But why did you need this? The only signals handled are from `ConstellationMgr`, not from `scm::ScmConstellation`.

xLPMG:
> > I didnt want to do this, but had to in order to make Constellations QObjects.
> 
> But why did you need this? The only signals handled are from `ConstellationMgr`, not from `scm::ScmConstellation`.

but the function "connect" is defined inside QObject

Edit: Im totally in favor of not changing to a vector of pointers to constellations. But I dont know how to call 'connect' otherwise.

10110111:
> but the function "connect" is defined inside QObject

But it is `static` :)

xLPMG:
> > but the function "connect" is defined inside QObject
> 
> But it is `static` :)

Oh.. then this would work?

`
	QObject::connect(constMgr, &ConstellationMgr::constellationLineThicknessChanged,
			[this](int v) { constellationLineThickness = v; });
			`

10110111:
> then this would work?

Sure. You could even call `QObject::connect` from free functions, it's completely independent of the caller being a `QObject`.

xLPMG:
> > then this would work?
> 
> Sure. You could even call `QObject::connect` from free functions, it's completely independent of the caller being a `QObject`.

I now ran into the problem of dangling pointers. The previous code used
`
scm::ScmConstellation &scm::ScmSkyCulture::addConstellation()
{
	scm::ScmConstellation constellationObj(id, lines, isDarkConstellation);
	constellations.push_back(std::move(constellationObj));
	return constellations.back();
}
`

which changes the address of the constellation and thus the captured lambda becomes invalid when doing:
`
QObject::connect(constMgr, &ConstellationMgr::constellationLineThicknessChanged,
			[this](int v) { constellationLineThickness = v; });
`
resulting in a seg fault.

A simple solution for this problem right now is to use
`
scm::ScmConstellation &scm::ScmSkyCulture::addConstellation()
{
    constellations.emplace_back(id, lines, isDarkConstellation);
    return constellations.back();
}
`

but should the address of the constellation change at any time, maybe because the vector does a reallocation (im not sure about the specifics here), this problem will appear again.

Could this even happen? What do you think?

Edit: Based on https://stackoverflow.com/questions/2447392/does-stdvector-change-its-address-how-to-avoid this could be a real issue, in which cases the current appraoch with pointers could be better.

10110111:
OK, I see. Yes, reallocation of a vector indeed invalidates all the pointers to its elements. Then the use of a vector of pointers indeed makes sense. Just please leave a comment near its declaration that explains why this has to be so.

xLPMG:
> OK, I see. Yes, reallocation of a vector indeed invalidates all the pointers to its elements. Then the use of a vector of pointers indeed makes sense. Just please leave a comment near its declaration that explains why this has to be so.

Alright. This can also be done either with or without using QObject. Is there any benefit from not using QObject?

10110111:
> Is there any benefit from not using QObject?

The intent is shown first of all. If you added `Q_OBJECT`, you're expected to need something from the `QMetaObject` framework for this object. If you inherit from `QObject`, you're expected to use its nonstatic (or even protected) methods or to upcast to `QObject` pointer or reference. If nothing of this is of any real use, there's no reason to add this (and there's some overhead to this, but this is not the most important point here).

xLPMG:
Alright, thank you :) ill remove it then

gzotti:
Is there actually necessity to use std::vector, not QList? At some point (many years ago) IIRC we had even banned all std:: stuff. Only when Qt is lacking some feature we should find other solutions.

10110111:
> Is there actually necessity to use std::vector, not QList?

I'd ask to the contrary: is there any necessity to use Qt containers here, especially if they are also moving towards being more or less like the std ones. E.g. `QList` is just the same as `QVector` since Qt 6. Moreover, its iterator validity guarantees are even more vague than those of `std::vector`: e.g. any modification of a `QList` invalidates all its iterators, while in `std::vector` only the modification that results in a reallocation does.

> At some point (many years ago) IIRC we had even banned all std:: stuff.

This used to be a typical behavior for Qt-based projects during the times of Qt 4, because the C++ standard libraries were very standards-incompliant (especially those from MSVC). Nowadays there's no reason to go the all-Qt way (except when you want, say, implicit sharing).

plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
```diff
@@ -60,11 +59,9 @@ void scm::ScmSkyCulture::removeConstellation(const QString &id)
 
 scm::ScmConstellation *scm::ScmSkyCulture::getConstellation(const QString &id)
 {
-	for (auto &constellation : constellations)
-	{
-		if (constellation->getId() == id) return constellation.get();
-	}
-	return nullptr;
+	auto it = std::find_if(constellations.begin(), constellations.end(),
```
10110111, line 57: Funny, although this is following the "best practices", it (together with the subsequent `return` line) does look more complicated than the naive for loop, and I think is not worth it.

PR #2619, opened 2022-09-01
Avoid screen blanking on Intel UHD (#2166 #2472 #2484)

uwes-ufo:
### Description
Avoid screen blanking on Intel UHD (#2166 #2472 #2484)
Screen flickering on Intel UHD Graphics is a long-known issue. The backlight turns off and on multiple times! The current workaround is to switch to Angle mode. But in Qt6 there is no more angle mode...
Perhaps the user should make their choice based on a configuration setting value like:
... if (confSettings->value("video/SwapBehaviorSingleBuffer", false).toBool()) ...

Fixes #2166
Fixes #2472
Fixes #2484

### Type of change
- [x] Bug fix (non-breaking change which fixes an issue)

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11
* Graphics Card: Intel UHD 620

## Checklist:
- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4678, opened 2025-12-01
Dynamic selection between allowing switching between exclusive and managed fullscreen mode and forcing managed fullscreen mode

uwes-ufo:
### Description

In addition to #2619, this pull request offers dynamic switching between exclusive and managed fullscreen modes. This can be done by right-clicking the fullscreen button or by pressing Shift+F11.

Currently, switching between exclusive and managed fullscreen modes results in a brief screen flicker. Because exclusive fullscreen mode offers limited window functionality, the screen flickers every time, for example, a tooltip appears. #2619 keeps the screen permanently in managed fullscreen mode, thus preventing the flickering. Exclusive fullscreen mode, on the other hand, offers better performance, which translates into potentially higher frame rates.

I know the implementation is somewhat unusual… I based it on this [Stack Overflow question](https://stackoverflow.com/questions/60854878/issue-with-blanking-screen-when-running-qt-application-in-fullscreen-mode).

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:


uwes-ufo:
The change to `src/gui/locationDialogGui.ui` doesn't seem to be related. I'll check it out...

10110111:
> Exclusive fullscreen mode, on the other hand, offers better performance, which translates into potentially higher frame rates.

The change looks quite hacky. Do you actually get any tangible benefits from this mode?

10110111:
> Graphics Card: <Manufacturer (likely Intel, NVidia, AMD?), Model (HD, Geforce, Radeon..., with model number), driver version?>

I'd rather have this filled...

uwes-ufo:
This PR addresses the same issue as #2619. It resolves issues #2166, #2472, and #2484. It's again Windows on an Intel UHD 620. [Here](https://github.com/Stellarium/stellarium/pull/2619#issuecomment-1242932669), you tested the issue and asked how it works. I don't know the exact details either, but in my opinion, the number of buffers used is irrelevant to this problem. The `single buffer` has the side effect of permanently putting the graphics card into managed fullscreen mode, thus preventing a switch between exclusive and managed modes, which would trigger the flickering.

> The change looks quite hacky. 

Ja! It's strange, at the very least. I even find it ugly.

> Do you actually get any tangible benefits from this mode?

- Managed mode has no blinking effect. 
- Exclusive mode is faster.

10110111:
> Exclusive mode is faster

How much faster?

uwes-ufo:
- 60 BPS to 40 BPS without atmo
- 40 BPS to 30 BPS with atmo

on my hardware.

alex-w:
> * 60 BPS to 40 BPS without atmo
> * 40 BPS to 30 BPS with atmo
> 
> on my hardware.

What about BPS without patch?

uwes-ufo:
- no `--single-buffer` ~> exclusive fullscreen ~> blink effect + sometimes no mouse pointer
- - without atmo - 60 fps
- - with atmo - 40 fps

- `--single-buffer` ~> managed fullscreen ~> no blink effect + reliable mouse pointer
- - without atmo - 40 fps 
- - with atmo - 30 fps

This pull request provides an alternative to `--single-buffer`, which can be toggled during Stellarium's execution.

10110111:
So basically, this PR makes it possible to switch between different broken states: one blinking, another 25% slower.

Have you tried reaching out to Intel? Maybe they could provide a real fix for their drivers (they might not even know of the blinking problem). No other vendor has this blinking problem, neither should Intel.

10110111:
I wonder if we can, instead of all this hackery, just force the tooltips to be parts of the scene, just as most of the other dialogs.

uwes-ufo:
> I wonder if we can, instead of all this hackery, just force the tooltips to be parts of the scene, just as most of the other dialogs.

That would be an elegant solution to the problem.

I've noticed something else: Sometimes there's no mouse pointer in exclusive fullscreen mode, but the mouse works when you use it blindly.

gzotti:
That's a very old FAQ. Do you have some software cursor enabled like "mouse trail"?

uwes-ufo:
I have two practically identical laptops; all the software is essentially cloned. On one computer, the mouse cursor disappears, but not on the other...

I can confirm that the problem with the mouse pointer is related to the mouse trail.

10110111:
Apparently, it is possible to have scene-based tooltips. For this `QGraphicsScene::helpEvent` in `StelGraphicsScene` should be overridden to show the tooltips in a custom way.

10110111:
Note that the default tooltips have some special features like animation of one tooltip position into another when moving cursor from one control to another (instead of disappearing and appearing in the new place), as well as fade-in and fade-out effect (at least under KWin, and we have similar animations for screen buttons). These shouldn't be lost when the custom tooltips are implemented. But of course, a proof of concept could ignore these.

10110111:
I've made a proof of concept for the in-scene tooltips, see #4704.

uwes-ufo:
#4704 looks promising.

I've restored the original tooltips in managed fullscreen mode for debugging purposes, for easier comparison.

alex-w:
@uwes-ufo @10110111 any news please?

10110111:
I'm not sure what `TopMost` class is for. Is it a remnant from the fullscreen mode management from the times before the custom tooltips?

uwes-ufo:
- "Topmost" switches the graphics card to managed fullscreen mode. I use managed fullscreen mode for standard views where speed isn't a factor; there is no flickering here, not even when using Alt-Tab. To better evaluate the differences between the two tooltip options, I have reactivated the previous tooltips. 

- I use exclusive fullscreen mode in cases where maximum frame rate is desired; the new tooltips don't flicker in this mode, though Alt-Tab still causes flickering.

alex-w:
Would you prepare video samples for before/after behavior on your environment?

uwes-ufo:
First exclusive full-screen mode, then managed full-screen mode.
https://github.com/user-attachments/assets/1f37671a-3d78-461b-ba65-6fd2e17128f1

I think the flickering is so noticeable because the display's backlight turns off when switching between the two full-screen modes. If the backlight remained on, you might not even notice the transition.

gzotti:
Can you explain to a 'normal' PC user why we should have two fullscreen modes to select at runtime? I was not aware of this difference. If --single-buffer does the trick where needed, why do we need this again? (EDIT: OK, see below)

I tried this on my gaming laptop (Win 11, Geforce 3070) on external screen. 

1) Master, 
- without CLI arg --single-buffer
   * no tooltips
   * Youtube pop-out screen not visible 

- with CLI arg --single-buffer
   * Tooltips work 
   * Youtube pop-out screen visible 

No flicker in any mode.


2) This branch, 
- without CLI arg --single-buffer
   * hollow box in icon ('managed'): no tooltips
   * full box in icon: tooltips
   * Same framerate in both cases, but maybe something else is currently throttling this to 62.5. (I have observed much more earlier.)
   * Youtube pop-out screen not visible in both modes

- with CLI arg --single-buffer 
   * Tooltips work in both modes
   * Youtube pop-out screen visible in both modes.

No flicker in any mode.

I conclude --single-buffer is all needed at least with a Geforce on Windows.

3) Small field laptop (N150 based, unspecified "Intel Graphics"): master works without flicker, tooltips show in windows. --single-buffer not needed.


4) Older Intel-only system (field laptop): Pentium Silver N5030/UHD Graphics 605.
- master: 
  * no --single-buffer: Indeed, nasty flicker when tooltips should appear.
  * Maximized window, non-fullscreen: No issue
  * --single-buffer: No flicker

- This PR: 
  * no --single-buffer: 10 fps in managed, 15 fps in exclusive mode. Hmm, that is interesting, indeed. I confirm Flicker when Alt-TAbbing into exclusive-fullscreen Stellarium, but one flicker is acceptable IMO.
  * Maximized window, non-fullscreen: no issues
  * --single-buffer: no discernible difference, no flicker, both around 10fps.


But I mostly agree with @10110111 to report to driver manufacturer about flickering instead of working around with strange kludges. In any case, I think it's better to have, at best, a config.ini flag or CLI arg to express system necessity, it's usually a one-time configuration., e.g. `main/flag_fullscreen_mode_exclusive`  (Or switchable by script command for debugging/testing. But no user will like to switch between the modes after finding out the better one.) This should be explained to the non-tech user in a new section 6.3 in the User Guide.

I wonder, do Linux and Macs also have two fullscreen modes?

uwes-ufo:
I stumbled upon the two fullscreen modes somewhat by accident while investigating the cause of the flickering. My initial solution was to implement the `--single-buffer` setting. `--single-buffer` appears to switch the graphics rendering to "Managed Fullscreen" mode. Without `--single-buffer`, the system defaults to "Exclusive Fullscreen," likely due to the higher frame rate. However, if certain conditions prevent Exclusive Fullscreen—such as another window being displayed simultaneously—the system is forced to revert to Managed Fullscreen, resulting in a drop in frame rate. 

Unfortunately, on certain hardware, the screen backlight appears to turn off briefly during this switch. 

- unset `--single-buffer` allows dynamic switching between exclusive and managed full-screen modes.
- set `--single-buffer` forces managed full-screen mode.

Currently, switching between these two settings (enabling/disabling `--single-buffer`) in Stellarium requires restarting the program; this pull request allows making the selection while the application is running.

>So basically, this PR makes it possible to switch between different broken states: one blinking, another 25% slower.

Indeed! But since no one is getting hurt...

PR #4682, opened 2025-12-04
SCM: Use artIntensity from ConstellationMgr

xLPMG:
Added a connection for artIntensity to ConstellationMgr::artIntensityChanged

### Type of change

- [ ] Housekeeping
- [ ] 
## Checklist:

### DIFF
--- a/plugins/SkyCultureMaker/src/ScmConstellation.hpp
+++ b/plugins/SkyCultureMaker/src/ScmConstellation.hpp
@@ -245,10 +245,10 @@ private:
 	QFont constellationNameFont;
 
 	/// The default color used for drawing the constellation
-	Vec3f defaultConstellationLineColor = Vec3f(0.5f, 0.5f, 0.7f);
+	Vec3f defaultConstellationLineColor = Vec3f(0.0f, 0.0f, 0.0f);
 
 	/// The default color used for drawing the constellation names
-	Vec3f defaultConstellationNameColor = Vec3f(0.5f, 0.5f, 0.7f);
+	Vec3f defaultConstellationNameColor = Vec3f(0.0f, 0.0f, 0.0f);
 
 	/// Holds the artwork of this constellation.
 	ScmConstellationArtwork artwork;


plugins/SkyCultureMaker/src/ScmConstellation.cpp
```diff
@@ -32,23 +32,28 @@ scm::ScmConstellation::ScmConstellation(const QString &id, const std::vector<Con
...
+		QObject::connect(constMgr, &ConstellationMgr::fontSizeChanged,
+		                 [this](int v) { constellationNameFont.setPixelSize(v); });
+		QObject::connect(constMgr, &ConstellationMgr::linesColorChanged,
+		                 [this](const Vec3f &c) { defaultConstellationLineColor = c; });
+		QObject::connect(constMgr, &ConstellationMgr::namesColorChanged,
+		                 [this](const Vec3f &c) { defaultConstellationNameColor = c; });
+	}
+	else
+	{
+		qWarning() << "SkyCultureMaker: Failed to load constellation settings, because the ConstellationMgr is null";
```
gzotti: You can easily give a qCritical() here, or even qFatal() and abort the program. The ConstellationMgr must exist.

xLPMG: on that note: is there some kind of guarantee that it exists for the program to even reach that point? In that case we could omit the guards entirely

xLPMG: Hm it seems like other parts of the Stellarium code dont check the return value of GETSTELMODULE at all

gzotti: The core modules are loaded ahead of the plugins. (See StelApp.cpp) When your plugin's init() is called, ConstellationMgr must be up and running.

gzotti: Ah, the connections should be set in init(), not the constructor.

xLPMG: > Ah, the connections should be set in init(), not the constructor.

The plugins init()? This might be possible for dialogs which are created there, but not really for classes that can have multiple instances and are created by user interaction, like constellations, artworks, etc..

gzotti: > Hm it seems like other parts of the Stellarium code dont check the return value of GETSTELMODULE at all

Indeed. Usually not finding a core module is just fatal when the module is trusted and required to be present. In rare cases like finding an optional plugin to add a button in the GUI (SSEditor) we must check the result, though.

gzotti: > > Ah, the connections should be set in init(), not the constructor.
> 
> The plugins init()? This might be possible for dialogs which are created there, but not really for classes that can have multiple instances and are created by user interaction, like constellations, artworks, etc..

Ah, didn't look closely enough. You set each constellation with individual colors and run multiple instances at once? I had expected drawing color and line width is a property of the plugin and are set just before drawing the currently edited constellation or maybe all new and the currently edited. 

For such small objects you don't need an init(), as they can only be created after program launch when the modules are certain to be initialized.

xLPMG: > > > Ah, the connections should be set in init(), not the constructor.
> > 
> > 
> > The plugins init()? This might be possible for dialogs which are created there, but not really for classes that can have multiple instances and are created by user interaction, like constellations, artworks, etc..
> 
> Ah, didn't look closely enough. You set each constellation with individual colors and run multiple instances at once? I had expected drawing color and line width is a property of the plugin and are set just before drawing the currently edited constellation or maybe all new and the currently edited.
> 
> For such small objects you don't need an init(), as they can only be created after program launch when the modules are certain to be initialized.

I understand what you mean and that might make more sense, yes. Then we also dont have the issue that the object address must not change, if we keep most properties in the SkyCultureMaker class instead of in each constellation, artwork, etc object.

I suggest that I address this issue in a separate PR :) i dont want to bloat this simple PR right before release preparations. Its not really a critical issue anyway, just an improvement of the code architecture

PR #4683, opened 2025-12-05
Expansion of skycultures through spatial and temporal extent

### LINKED ISSUE #4636: SCM: "Test only. The result does not yet comply to Stellarium's formatting rules."
As of release 25.3, sky cultures created with the SCM did not yet comply with Stellarium's formatting rules. 

The expected behavior of the SCM would be full compliance with the guidelines and rules, so that the warning message (as stated in the title of this issue) can finally be removed. Quite a few changes have been made to the SCM since the message was added, which made me wonder when the message can be removed.

With this issue and your help, I hope to collect the concise requirements that need to be fulfilled by the SCM so that it does comply with Stellariums rules. (Maybe this is the case already, then this issue is trivial)

mRaetz:
### Description

This extension adds spatial and temporal dimensions to sky cultures. These are used to allow users to make interactive selections. The selection menu has been redesigned for this purpose. In addition to a list, users can now explore and select the available cultures using a time-resolved map. 
All new cultures in the future should have both a temporal and spatial dimension, which is why the SkyCultureMaker plugin has been updated as well. In addition to a region selection feature, a simple digitization tool has been implemented that allows users to digitize the culture territories themselves. Furthermore, a few minor changes were made that could be described as bug fixes. For example, a problem with pressing the ESC key in ScmSkyCultureDialog was fixed, as well as the previously invisible button for the artwork tooltips in ScmConstellationDialog.

- addition of spatial of temporal dimension to skyculture def
- new menu for skyculture selection that allows exploration of the available cultures
- integration of spatial / temporal extension into the SkyCultureMaker plugin
- small changes / bugfixes to the SkyCultureMaker plugin

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

This extension has been manually tested on a desktop PC (Windows 10) and a laptop (Windows 11). In both cases, there were no compatibility issues. However, no statement can be made regarding correct functionality on other operating systems.

Test specification:
- OS: Windows 10 / 11
- CPU: AMD Ryzen 7 2700x
- GPU: NVIDIA GeForce GTX 1660 Ti

## Checklist:


alex-w:
Oops... @mRaetz please resolve conflicts

plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
```diff
@@ -45,13 +55,29 @@ scm::ScmConstellation &scm::ScmSkyCulture::addConstellation(const QString &id,
...
+}
+
 void scm::ScmSkyCulture::removeConstellation(const QString &id)
 {
 	constellations.erase(remove_if(begin(constellations), end(constellations),
 	                               [id](const std::unique_ptr<ScmConstellation> &c) { return c->getId() == id; }),
 	                     end(constellations));
 }
 
+void scm::ScmSkyCulture::removeLocation(int id)
```
xLPMG, line 70: Personally I would prefer if `removeLocation` used the same pattern as `removeConstellation`. Its a modern c++ pattern that avoids iterator invalidation issues and won't break if multiple items match (although in practice that shouldnt happen with ids anyway..). It might give the compiler more freedom for optimizations as well :)

plugins/SkyCultureMaker/src/ScmSkyCulture.hpp
```diff
@@ -60,9 +67,15 @@ class ScmSkyCulture
 									   const std::vector<ConstellationLine> &lines, 
 									   const bool isDarkConstellation);
 
+	/// Adds a constellation to the sky culture
```
xLPMG:
```suggestion
	/// Adds a location to the sky culture
```

plugins/SkyCultureMaker/src/ScmSkyCulture.cpp
10110111: Right. In fact, `i++` after `location.erase(i)` has undefined behavior.

xLPMG:
I compiled this on my machine and your code generates a few new warnings that need to be fixed. Some examples include

```
ScmSkyCultureDialog.cpp:659:69: warning: unused parameter 'column' [-Wunused-parameter]
  659 | void ScmSkyCultureDialog::selectLocation(QTreeWidgetItem *item, int column)
      |                                                                     ^
ScmMultiselectionComboBox.cpp:131:12: warning: variable 'currentModel' set but not used [-Wunused-but-set-variable]
  131 |         if (auto *currentModel = model)
      |                   ^
SkycultureMapGraphicsView.cpp:452:8: warning: unused variable 'maxDuration' [-Wunused-variable]
  452 |         qreal maxDuration = 2000;
      |               ^~~~~~~~~~~
SkyculturePolygonItem.cpp:30:4: warning: field 'isHovered' will be initialized after field 'lastSelectedState' [-Wreorder-ctor]
   30 |         , isHovered(false)
      |           ^~~~~~~~~~~~~~~~
      |           lastSelectedState(false)
   31 |         , lastSelectedState(false)
      |           ~~~~~~~~~~~~~~~~~~~~~~~~
      |           isHovered(false)
```

If these dont appear in your console, maybe try deleting the build folder and doing a fresh build.

Also now that I see this warning: the usual name scheme was to use camel case for "SkyCulture", but your files use "Skyculture". Its only a cosmetic issue, but it would still be nice if the same naming scheme is used everywhere :))

mRaetz:
Thanks for the tip and for pointing out the incorrect names. I didn't even notice the mistake with the lowercase c.

Another thing i noticed when looking over the files is that one of the buttons in ScmHideOrAbortMakerDialog is incorrectly translated in german. The buttons are supposed to be distinguishable (hide - abort - cancel) but in the current version they are not (Ausblenden - Abbrechen - Abbrechen)
 
{9A74251C-C552-4516-9271-54471F9B0311}

I'm also not sure how to solve the problem with the SvgWidgets module. Apparently, there was a change in Qt in this regard, but even if there is a difference between Qt5 and Qt6, I don't understand why the checks for Linux (amd64; qt6; core) / Linux (amd64; qt6) failed when they passed for Windows and macOS with qt6.

alex-w:
> I'm also not sure how to solve the problem with the SvgWidgets module. Apparently, there was a change in Qt in this regard, but even if there is a difference between Qt5 and Qt6, I don't understand why the checks for Linux (amd64; qt6; core) / Linux (amd64; qt6) failed when they passed for Windows and macOS with qt6.

Please add missing package in linux boxes

10110111:
> The buttons are supposed to be distinguishable (abort - hide - cancel) but in the current version they are not (Ausblenden - Abbrechen - Abbrechen)

Actually, even in English "Abort" and "Cancel" require too much consideration to figure out what one should press. What if we change them to more explicit "Drop progress and close" and "Don't close"?

The current situation reminds me of the MSVC's age-old "Debug error!" dialog box with the stock "Abort", "Retry", "Ignore" buttons and the "(Press Retry to debug the application)" kludge instead of having sensible button labels.

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
```diff
@@ -0,0 +1,386 @@
...
+#include <qguiapplication.h>
+#include <qjsonarray.h>
+#include <QGraphicsSvgItem>
+#include <qscrollbar.h>
+#include <QFileDialog>
+
+#include <QJsonObject>
+#include <QJsonDocument>
+#include <QtMath>
+#include <cmath> // for M_PI
```
10110111, line 30: `<QtMath>` is documented to define `M_PI` too. But why on Earth use `QtMath` at all? For `qPow`? How is it better than `std::pow`?

Besides the apparent uselessness, Qt is constantly making some crazy assumptions, like `qFloor` being templated (so that it could support large numbers) but returning a mere `int` (throwing away most of supportable integers), downcast from a call to `std::floor` (basically)!

So I'd advise to avoid `QtMath`, except maybe for the unique functions like the variadic `qHypot` overload or `qNextPowerOfTwo`, etc.

10110111:
I had to apply the following patch to build without precompiled headers on Ubuntu 20.04 and Qt 6.4.1:
```diff
diff --git a/src/gui/SkyCultureMapGraphicsView.cpp b/src/gui/SkyCultureMapGraphicsView.cpp
index c5fcee207a..edaa5faaec 100644
--- a/src/gui/SkyCultureMapGraphicsView.cpp
+++ b/src/gui/SkyCultureMapGraphicsView.cpp
@@ -20,6 +20,7 @@
 #include "SkyCultureMapGraphicsView.hpp"
 #include "SkyCulturePolygonItem.hpp"
 #include "StelLocaleMgr.hpp"
+#include "StelFileMgr.hpp"
 #include "StelSkyCultureMgr.hpp"
 #include <qjsonarray.h>
 #include <QGraphicsSvgItem>
@@ -27,7 +28,10 @@
 
 #include <QJsonObject>
 #include <QJsonDocument>
+#include <QWheelEvent>
+#include <QMouseEvent>
 #include <QtMath>
+#include <QFile>
 
 SkyCultureMapGraphicsView::SkyCultureMapGraphicsView(QWidget *parent)
     : QGraphicsView(parent)
```

10110111:
So, I've finally tried this PR. Here's my critique:

0. See the above comment.
1. The separate "Sky Culture Description" looks quite out of place as a category among the other tabs. Would it be more appropriate to make it a subtab inside the "Sky culture" tab?
2. The labels like "current Time", "min Year" etc. should be capitalized in a way to match the rest of the GUI, i.e. "Current time", "Minimum year" etc. (note "minimum" instead of "min", because there's no reason to shorten it, since we have plenty of space in this row).
3. Why does the "current Time" spinbox look so different from all the other spinboxes? The -/+ buttons seem to do exactly the same as the usual down/up buttons.
4. The "Enter Culture..." filter behaves strangely: I type `b`, it shows all SCs that contain a letter "b" in their names, but if I type `be`, almost all SCs reappear (but some are still filtered out).
5. I can move the world map by half its size up and down, showing useless space above or below, which makes no sense to me.
6. Sky Culture Description contains most of the settings, which is not obvious from tab name. This implies that the name should rather be something like "Sky Culture Details".

mRaetz:
> The separate "Sky Culture Description" looks quite out of place as a category among the other tabs. Would it be more appropriate to make it a subtab inside the "Sky culture" tab?

Originally, the map and description were in the same tab (a stackedWidget allowed users to switch between them). On the suggestion of GZotti / SMH, the description was later moved to this extra tab.

> Sky Culture Description contains most of the settings, which is not obvious from tab name. This implies that the name should rather be something like "Sky Culture Details".

Yes, the name could definitely be better. If the description stays in its own tab, I would also prefer this name.

> Why does the "current Time" spinbox look so different from all the other spinboxes? The -/+ buttons seem to do exactly the same as the usual down/up buttons.

This was purely a design decision. I felt that buttons on the left and right would better convey the connection with the horizontal slider. The normal up/down buttons rotated and placed on the left/right looked a bit strange, which is why I used -/+ buttons. If this disturbs the overall design too much, the normal up/down buttons on the right side of the spin box can be used instead.

> The "Enter Culture..." filter behaves strangely: I type b, it shows all SCs that contain a letter "b" in their names, but if I type be, almost all SCs reappear (but some are still filtered out).

Yes, that's right. I initially used a strict query (toLower(filterString) must be contained in the word). While testing, I noticed that some cultures are difficult to find with this query. Typos are an obvious problem, but special characters can also prevent the filter from matching the word. A concrete example of this is as follows: In the German version, the indigenous culture of New Zealand is referred to as Māori. However, the ā is a very rare character, so it must be assumed that the user will enter Maori (normal a) in the search bar. Therefore, I tried to use a kind of string similarity measure to solve this problem. This is not perfect, but most importantly, it does not filter individual characters (a deviation of at least one character should be allowed to avoid the problems described above). Therefore, it is only used for 2 or more characters, which leads to the behavior described. I am sure there is a better solution for the whole thing, but the current implementation has basically done its job. However, I would also be open to using a simple, strict matching (via contains).

> I can move the world map by half its size up and down, showing useless space above or below, which makes no sense to me.

If the scene is the same size as the map, navigation of the map becomes restricted. On the one hand, you could then only view the edge of the map (depending on the zoom level) at the edge of the window, but most of the time you want the relevant part to be in the center of your field of view. On the other hand, there is a problem with smooth zooming after selecting a culture from the list. Since during this zoom only the scene is moved to create the illusion of smooth movement, it occurs, especially in the border areas of the map, that you reach the boundary of the scene, resulting in a rigid movement. This can be solved by setting the extent of the scene to a multiple of the map.

10110111:
> This was purely a design decision. I felt that buttons on the left and right would better convey the connection with the horizontal slider.

I don't think it's a good idea, because it makes widget style inconsistent and makes it seem that the two spinboxes near each other are actually different kinds of widgets. We already have similar combinations of a slider and a (normal) spinbox, and they don't cause any confusion. See e.g. _Labels and Markers_ in the _DSO_ tab of the _View_ dialog, or _Shooting stars_ in the _Sky_ tab.

> Typos are an obvious problem, but special characters can also prevent the filter from matching the word.

You could strip diacritics from both the search term and the list of names before doing the filtering. See e.g. [here](https://stackoverflow.com/a/14009539) on how to do it.

> most of the time you want the relevant part to be in the center of your field of view

Even if we go this way, it shouldn't be possible to move the edge of the map farther away than to the center of the window. Currently I can move the whole map away when scaling up so as to fit Greenland in the window, and may have to pan several window heights to get the map back into the viewport.

mRaetz:
I apologize for the long break on my part. A few things came up, but now most things should be fine.

The latest changes include:
- compatibility fixes for Qt5 on Linux / Mac
- style changes to GUI (Spinbox, upper case labels, ...)
- moved skyculture details (description) and map into the same tab in viewDialog (accessible via subtab)
- revision of the search bar algorithm (now simply stripping diacritics before matching the strings)
- limited the movement of the map by the user (protection against panning into void)
- clean up polygons before exporting in SCM (to prevent tedious inspection when reviewing user contribution, all overlapping polygons are merged into one automatically pre export)
- updated basemap (now includes some major lakes and revised river lines)

plugins/SkyCultureMaker/src/types/Region.hpp
```diff
@@ -0,0 +1,102 @@
...
+	CENTRAL_AMERICA,
+	NORTHERN_AMERICA,
+	SOUTH_AMERICA,
+	CENTRAL_ASIA,
+	EASTERN_ASIA,
+	SOUTH_EASTERN_ASIA,
+	SOUTHERN_ASIA,
+	WESTERN_ASIA,
+	AUSTRALASIA,
+	MELANASIA,
```
gzotti: Melanesia?

alex-w: Yes, it should be Melanesia

skycultures/lokono/index.json
```diff
@@ -1,6 +1,6 @@
 {
   "id": "lokono",
-  "region": "Southern America",
+  "region": "South America",
```
gzotti, line 3: What is the reason to deviate from the list documented and published before in the User Guide, section 9.2.1?

src/gui/ViewDialog.hpp
```diff
@@ -139,6 +146,14 @@ private slots:
 	QHash<QString/*planet English name*/, PlanetSurveyPack> planetarySurveys;
 	QHash<QString/*survey URL*/, QTreeWidgetItem*> surveysInTheList;
 	QString selectedSurveyType;
+
+	// list of special characters and their 'normal' counterpart for filtering (skycultures / the culturesListWidget)ĀA
+	// (when a new special character is added to 'specialCharString', 'normalCharList' must also be expanded)
+	QString specialCharString = "ŠŒŽšœžŸ¥µÀÁÂÃĀÄÅÆÇÈÉÊËÌÍÎÏÐÑÒÓÔÕÖØÙÚÛÜÝßàáâãāäåæçèéêëìíîïðñòóôõöøùúûüýÿ";
```
gzotti: There is at least a "dotless i" to be included. But see Vietnamese for a few more suggestions :-)

skycultures/lokono/index.json
mRaetz: Perhaps it is not necessary, but since the regions are supposed to follow the United Nations Geoscheme UN M49, I have renamed the region to remain faithful to the current version, in which there is no “Southern America” (but "South America" instead).
(https://unstats.un.org/unsd/methodology/m49/)
{2BBA80F8-684B-4CA0-849A-57C7E4B339BB}

gzotti: OK I see it. Not sure who introduced the list in its current form, but now we have used it for a year already and IMHO it should not be changed unless really wrong (like "Suothern Amreica"). Same goes for "World" vs "Global". There are several projects already utilizing tens of skycultures in the form documented in the User Guide. Other opinions, @10110111 , @alex-w, (probably) last chance to change, or keep using the current 2025 spelling?

10110111: I think backwards compatibility is better than pursuit of all the changes in wording upstream (although "Southern America" does sound odd).

src/gui/ViewDialog.hpp
gzotti: Maybe this helps: https://share.google/aimode/o2Yk9DvqGZYorAEnj

mRaetz: I'm not entirely sure anymore because I didn't write down the results back then, but I experimented with the normalization methods before the current implementation, and none of them delivered a completely satisfactory result. (I believe it was mostly about the ligatures)
In any case, it will probably be difficult to cover all special characters in this way. I'm not quite sure how to handle this yet, whether all existing special characters should appear in the list as a preventive measure, or whether it should be expanded in the future if problems arise. (as in minor problems because it is a QoL feature that won't cause any issues apart from inconveniences)

gzotti:
In the SCM GUI, I see a few weird things.

- you can select a classification. It seems you did not write this code, but IMHO we should make sure that NONE is not an option here. A newly created SC should be personal for a start, and authors can select something else where applicable. 
- You can apparently select more than one region. The region entry in index.json only has one string. Huh?

The Constellation Editor/Information is missing a line for the (optional, scientific) Transliteration (see User Guide 9.2.4). 

The polygon drawing works well but still could become more intuitive. Of course it depends on which other software you are using to use the same keys.  I had expected LMB draws, and RMB finalized one polygon, and needed the help button.  You could remove the Alt combination, and use ESC as panic button to remove last-pressed key. Remove full polygon is available in the dedicated button, so may also not be necessary on a key. Or compare to the existing constellation drawing in the same plugin: RMB to draw, double click to finish. This could be harmonized.

Who is going to write the detailed chapter for the User Guide? The logic behind the temporal polygons needs explanation.

EDIT: I created a sample SC. After storing, it seems Skycultures are reloaded (short blinking of constellation lines).  Unfortunately, the program then exited silently. Also after relaunch, calling up the view menu to load the new SC failed with silent exit. Does it work for you?

skycultures/lokono/index.json
alex-w: It may be my spelling error (South America) due I filled the list of regions - I'll fix it in the master.

mRaetz:
1) Classification Combobox:
> you can select a classification. It seems you did not write this code, but IMHO we should make sure that NONE is not an option here. A newly created SC should be personal for a start, and authors can select something else where applicable.

I am not responsible for the logic behind the classification combo box, but as far as I can see, it works as intended. The default setting is “none,” and the user can select an appropriate classification. If the user doesn’t do this, a pop-up window will appear when exporting Skyculture, informing them that a classification != none must be selected.

2) Multi-Region Skycultures:
> You can apparently select more than one region. The region entry in index.json only has one string. Huh?

In my discussions with @sushoff , it became clear that it would be desirable for a skyculture to be able to have multiple regions (due to the use of the UN geoscheme, it is possible that the occupied area of a culture could extend across multiple regions). I have modified the code in several places so that a JSONArray is used instead of a JSONString. For legacy reasons, the old format will continue to be supported, but sooner or later (provided that the SCM is always used to create new cultures), all cultures should have the new region format.

3) Constellation Editor missing field
> The Constellation Editor/Information is missing a line for the (optional, scientific) Transliteration (see User Guide 9.2.4).

This appears to be an error on the part of the original SCM team. I have no idea whether this field was simply overlooked or whether there was a reason not to implement it.

4) Controls of polygon digitizing

I tried to base the controls on those in QGIS, simply because they are already established there. Originally, a new point could only be set with LMB, but based on feedback that the Constellation Editor uses RMB, I also allowed RMB to be used to set a new point. Personally, I prefer the option where you add a point with LMB and finalize the polygon with RMB. (I'm not a fan of the controls in the Constellation Editor, but my personal preferences shouldn't matter here. If there is a desire for uniform controls, I will be happy to implement that.)
It's true that a polygon can be deleted once it has been digitized, but I don't really see why it would be necessary to remove the option to reset the digitization process (in the end its just an optional QoL feature). Again, this is just my personal opinion, because that's how I've come to use it in other programs (such as QGIS or ArcGIS), but I think using the backspace key to delete the last point is more intuitive than using ESC. 

5) User Guide
> Who is going to write the detailed chapter for the User Guide? The logic behind the temporal polygons needs explanation.

I suppose I have to at least help the person writing the chapter in the form of a detailed explanation (if not more).

6) 
> Unfortunately, the program then exited silently. Also after relaunch, calling up the view menu to load the new SC failed with silent exit.

I have never encountered the bug described above during testing and have not been able to replicate it. Apart from that, to my knowledge, only the list of sky cultures in SkyCultureMgr is updated, but not the GUI elements in ViewDialog (populateLists is not called again and there is no function (yet) that would update the map including the polygons). When opening Stellarium again, all new cultures are loaded and displayed as expected (in the list and on the map).
To be honest, I always closed and reopened Stellarium during all tests. It didn't even occur to me that it might be desirable to be able to use the exported culture directly in ViewDialog.

gzotti:
> 1. Classification Combobox:
> 
> [...] If the user doesn’t [select a valid option], a pop-up window will appear when exporting Skyculture, informing them that a classification != none must be selected.

OK, seen. 
 
> 2. Multi-Region Skycultures:
> 
> > You can apparently select more than one region. The region entry in index.json only has one string. Huh?
> 
> In my discussions with @sushoff , [...], all cultures should have the new region format.

What about the other projects utilizing the new SC format? @10110111 is S-M able to cope with this? The only format relevant as of 25.4 is the one from the User Guide.

 
> 3. Constellation Editor missing field
> 
> > The Constellation Editor/Information is missing a line for the (optional, scientific) Transliteration (see User Guide 9.2.4).
> 
> This appears to be an error on the part of the original SCM team. I have no idea whether this field was simply overlooked or whether there was a reason not to implement it.

Must have been overlooked, and should be added.

> 4. Controls of polygon digitizing
> 
> 
> I tried to base the controls on those in QGIS, simply because they are already established there. Originally, a new point could only be set with LMB, but based on feedback that the Constellation Editor uses RMB, I also allowed RMB to be used to set a new point. Personally, I prefer the option where you add a point with LMB and finalize the polygon with RMB. (I'm not a fan of the controls in the Constellation Editor, but my personal preferences shouldn't matter here. If there is a desire for uniform controls, I will be happy to implement that.) [...]

For me it appears desirable to have similar controls for similar actions in a single plugin. @alex-w ? @10110111 ?

> It's true that a polygon can be deleted once it has been digitized, but I don't really see why it would be necessary to remove the option to reset the digitization process (in the end its just an optional QoL feature). Again, this is just my personal opinion, because that's how I've come to use it in other programs (such as QGIS or ArcGIS), but I think using the backspace key to delete the last point is more intuitive than using ESC.

Sure, ESC is uncommon. Backspace is OK for me.

 
> 5. User Guide
> 
> > Who is going to write the detailed chapter for the User Guide? The logic behind the temporal polygons needs explanation.
> 
> I suppose I have to at least help the person writing the chapter in the form of a detailed explanation (if not more).

I can do some formatting, but the operational instructions should come from someone who has designed this or who wants other people to use it, else nobody will know what to do with it. @sushoff ? 
 
> > 6. Unfortunately, the program then exited silently. Also after relaunch, calling up the view menu to load the new SC failed with silent exit.
> 
> I have never encountered the bug [...]

For now my creation seems it work at least after restart. But before merge the plugin should survive storing SC and reloading all SCs, then loading and showing the newly created one.

10110111:
> What about the other projects utilizing the new SC format? @10110111 is S-M able to cope with this? The only format relevant as of 25.4 is the one from the User Guide.

S-M only supports one region, and supporting multiple will require some tricks. Particularly, the region is used for splitting the list of SCs into sections (=regions), and I don't know how it could use multiple regions. OTOH, the SCs there use different regions from those in desktop Stellarium anyway (a different split of the world that e.g. makes Arabic SCs part of Middle East region rather than Western Asia).

mRaetz:
Does that mean there are potential compatibility issues if a Skyculture has multiple regions? Or is that fine because other regions are used?

10110111:
> Does that mean there are potential compatibility issues if a Skyculture has multiple regions? Or is that fine because other regions are used?

Honestly, I don't know. Are there any real SCs (possibly not yet included in Stellarium) that want to have multiple regions? If not, I'd leave this at single region per SC for now.

gzotti:
From discussion above, apparently @sushoff asked @mRaetz for it. But we were not asked for it, and also not for the change in wording ("World", "South America") that deviates from written instructions in the User Guide. If our aim was designing an SC format for all flavours of Stellarium, all parties have to agree on changes! Some sides may decide to ignore an entry when it cannot be used or comes in an unexpected format (array, not string), but even ignoring a wrong format (or fallback behaviour, like using the first-found string in the array) must be implemented.

mRaetz:
I would like to clarify that although I was commissioned by Ms. Hoffmann to make these changes and worked closely with her, most of the design decisions and implementations were ultimately made by me. 
I implemented the more problematic format changes discussed above without knowing that there were any dependencies between the different Stellarium versions. (For example, I cannot remember the reason for changing “world” to “global.” This probably happened at the very beginning of the development process for no particular reason.)
Since I realized that changing the format carelessly could lead to complications, I am absolutely willing to reverse all problematic changes.

mRaetz:
> Are there any real SCs (possibly not yet included in Stellarium) that want to have multiple regions?

To be honest, this is beyond my knowledge. There are certainly cultures to which this would apply, but I am unable to state whether they are available in Stellarium or will be added in the near future.

Is it an acceptable solution to leave the option for multiple regions implemented, but to continue working with only one region (and therefore strings) for now? Then the infrastructure would already be in place if it became necessary to cover several regions.

For example: 
- When reading the skyculture in `StelSkyCultureMgr` an array is checked for; if it is not found, the string is packed into an array. 
- The region selection in SCM only allows only one region at a time. If only one region is selected, it is saved as a JsonString; if multiple regions are selected, it is saved as a JsonArray

However, I can also understand if this is considered unnecessary overhead, because only one region is to be used anyway.

gzotti:
If @sushoff has a requirement for an array and can explain, she should discuss it with esp. @xalioth and @10110111, and @alex-w and I should then be informed with a conclusion. Somebody must change the specs, and somebody must implement the changes. If we don't need it now but maybe in 2 or 5 years, you can leave it in the code, either as clause that actually is never called when using the current files, or just commented out, but in any case documented which says something like "Prepared for a possible extension of region into an array." The developers of dependent (data-consuming) programs can then deal with it. A simple "don't crash" solution while just reading a new-format file is to take the first string in the array, but else keep it as one region. This only changes the file parser. A full implementation will have to deal with actually having more than one region in other parts of the program. Similarly, changing a name entry can be mitigated in the reader (tolerating misspelled entries or "old values" is something that has a tradition here...)

skycultures/lokono/index.json
sushoff: sorry for my long silence, I'm trying to catch up... I don't think, we can't change an error that had been used for a year (a year is nothing on the timescales of the history of astronomy)

sushoff:
polygon edits/ harmonize LMB/RMB etc. - yes, absolutely!!! this was also the feedback from the test users that were asked for the BA-thesis. It's only that their feedback came too late before the submission date of the thesis, so that it couldnt be implemented. Yet that should be done. 

ESC for backward: oh, yes, please avoid that - Ctrl+Z is common

Multi-Region SCs: absolutely, this is a necessity, as cultures may have moved over time. The logic is clear, in my view, from historical facts. you're right @gzotti, that index.json allows only one - didn't think of that, the issue only emerged from facts. Q: considering index.json as "current" state. is that really a problem/ conflict, if the geo.poly submenu uses a different adoption? in other words: can we separate the index-definition with one region and the geo.poly.menu with more than one?

User Guide: @gzotti, I know, I had promissed you to provide some text - I keep my promisses, although, I haven't done so yet, it is on my list. (you know, I become silent - work-focused - with growing workload)

I have not asked for change of wording - that was just a development that occured from harmonisation. I am rather emotionless concernign "world" vs "global" but for the other terms, @mRaetz found some inconsistencies in the existing (free-style written) description.md files. If - say - some of the authors wrote "South Am." and others wrote "Southern Am.", I had suggested to follow the UN-terms. If I remember correctly, there was at least one region, where we/ he had then discovered that the majority had used the "wrong" term and it would have been less effort to/ fewer updates to harmonize all, but - well - why not doing it correctly, when we apply changes anyway.

mRaetz:
The latest changes should resolve most of the issues discussed above.

- Until further details are clarified, skyculture files created by SCM will be exported with one region only (and stored as a string accordingly).
- The list and map in ViewDialog are updated after a skyculture is exported in SCM.
- The missing “transliteration” field in SCM has been added.
- The controls have been modified to match those used in the constellation editor. 
(I would still like to point out that using these controls to digitize polygons is generally questionable. While the constellations snap to specific fixed points and it is also possible not to target anything, this option does not exist for polygons. Every right-click sets a new point. It is much easier to accidentally place a point when using RMB and double-RMB to digitize polygons than when drawing constellations.)

sushoff:
@gzotti @10110111 @alex-w is it all resolved? I am checking the new SCs that will go into the new release; what about this PR?

sushoff:
see #4636

mRaetz:
In response to @gzotti in #4636:

The missing territory files are actually optional; thanks for pointing that out. I’ve changed it to qWarning. 
Since the program crashed again for you: During development, I only used the skycultures already available in Stellarium for testing. There don't seem to be any issues with those. Could you send me your collection of SCs (or at least a part if it) so I can investigate the source of the error? (Assuming this collection of yours differs from the skycultures on the main branch.)

gzotti:
It seems I found the issue caused by something already mentioned. One of my own SCs is in region  "Southern America" which is the region label according to the User Guide. Changing it to "South America" loads it. It would have been wise to implement this in "tolerance mode" (accept both), not just break everything. 

@10110111 @alex-w do we change the region name to "South America"? Or allow both forms? What about St-Web/St-M?

10110111:
If we go for switch, I think we should accept both, but emit a warning when
the obsolete one is found. And of course, the SCM should emit the new one.
The python validator could check this too.
S-M is irrelavant since it uses its own regions anyway. S-W doesn't seem to
support switching SCs at all.

gzotti:
> If we go for switch, I think we should accept both, but emit a warning when the obsolete one is found. And of course, the SCM should emit the new one. The python validator could check this too. S-M is irrelavant since it uses its own regions anyway. S-W doesn't seem to support switching SCs at all.

Yes, then please @mRaetz add something to accept "Southern America" identical to "South America", just emitting a warning that this key should be updated to "South America". Some authors have already published their stuff which works in version 25.1 and later, and should not break now. And just change the description key in the User Guide, 9.2.1.

gzotti:
The other question was, who will add all the polygons for the existing SCs so that the map selection becomes the awesome feature it now only promises to become? 
While we are changing, when viewing a territory.json I see the keys are stored alphabetically, as usual. However,  when wanting to edit this file in a text editor, this gives an eye-discomforting sectioning per polygon: endTime, geometry, startTime. If you could rename startTime to beginTime, it would be placed into logical order. 

Another issue is when adding a polygon in March 2026.  endTime=2026. What is the interpretation if I look into this in 2030? Is it useful to add special values like "origins" (as hypothetical unknown before date) and  "present", or is it acceptable to just leave endtime away to interpret as "now"?

On the technical side: There is the GeoJSON fileformat. Wouldn't it be useful to follow this (expandable) simple JSON-based standard, that might make editing with other tools easier? (Imagine editing a detailed shapefile over a detailed map in QGIS and exporting in GeoJSON. We could have a "master shapefile" with per-shape fields "skyculture", "beginDate", "endDate" in which we collect all world entries.) Each current polygon would be a feature in this FeatureCollection, with a Geometry of type polygon, with properties beginTime and EndTime. This (utilizing GeoJSON) would be a change I'd like to see for 26.2 (while maps are not yet really useful in 26.1) when we can accept the rest now. The QGIS map is something for the future (2026-2027), as it should come with its own import/export QGIS plugin.

mRaetz:
Regarding the change from startTime to beginTime: Should this change apply only to the JSON file or throughout the entire code?

>Another issue is when adding a polygon in March 2026. endTime=2026. What is the interpretation [...]

I'm not sure if I understand this correctly, but is the question about how existing skycultures are handled?
If so: Currently, it is possible to not assign a fixed endTime to skycultures or polygons but a special character (this happens automatically in SCM if a year greater than the current year is entered). If such a polygon is found during reading, the current year is assigned to it directly in the skyCultureExplorer (based on the example given: in the year 2030, the year 2030 is displayed as the endTime). 

Actually good point, there’s no reason not to use GeoJSON. At first, I worked with regular JSON because it wasn’t clear whether the polygons should be saved in a separate file or included in the index.json file. But am I understanding this correctly: instead of multiple territory.json files in the respective SkyCultures folders, there would be a single large GeoJSON file that is edited or expanded as needed?

gzotti:
> Regarding the change from startTime to beginTime: Should this change apply only to the JSON file or throughout the entire code?

I meant, if curious users ever want to study the JSON files, the files are IMO easier to read for humans if the entries start with beginTime, then followed by endTime, then the polygon data. When switching to GeoJSON, the same names would apply to the properties section there. Renaming inside the code would also be practical (and easy in QtCreator, just rightclick-rename the variable and it does so everywhere.


> > Another issue is when adding a polygon in March 2026. endTime=2026. What is the interpretation [...]
> 
> I'm not sure if I understand this correctly, but is the question about how existing skycultures are handled? If so: Currently, it is possible to not assign a fixed endTime to skycultures or polygons but a special character (this happens automatically in SCM if a year greater than the current year is entered). 

Maybe make it "... if a year equal or greater than the current year...". Hardly anybody would think of it. And then there is the question whether this should be a string (to accept a special character) or an int (to allow math on it). Also for the (Geo)JSON property.

> If such a polygon is found during reading, the current year is assigned to it directly in the skyCultureExplorer (based on the example given: in the year 2030, the year 2030 is displayed as the endTime).

OK, so just "greater or equal" should work. 

> 
> Actually good point, there’s no reason not to use GeoJSON. At first, I worked with regular JSON because it wasn’t clear whether the polygons should be saved in a separate file or included in the index.json file. But am I understanding this correctly: instead of multiple territory.json files in the respective SkyCultures folders, there would be a single large GeoJSON file that is edited or expanded as needed?

No, I'd keep it as one GeoJSON per SC. Just for advanced editing/maintenance, we would need a new project that includes a master shapefile ("database") and a QGIS plugin that can import "basic" GeoJSONs from SCM into the shapefile, can improve the polygons by editing over a hi-res map (or even re-work and reproject users' (contributors') existing mappings from other EPSG projection systems, ...), and re-export the polygons back to the skycultures folder. I just feel that the editing functionality of QGIS and the chance in it to use higher-res maps to plot or construct polygons upon will be a better path than to include half a GIS into Stellarium. There may be other applications or links between Stellarium and QGIS of course. This still leaves the topic open who will define the shapefile, who will write the plugin with import and export functions, and who defines the polygons for 50+ skycultures. But this won't happen this week anyhow...

alex-w:
@gzotti @sushoff is this pull request OK for merge?

gzotti:
I did not yet check the GeoJSON. Hopefully tonight...

mRaetz:
_Just to avoid any confusion:_ 
the latest commit only contains (commented-out) code for writing GeoJSON. I wasn't sure if I had understood the conversation correctly and whether we actually want to switch directly to GeoJSON. If that's what you want, I'm happy to update the reading part as well (I just didn't want to make the code too messy for now). 

On this note (something I noticed yesterday): some cultures on the main branch (at least one, namely norse_edda) already have a territory file. However, this was created using one of the SCM prototypes I provided to @sushoff  during the development process. Since some things have changed since then, this skyculture isn’t being read correctly by the skyculture explorer. (I guess if we switch to GeoJSON, it would have to be redone anyway.)

sushoff:
@mRaetz ok, thanks for the note - so I'd need someone to create polygons ... :-) Good that I didn't have time for it so far ;-) thanks for notifying me!

gzotti:
I have a preference to use existing applicable standard formats and switch before the old format is being circulated. If it's only norse_edda, this can be converted manually.  The advantage is using other tools which already exist and which may have special capabilities that facilitate editing. Although I have not used it by now, QGIS has GeoJSON capabilities, at least via dedicated plugins. A very easy-to-use interface for drawing is geojson.io. In the right part of the Window the created GeoJSON is displayed, and you would have (for now) to add the beginDate/endDate properties manually. Maybe the territory file should be named `territory.geojson` to make it easier to understand.

gzotti:
I have just pushed an editor-converted territory.geojson of norse_edda to master. It loads in QGIS, so should be OK. The geojson file and map use still has to be described in the user guide.  Here we can write that only polygon features are supported, and other feature types ignored. The older format should then go away without sorrow. I can imagine authors will want to start editing the polygons in SCM but may want to refine later. And adding the missing polygons later this year should be much easier with this. (with or without bespoke QGIS import/export plugin)

mRaetz:
> I have a preference to use existing applicable standard formats and switch before the old format is being circulated

That sounds to me like we're going to use GeoJSON directly? If so, I'll push the changes so that the territory.geojson files are read correctly. (?)

> I have just pushed an editor-converted territory.geojson of norse_edda to master.

It's looking pretty good, but im afraid those weren't the only issues. The endTime is still an int (while the current implementation uses a string), and the entries for the global beginTime and endTime are missing from the index.json file. 

Another thing (that isn't really due to the format, but it's probably something that could have been avoided): 
There's an artifact between the polygons that likely has no business being there, but is instead a result of the digitization process.
{C0A00499-F694-4BBE-A7AF-6D85AC7A2EF6}

gzotti:
> > I have a preference to use existing applicable standard formats and switch before the old format is being circulated
> 
> That sounds to me like we're going to use GeoJSON directly? If so, I'll push the changes so that the territory.geojson files are read correctly. (?)

If nobody objects by tomorrow (@alex-w , @10110111 , @sushoff ?), I see no point in keeping the self-made format. 

> > I have just pushed an editor-converted territory.geojson of norse_edda to master.
> 
> It's looking pretty good, but im afraid those weren't the only issues. The endTime is still an int (while the current implementation uses a string), 

Ah, you can change the property type then. As I understand the properties are a dictionary where the data types are defined as required. However, you could also just use 10.000 as special marker value. 

> and the entries for the global beginTime and endTime are missing from the index.json file.

Where should it go? It is not described in the specifications (User Guide), so I know nothing about it. 
 
> Another thing (that isn't really due to the format, but it's probably something that could have been avoided): There's an artifact between the polygons that likely has no business being there, but is instead a result of the digitization process. 

Right, but it was already there. Maybe we can repair this into a 4th polygon feature later. I was not aware of this file until 2 hours ago...

alex-w:
> > That sounds to me like we're going to use GeoJSON directly? If so, I'll push the changes so that the territory.geojson files are read correctly. (?)
> 
> If nobody objects by tomorrow (@alex-w , @10110111 , @sushoff ?), I see no point in keeping the self-made format.

+1 for using standard formats

mRaetz:
- The GeoJSON files should now be read correctly. 
- I've “fixed” the norse_edda culture (added beginTime / endTime in index.json and removed the artifact we talked about yesterday).  
- I also changed the data type of endTime from string back to integer. The new special value in the territory.geojson file is now 9146.

gzotti:
9146... do you mind sharing any deeper explanation for this? (It will be a documented code, so whatever it is. Just that users will be curious.)

mRaetz:
previously it was "∞", the symbol for infinity. alpha to num: I = 9, N = 14, F = 6 (INF or 9146)

gzotti:
Argh. I tested the PR and just wanted to approve this as it now really works (at least I could not find new problems. I have not tested image aligning since February though. This should not have changed presumably). I am now reading of a rebase problem, but a squash/merge should work here as well. It's one big feature change. I will add user changes to the SUG on weekend, hoping I don't mis-describe the new features.

10110111:
Oops, this breaks the Surveys tab:

Screenshot_2026-03-27_19-59-44

alex-w:
Damn... The tab _Sky Culture_ has troubles too

gzotti:
Seems some widget definition was duplicated in the survey tab. Going into it now...

EDIT: Fixed. One lengthy widget duplicate removed.

10110111:
Why do we have `minimumSize=1500` in `culturesListWidget`? This makes the window unnecessarily huge.
Also, I get this warning multiple times when using Qt Designer on `viewDialog.ui`:
```
Could not create pixmap from :/graphicGui/uieSearchBoxBackground.png
```

10110111:
What is the map supposed to be used for? I see cities marked on it, but no symbols for the location of at least the current SC.

mRaetz:
> Why do we have minimumSize=1500 in culturesListWidget?

I'm not entirely sure anymore, but I think the list was so “thin” that some names were not easily readable or couldn't be distinguished from others. But maybe that's unnecessary and should be removed. I've never had a problem with the size on my screen, but I can imagine it might be an issue on smaller screens.
As for the error in QtDesigner, I can't say anything about it because I don't get that error message. (I'm using version 6.5.8).

> What is the map supposed to be used for? I see cities marked on it, but no symbols for the location of at least the current SC.

Are we talking about how the territory of skycultures is represented? This is done using polygons that currently do not exist. For skycultures like “modern,” these will likely never exist, since there isn't really a definable territory for them.

10110111:
> This is done using polygons that currently do not exist.

Ah, that explains it.

> As for the error in QtDesigner, I can't say anything about it because I don't get that error message.

Yeah, somehow Stellarium itself also finds the file successfully.

gzotti:
Map: Each SC can now have a GeoJSON file territory.geojson that shows its location as click-selectable polygon. The time slider below would show the localisation of the SCs as coded in beginTime/endTime properties. The polygon can be created when defining a SC with the SCM plugin, and probably edited/refined with general GIS tools like QGIS.  So far, only Norse-Edda has it, but we should add those files to all non-"World" SCs in the upcoming months to populate the map (#4842). I have disabled a warning for missing polygon for now.

axd1967:
@mRaetz Very nice but I never understood why project managers accept a change without corresponding user manual updates. 

There is even a checkbox... it should be checked...

It is part of the development effort that should not be left to others to fix. Whi else will do it? Who has the best knowledge of what changed? At least a few sentences should be added, even if it's only a "TODO", "update screenshot here", add some basic explanation...

I know by experience that developers tend to underestimate the value of the UG... They like to code, and leave user manual updates to others.

gzotti:
@axd1967 project managers have decided to accept this as-is for 26.1 and add notes into the SUG only when at least a few map polygons have been added so that it's worth discussing them. (Unfortunately I have not seen _someone_'s PR with ~35 GeoJSONs by now - maybe _anyone_ could do it instead...) But generally I agree, new features are best described by the respective developer which is why the SUG is a "living" part of the git and not a separate project as it was 15-10 years ago.

sushoff:
@axd1967 you know, unpaid project managers just want to punish the world ... 

@gzotti is right: we are currently working on adding data:  it's correct that there hasn't been any PR yet, but just a couple of hours ago, we discussed it (there had been some delay in the past ~2 weeks due to illness of humans who are unfortunately not able to work always at same pace in spare time activities).

PR #4700, opened 2025-12-21
SUG: foreword for 25.4

gzotti:
### Description

Just a short editorial. Did I forget something? Objections?


10110111:
It may be worth noting that the narration feature is not the accessibility reading feature that some users asked about, so those who are excited about it should be aware of this.

PR #4704, opened 2025-12-26
Avoid showing a separate-window tooltip

10110111:
On many Intel GPUs under Windows, when in full-screen mode, opening of a popup window for the tooltip by QToolTip switches the screen from exclusive full screen mode to a managed one, which results in flicker.

This patch instead shows the tooltip as a QGraphicsScene item, which avoids creating of a separate window.

This PR replaces #4678.

The implementation is currently incomplete. What's left to do is:

- [ ] Fix positioning of the tooltip when there's too small space to the right of the mouse cursor to fit the whole text. (Here a `QToolTip` would be positioned to the left of the cursor, and also clamped in x position by the screen edge.)
- [ ] Fix oversized tooltips after having showed a tooltip with e.g. two lines (e.g. the tooltip for "Skylight Details", then for "Limit magnitude") or a long line (e.g. the tooltip for "Use DE440", then for "Store language settings").
- [ ] Fix failure to remove the tooltip in some cases (e.g. hover over "Skylight Details" button, after the tooltip appears click the button, then press <kbd>Esc</kbd>).

### DIFF
--- a/src/gui/StelGuiItems.cpp
+++ b/src/gui/StelGuiItems.cpp
@@ -91,7 +91,9 @@ void StelToolTip::showToolTip(const QPoint& scenePos, const QString& text)
 	}
 
 	label->setText(text);
-	setPos(scenePos);
+	// The shift avoids clicking the tooltip instead of the control it's annotating
+	const QPoint shift(2, 2);
+	setPos(scenePos + shift);
 	updateGeometry();
 	setVisible(!text.isEmpty());
 }


alex-w:
The placement of tooltips are weird now

10110111:
> The placement of tooltips are weird now

What exactly is weird? Can you show some screenshots before/after this change?

alex-w:
> > The placement of tooltips are weird now
> 
> What exactly is weird? Can you show some screenshots before/after this change?

The current branch:
stellarium-030
stellarium-031
stellarium-032

10110111:
> The current branch:

The first screenshot shows the known problem with oversized tooltip (second in my list in the OP).

What does the second one show?

How did you get to the state of the third screenshot?

alex-w:
> The first screenshot shows the known problem with oversized tooltip (second in my list in the OP).

No, I said about positioning problem. The mouse pointer is over "UGC" label

> What does the second one show?

Same place for mouse pointer, but in fillscreen mode.

> How did you get to the state of the third screenshot?

The mouse pointer is localed over middle checkbox for "Horizon line"

PR #4148, opened 2025-02-23
TC plugin: add Alpaca

gzotti:
### Description

This branch shall be used to add support for ASCOM/Alpaca

Motivation: Our ASCOM support was reportedly prone to frequent crashes when built with Qt6. 
Meanwhile ASCOM7 is out, which seems to work again. 

Alpaca, a platform independent IP based protocol, is the new player in the ASCOM universe. We also should observe the EOL of .NET3.5 in early 2029.  So, replacing COM with that seemed the longer-time solution. 

The Alpaca standard is platform independent, so it would be a new client for all platforms. When Alpaca works and ASCOM still occasionally fails, we can then bury it. 

However, _one part of the Stellarium telescope protocol seems problematic_. The isAlive test which is called every frame exchanges a data package to the connected device. May be OK for a private serial line, but maybe not in the local IP network, connecting to semi-remote instruments controlled by miniPCs or low-power Raspberries. I cannot say at this time whether this data exchange can be moved to a lower-priority e.g. timer-triggered call. **Should therefore be developed and field tested by somebody with access and regular use of such hardware.** This change may affect all other clients. 

Fixes # (issue) [uncountable]

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update
- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11
* Graphics Card: Probably irrelevant

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4776, opened 2026-02-17
Fix win11 arm64

### LINKED ISSUE #4772: Finalize adoption of WIndows/Arm64 architecture
Thanks to a device donation from Microsoft we have finally an Arm64 (Surface laptop) to test and finetune some parts of the program which we could not test so far. 

I observed the following issues in 25.4 where the Arm/64 Windows download differs from Intel x86_64 Windows, and thought of the following areas to investigate. Please add where you see need.

 -[ ]-  Upgrade gray splash (indicating "unfinished"). This should be fixed last!
 -[ ]-  Sign the installer
 -[ ]-  Make ShowMySky available! This is a pity.
 -[x]-  Test Star catalog download
 -[ ]-  Test DE4xx ephemerides
 -[x]-  Test Spout
 -[x]-  Test Speech output
 -[ ]-  Test Media output
 -[ ]-  Test RemoteControl, RemoteSync
 -[ ]-  Test online DSS and HiPS
 -[ ]-  Test offline DSS and locally installed HiPS
 -[ ]-  Test Telescope Control plugin: LX200 native/Serial and ASCOM7.1. 
 -[ ]-  Test Scenery3D
 -[ ]-  Test rendering quality of Sun (texture size) and esp. Moon.

gzotti:
### Description

This shall include fixes identified during testing work on Arm64 hardware donated by Microsoft.

Apart from the tests listed in #4772, getting ASCOM to play nice seems the largest issue. :-(

Fixes #4772 (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Windows 11 on ARM64 25H2
* Graphics Card: Qualcomm Adreno X1-85 

## Checklist:


plugins/TelescopeControl/src/common/ASCOMSupport.cpp
```diff
@@ -61,16 +62,21 @@ int ASCOMSupport::getASCOMMajorVersion()
 	IDispatch* utilDispatch;
 
 	initResult = OleInit(COINIT_APARTMENTTHREADED);
-	hResult = OleCreateInstance(L"ASCOM.Utilities.Util", &utilDispatch);
+        hResult = OleCreateInstance(L"ASCOM.Utilities.Util", &utilDispatch);
```
10110111: Why this change to spaces?

plugins/TelescopeControl/src/common/ASCOMSupport.cpp
```diff
@@ -61,16 +62,21 @@ int ASCOMSupport::getASCOMMajorVersion()
...
 
 	// OLE Problem
 	if (FAILED(hResult) || !initResult) {
+                qCCritical(Telescopes) << "ASCOM getASCOMMajorVersion(): initResult=" << initResult
```
10110111: Don't you have to use `qCCritical(Telescopes).nospace()` to avoid `0x 1234` instead of `0x1234`?

Also, you're using spaces instead of tabs.

10110111 (commented):
So this doesn't seem to "fix" anything, just tweaks logging and the docs. Am I missing something?

10110111:
> getting ASCOM to play nice seems the largest issue

If you have access to the actual hardware, could it be better to switch to Alpaca and drop the COM-based ASCOM? This would bring support for these devices to Linux (and possibly macOS).

gzotti:
Please ignore this PR for now. It is a draft, mostly used for exchanging between two systems. Yes, spacing is off, I had not installed the style file on the new system. Yes, Alpaca would be nice, but see other PR, need deeper insight into Stellarium's underdocumented TelescopeControl plugin. In just these days it may even be an issue between MS and ASCOM with some problems in .NET3.5.

plugins/TelescopeControl/src/common/ASCOMSupport.cpp
gzotti: What's worse is the double quotation marks around the QString::number(). Is there another tweak against that?

10110111: Just append `.noquote()` to the `.nospace()` :)

plugins/TelescopeControl/src/common/ASCOMSupport.cpp
gzotti: I have loaded the .clang-format file now. When I select a region and right-click "Auto-Indent Selection", there are tabs at the head of lines. As soon as I store the file, there are space characters. Same nonsense on every new system, and doing this only once per year I have forgotten the switches.  :-( Will find out.

BTW, is .clang-format or still the older xml file the right? Why are there both? Could we delete one?

10110111: > BTW, is .clang-format or still the older xml file the right? Why are there both? Could we delete one?

In Ubuntu 22.04 the `qtcreator` package is still v6.0.2, which needs an XML file. Modern QtCreator (that I built from sources) wants a clang-format file. So we need both. (And ideally, also update the Coding Style docs to include the clang-format file, in addition to the XML.)

gzotti: OK, then we need both, yes, and I hope they are equal in rules. (My 2012 laptop must stay on U22.04LTS until 2032 or until nouveau supports a Geforce 580...) 
Just the auto-conversion from tab to space on store is annoying, I must find the relevant switch.

gzotti:
Of course, if anyone has time/knowledge to go into Alpaca (which works on all platforms), the relevant branch #4148 is rebased and open for continuation.

gzotti:
With some help from the ASCOM developers we found that we need a sidecar file `stellarium.exe.config` to be installed next to stellarium.exe:

```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
  <startup useLegacyV2RuntimeActivationPolicy="true">
    <supportedRuntime version="v4.0"/>
    <supportedRuntime version="v2.0.50727"/>
  </startup>
</configuration>
```

Where should this go in the repo (inside ASCOM sources?), and how do I ensure it is being copied to the install folder? In addition, apparently this use of ASCOM based on .NET requires a signed program to run without interference of 'Smart App Control'. 

I am in the process of upgrading qDebug() etc into qCDebug(Telescopes) and decide what's  debug and what's a warning, but this should go into all parts of the plugin, will take a few days longer. We can then finetune also this plugin, with qCWarning as default.

alex-w:
@gzotti it’s an easy - please see similar files in `data` directory + `iss` files in `cmake` directory (of course if you want to put this file when telescope control plugin is enabled then you need to update `CMakeLists.txt` file too)

gzotti:
Now I am back at the coding style issue. Using the .clang-format, indentation works with tabs. When I store the file, tabs are replaced by 8 spaces.  Where is the switch to fix that nonsense?

EDIT: Bah, found it. Preferences/Text Editor/Cleanups Upon Saving. Disable "Clean indentation".

alex-w:
> EDIT: Bah, found it. Preferences/Text Editor/Cleanups Upon Saving. Disable "Clean indentation".

Hmm… Should be xmi file updated too?

alex-w:
@gzotti I fear patch for `stellarium.exe.config` should be a bit different

10110111:
> Should be xmi file updated too?

What do you mean? The XML is short and to the point. "Cleaning" is a separate config option (rather stupid in that it doesn't follow the coding style configuration).

10110111:
How can we refer to the `.clang-format` file in `doc/codingConventions.doxygen` without having to maintain a copy of it in the `doc` directory, but having it available on the website?

gzotti:
> @gzotti I fear patch for `stellarium.exe.config` should be a bit different

I know. The file is where we said, but the copying to destination is not there yet.

gzotti:
> How can we refer to the `.clang-format` file in `doc/codingConventions.doxygen` without having to maintain a copy of it in the `doc` directory, but having it available on the website?

Do we need to actually show its contents? Else, a link to Github should suffice. But a notice esp. also in the BUILDING instructions for newcomers that in addition to this file being used by QtCreator, developers must also disable the switch mentioned above. This is more a QtCreator issue, but we have to live with it. When the last LTS version that depends on the old XML style file is EOL, we can at least remove that one.

gzotti:
Thanks. But IMHO it should be behind "if ENABLE_PLUGIN TelescopeControl and ASCOM" (or something similar that works :-).

gzotti:
I think the AppVeyor for Windows (all flavours) could be upgraded to Qt6.10 if possible. On Windows the libraries are packed with the program, so we can take the latest.

alex-w:
> Thanks. But IMHO it should be behind "if ENABLE_PLUGIN TelescopeControl and ASCOM" (or something similar that works :-).

Yes, we can do it, but I'm not sure that we need overengineering here...

alex-w:
> What do you mean? The XML is short and to the point. "Cleaning" is a separate config option (rather stupid in that it doesn't follow the coding style configuration).

I mean should we update `stellarium-ide.xml` file or not?

alex-w:
> I think the AppVeyor for Windows (all flavours) could be upgraded to Qt6.10 if possible. On Windows the libraries are packed with the program, so we can take the latest.

You can try it

P.S. You want to drop support 32-bit (Qt5) packages?

10110111:
> I mean should we update stellarium-ide.xml file or not?

No, it's completely correct.

I've updated the relevant docs (see 103511192487a409bd9dbdffd2f852b880a58be7).

gzotti:
> P.S. You want to drop support 32-bit (Qt5) packages?

I thought we had dropped 32bit Windows with the new star catalog? At least there is no download for it, right? Actually, the ARM-related changes should not affect Qt5 builds. (Maybe the config file is useful also here.) I meant, all Appveyors now running Qt6.8 (or lower?) could be updated to 6.10 if possible. I'll try...

alex-w:
> > P.S. You want to drop support 32-bit (Qt5) packages?
> 
> I thought we had dropped 32bit Windows with the new star catalog? At least there is no download for it, right? Actually, the ARM-related changes should not affect Qt5 builds. (Maybe the config file is useful also here.) I meant, all Appveyors now running Qt6.8 (or lower?) could be updated to 6.10 if possible. I'll try...

No, we still publish win32 packages. The current branch can’t be compiled with Qt5 (type conversion issues).

P.S. Maybe it’s time to drop support win32 packages (W7+), but we should resolve issues for compilation with Qt5.

P.P.S. Qt 6.8+ requires Windows 10+.

gzotti:
Type conversion issues in Qt5: Ah yes, I think this can be fixed.

But I see we have: Qt5/Win64 for Win7+, Qt6/Win64 for Win10+, and Qt6/Arm64 (so far Win10+, but we should officially rather write Win11 now). I see no packages for 32-bit Windows. (Although the "Windows Platform" is named Win32, yes that's confusing. This is from when 32bit was new...)

>Qt 6.8+ requires Windows 10+.

Yes, but we already used 6.8.  6.10 still supports W10 1607 and later, which is really not much to ask. W7 stays at Qt5. (Still 5.12. Will we ever use 5.15? If not, I could delete it...)

alex-w:
> But I see we have: Qt5/Win64 for Win7+, Qt6/Win64 for Win10+, and Qt6/Arm64 (so far Win10+, but we should officially rather write Win11 now). I see no packages for 32-bit Windows. (Although the "Windows Platform" is named Win32, yes that's confusing. This is from when 32bit was new...)


Oops… you are right - we have 64-bit packages only for Windows

alex-w:
> Still 5.12. Will we ever use 5.15? If not, I could delete it...

Qt 5.15.2 has serious regression in localization support - we may switch to use Qt 5.15, but for English-only package or via workaround for downloading external translations…

gzotti:
5.15.2 fixed most of 5.15.[01]'s crazy memory hunger for compiling translations. It worked for me (5.15.0 failed), but is the CI memory limited? 

EDIT: Just tried again. At some point it consumed about 30 from my 64 GB on 20 threads. So OK, I don't know if a 4 or 8 or even 16GB build system can succeed.

gzotti:
It seems I can only use a Qt6.9 image in AppVeyor. This may miss QWebEngine (used in OnlineQueries) which comes with 6.10 only (for Arm). Temporary replacement just opens in the system browser. But why is the scConverterEnabled=false ? Even if it's an x64 command line tool (if I understand the after_test section in the appveyor.yaml) it should work by emulation layer, or not?

alex-w:
> But why is the scConverterEnabled=false ?

Because this is arm64

gzotti:
Even if it's an x64 command line tool (if I understand the after_test section in the appveyor.yaml) it should work by emulation layer, or not?

alex-w:
> Even if it's an x64 command line tool (if I understand the after_test section in the appveyor.yaml) it should work by emulation layer, or not?

We used win64 libraries and produced win64 CLI application for converter. Mixing win64 and arm64 libraries/applications is not very good idea.

Of course you can download win64 converter on arm64 device and check how it works or not.

gzotti:
Even worse issue: Wanted to install Qt6.9 locally to try what we can do with this latest AppVeyor config, but apparently this comes without qtopengl (???) Are you in contact with someone there, or should I try to ask for a Qt6.10 environment?

gzotti:
Coveralls CI is failing. Any ideas? Other than that, I think it's ready. 

But what was the keyword again to trigger building a downloadable signed (?) beta installer? With Qt6.9 it will lack QtWebEngine but may work otherwise.

alex-w:
> Coveralls CI is failing. Any ideas? Other than that, I think it's ready.

It's internal server error - please ignore Coveralls at the moment

CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ ELSE()
      SET(ISS_MIN_WIN_VERSION "6.1")
 ENDIF()
 # Define required version for MSVC
-# Details: https://docs.microsoft.com/ru-ru/windows/win32/winprog/using-the-windows-headers
+# Details: https://docs.microsoft.com/en-en/windows/win32/winprog/using-the-windows-headers
```
10110111: Shouldn't this be `en-us` rather than `en-en`? (Similarly below.)

src/main.cpp
```diff
@@ -471,8 +471,8 @@ int main(int argc, char **argv)
 
 	const auto virtSize = QSize(confSettings->value("video/screen_w", screenGeom.width()).toInt(),
 								confSettings->value("video/screen_h", screenGeom.height()).toInt());
-	const auto size = QSize(std::lround(virtSize.width()/pixelRatio),
-							std::lround(virtSize.height()/pixelRatio));
+        const auto size = QSize(std::lround(virtSize.width()),
```
10110111: This will break on the systems where pixel ratio is detected correctly. See the commit where the division was introduced: f74171abf02c58cadaa964a30c128540dca74689. If that code doesn't work for you, the scaling detection should be fixed instead (I think it should be simpler on Windows than on X11 or Wayland, no?).

gzotti: Yes, this change has to be tested more, indeed. I think on my AMD64 Windows I saw the same problem on a 4k screen, so it may rather need a  `#ifdef OS_WIN ` clause. Just that I usually run Stellarium in Fullscreen, never used scaling on my screens so far and the new device has a fine screen with scaling that showed a shrinking window. Will test on 2-3 screens and AMD64 (Win/Linux) and ARM64 (Win) on Saturday.

CMakeLists.txt
gzotti: Ah, indeed. en-en forwards to en-us, though.

alex-w:
Hi @hmartinez82, could you help us for building arm64 package of Stellarium for Windows?

hmartinez82:
@alex-w Where's the build failure? Or is it a runtime issue?

gzotti:
We need a build environment for Qt6.10 on Windows/Arm64 on AppVeyor. They only provide 6.9, but allow self-configured build environments. I will start tonight with trying that, but maybe we need some help.

alex-w:
I think at the moment we need a stable building arm packages with Qt 6.9

hmartinez82:
I see. I'll try to help once you get the env up and running. @gzotti  When was the dependency on QtWebEngine introduced? I'm asking because we have had Windows ARM64 builds of Stellarium Qt 6 for a long while now. 

Sorry if I'm not up to date with the changes of the past year, but what big changes happened that is making the ARM64 build this difficult? (Besides WebEngine)

gzotti:
The OnlineQueries plugin (2022?) has always preferred to use QWebEngine when available. Qt6.10 finally has it also for Arm64, and I have finally a Surface to test.  Now ASCOM is also working, and we should finally CI-build a signed fully operational installer package.

gzotti:
I asked Google AI for assistance. https://share.google/aimode/dOV4XOrDHM3mNpmXm. But I don't know, is the result one entry in the "matrix", or would the other recipes (for amd64, Qt5, ...)  also change?

hmartinez82:
Just a heads up regarding `aqt` Qt 6.10+, QtWebEngine and ARM64:
https://github.com/miurahr/aqtinstall/issues/967

alex-w:
Qtwebengine and qtpdf has separate installation since Qt 6.8

src/main.cpp
gzotti: My change is needed on Windows on AMD64 and Arm64 alike. I had never seen it on my always 100% only screens. So, more ifdefs.

gzotti:
Seems this new appveyor config is so invalid it's not even listed here. @alex-w do you have any first-sight ideas? (AppVeyor config is new to me, and a bit overwhelming. )

gzotti:
OK, I finally came to the points: 
- Inno fails to pack when qtstuff subdirectory is empty.
- Apparently it is windeployqt that does not fill qtstuff. 
- The installer lacks the Qt libraries. It just works on my test system where Qt libs are in PATH. 

@alex-w How can we bring AppVeyor to fill the Qt libraries (and whatever else is required) into qtstuff?

alex-w:
> OK, I finally came to the points:
> 
> * Inno fails to pack when qtstuff subdirectory is empty.

This is correct behaviour of InnoSetup

> * Apparently it is windeployqt that does not fill qtstuff.

Yes. The windeployqt working good on my PC, but I not on AppVeyor instance - at least I see installed arm64 DLL's from Qt 6.8 & Qt 6.9.

> * The installer lacks the Qt libraries. It just works on my test system where Qt libs are in PATH.

[windeployqt.err](https://ci.appveyor.com/api/buildjobs/jwi63iyfdrg83n3b/artifacts/windeployqt.err) has error in PATH - maybe some scripts is not called or some PATH is wrong?

> @alex-w How can we bring AppVeyor to fill the Qt libraries (and whatever else is required) into qtstuff?

I'll check couple ideas tomorrow for this...

gzotti:
Maybe a manual call of windeployqt with some extra arguments (https://doc.qt.io/qt-6/windows-deployment.html)? (I thought about manually copying the libraries, but it's quite a crazy lot...)

gzotti:
I don't think this changes anything, as the host tools were found. On the doc page (https://doc.qt.io/qt-6/cmake-variables-qtcore.html) I don't see something like QT_TARGET, but maybe the following can be enabled in the next run and it may give some insights into what it is doing.

[QT_ENABLE_VERBOSE_DEPLOYMENT](https://doc.qt.io/qt-6/cmake-variable-qt-enable-verbose-deployment.html)

alex-w:
Maybe - I just compare with locale instance where it works good.

P.S. Hmm... I disabled locally some stuff with external dependencies - it may affected the building process...

P.P.S. Just a test locally right now with same config as for AV

alex-w:
@gzotti please check installer from AV

gzotti:
Same size package, same windeployqt.err... Next try.

alex-w:
> Same size package, same windeployqt.err... Next try.

What you see within installer?

gzotti:
> > Same size package, same windeployqt.err... Next try.
> 
> What you see within installer?

- unsigned (like my previous try; I think it's intentional, right?)
- Having installed my previous build in Stellarium_26.0 and yours in Stellarium_26.0_AW, doing a diff on WSL:

```
diff -rq Stellarium_26.0*
Files Stellarium_26.0/ShowMySky-Qt6.dll and Stellarium_26.0_AW/ShowMySky-Qt6.dll differ
Files Stellarium_26.0/stellarium.exe and Stellarium_26.0_AW/stellarium.exe differ
Files Stellarium_26.0/translations/stellarium/fr.qm and Stellarium_26.0_AW/translations/stellarium/fr.qm differ
Files Stellarium_26.0/unins000.dat and Stellarium_26.0_AW/unins000.dat differ
Files Stellarium_26.0/unins000.exe and Stellarium_26.0_AW/unins000.exe differ
```

The program folder in the AV builds contains no Qt6*.dll, no iconengines, multimedia etc. subdirs, i.e. everything that should have been in qtstuff.

alex-w:
Interesting. Please check architecture of ShowMySky-Qt6.dll

gzotti:
/mnt/c/Program Files$ file Stellarium_26.0*/ShowMySky-Qt6.dll
Stellarium_26.0/ShowMySky-Qt6.dll:    PE32+ executable (DLL) (GUI) Aarch64, for MS Windows, 6 sections
Stellarium_26.0_AW/ShowMySky-Qt6.dll: PE32+ executable (DLL) (GUI) Aarch64, for MS Windows, 6 sections

 ls -l Stellarium_26.0*/ShowMySky-Qt6.dll
-r-xr-xr-x 1 gzotti gzotti 276480 Mar 10 15:07 Stellarium_26.0/ShowMySky-Qt6.dll
-r-xr-xr-x 1 gzotti gzotti 276480 Mar 11 10:35 Stellarium_26.0_AW/ShowMySky-Qt6.dll

Same size, same arch. Maybe some internal timestamp? Anyhow, qtstuff is missing...

gzotti:
@hmartinez82 I have given up on an AI-assisted Qt6.10 config for now. I have raised the Qt6.8 to Qt6.9, but it seems windeployqt fails to find qconfig.pri, and therefore qtstuff remains empty. We tried to call windeployqt later in the process, but of course it likewise failed. Do you have any ideas on this? Or do we have to revert to Qt6.8 also for ARM64? (We had lost QtWebEngine anyhow. So if at least Qt6.8 builds a signed Arm package, we take that until appveyor can provide a working Qt6.10+ environment.

alex-w:
> Or do we have to revert to Qt6.8 also for ARM64? (We had lost QtWebEngine anyhow. So if at least Qt6.8 builds a signed Arm package, we take that until appveyor can provide a working Qt6.10+ environment.

Latest signed package for ARM64 was with Qt 6.5

gzotti:
Ouch. Is Stellarium the only project that wants a working AppVeyor CI system for Win/Arm with Qt6.10? Or 6.8? I understand they want to support LTS only, but 6.8 IS an LTS.

gzotti:
We are back at square 1 :-(  qtstuff is still empty, windeployqt cannot call qtpaths correctly: 

> Warning: Unable to read \mkspecs\qconfig.pri: The system cannot find the path specified.

The files are both in arm64 and win64 dirs, but how to tell windeployqt?

alex-w:
@gzotti could you share or send to me qtwindeploy from Qt 6.5?

gzotti:
I had a similar idea this morning. You said last successful (and signed!) win/arm64 installer build in CI was on Qt6.5. I wanted to have Qt6.10 because of QWebEngine. If that cannot be built on AppVeyor currently, maybe it is even irrelevant whether we have Qt6.5 or 6.8 or 6.9. Or do you think the actual windeployqt from Qt>6.5 is broken, and you want to "inject" the one from Qt6.5? Speech output needs Qt6.4+. Were there other features which need Qt>6.5?

Else I can send this tonight.

alex-w:
Hmm... what about switch to use Qt 6.5 for arm64 - I can find latest working config for it and apply here. I need couple, maybe three hours for experiments.

gzotti:
This is what I mean. Unless we are losing more than just WebEngine, it may be OK to keep win/arm64 at Qt6.5 until appveyor upgrades the CI to a working 6.10 (or even 6.12 LTS later this year).  For experiments, comment away (don't delete yet) all the experimental stuff I had added to src/CMakeFile and .appveyor.yml here. (I also modified the second windeployqt call in src/CmakeFile, not sure if this is needed.) I would just have liked to understand which program argument to change so that the paths are properly resolved. windeployqt.err always contained the same message, "cannot resolve \mkspecs\qconfig.pri" or so. The absolute path without Qt base directory is suspicious. Presumably something stupidly small has changed...

.appveyor.yml
```diff
@@ -66,17 +66,19 @@ environment:
...
     scConverterEnabled: false
-  - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
-    qtver: 6.5
-    qtbin: msvc2019_arm64
-    msvcname: Visual Studio 17 2022
-    cmake_args: -A ARM64 -DSTELLARIUM_BUILD_ARM64=ON -DQT_HOST_PATH=C:/Qt/6.5/msvc2019_64 -DQt6LinguistTools_DIR=C:/Qt/6.5/msvc2019_64/lib/cmake/Qt6LinguistTools -DENABLE_QTWEBENGINE=OFF
+  - BUILD_NAME: Arm64 Qt6.5/2022
+    APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2022
+    QT_VERSION_MAJOR: 6
+    QT_BASEDIR: C:\Qt\6.5\msvc2019_arm64
```
gzotti, line 75: Is this and the path in CMAKE_ARGS really msvc2019, not 2022?

alex-w: Please check https://www.appveyor.com/docs/windows-images-software/

alex-w:
@gzotti please check the package on real hardware

gzotti:
Qt6.9: libraries missing. (qtstuff...)
Qt6.5: works, gray splash, but without qttexttospeech. In the logfiles it is mentioned to be missing - is it unavailable in Qt6.5 for Arm or can it be added?

alex-w:
No, we can’t install extra modules for Qt 6.5 here. We should disable this feature via cmake (cmake_args) for arm64

gzotti:
Hmm, sad decision. But probably better a signed installer that works with ASCOM than an unsigned with speech output that strikes with ASCOM telescope driving. We will then (semi-)automatically upgrade as soon as AppVeyor steps up.

alex-w:
OK, could you prepare the branch to merge?

gzotti:
Yes, will do tonight.

src/CMakeLists.txt
```diff
@@ -692,14 +694,17 @@ IF(WIN32)
 		    COMMAND \"${WINDEPLOYQT_COMMAND}\"
 			 --dir \"\${CMAKE_INSTALL_PREFIX}/qtstuff\"
 			 --no-translations
-			 ${VIRTUAL_KBD_SETTING}
-			 --compiler-runtime
+                         --verbose 2
```
gzotti: I cannot say whether these added lines were useful. In addition, I just see a tabs/space mismatch again. My QCreator editor just does not use tabs here.

gzotti:
I squashed a few commits where mostly typos and early build failures were involved. Should I squash the numerous appveyor-and-cmakefile-only commits into just one?

alex-w:
> I squashed a few commits where mostly typos and early build failures were involved. Should I squash the numerous appveyor-and-cmakefile-only commits into just one?

At next step you want to use squash or rebase?

gzotti:
I think we should rebase-merge the branch to keep the few interesting "content" commits (changes for Arm integration, ASCOM stuff, windows screen scaling change) separate. Of course, some more of the 18 last commits (appveyor and CMakeFiles only) commits can still be squashed before merge when the result of an experimental change in the CI settings was just a failed CI run. In the end I would keep most comments in the appveyor file for reference when continuing raising Qt version for the arm build after the 26.1 release. When Arm builds on Qt6.10 I will be more than happy to remove the now-commented out "poke the CI for info what's happening" parts.

What about the two windeployqt runs in src/CMakeLists.txt? Should they have the same arguments?

alex-w:
> What about the two windeployqt runs in src/CMakeLists.txt? Should they have the same arguments?

this is two different cases for windeployqt ;)

gzotti:
I can see that. GENERATE_STELMAINLIB=false for MSVC, so the second call is almost never executed (only when doing private builds on MSYS? But never on appveyor/MSVC configs), and therefore maybe some arguments have not been added/updated in the last few years. Still, is it useful to add qt_paths, or set verbosity level?

alex-w:
> Still, is it useful to add qt_paths, or set verbosity level?

Yes, of course it’s useful - in any way (MSVC or MinGW/MSYS) windeployqt is called for packaging (MinGW/arm64 is working too)

gzotti:
What's the problem in the Mac/Qt5 build now?

alex-w:
> What's the problem in the Mac/Qt5 build now?

Troubles in qt5 package, please ignore it at the moment

gzotti:
OK. Should we do one final test before merge?  Else it's ready IMO.

alex-w:
It's ready to merge IMHO

PR #4779, opened 2026-02-19
Added Chinese_Xianglin_Star_Chart and Manchu sky cultures

Guanjin0562:
### Description

Added Chinese_Xianglin_Star_Chart and Manchu sky cultures.

The **Chinese_Xianglin** sky culture originates from a 17th-century Chinese antique star chart. The English translations of its constellation names are consistent with those of other Chinese sky cultures. Its classification parameter is "single".

The **Manchu** people are an ethnic minority in Northeast China. This sky culture is derived from ethnographic literature. All star names in the source material are recorded in Chinese, with no corresponding Manchu terms provided. Therefore, I believe it is necessary to include both Chinese and Manchu (reconstructed from Chinese pronunciation) in the "native" field. Its classification parameter is "ethnographic".

- [x] Did you remember to update `skycultures/CMakeLists.txt` file respectively to changes in sky cultures?
- [x] Did you remember to define classification parameter in sky cultures (see `index.json` file)?
- [x] Did you remember to define license parameter in sky cultures (see `description.md` file)?
- [x] Did you remember to define region parameter in sky culture (see `index.json` file)?

Fixes # (issue)
None.

image

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

The two sky cultures can be displayed normally, including constellations, names, illustrators and descriptions.
The sky cultures do not affect other code files or the program's operation.

**Test Configuration**:

## Checklist:


Guanjin0562:
The **Chinese_Xianglin​** sky culture originates from a 17th-century Chinese antique star chart. The English translations of its constellation names are consistent with those of other Chinese sky cultures. Its classification parameter is "single".

The **Manchu**​ people are an ethnic minority in Northeast China. This sky culture is derived from ethnographic literature. All star names in the source material are recorded in Chinese, with no corresponding Manchu terms provided. Therefore, I believe it is necessary to include both Chinese and Manchu (reconstructed from Chinese pronunciation) in the "native" field. Its classification parameter is "ethnographic".

alex-w:
Thank you very much for your contribution!

I have a question - is it possible to add an IPA data to Chinese Xianglin​ sky culture?

Guanjin0562:
> Thank you very much for your contribution!
> 
> I have a question - is it possible to add an IPA data to Chinese Xianglin​ sky culture?

Of course, it is possible, and I noticed that the Chinese Song dynasty sky culture already has IPAs with identical star names, so we can simply migrate and use them. However, this might take some time.

skycultures/chinese_xianglin/description.md
```diff
@@ -0,0 +1,40 @@
...
+The chart is organized according to the classic Three Enclosures and Twenty-Eight Lunar Mansions​ system. It comprises 31 individual sectional maps​ that, when assembled, form a complete all-sky dome centered on the North Celestial Pole. These maps clearly delineate key celestial reference lines: the Celestial Equator, the Ecliptic, and the Circles of Perpetual Visibility and Invisibility.
+
+In its depiction of constellations, the chart adheres to the traditional “Three Schools” convention: stars from the Shi​ and Wuxian​ schools are denoted with outlined circles, while those from the Gan​ school are marked with solid dots.
+
+Crucially, the chart introduces a significant innovation within this traditional framework: it departs from the common Chinese practice of not indicating stellar brightness.​ This is achieved by filling in the outlined circles​ representing particularly bright stars (e.g., ε Boo​ and β And) that belong to the Shi or Wuxian schools, thereby visually distinguishing them from fainter stars within the same constallation.
+
+![](illustrations/xianglin.png)
+
+*The Xianglin Star Chart (composited from 31 sectional maps)*
+
```
gzotti, line 28: You could add a section `### Constellations` here which could describe the figures a little more. Things you would like to tell when you explain the sky to others.

gzotti:
Thanks a lot! 
I wonder whether the illustrations are really useful though, as they are only connect-the-dot patterns which the linesets provide already. Or is there something I overlooked in my brief first fly-over? Why are there sometimes 2 versions?

Guanjin0562:
Regarding the **IPA** for the **Chinese Xianglin** sky culture, I have already discussed this with Sun Shuwei. I will conduct a more systematic organization and add them. If necessary, I may revise the existing IPAs in the **Chinese Song Dynasty** sky culture to ensure consistency.

For the description of the **Chinese Xianglin** sky culture, I believe we can follow the approach used for **Chinese Chenzhuo**—briefly describing a few distinctive constellations as examples. There is no need to list them all, as the complete constellation table is already available in the **Chinese** sky culture. It would be sufficient to highlight only the notable differences between them in this specific sky culture.

The constellation **illustrations** in the **Manchu** sky culture are indeed not essential. They are all hand-drawn star charts from field research. When creating the star lines, I referred to these images. However, some of them deviated too significantly from the actual starry sky, so I had to apply some stretching and adjustments, which resulted in two versions. The images in the description section are all the original ones. I can remove the unnecessary illustrations from the sky view and modify the filenames, displaying only all the original images in the description.

gzotti:
Of course the hand drawings give some personal/historical touch. If they fit well (did not check yet) you can leave them in. But i know sometimes adjusting a drawing with star symbols is a nightmare.

gzotti:
Hmm, the drawings don't really fit well, unfortunately. These should be re-made probably based on screenshots of Stellarium. Some distortions are unavoidable, but this cannot be presented to an auditorium of laypeople.

You could fully remove the drawings and emulate this style with our lines: zero-length sections (only one star) are drawn by circling the star. 


On my European Windows the first name (sorry for my ignorance, what character set is this?) is only displayed as empty blocks. We used a certain tweak for Cuneiform. Can you identify the character script name in https://doc.qt.io/qt-6/qchar.html?

gzotti:
Three issues:

1. Given its close connection, the skyculture should be named "Chinese Manchu". Then is is sorted next to its kin.

2. The logfile reports:

> [   468.425][WARN] Empty asterism lines array found for asterism "AST manchu 063" ("??")
> [   468.426][WARN] Empty asterism lines array found for asterism "AST manchu 092" ("??")
> [   468.426][WARN] Empty asterism lines array found for asterism "AST manchu 150" ("??")
> [   468.426][WARN] Empty asterism lines array found for asterism "AST manchu 166" ("??")
> [   468.426][WARN] Empty asterism lines array found for asterism "AST manchu 170" ("??")
> [   468.427][WARN] Empty asterism lines array found for asterism "AST manchu 264" ("???(??)")

3. The structure of description.md does not follow the mandatory structure described in the User Guide 9.1. 

- Introduction – a summary that could ideally be shown on a small screen entirely.
- Description – this can be structured with further subsections: non-astronomical cultural information (place, time, ethnicity, livelyhood, ...), astronomy, ...
- Constellations: structured list as described so that the speech engine can work.
- References 
- Authors 
- License

10110111:
> On my European Windows the first name (sorry for my ignorance, what character set is this?) is only displayed as empty blocks.

It's the [Manchu script](https://en.wikipedia.org/wiki/Manchu_script). For the first character in ᡤᠠᠰᡥᠠ I get this on Ubuntu 22.04:
```
$ fc-list :charset=1864
/usr/share/fonts/truetype/noto/NotoSansMongolian-Regular.ttf: Noto Sans Mongolian:style=Regular
```
BTW, does it not display even in the browser for you?

10110111:
This script is another example of a script that will break in gravity labels, e.g. ᡤᠠᠰᡥᠠ will look like so with the letters separated by spaces: "ᡤ ᠠ ᠰ ᡥ ᠠ".

gzotti:
I see the chars here in the github website. But the Qt font database needs a little nudge. I just added "Mongolian Baiti" to the list. (Qt6.8+ only.)

Gravity labels: Yes, this feature needs a rewrite for many non-Latin scripts. :-(

sushoff:
> The constellation **illustrations** in the **Manchu** sky culture are indeed not essential. They are all hand-drawn star charts from field research. When creating the star lines, I referred to these images. However, some of them deviated too significantly from the actual starry sky, so I had to apply some stretching and adjustments, which resulted in two versions. The images in the description section are all the original ones. I can remove the unnecessary illustrations from the sky view and modify the filenames, displaying only all the original images in the description.

thanks @Guanjin0562  for this amazing contribution! 

As your hand drawn images serve as a reference in scholarly work, I recommend to leave them in. 

@gzotti 
As missing fonts keep causing issues, might it be possible to link any public font repo to local Stellarium installations? (e.g. phabricator or googlefonts ...) Or to ask the question the other way round: can we make fonts "loaded" on purpose when a SC is called that needs them?

gzotti:
It seems we don't need to install any extra fonts. Qt has its own font database, but it may not know which fonts to apply to more exotic script systems. We can at this point not say when the next skyculture will be brought to us or which script it may use. It seems @10110111 can see the glyphs out of the box on Linux, but on Windows we must help Qt a bit in code and tell it "use font XY for xy script". I have added a fitting font for Mongolian script in master some days ago. (see above.) Of course someone could work into the future and go through the list linked in the discussion above, and decide on a stock Windows11 install which fonts to use for all these script systems, just in case.

sushoff:
sorry for not being clear: I did not think of "installing own fonts", but allowing them to be pulled if necessary. 

background thought: fonts are continually developed further - our knowledge on scripts grows, laypeople will be happy with "any" cuneiform-like font that happens to float around in any public library (might that be Qt or GoogleFonts or anything else), but specialists who wish to create illustrations for their research papers may want a specific font. Not Old Babylonian but NeoAssyrian (is the most frequent issue). Of course, not all amateur astronomers and school teachers will need that - or even be able to distinguish it. So, it would be nice of us to provide the specialists with the option to pull (download or use-by-hyperlink ...) a specific font if required.

gzotti:
Do you have a working example where the stock fonts don't work? Or can you just create a text list that lists, per QChar::script entry in the list linked above, a font name that should be loaded (not the font file name but names like "Segoe UI Historic"). You can just load whatever fonts you find and have installed. We could then add such list to config.ini (in 26.2, earliest). The default list should list only built-in fonts, like this (my current list)

```
QChar::Script_Cuneiform: "Segoe UI Historic"
QChar::Script_EgyptianHieroglyphs: "Segoe UI Historic"
QChar::Script_Mongolian: "Mongolian Baiti"
```

The affected users would then be free to configure what they want. If such specialized fonts come with Cuneiform and Latin glyphs, you can set them interactively as main font in the Stellarium GUI.  (But you don't want a GUI in Cuneiform, right?) Unfortunately, some of these fonts have weird metrics settings and irritate the GUI. The database entries should provide a way to use a font with good metrics as GUI and standard font, and switch in the characters from those extra fonts when needed. 
Currently this feature is only used on Windows. I know nothing about current font management on Linux or Macs, but understood from @10110111 that the intended glyphs from the Unicode character locations were correctly displayed on Linux.

alex-w:
Glyphs in both sky cultures are OK on macOS

Guanjin0562:
I've noticed the issue with the Manchu characters. However, I must point out that even if these characters appear to display correctly on your computer, they are most likely incorrect. This is because in Unicode, Mongolian and Manchu share the same code points. Even though some glyphs differ in appearance, they are assigned the same characters—a well-known point of criticism for Unicode.

Resolving this issue relies on fonts. However, given that most fonts default to Mongolian glyphs, correctly displaying Manchu text may even require dedicated Manchu fonts. This is a very challenging problem, and I’m not sure how to solve it yet. That said, if we plan to add other, more niche sky cultures that require special scripts in the future, the font issue will likely need to be addressed.

gzotti:
Uh, that's bad. My assumption was that Unicode finally has unique code points per script (also Old Babylonian/NeoAssyrian, above). For me a "font" has "glyphs" which are the implementation of "character" which may differ in artistic or evolutionary style (e.g. in Latin: Serif vs. Sans Serif, typewriter,  Blackletter (but has distinct code points already), etc.; Arab has a Kufi variant which are still the same alphabet, ...) but not actual meaning. **A**, _A_, _**A**_, `A` are all "A".

So, Manchu, users of NeoAssyrian, etc unite and request unique code points for your scripts in Unicode, and ask foundries to create your glyphs correctly!

10110111:
What is the actual difference between the Manchu and Mongolian script? Can you show an example of a correct vs incorrect rendering of some words/phrases? [Some discussions](https://tex.stackexchange.com/q/87259/30726) say that the difference is just an addition of some diacritics to the Mongolian letters.

Guanjin0562:
Thank you all for the suggestions over the past few days. I have made some modifications, and the details are as follows:

Regarding the **Chinese Manchu** sky culture:

1. Changed the sky culture name from "Manchu" to "Chinese Manchu".
1. Removed several empty constellations that were causing errors.
1. The constellation illustrations are temporarily retained, but I have removed the 3 images I manually stretched. The remaining ones are completely consistent with the source literature.
1. Considering that Manchu script might not display correctly, I have changed all `"native"` fields to Latin transliteration, which is a common practice when Manchu cannot be displayed in China. If there is an opportunity to improve this in the future, it can be reverted.
1. In the `description.md`, the Manchu spellings are displayed as images (as Manchu must be written vertically), similar to the Japanese sky culture. The Manchu font used (Alari Šanyan Manchu) has been authorized by its creator.

Regarding the **Chinese Xianglin** sky culture:

1. Added IPA for the constellations.
1. Improved the `## Constellations` section in description.md, listing the constellations that differ from the "Chinese" sky culture. For the full list, please refer to the Chinese sky culture.

The other changes are minor, such as supplementary references, acknowledgments, etc.

Please review my modifications!


Regarding the difference between Manchu and Mongolian scripts mentioned earlier, I used the letter 'i' as an example. Please refer to the image below. For both fonts, I used the same input method to input the same characters.

image


Incidentally, I also created a work on the sky culture of the Yi ethnic minority in China. Their alphabet is even rarer; although it's included in Unicode, it can't be displayed in Stellarium on my local computer. For example, it looks like this: ꄡꀨ ꆫꐪ.

gzotti:
Sounds good, but I am afraid handling vertical script exceeds my experience. Did we open Pandora's box here? We need some developer experienced in multilingual multiscript multifont handling, including mixing all edge cases and developing Gravity Labels. (Actually this looks even simpler for Manchu, just write "vertically" along an azimuth...) Regarding Latin transliteration, this is what the Transliteration field should be used for. 

For the Yi SC, is there a QChar::Script_XXX constant in the list linked above, and can you identify a font that is installed in Windows or Linux that has the relevant characters? Or is there at least an additional free font that can be downloaded and installed from somewhere? As said, Qt has a mechanism, but may need some help for unusual script systems.

alex-w (approved):
Thanks!

gzotti:
The "Divine Eagle" constellation is drawn with stars down to mag 9 or so, most likely to show an idea of its imagined shape, but such stars are way beyond human eyes. Are there historical records for it that use only naked-eye stars?

gzotti:
Maybe I missed it, but the description should explain why there is a transcribed first constellation name, then a Chinese (of which only few non-Chinese users will understand that it is not the original Manchu name), then again the same string in the Pronounce tag. Would the Pronounce and IPA tags now fit to the Chinese or Manchu strings? Or are they equivalent? If the Chinese strings are translations for Chinese users but are not Manchu language, these should be provided in the Translation system. Then the Original should be the Manchu Unicode strings, and a proper font identified. If a font is missing, users can still read Pronounce and IPA (which should represent Manchu), and just configure away display of the malformatted "original" strings.

gzotti:
The Constellations chapter in description.md is also not formatted translation, display and speech output friendly as documented in the Guide, ch.9.1. It should not be a table but a list of `#####`-sections each describing one constellation. General remarks (and things like the last addendum) should go into the `## Description` section. The HTML tables are legacy of old SCs but we should not use them in new constributions.

Guanjin0562:
The "Divine Eagle" (Gas'ha) constellation does include some symbolic lines with no actual stars. In fact, some constellations in star charts use imaginary lines for the constellation shape regardless of real stars—another example is the Rat constellation (Leo). Should I change the specific HIP codes for these lines to coordinates to indicate the absence of actual stars?

For the native field: Manchu is the culturally native name, with IPA/pronounce being Manchu transliterations. However, I believe Chinese characters are the academically authoritative native names and equally important. All source materials document these names in Chinese characters, and no academic papers provide Manchu spellings (the Manchu forms used here were reconstructed in collaboration with several Manchu speakers). Thus, I consider Chinese characters to be the more authoritative source. Of course, if you feel these Chinese characters should be classified as Chinese translations and excluded from index.json, that is perfectly acceptable—please kindly let me know. If retained, I will add the following text to the description:

```
### Notes on Script Representation

All traditional Manchu constellation names in this culture originate from the Manchu language. However, because the Unicode standard shares code points between the Manchu and Mongolian scripts, and most fonts default to displaying Mongolian glyphs, to ensure display consistency and avoid confusion, Stellarium does not provide Manchu script characters as their `native` names. Instead, it uniformly provides the Möllendorff Latin transliteration, which aligns with the `transliteration` names. Furthermore, since the original ethnographic research materials recorded these names and descriptions exclusively in Chinese, the authoritative original Manchu spelling is not directly attested. Therefore, the corresponding Chinese characters are also appended to the `native` names for reference.

To preserve and present the authentic visual form of the Manchu script, the Manchu spellings for the 19 core traditional constellations have been reconstructed and are displayed as images alongside their constellation illustrations in the "Constellations" section below.
```



I will revise the Constellations section to the correct format below. Please let me know if you would prefer the Chinese characters to be removed here:

```
##### Seven Maidens

<br>
<img src="illustrations/manchu_script/01Narhv_manchu.png" width="40">

*Nadan Narhū* (Chinese: 那丹那拉呼, English: Seven Maidens) corresponds to ...

##### Divine Eagle

<br>
<img src="illustrations/manchu_script/08Gasha_manchu.png" width="40">

*Gasha* (Chinese: 嘎思哈, English: Divine Eagle) is a ...
```

I have removed the constellation illustrations as I noticed they are sourced from an academic paper and may be subject to copyright restrictions.

Additionally, after reviewing the User Guide, I understand that Manchu constellation `id` should not use numbers. However, 3–4 letter abbreviations do not align with our usage conventions (they offer no mnemonic value, and abbreviation practices are difficult to standardize—this convention may be more prevalent in Western contexts). I would greatly appreciate your approval to set abbreviations as core words (e.g., Gasha, Narhu), which average 5–7 characters in length.

(Off-topic note: The User Guide uses Tibetan as an example for pronunciation—could you confirm if the Tibetan SC uses Wylie transliteration? I noticed the letter ö is included, which seems inconsistent with standard Wylie transliteration rules)

Guanjin0562:
> Maybe I missed it, but the description should explain why there is a transcribed first constellation name, then a Chinese (of which only few non-Chinese users will understand that it is not the original Manchu name), then again the same string in the Pronounce tag. Would the Pronounce and IPA tags now fit to the Chinese or Manchu strings? Or are they equivalent? If the Chinese strings are translations for Chinese users but are not Manchu language, these should be provided in the Translation system. Then the Original should be the Manchu Unicode strings, and a proper font identified. If a font is missing, users can still read Pronounce and IPA (which should represent Manchu), and just configure away display of the malformatted "original" strings.

I understand your point regarding Manchu characters. If you think it's acceptable, I can restore the Manchu Unicode in the native field, and then let the user configure the Manchu font or not display the native name.

gzotti:
I am by far no expert in any eastern-Asian sky culture. If Manchu is now written in Chinese script, but older Manchu script still available, a Manchu-Chinese double name is IMHO OK. But I don't know how to best then write the other fields. Maybe separated with hyphen or slash? 

Using 5-7-Latin-char names as ID: OK, no problem. They _must_ be unique, and _should_ be short to not clutter the screen, that's at least the idea.

The Tibetan SC should use Wylie in the 'transliteration' fields, but uses "simple pronounce" as found in non-Tibetologist expert textbooks in the 'pronounce' fields, I hope this is where you saw an ö.  It also "should" have genuine Tibetan in the original strings, but I can provide neither this nor Wiley by myself, sorry. If anybody can improve our Tibetan SC, go for it please.

I cannot say what the best way to show the SC names is. I cannot discuss "small" misrepresentations in a few Unicode code points. If it looks sufficiently OK for you, Manchu glyphs (even if borrowed from Mongolian) 'look' more authentic to me than just Chinese. If it is too erroneous to be useful, you may want no not even show it.  For anything but decorative value I must resort to whatever pronunciation, transliteration, or IPA is made available to users. Maybe @10110111 , @alex-w and @sushoff can also share their thoughts?

alex-w:
The both SC are OK for me

Guanjin0562:
1. The structure of the "Constellations" chapter in the `description.md` file has been revised to use the requested #####heading format for each entry.
1. The `native` field for constellation names has been updated to use the Manchu script (Unicode).
1. The corresponding Chinese characters have been removed from the `native` field. As a compromise, the Pinyin (Romanization) of the Chinese names as they appear in the source materials​ has been added to the `byname` field for reference.
1. The `description.md` file has been updated with a corresponding explanation regarding the script usage, font requirements, and the rationale for the `native``byname` fields.

I was not able to run the Python test utility mentioned in the Guide successfully. I will investigate this further.

Regarding the line data for asterisms like "Gasha" (Divine Eagle) that seem to mix real stars with symbolic lines: I understand a single constellation cannot use both HIP numbers and pure coordinate pairs in its definition. For now, I have left the data as is. Perhaps the solution is to split such a constellation into two separate logical parts?

sushoff:
> Regarding the line data for asterisms like "Gasha" (Divine Eagle) that seem to mix real stars with symbolic lines: I understand a single constellation cannot use both HIP numbers and pure coordinate pairs in its definition. 

Well, all HIP numbers have coordinates: The sky culture maker-plugin gives an output that runs correctly (although the layout needs to be beautified for human-readability). 

> For now, I have left the data as is. Perhaps the solution is to split such a constellation into two separate logical parts?

Guanjin0562:
> Well, all HIP numbers have coordinates: The sky culture maker-plugin gives an output that runs correctly (although the layout needs to be beautified for human-readability).

I see. However, constellations connected solely by coordinates are treated as "dark constellations" and are displayed differently from those bound to actual stars. More importantly, they cannot be linked to specific stars and therefore cannot account for proper motion over time.

sushoff (commented):
great SC, well done! thank you so much <3

sushoff:
> > Well, all HIP numbers have coordinates: The sky culture maker-plugin gives an output that runs correctly (although the layout needs to be beautified for human-readability).
> 
> I see. However, constellations connected solely by coordinates are treated as "dark constellations" and are displayed differently from those bound to actual stars. More importantly, they cannot be linked to specific stars and therefore cannot account for proper motion over time.

Oh, interesting aspect ... indeed, we discussed that dark constellations have to be propagated with precession, but of course, no pm is given.

Guanjin0562:
image

The most practical approach seems to be splitting the asterism into two parts in the data: one with HIP-bound lines, and another for the purely symbolic lines defined by coordinates. Would an entry with no common_namecause any issues?

Ideally, I'd still prefer a way to mix HIP numbers and coordinate pairs within a single constellation entry, and to tag it appropriately (coordinate-based lines aren't necessarily "dark"). Using only coordinates for bright stars is less precise and doesn't account for proper motion.

gzotti:
The dark constellations explicitly are formed from dark clouds seen against the bright Milky Way, so the lines should not involve stars and their proper motion. What you want to show is (probably) what the authors of the classical European copper print atlases were "imagining" with their artwork, without showing connecting telescopic stars when they were engraving figure details and outlines.  When I see Cygnus in a dark sky, I can imagine its wings and feathers although I only see the stick figure "skeleton". The outline of a swan figure would be given in the artwork layer, and not by a connect-the-dot drawing of whatever mag10 stars are sitting on the imagined outline. One question is, is there a tradition (at least a few decades, necessarily from the telescopic era) in those outlines, did you invent these mag6/mag8 star chains to outline the figure artistically, or is there a publication that already has them?

sushoff:
> The dark constellations explicitly are formed from dark clouds seen against the bright Milky Way, so the lines should not involve stars and their proper motion. What you want to show is (probably) what the authors of the classical European copper print atlases were "imagining" with their artwork, without showing connecting telescopic stars when they were engraving figure details and outlines. When I see Cygnus in a dark sky, I can imagine its wings and feathers although I only see the stick figure "skeleton". The outline of a swan figure would be given in the artwork layer, and not by a connect-the-dot drawing of whatever mag10 stars are sitting on the imagined outline. One question is, is there a tradition (at least a few decades, necessarily from the telescopic era) in those outlines, did you invent these mag6/mag8 star chains to outline the figure artistically, or is there a publication that already has them?

No, this is not the question. 

The problem is that sometimes it may make sense for your 'cygnusses' to use both stars and coordinates to draw a stick figure that helps the modern viewer to imagine the constellation in the same way as the ancients. Technology is made to provide options and must not limit what can be done.

Guanjin0562:
The lines in question are traced from hand-drawn star charts created during ethnographic field research in the last century. This is the sole source for these outlines, and the researcher who created them has since passed away. I had initially included these original drawings in the description for reference, but have since removed them.
You are correct that these specific connecting lines are, in a sense, artistic interpretations from the source material, and there are no actual stars that perfectly match the precise paths shown. However, these roughly sketched lines form a significant part of the overall figure for some of these asterisms. Omitting them entirely would severely disrupt the intended shape and cultural interpretation of the constellation as documented.

gzotti:
OK, omitting the "idea" of the shape is a loss, but that makes the drawing of "wing is somewhere over there" eligible in the artwork layer. Deviations from this rule are possible if sufficiently explained in the text. ("We want to show XY because...", "lines based on the fieldwork by..."). Else, without such explanation, it appears the Manchu have seen an outline of the figure formed by mag8 stars, which is just not plausible.

sushoff:
I think we/ I start over-engineering... 

As far as I understand @Guanjin0562 , he found a workaround for Manchu SC. So, we can - in my view - just accept the SC as is for the March release. 

However, @gzotti, I think we could think about providing more technical options in future releases - with no urgency, of course... pretty much as for the "sky culture explorer" in the other threat ...

gzotti:
It is also that somebody relevant here once told me the Chinese SCs traditionally have used stick figures only (of stars visible with the unaided eye). Of course, any SC author can decide to fill the sky with patterns as needed.  However, only brighter stars (mag<5 with rare extensions into mag6) make sense to me for "traditional" SCs. And sorry, yes, I am just projecting that presumably many other users will have the same expectation. Questions will go to us, and we will have to write or at least coordinate the answers. So, better clarify now.

skycultures/chinese_manchu/description.md
```diff
@@ -54,6 +54,8 @@ The stellar nomenclature in this Manchu sky culture comprises two distinct layer
 
 Stellarium includes over 20 of these shamanistic constellations. It also uses asterism lines to represent the *xingguans*(Chinese costellations) from official Qing dynasty star catalogs (such as the *Yixiang Kaocheng*).
 
+It should be noted that some of the connecting lines in the traditional Manchu asterisms—especially those representing imaginative or symbolic parts of the figures—are artistic interpretations from historical hand-drawn charts and do not correspond precisely to actual stars. In order to preserve the overall shape of these constellations as traditionally depicted, lines occasionally link to stars fainter than 6th magnitude. This does not imply that the original culture recognized or included such faint stars; it is merely a graphical means to render the envisioned form.
```
alex-w: typo: `figures—are` -> `figures — are`

alex-w: typo: `asterisms—especially` -> `asterisms — especially`

10110111:
No, in English the m-dash is not separated from the words—at least not in
this use case.

alex-w:
> No, in English the m-dash is not separated from the words—at least not in this use case.

English English or American English?

skycultures/chinese_manchu/description.md
sushoff: that's right: https://en.wikipedia.org/wiki/Wikipedia:Hyphens_and_dashes

sushoff: https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style

Guanjin0562:
> It is also that somebody relevant here once told me the Chinese SCs traditionally have used stick figures only (of stars visible with the unaided eye). Of course, any SC author can decide to fill the sky with patterns as needed. However, only brighter stars (mag<5 with rare extensions into mag6) make sense to me for "traditional" SCs. And sorry, yes, I am just projecting that presumably many other users will have the same expectation. Questions will go to us, and we will have to write or at least coordinate the answers. So, better clarify now.

You are right. For almost all traditional SCs, there is a tendency to use stars brighter than magnitude 5, and it's almost impossible for them to include stars fainter than magnitude 6.5, which are invisible to the naked eye. However, for some specific SCs, this general rule might not strictly apply.
Take the Chinese system, for example, which includes 1464 stars. This sheer number means it inevitably incorporates a significant quantity of faint stars. One example is HIP 113445, a star of magnitude 6.44. It was listed as a component of the constellation "Thunder and Lightning" as far back as the 3rd century when the constellation system was being established (refer to the Chinese Chenzhuo SC). I also find it hard to understand, but all evidence points to this. The only plausible explanation is that the creators of these asterisms, in pursuit of a visually pleasing shape, incorporated some faint stars into the patterns. This practice is actually quite common in the Chinese constellation system.

skycultures/chinese_manchu/description.md
alex-w: Oops... @Guanjin0562 sorry, this is my fault

sushoff:
> It is also that somebody relevant here once told me the Chinese SCs traditionally have used stick figures only (of stars visible with the unaided eye). Of course, any SC author can decide to fill the sky with patterns as needed. However, only brighter stars (mag<5 with rare extensions into mag6) make sense to me for "traditional" SCs. And sorry, yes, I am just projecting that presumably many other users will have the same expectation. Questions will go to us, and we will have to write or at least coordinate the answers. So, better clarify now.

that's right ... the faint stars are rather brave identifications and I would be VERY careful with them in historical stick figures (with regard to their use in transdisciplinary research = use of maps by users without knowledge of the culture). However, I think, this does not really apply to the problematic cases here - this needs to be clarified in the description: you're right.

skycultures/chinese_manchu/description.md
gzotti: OK, good enough for me!

skycultures/chinese_xianglin/description.md
```diff
@@ -0,0 +1,101 @@
...
+<br><img width="500" src="illustrations/xianglin4.png"/>
+
+*Bond, Stomach, Hairy Head, Net, Turtle Beak, Three Stars Mansions*
+
+<br><img width="500" src="illustrations/xianglin5.png"/>
+
+*Well, Ghosts, Willow, Star, Extended Net, Wings, Chariot Mansions*
+
+## Constellations
+
```
gzotti, line 54: Also here: The technical structure of this section is important. The introductory text must be moved above into the `## Description`.  Hmm, actually all of this should be moved up. The chapter `## Constellations` should consist of `##### constellation A` elements only.

Guanjin0562: So, in the Constellation section, do I really need to write an introduction to all 283 constellations, each with a five-level heading `#####`?

10110111: If the goal of this section is to simply give some common information rather than describe each (or some) of the constellations, you can just make it `### Constellations`. The level-2 heading is reserved for the section that contains per-constellation descriptions.

Guanjin0562: I understand that the `## Constellations` section is used for audio explanations and therefore requires strict formatting. What I'm more concerned about is the specific content of that section, specifically the introduction to each constellation. I may not be able to complete the introductions for all 283 constellations.

10110111: This section isn't mandatory.

gzotti: The text in the `##### constel ` inside `## Constellations` is used for display and audio output. If the program cannot identify this structure, these output elements just don't show up. The system allows very detailed explanations or storytelling, it is up to you to make use of it.

Guanjin0562: Do I have to cover all the constellations? Can I complete a part first and gradually add more later?

gzotti: You don't "have to". Maybe start with important ones, and gradually add. You can also say OK for now and add/improve later. Just add info in the right sections please.

skycultures/chinese_xianglin/description.md
```diff
@@ -87,6 +87,8 @@ The constellations in the *Xianglin* Star Chart fully adhere to the traditional
 
 While omitting the constellations mentioned above, the official system added new constellations near the southern celestial pole, which are not present in the *Xianglin* Star Chart.
 
+In the following `Constellations` chapter, all 283 Chinese traditional constellations will be introduced. However, this introduction is *not yet complete*, as only basic information about the constellations, such as the number of stars and their locations, has been compiled. The cultural significance behind each constellation has not been fully explored in this chapter.
```
Guanjin0562: I have added a basic introduction for each constellation, following the classic text structure used in ancient Chinese books to introduce constellations. I have manually checked to ensure that the English version of each constellation corresponds to the index.json.

gzotti:
In description.md, you are using zero-width spaces, usually but not always after a `*starred*[ZWS]` group. (I see these spaces in Notepad++. )  This should not be needed in the MD and it creates problems for line wrapping.  Do you have a real reason for them, or else please remove them.

gzotti:
Also, I see warning messages concerning constellations "Wadan (The Celestial Altar)" and "Star of Wisdom".

skycultures/chinese_manchu/description.md
```diff
@@ -73,137 +73,137 @@ To reliably present the authentic visual form of the Manchu script, the reconstr
...
+This constellation is recorded in another work by Fu Yuguang, but no corresponding modern constellation is specified in the source material. Based on the provided star chart and descriptions such as "southern sky" and "bright stars," a plausible speculation is that it may also point to the Corvus constellation. This would mean it occupies the same region of the modern sky as the *Wadan* star, and could potentially represent an alternative mythological interpretation or naming of the same stellar group within a different regional or transmission lineage. Another view suggests a possible correspondence to the four stars θ, η, γ, and δ in Cancer (the additional stars included in the chart might represent the M44 cluster). However, this interpretation is difficult to reconcile with the descriptions "bright stars" and "southern sky."
```
Guanjin0562: The position of the Star of Wisdom constellation is unclear, hence it has not been added to `index.json`, but there is a description of it in `description.md`.

alex-w:
@gzotti @10110111 @sushoff is this PR OK for you now?

gzotti:
I have not tested it since my last comments. Are the warnings gone?

alex-w:
> I have not tested it since my last comments. Are the warnings gone?

I see in the log:
```
[   157.304][DBG ] convertMarkdownLevel2Section: cons not found:  "Star of Wisdom"
[   157.309][INFO] Loaded 312 / 312 asterism records successfully for culture chinese_manchu
[   157.311][DBG ] convertMarkdownLevel2Section: cons not found:  "Star of Wisdom"
```

Guanjin0562:
The warning message is:

> [DBG ] convertMarkdownLevel2Section: cons not found: "Star of Wisdom"

Sorry my previous reply was incomplete. The constellation "Star of Wisdom" is documented in the `## Constellations` section of `description.md`, but it is not included in `index.json` because its position is historically disputed. I only intended to describe it in the description text.
As far as I know, this warning should not affect the program's execution. Could you confirm if any other functionality I’m unaware of would be impacted by this?

By the way, the issue with the Wadan constellation has been resolved, with no warnings

10110111:
> Could you confirm if any other functionality I’m unaware of would be impacted by this?

The entries in the description are like values in a key-value structure, where the keys are the entries in the JSON. If you try to look for a description of this constellation in e.g. the narration feature, it will not be available due to its nonexistence in the JSON file.

If at some point Stellarium's UI starts behaving more like that of Stellarium Mobile, which hides all the constellation descriptions in the text and only shows them in the info panel of the selected constellation, this description of a non-existing constellation will never be shown.

gzotti:
I think the description of a constellation that cannot be localized should go into the overall `## Description`  section. Section `## Constellations`  should list concrete visible items.

@10110111 about constellation selection: this is currently still not really intuitive, admittedly. Constellation selection currently goes only via F3 panel. Not sure if any Ctrl+Alt+Click (or whatever) on a star should select the constellation instead?

For users with more than one screen, I wonder if we could make the view dialog configurable so that it could be shown on a second screen, and users could read the text without covering the screen. (Of course the RemoteControl webbrowser UI can be used now.)

10110111:
> For users with more than one screen, I wonder if we could make the view dialog configurable so that it could be shown on a second screen, and users could read the text without covering the screen.

And we are back to the discussion in #4731?..

gzotti:
It may be less. We have already created a StelDialog for the OnlineQueries (also used by the SCM) which IIRC just has a separate window. It might be possible to extend the StelDialogs generally to use either in-Mainview or separate windows.

PR #4804, opened 2026-03-05
Better Moon phase at wide FOV

### LINKED ISSUE #914: Realistic moon phase
A realistic moon phase show, as seen in the sky, is missing.
Now, the user can see the realistic lunar phase, only by zooming in, on the moon.

Atque:
### Description
This PR draws the Moon better at wider FOV's. It takes advantage of the `eclipsePush` and applies when the Moon is not in the Earth's shadow, making it FOV depentent. The push effect starts at roughly 9 degrees and reaches 4x at roughly 87 degrees. It is color corrected so that we keep the extinction reddening effect even at wider FOV's. The push effect starts at roughly 9 degrees and reaches 4x at roughly 87 degrees.

Fixes #914 (a better fix IMHO)

bild

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Zooming in and out at different phases and altitudes of the Moon

**Test Configuration**:
* Operating system: Windows 11 and Ubuntu 24.04.4 LTS
* Graphics Card: RTX 4070 Ti Super
* Qt 6.10 and 6.4.2

## Checklist:

### DIFF
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -4432,38 +4432,59 @@ void Planet::drawSphere(StelPainter* painter, float screenRd, bool drawOnlyRing)
 	{
 		GL(normalMap->bind(2));
 		GL(moonShaderProgram->setUniformValue(moonShaderVars.normalMap, 2));
+
+		// Compute FOV-based push (always active, not just during eclipse)
+		const double Lambda = getEclipticPos().norm();
+		const double tau    = atan(getEquatorialRadius() / Lambda) * M_180_PI;
+		const double fov    = StelApp::getInstance().getCore()->getMovementMgr()->getCurrentFov();
+		const float sizeRatio = static_cast<float>(tau / fov);
+
+		const float MIN_RATIO = 0.003f;
+		const float MAX_RATIO = 0.03f;
+
+		GLfloat push = 1.0f;
+		if (sizeRatio < MIN_RATIO)
+		{
+			push = 6.0f;
+		}
+		else if (sizeRatio < MAX_RATIO)
+		{
+			float t = (sizeRatio - MIN_RATIO) / (MAX_RATIO - MIN_RATIO);
+			push = 1.0f + 3.0f * (1.0f - pow(t, 1.0f/6.0f));
+		}
+
 		if (!rData.shadowCandidates.isEmpty())
 		{
 			GL(texEarthShadow->bind(3));
 			GL(moonShaderProgram->setUniformValue(moonShaderVars.earthShadow, 3));
-			// Ad-hoc visibility improvement during lunar eclipses:
-			// During partial umbra phase, make moon brighter so that the bright limb and umbra border has more visibility.
-			// When the moon is half in umbra, we start to raise its brightness. Near edge of totality we try to simulate the apparent super-bright edge.
-			static const double tweak=1.015; // 1.00 to have maximum push only in full umbra. 1.01 or even 1.02 looks better to show a brilliant last/first edge
-			GLfloat push=1.0f;
 
-			// Like in getVMagnitude() we must compute an elongation from the aberrated sun.
+			static const double tweak=1.015;
 			PlanetP sun=ssm->getSun();
 			const Vec3d obsPos=parent->eclipticPos-sun->getAberrationPush();
 			const double observerRq = obsPos.normSquared();
 			const Vec3d& planetHelioPos = getHeliocentricEclipticPos() - sun->getAberrationPush();
 			const double planetRq = planetHelioPos.normSquared();
 			const double observerPlanetRq = (obsPos - planetHelioPos).normSquared();
-			double aberratedElongation = std::acos((observerPlanetRq  + observerRq - planetRq)/(2.0*std::sqrt(observerPlanetRq*observerRq)));
-			const double od = 180. - aberratedElongation * (180.0/M_PI); // opposition distance [degrees]
-
-			// Compute umbra radius at lunar distance.
-			const double Lambda=getEclipticPos().norm();                             // Lunar distance [AU]
+			double aberratedElongation = std::acos((observerPlanetRq + observerRq - planetRq)/(2.0*std::sqrt(observerPlanetRq*observerRq)));
+			const double od = 180. - aberratedElongation * (180.0/M_PI);
 			const double sigma=ssm->getEarthShadowRadiiAtLunarDistance().first[0]/3600.;
-			const double tau=atan(getEquatorialRadius()/Lambda) * M_180_PI; // geocentric angle of Lunar radius [degrees]
 
-			if (od<tweak*sigma-tau)     // if the Moon is fully immersed in the shadow
-				push=4.0f;
-			else if (od<tweak*sigma)    // If the Moon is half immersed, start pushing with a strong power function that make it apparent only in the last few percents.
-				push+=3.f*(1.f-pow(static_cast<float>((od-tweak*sigma+tau)/tau), 1.f/6.f));
+			float eclipsePush = 1.0f;
+			if (od<tweak*sigma-tau)
+			eclipsePush = 4.0f;
+			else if (od<tweak*sigma)
+			eclipsePush = 1.0f + 3.f*(1.f-pow(static_cast<float>((od-tweak*sigma+tau)/tau), 1.f/6.f));
 
-			GL(moonShaderProgram->setUniformValue(moonShaderVars.eclipsePush, push)); // constant for now...
+			push = std::max(push, eclipsePush);
 		}
+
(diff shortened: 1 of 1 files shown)


src/core/modules/Planet.cpp
```diff
@@ -4432,38 +4432,59 @@ void Planet::drawSphere(StelPainter* painter, float screenRd, bool drawOnlyRing)
...
+		{
+			float t = (sizeRatio - MIN_RATIO) / (MAX_RATIO - MIN_RATIO);
+			push = 1.0f + 3.0f * (1.0f - pow(t, 1.0f/6.0f));
+		}
+
 		if (!rData.shadowCandidates.isEmpty())
 		{
 			GL(texEarthShadow->bind(3));
 			GL(moonShaderProgram->setUniformValue(moonShaderVars.earthShadow, 3));
-			// Ad-hoc visibility improvement during lunar eclipses:
```
alex-w, line 4439: Why important comments are removed?

alex-w:
Did you tested look of lunar eclipses?

alex-w:
I see artifacts at daylight time, but yes, now users can see phases of the Moon in wide range of FOV. It's good for me, but please restore important comments in the code.

gzotti:
A New Moon at FOV 40 looks like a diamond ring now. Intended?
image

A 1-day old moon is in fact not visible like that.  I would accept it only with a toggle that restores the previous behaviour (and yes, the comments that describe why and what)

alex-w:
@gzotti please try set FOV 60-70 degrees and check it again - it’s artifacts IMHO

gzotti:
Yes, artifacts. I see bright pixels at the sun-opposing side of the lunar disk even at 90° fov. Do we want them? It's not an improvement. 

Uh, I see this ring also now in 25.4.  So please, what changed? @Atque please show before/after comparison. A New Moon is hard to see, but ashen moonlight is already in use to push visibility, modelled from some papers. 

OK, with 25.4 in a parallel window, with fov around 100°, I see a 4-days old moon on a dark sky as tiny crescent in 25.4, but a practically full disk in this PR. Both are surrounded with the Halo which has always been a bit problematic. But I prefer the 4-day-crescent over a full disk that at least suggests a gibbous moon.

A week later, the almost full moon is then displayed brighter than the current, which is already very bright. This increase has it now overexposed.

10110111:
I think we can reduce the artifacts by choosing a finer mesh in the coarsest case (this of course is a separate change from this PR):
```diff
diff --git a/src/core/modules/Planet.cpp b/src/core/modules/Planet.cpp
index 8418447f08..46bfe0a624 100644
--- a/src/core/modules/Planet.cpp
+++ b/src/core/modules/Planet.cpp
@@ -4354,7 +4354,7 @@ void Planet::drawSphere(StelPainter* painter, float screenRd, bool drawOnlyRing)
 
     // Draw the spheroid itself
     // Adapt the number of facets according with the size of the sphere for optimization
-    const unsigned short int nb_facet = static_cast<unsigned short int>(qBound(10u, static_cast<uint>(screenRd * 40.f/50.f * sqrt(sphereScaleF)), 100u));   // 40 facets for 1024 pixels diameter on screen
+    const unsigned short int nb_facet = static_cast<unsigned short int>(qBound(18u, static_cast<uint>(screenRd * 40.f/50.f * sqrt(sphereScaleF)), 100u));   // 40 facets for 1024 pixels diameter on screen
 
     // Generates the vertices
     Planet3DModel model;
```

Atque:
> Yes, artifacts. I see bright pixels at the sun-opposing side of the lunar disk even at 90° fov. Do we want them? It's not an improvement.
> 
> Uh, I see this ring also now in 25.4. So please, what changed? @Atque please show before/after comparison. A New Moon is hard to see, but ashen moonlight is already in use to push visibility, modelled from some papers.
> 
> OK, with 25.4 in a parallel window, with fov around 100°, I see a 4-days old moon on a dark sky as tiny crescent in 25.4, but a practically full disk in this PR. Both are surrounded with the Halo which has always been a bit problematic. But I prefer the 4-day-crescent over a full disk that at least suggests a gibbous moon.
> 
> A week later, the almost full moon is then displayed brighter than the current, which is already very bright. This increase has it now overexposed.

IMO, when looking at the Moon in the night sky, it IS rather overexposed. At least I can't see the maria at full moon. So I don't think this is a problem.

The goal with this PR is to simulate lunar phases with Moon set at 1x size, and FOV at "large", and halo for moon = false.


> I think we can reduce the artifacts by choosing a finer mesh in the coarsest case (this of course is a separate change from this PR):
> 
> ```
> diff --git a/src/core/modules/Planet.cpp b/src/core/modules/Planet.cpp
> index 8418447f08..46bfe0a624 100644
> --- a/src/core/modules/Planet.cpp
> +++ b/src/core/modules/Planet.cpp
> @@ -4354,7 +4354,7 @@ void Planet::drawSphere(StelPainter* painter, float screenRd, bool drawOnlyRing)
>  
>      // Draw the spheroid itself
>      // Adapt the number of facets according with the size of the sphere for optimization
> -    const unsigned short int nb_facet = static_cast<unsigned short int>(qBound(10u, static_cast<uint>(screenRd * 40.f/50.f * sqrt(sphereScaleF)), 100u));   // 40 facets for 1024 pixels diameter on screen
> +    const unsigned short int nb_facet = static_cast<unsigned short int>(qBound(18u, static_cast<uint>(screenRd * 40.f/50.f * sqrt(sphereScaleF)), 100u));   // 40 facets for 1024 pixels diameter on screen
>  
>      // Generates the vertices
>      Planet3DModel model;
> ```

Are the meshes in your forks ready for use? If so, we might combine them into this PR.

gzotti:
Ah finally I see: visibility improves with halo switched off. 
But it is really too bright IMHO. This is with 4x magnification. Left is 25.4. It looks like a Total Lunar Eclipse, not a New Moon with ashen light.
Screenshot (6)

This mode can at best be placed behind another toggle, "brighten Moon for large FoV" or so. Seriously, I have balanced lunar brightness more than once, and it is difficult to test with all circumstances, with twilight or even light pollution enhancing full moon, etc. I also know the eclipse push is not perfect yet, as there is a point of sudden brightness increase, which I also see here at one zoom phase. It's far from easy. 

Full Moon in large fov is blindingly bright indeed. Still, we have the "increase moon size" for large fovs, which show a kind of "attention based scaling". (Observing with unaided eye, the Moon looks larger while we look at it.) With fov around 35°, a gibbous moon now is burned-out, scaled or not, while it shows maria in 25.4.

Atque:
@gzotti Do you think the current solution is acceptable if made togglable?

10110111:
> Are the meshes in your forks ready for use? If so, we might combine them into this PR.

No, and even if they were they definitely must not be combined with this PR, since their introduction is a completely unrelated change.

gzotti:
> @gzotti Do you think the current solution is acceptable if made togglable?

I am actually unsure whether we should have yet another new mode that has visible problems under many operational conditions. "Making the Moon more visible" can already be done with:
- planet marker. Circle that is visibly not a celestial item.
- moon halo. Needed when moon sphere is too small to be even plotted. A few years ago we dimmed it so that a tiny sphere shows a distinct phase.
- moon scaling. Helps with the "attention scaling" mentioned above.

This mode adds brightness, and IMHO often too much of it to see phase or so that maria are burned out, which is ugly.  If @alex-w and @10110111 like it, I can accept it behind a toggle which I most likely would permanently leave off. 

I think (looking into lunar appearances when not scaled up) there is room for improvement in the appearance of the halo. It should not appear as diffuse bright point centered on a lunar disk. It has also been proposed to adapt upscaling with fov. There are always ideas, but also real life...

gzotti:
@10110111 with more faces it gets better, indeed. Maybe even 32 minimum. Can you push that to master?

10110111:
I've pushed 20. It seems sufficient, especially when 8× multisampling is on.

Atque:
Just a thought before proceeding: Would FOV dependent Moon enlargement be more usable you reckon? So that the Moon keeps the same diameter WRT the viewport until it reaches 1x size when zoomed in enough? "Enlarge Moon at large FOV's", and making it toggleable?

gzotti:
Yes, I'd consider using this. But the adaptive scale should be an option. Some may want it permanently. Not sure how to set it up, like "min lunar diameter in pixels?" Or "percent of vertical window size"? It should work well with RemoteSync'ed systems, with and without display scaling. The actual scaling factor should of course then be displayed in the screen label.

Atque:
> Yes, I'd consider using this. But the adaptive scale should be an option. Some may want it permanently. Not sure how to set it up, like "min lunar diameter in pixels?" Or "percent of vertical window size"? It should work well with RemoteSync'ed systems, with and without display scaling. The actual scaling factor should of course then be displayed in the screen label.

All right, I'll try that in a new PR.

PR #4807, opened 2026-03-08
Replace deprecated constructor for QMouseEvent and avoid unnecessary …

pkrawczun:
…implicit conversion from QPoint to QPointF

### Description

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/src/core/StelApp.cpp
+++ b/src/core/StelApp.cpp
@@ -1170,8 +1170,13 @@ void StelApp::handleClick(QMouseEvent* inputEvent)
 	if (viewportEffect)
 		viewportEffect->distortXY(x, y);
 
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
 	QMouseEvent event(inputEvent->type(), QPointF(x*devicePixelsPerPixel, y*devicePixelsPerPixel), inputEvent->globalPosition(), inputEvent->button(), inputEvent->buttons(), inputEvent->modifiers());
 	event.setAccepted(false);
+#else
+    QMouseEvent event(inputEvent->type(), QPoint(qRound(x*devicePixelsPerPixel), qRound(y*devicePixelsPerPixel)), inputEvent->button(), inputEvent->buttons(), inputEvent->modifiers());
+	event.setAccepted(false);
+#endif
 	
 	// Send the event to every StelModule
 	for (auto* i : moduleMgr->getCallOrders(StelModule::ActionHandleMouseClicks))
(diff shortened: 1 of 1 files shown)


src/core/StelApp.cpp
```diff
@@ -1170,8 +1170,13 @@ void StelApp::handleClick(QMouseEvent* inputEvent)
 	if (viewportEffect)
 		viewportEffect->distortXY(x, y);
 
-	QMouseEvent event(inputEvent->type(), QPoint(qRound(x*devicePixelsPerPixel), qRound(y*devicePixelsPerPixel)), inputEvent->button(), inputEvent->buttons(), inputEvent->modifiers());
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+	QMouseEvent event(inputEvent->type(), QPointF(x*devicePixelsPerPixel, y*devicePixelsPerPixel), inputEvent->globalPosition(), inputEvent->button(), inputEvent->buttons(), inputEvent->modifiers());
+	event.setAccepted(false);
```
10110111: This common statement could be moved out of the `#if`-`#endif`.

src/core/StelApp.cpp
```diff
@@ -1170,8 +1170,13 @@ void StelApp::handleClick(QMouseEvent* inputEvent)
...
+#else
+    QMouseEvent event(inputEvent->type(), QPoint(qRound(x*devicePixelsPerPixel), qRound(y*devicePixelsPerPixel)), inputEvent->button(), inputEvent->buttons(), inputEvent->modifiers());
```
10110111: Please restore the tab that you replaced with spaces.

10110111 (changes requested):

10110111:
Shouldn't the global position also be multiplied by `devicePixelsPerPixel`, as it's done in `StelApp::handleWheel`?

pkrawczun:
> Shouldn't the global position also be multiplied by `devicePixelsPerPixel`, as it's done in `StelApp::handleWheel`?

But aren't wheel events different and scaling them by devicePixelsPerPixel is essentially a preference to keep zoom speeds consistent? I should actually test this empirically on a 4K screen to see what happens.

10110111:
> I should actually test this empirically on a 4K screen to see what happens.

You don't have to use a 4K screen, screen scaling is a user-level option that you can just as well enable on FullHD screens.

pkrawczun:
> > I should actually test this empirically on a 4K screen to see what happens.
> 
> You don't have to use a 4K screen, screen scaling is a user-level option that you can just as well enable on FullHD screens.

You're right. I just wanted to remove a deprecated piece of code.

gzotti (changes requested):
According to https://doc.qt.io/qt-6/qmouseevent-obsolete.html, also the newer proposed version is already deprecated. Please update to current standard as of Qt6.5 and later.

10110111:
> According to https://doc.qt.io/qt-6/qmouseevent-obsolete.html, also the newer proposed version is already deprecated. Please update to current standard as of Qt6.5 and later.

How so? The only constructor that's deprecated is the one that doesn't take the global position, while the new code here uses the one with the global position.

gzotti:
Oops, seems I was not looking straight. Then it's fine.

10110111 (approved):

PR #4812, opened 2026-03-11
Scenery3d key doc

10110111:
It appears that navigation keys of Scenery3D are different on PC and Mac, and this difference isn't mere terminology. It's [specially done](https://github.com/Stellarium/stellarium/blob/master/plugins/Scenery3d/src/Scenery3d.cpp#L124) to work around https://bugreports.qt.io/browse/QTBUG-36839. This PR documents the keys for the Help dialog, as well as describes the difference in the User Guide.

Please check that the Help dialog shows the correct info on macOS.

Also, the bug is now closed with the fix version being Qt 5.9. Should we maybe simply undo the PC/Mac difference and thus avoid all these special cases in the User Guide and conditional compilation in C++?

### DIFF
--- a/guide/plg_scenery3d.tex
+++ b/guide/plg_scenery3d.tex
@@ -33,10 +33,10 @@ sundials and other shadow casters in action, etc.
 You activate the plugin with the \emph{circular enclosure} button \guibutton{0.6}{bt_scenery3d_off.png} at screen
 bottom or by pressing \key{\ctrl+W}. A right-click on that button \newFeature{0.21.2} (or \key{\ctrl+\shift+W})
 opens the settings dialog. Once loaded and
-displaying, you can walk around pressing \key{\ctrl} plus cursor keys. Change eye
-height with \key{\ctrl+Page\,\arrowkeyup\,}/\key{\ctrl + Page\,\arrowkeydown\,} keys. Adding \key{\shift} key increases speed by 10,
-adding \key{\Alt} multiplies by 5 (pressing both keys multiplies by 50!). If you release \key{\ctrl} before
-the cursor key, animation will continue. (Press \key{\ctrl}+any cursor key to stop
+displaying, you can walk around pressing \key{\ctrl} on PC or \key{Option} on Mac plus cursor keys. Change eye
+height with \key{\ctrl+Page\,\arrowkeyup\,}/\key{\ctrl + Page\,\arrowkeydown\,} keys on PC or \key{Option+Page\,\arrowkeyup\,}/\key{Option + Page\,\arrowkeydown\,} keys on Mac. Adding \key{\shift} key increases speed by 10,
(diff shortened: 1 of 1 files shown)


alex-w (commented):
Why you do not use `hotkeyTextWrapper`?

10110111:
> Why you do not use hotkeyTextWrapper?

Well, first of all, because it won't work for abstract "arrow keys" or "PgUp/PgDown", second, `Pan view around the sky` description also doesn't use it (and I used it as the reference).

alex-w:
> > Why you do not use hotkeyTextWrapper?
> 
> Well, first of all, because it won't work for abstract "arrow keys" or "PgUp/PgDown", second, `Pan view around the sky` description also doesn't use it (and I used it as the reference).

OK, but we use hook for similar case in Telescope Control plugin already:
```
#ifdef Q_OS_MACOS
	QString modifierNameCtrl = "⌘"; // Command key
	QString modifierNameAlt = "⌥"; // Option key
#else
	QString modifierNameCtrl = "Ctrl";
	QString modifierNameAlt = "Alt";
#endif
```

10110111:
OK, I've replaced names with symbols.

alex-w:
It's OK for me, but just interesting:
a) Did you tried to use `QKeySequence(Qt::ALT).toString(QKeySequence::NativeText)` for baseMultiplier?
b) OK for arrow keys, but what about `hotkeyTextWrapper("Ctrl+Page Up") + "/" + hotkeyTextWrapper("Ctrl+Page Down")`?

10110111:
a) results in `Ctrl+` for `Qt::CTRL`, which is bad for the base multiplier that's simply `Ctrl` (Command).
b) seems OK, since it's more consistent with the description of the general PgUp/PgDown hotkeys.

10110111:
Maybe I shouldn't use `q_()` with "Alt" and "⌘" though... should I?

alex-w:
> Maybe I shouldn't use `q_()` with "Alt" and "⌘" though... should I?

I think we shouldn't use `q_()` for these strings

P.S. Alt/⌥

alex-w:
Hmmm... a bit mix now for shortcuts - please review it

10110111:
> a bit mix now for shortcuts

Not sure what you mean. Seems to reflect reality to me. At least on PC the keys work as specified. Do they not work on macOS?

gzotti:
I am a bit surprised to see plugin documentation in the main Help dialog. Shouldn't it at least depend on plugin availability (loaded state)? (Even cleaner: a doc hook in the StelPluginInterface class in which plugins can report features worth mentioning when asked by HelpDialog.)

10110111:
> Shouldn't it at least depend on plugin availability (loaded state)?

It does. A group list is retrieved from ActionMgr, where a plugin that isn't loaded won't register its shortcuts. Then, when processing the particular group, we can add the missing items.

But I agree, ideally this should be a separate API call to the plugin to get the shortcuts help text.

Issue #4782, opened 2026-02-24
A better texture of the Moon was released by NASA CGI Moon Kit


Atque:
A better Moon texture has been released by NASA using the CGI Moon Kit, released December last year.

https://svs.gsfc.nasa.gov/4720/

It can be found here. I haven't tested it myself yet, and it likely needs some kind of shading balancing, gamma correction and perhaps desaturation depending on taste. The most obvious optimization is the better coverage near the poles.

10110111:
From the preview image it looks like the poles have become blurrier, and
they now have stripe artifacts similar to those I've seen in the LOLA
reflectance map.
And the midlatitudes, judging by the description, have simply been made
more saturated etc. - the changes we would undo anyways.
I'll inspect the texture in detail later, but at the first glance it
doesn't seem to an improvement for our purposes.

10110111:
Yes, indeed it looks like they used some blurry version of the LOLA reflectance, which provides neither details nor color hues. I wonder why they didn't use the empirically-normalized WAC 643 nm data for ±(60°..90°) latitude available at 304 pixels/degree. I used them for my Moon HiPS, and it appears to be quite usable (aside from the same lack of hues, but these I don't know where to get from).

I suppose I should actually use the same process as I did for the creation of the HiPS to make our own version of the texture.

Atque:
I thought the normal + shadow maps would hide the blurriness, but maybe they're not enough. I think the color hue is better near the poles in this new version though. In our current texture, the missing data is more obvious.

10110111:
> I think the color hue is better near the poles in this new version though.

No, it isn't! Just try opening the texture and decompose colors into HSV, and see that Hue is constant, so there's not hue information in this texture.

Atque:
Oh, I'm just checking visually. :)

10110111:
Does the connection of the polar regions in [my HiPS](http://86.106.181.97/hips/hipslist) look better than any of these textures to you?

Atque:
> Does the connection of the polar regions in [my HiPS](http://86.106.181.97/hips/hipslist) look better than any of these textures to you?

This one? Yes, looks better. I can't see any blur or missing coverage. Cratering looks a bit weird, but I guess that's because we don't have a mesh.

<img width="969" height="724" alt="Image" src="https://github.com/user-attachments/assets/1dd00c4e-c044-4742-969c-4249177057b0" />

10110111:
[...]

PR #4816, opened 2026-03-13
Update albedo texture of the Moon

10110111:
This replaces the albedo texture of the Moon with my own processing of the LROC WAC-based normalized maps, as described [here](https://github.com/10110111/map-bumper/blob/master/hips-howto.mdown) but in equirectangular projection instead of HiPS and with additional color balancing to make the texture grayish.

This makes the poles much less different in apparent quality from other regions, and avoids the blurry version of them suggested in #4782.


gzotti:
Funny that you write 

> color-balanced to yield the grayish look of the texture

I thought this also just pushes through the reddish hue you wanted to add in #3041...

10110111:
Well it does have a little bit of redness compared to the old image, but I
think it's rather subtle, so can be ignored. If you want, I can try to make
it as neutral as possible.

Atque:
Wow, this looks very good! It is practically impossible to see the "bad area" above 70 lat now.

10110111:
For reference, here's the original image before gray-balancing:

moon_4k-new-orig

And here's how it'd look if we strive to reach maximal neutrality (here I used the criterion that GIMP's resize to 1px×1px yields identical values in RGB channels):

moon_4k-new-grayer

gzotti:
Yes, I prefer the neutral image. It's just IMV not a pink disk when seen small enough. Neutral, or yellowed/orange on the horizon. On beige/reddish hues in closeup, see discussion in #3041, I can imagine a fov-based hue adaptation (actually similar to white balance adjustment in photography) when the Moon fills a larger part of the screen. But that's beyond this texture change. (And yes, texture has nice detail!)

PR #4835, opened 2026-03-22
Add support for proper lunar mesh instead of the ellipse

10110111:
This is a WIP implementation of the lunar mesh, which helps us get

1. The proper silhouette of the Moon, and
2. Better look of the rim during full Moon due to the back-faced parts of craters and mountains not being visible.

What's currently missing:

1. ~~This should obey `flagUseObjModels` and resort to the old sphere-based rendering when it's disabled. Or maybe add a separate feature flag, since the other models are much more lightweight?~~
2. Surveys are still rendered as spheres (and I don't think this will be changed in this PR).

This implementation includes a 44 MiB lunar mesh in a custom binary format optimized for compressibility, so its 7zip archive is only 7 MiB. This archive is what's contained in the repo, and it gets extracted on installation, so the installers will get the 44 MiB input. I hope that in particular Inno Setup will be able to achieve a similar compression. Of all compressors only 7zip and lzma/xz achieve about 7 MiB size, others (bzip2, gzip) appear to be much less efficient (13 MiB and 21 MiB, respectively). I chose 7zip because CMake supports it out of the box.

The worst-case performance of displaying only the Moon fullscreen at 0.3° FoV on my ASUS ProArt PX13 (with AMD Radeon 890M) dropped from 42 FPS to 33 FPS, which seems fair for the improvement achieved. On Zenbook UX333F with Intel UHD Graphics 620 the drop is from 16 FPS to 4.4 FPS, which is why it's desirable to fix the ability to turn the feature off.

I don't intend for this to get into 26.1, since it's likely to break some other aspect of planet rendering, so I suppose it should aim at 26.2.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -42,8 +42,10 @@ environment:
     VSPATH: C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\SDK\ScopeCppSDK\VC\bin
     QT_VERSION_MAJOR: 5
     QT_BASEDIR: C:\Qt\5.12\msvc2017_64
-    CMAKE_GENERATOR: Visual Studio 15 2017 Win64
-    CMAKE_ARGS: --
+    CMAKE_GENERATOR: Visual Studio 15 2017
+    CMAKE_ARGS: -A x64
+    cmakeURL: https://github.com/Kitware/CMake/releases/download/v3.18.6/cmake-3.18.6-win64-x64.zip
+    cmakeBaseName: cmake-3.18.6-win64-x64
     exiv2url: https://github.com/10110111/exiv2/releases/download/ver0.28.0-final/exiv2-0.28.0-2017msvc64.zip
     exiv2baseName: exiv2-0.28.0-2017msvc64
     scConverterEnabled: false
@@ -113,6 +115,10 @@ before_build:
 
   - ps: if($env:exiv2url -ne $null) { appveyor DownloadFile $env:exiv2url -FileName c:\$env:exiv2baseName.zip }
   - ps: if($env:exiv2url -ne $null) { 7z e c:\$env:exiv2baseName.zip -spf -oc:\ }
+  - ps: if($env:cmakeURL -ne $null) { appveyor DownloadFile $env:cmakeURL -FileName c:\$env:cmakeBaseName.zip }
+  - ps: if($env:cmakeURL -ne $null) { 7z e c:\$env:cmakeBaseName.zip -spf -oc:\ }
+  - ps: if($env:cmakeURL -ne $null) { $env:PATH="c:\$env:cmakeBaseName\bin;$env:PATH" }
+
   - if [%PUBLISH_BINARY%]==[true] appveyor DownloadFile https://github.com/Stellarium/stellarium-data/releases/download/guide/guide.pdf -FileName c:\stellarium\guide\guide.pdf
 
   - ps: if($env:INSTALL_CONVERTER_DEPS -eq "true") { appveyor DownloadFile $env:gettextURL -FileName c:\$env:gettextBaseName.zip }
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.16)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.18)
 # CMake 3.20+ is required if x-compiling for Windows on ARM
 MESSAGE(STATUS "Found CMake ${CMAKE_VERSION}")
 
--- a/data/shaders/planet.frag
+++ b/data/shaders/planet.frag
@@ -35,6 +35,7 @@ uniform mediump float skyBrightness;
 uniform bool hasAtmosphere;
 uniform bool hasNormalMap;
 uniform bool hasHorizonMap;
+uniform bool texCoordsFromFragment;
 
 uniform int shadowCount;
 uniform highp mat4 shadowData;
@@ -171,6 +172,15 @@ lowp float outgasFactor(in mediump vec3 normal, in highp vec3 lightDir, in mediu
     return opac;
 }
 
+vec4 sampleTexDxDy(sampler2D sampler, vec2 texCoord, vec2 texDx, vec2 texDy)
+{
+#ifdef textureGrad_SUPPORTED
+    if(texCoordsFromFragment)
+        return textureGrad(sampler, texCoord, texDx, texDy);
+#endif
+    return texture2D(sampler, texCoord);
+}
+
 void main()
 {
 #ifndef IS_MOON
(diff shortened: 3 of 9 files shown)


gzotti:
Sounds terrific, even before trying. But yes, a separate switch please for the "lighter" systems (2010 field laptops, Raspberries, ...).

Atque:
Very nice! I'm not sure why, but for some reason the Earthlight is broken:

bild

alex-w:
Hmm... I see no changes when toggle support 3D models for SSO

10110111:
> I see no changes when toggle support 3D models for SSO

Of course, I said in the OP that this switch is currently missing.

alex-w:
> > I see no changes when toggle support 3D models for SSO
> 
> Of course, I said in the OP that this switch is currently missing.

Oh, yeah! My eyes missed this part...

10110111:
OK, so now the toggle between the sphere and the proper mesh should work.

10110111:
The earthshine still seems to look different from `master`.

10110111:
The earthshine is fixed now too.

gzotti:
Seems I'm missing something.

> [   382.970][CRIT] Failed to load model of the Moon: cannot find the model file

The moon is not rendered (stars are visible), but the Nomenclature marks (crater ovals) are there.

---

Ah, reading helps... :-)  unpacked the mesh file, restarted. 

WOW! That lunar rim is simply insane! And now somebody has to observe occultations....

I still should test with RPi3/4 and my smaller field laptops.

10110111:
> Ah, reading helps... :-) unpacked the mesh file, restarted.

Wait, what? Unpacking should happen automatically at CMake stage. Did it not work for you?

gzotti:
I did not investigate at which stage CMake would unpack. After switching to the branch, QtCreator detects new files and triggers CMake to prepare compilation (with previously configured plugins, NLS, PCH, etc.), then I compile and run inside QtCreator. No install step, if that's where unpacking might happen.

gzotti:
It works on my smaller notebook (Intel GPU). However, I noted that when nomenclature (crater labels) is on, they are now further north than the lunar shape. Switching off 3D all is fine. I know a long time ago I added a figure offset from some source (explanatory supplement, Meeus, ...?) because the Moon is not a fully homogeneous sphere, but cannot find it (a comment should give a clue.). Or is your object an off-center sphere?

10110111:
No, it was created as a well-centered reference sphere with the height map
applied. Theoretically this could shift its center of mass.

gzotti:
Found it at end of EphemWrapper::get_lunar_parent_coordsv(). So this patches the position so that a sphere with asymmetrical center of gravity gets centered on the right spot. Not sure how to adjust your Lunar mesh. But what we can do is compare Bayley's Beads (2nd/3rd contact at Total Solar Eclipses) with your rim craters. Likewise, I have a few planet occultations; I hope with accurate-enough timestamps.

10110111:
> But what we can do is compare Bayley's Beads (2nd/3rd contact at Total Solar Eclipses) with your rim craters.

Do you have any well-geopositioned and well-timed photos for these?

gzotti:
I have a few. 1999, 2001, 2006, 2008, 2017, 2024. Start on weekend...  
EDIT: 
2006 just not in focus at begin (2nd contact), and 3rd contact too overexposed. 
2008 I see some possible matches, but also deviation, esp with recorded observation time. 
2017: ingress featureless. Egress very few bright spots.
2024: ingress featureless (or 1/2s too late), egress has 4 beads, should have nothing according to Stellarium. 

Needs more test with shifted/unshifted center.
Anyone else?

gzotti:
I think it's time to declare Raspberry3 no longer supported. It may still work under some circumstances, but it's no real fun, just too little memory. Probably ending with 24.4. On the RPi4 I get otherwise nice performance, but this branch has a shader error. See log:
[log.txt](https://github.com/user-attachments/files/27053633/log.txt)
 
The fallback to just keep the moon as sphere is OK, but if there is no principal obstacle, having the mesh would be great

10110111:
OK, the GLES2 shader issue is now fixed.

And now I also see what shift you mean. It's so huge, but somehow I totally missed it. It's most likely a bug in my refraction code. Without atmosphere everything works OK. I don't remember seeing this when I was developing the refraction on GPU part, this might have been broken during one of the rebases...

gzotti:
Thanks, RPi4 is happy. If this was a GLES2 issue, it would have affected Qt5/ANGLE which I have also not tested yet. 

Funny, Now _I_ cannot reproduce the large shift (pre-last-commit)! refraction makes no difference.

10110111:
> Funny, Now I cannot reproduce the large shift (pre-last-commit)! refraction makes no difference.

Haha, yeah. Scroll two hours back, and you'll see it :)

gzotti:
Just testing on an older i3-5xxx NUC (Ubuntu 24.04). Works well, but also has the refraction/shift problem with atmosphere.

gzotti:
I wonder about the coefficients in your sine variant. The usual Taylor series is x-x³/3!+x^5/5!-... , why do they deviate here?

10110111:
> The usual Taylor series is x-x³/3!+x^5/5!-... , why do they deviate here?

Taylor series, while being a great tool for calculus, is often a bad choice for real-life approximations due to its very non-uniform scaling of error: O(x<sup>N</sup>). The approximation used in my code is close to a minimax polynomial instead, which can be approximated by a truncated Chebyshev series.

See the comparison of the errors between my approximation and the Taylor polynomial of the same order:

sines

Or, in the logarithmic scale, to see the full scale of the error of the Taylor polynomial at the domain edges:

sines-log

gzotti:
Ah right, I forgot about the [-1, 1] limit. Interesting "ugly" series. Just found: https://gist.github.com/publik-void/067f7f2fef32dbe5c27d6e215f824c91. Gives similar but slightly different series. 

For whichever reasons, currently mesh and sphere behave well with/without atmosphere and with Nomenclature (which should not use the new formula, right?)

10110111:
> Gives similar but slightly different series.

Actually, their series are quite interesting in the fact that they optimize the _relative_ error, rather than absolute that mine does. But in any case, we can't expect anything much better than 10<sup>-7</sup> error, since it's about the precision limit of `float`.

> For whichever reasons, currently mesh and sphere behave well with/without atmosphere and with Nomenclature (which should not use the new formula, right?)

Nomenclature is not affected by my change, it continues to apply refraction as it used to do (on the CPU, I suppose). As for the mesh, as I said above, its position error depends on time, I still haven't investigated what exactly is going on here.

10110111:
OK, after several hours of debugging it appears that on Intel the error in asin(sin(x)) for x<80° reaches 0.03°, on AMD about 0.02°, while on NVIDIA, where the issue doesn't exist, it's less than 0.0006°.

Ugly stuff, but so is the reality with Intel and, surprisingly, AMD, which I thought would be on par with NVIDIA quality-wise.

@gzotti What GPUs did you see this problem on?

gzotti:
Ah, I wrote above

> It works on my smaller notebook (Intel GPU). However, I noted ...

That is an Intel N150 CPU. later I wrote I cannot see it any longer, this was my main (Geforce) system. Also the RPi4 was OK. Then the Intel NUC i3-5xxx again showed it. I have another older Intel (Pentium Silver?) field notebook which I have to re-check. Would it help to add one more term to the series?

gzotti:
Weird. On the Pentium Silver N5030 on Win11 (with Intel UHD605, driver 31.0.101.2137) the effect is there but hardly discernible. On the NUC (Ubuntu 24.4, Intel HD5500 (BDW GT2), Mesa 25.2.8-0ubuntu0.24.04.1) it is shouting into my face.
BTW tested Adreno GPU (Windows for ARM): All OK. 

OK, one more candidate, i5-4xxx, Win10, booting up to build... EDIT: HD Graphics 4600, Driver 20.19.15.5171. Works without any visible problem (observed with Lunar altitude near 40° and 70°). 

I would rather not want to build on a 2010 first-gen Core-i5 (Win10, Qt5; test system for weekly/beta installers only), would have to setup environment first. Or you can build an installer for that? (Win/Qt5)

10110111:
> Would it help to add one more term to the series?

No, this time it's `asin` that needs a custom implementation.

> Or you can build an installer for that?

No need now. You've confirmed what I wanted to know.

alex-w:
Any news?

gzotti:
@10110111 is it ready for review, all sine issues fixed, or should we wait with it?

10110111:
It's ready for review

alex-w:
I don't like the performance regress, but I like the effects. Is it possible optional use it - when using 3D models is enabled for example?

10110111:
> Is it possible optional use it - when using 3D models is enabled for example?

It is already optional.

10110111:
Hmm, something goes wrong in GLES mode: the Moon disappears when zoomed in enough...

alex-w (approved):
macOS: everything OK

10110111:
> Hmm, something goes wrong in GLES mode: the Moon disappears when zoomed in enough

Yeah, and additionally there are precision issues when atmosphere is on. Apparently I need to add some `highp` qualifiers somewhere.

alex-w:
macOS 26.5: everything OK

10110111:
So I guess this can be merged now if there're no objections. Just please keep the commits separate (rebase).

alex-w:
It's OK for me for merge

alex-w:
Please do merge by yourself again

10110111:
Will this get into RC1?

alex-w:
> Will this get into RC1?

Yes, I plan to publish RC1 tonight

PR #4839, opened 2026-03-25
Better balance for stellar visibility in twilight and moonlight

### LINKED ISSUE #4291: Objects visible too early in twilight
I know this is a feature that has been mentioned, but not fully addressed in a feature request.

Visibility of stars, the Milky Way, Zodiacal light and DSO's needs to be calibrated, because now we have full stellar visibility at solar altitude -8, and full visibility of MW, ZL and DSO at -12.

AFAIK visibility is linked to luminance in the code, but maybe it would be better so link it strictly to solar altitude. Full visibility would then be reached at solar altitude = -18, according to consensus.

Atque:
### Description
A long lasting problem with Stellarium is unrealistic appearance of stars during early twilight and moonlight. This fix brings better balancing and (to my layman eyes) more realistic stellar appearance, but it's purely empirical and does not bring any better physical model. All stars become fully visible at about -18 degrees solar altitude, instead of the current -8, which I find "reasonably good". See the screenshots below and notice the difference.

Although not many lines have changed (most of the new lines are just comments), do test it thoroughly. I might have missed some special-case where we might need calibration. I've tested for different moon phases, solar eclipse and of course different twilight steps, and it seems fine.

It should work equally well for ShowMySky and Preetham.

Opinions are most welcome, but bear in mind that this is chiefly a visual fix, to better estimate stellar visibility during early twilight stages and full moon.

Fixes #4291 

### Screenshots:
Both screenshots below are at the end of nautical twilight (Sun at -12 degrees).

This branch (greatly reduced stellar visibility):
stellarium-024

`master` branch (full stellar visibility!):
stellarium-025

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Ubuntu 24.04.4 LTS
Qt 6.4.2

## Checklist:

### DIFF
--- a/src/core/modules/MilkyWay.cpp
+++ b/src/core/modules/MilkyWay.cpp
@@ -327,7 +327,7 @@ void main(void)
 	const float nelm = StelCore::luminanceToNELM(drawer->getLightPollutionLuminance());
 	const float bortleIntensity = 1.f+(15.5f-2*nelm)*atmFadeIntensity; // smoothed Bortle index moderated by atmosphere fader.
 
-	const float lum = drawer->surfaceBrightnessToLuminance(12.f+0.15f*bortleIntensity); // was 11.5; Source? How to calibrate the new texture?
+	const float lum = drawer->surfaceBrightnessToLuminance(14.f+0.15f*bortleIntensity); // was 11.5; Source? How to calibrate the new texture?
 
 	// Get the luminance scaled between 0 and 1
 	float aLum =eye->adaptLuminanceScaled(lum*fader->getInterstate());


src/core/modules/MilkyWay.cpp
```diff
@@ -327,7 +327,7 @@ void main(void)
 	const float nelm = StelCore::luminanceToNELM(drawer->getLightPollutionLuminance());
 	const float bortleIntensity = 1.f+(15.5f-2*nelm)*atmFadeIntensity; // smoothed Bortle index moderated by atmosphere fader.
 
-	const float lum = drawer->surfaceBrightnessToLuminance(12.f+0.15f*bortleIntensity); // was 11.5; Source? How to calibrate the new texture?
+	const float lum = drawer->surfaceBrightnessToLuminance(14.f+0.15f*bortleIntensity); // was 11.5; Source? How to calibrate the new texture?
```
10110111, line 330: This comment first appeared in 282e148b68a. Did you check the look in the moonlight?

Atque: Only briefly after that commit. Does it look bad to you?

10110111: I haven't tried this change at all, just looked through the code. I don't have much clear and unpolluted skies here to observe stars in twilight, so I can't check how well this change corresponds to reality.

Atque: This is not really meant to be fully physically correct, only a better solution than what Stellarium currently provides.

gzotti:
Thanks for also taking this up. The so far latest act of balancing was needed urgently and maybe rushed when SMS came just for V1.0 in 2022 and showed different sky brightness values (and IIRC therefore different star appearance) compared to the Preetham/Schaefer sky. This is still an unsolved mystery. Unfortunately my last week under pristine skies was in 2017. It is also always difficult to model "remembered view" against "photographic record" when the latter can just show so much more than you may have seen with imperfect eyes. Also, while there is a bright belt of twilight over the solar azimuth, how many stars do you see 60° above when the eye is not blinded?   For a really long-lasting fix we should (and this is why I have not taken this up yet) step a bit further back and re-investigate actual computed sky brightness values according to our "classical" Schaefer implementation, SMS, and sky/twilight brightness values from the literature. (This has also be requested as extension of the mouse cursor plugin, see #1404.) Then don't forget the effects of different settings for light pollution and extinction. Most users still think in "Bortle Scale", which we have officially declared deprecated but still support in the GUI for convenience, as it is a nice and simple site descriptor. It would be great to have an artificial "Sky Quality Meter" estimate computed from our sky models that could give an equivalent reading to the physical device.

Atque:
Yes indeed, a better solution would of course be per-pixel luminance values, and draw stars and the Milky Way/Zodiacal light according to that. But I was bothered enough, as a resident of 58° N, by midsummer night skies showing just as many stars as December nights, that I thought a "quick fix" could at least be a start. Looking directly into the twilight arc still draws fewer stars than looking at zenith.

Atque:
@alex-w @gzotti @10110111 any further opinions on this?

gzotti:
We had probably too many quick fixes here. Your solution shows too few stars IMO. As said, this needs more systematic studies that involves all parameters but esp comparing the global sky brightness estimates of both sky models against scientific publications (so, we should extend the mouse pointer coordinate plugin and write scripts to sample the background sky brightness in an azimuthal grid compatible to measurements), estimates around ZL and Milky Way visibility, light pollution studies, etc. We have about 8 weeks to study and find better brightness behaviour curves before 26.2 freezes. Are you interested in German papers?

Atque:
That's quite an ambitious goal, especially within 8 weeks, and while I agree to that being a better solution, it's really out of the scope of this PR. Validating such a model will be difficult, since I guess none of us has easy access to Bortle 1 skies? And no, my German is far too rudimentary to read scientific papers or literature, sorry.

My intention with this is to match what can and cannot be seen with the naked eye during twilight and moonlight _better_ than what Stellarium currently displays.

PR #4843, opened 2026-03-28
add support to non-standard regions in culturesList

### LINKED ISSUE #4841: Crash when trying to reach View dialog
Latest master `1837cf1` seems to generate a crash when trying to reach the View dialog (F4). I'm on Ubuntu 24.04.4 LTS and build using Qt 6.4.2.

Bonus, I got a building error, requiring Qt package Qt6SvgWidgets. I guess that'd be necessary to mention in `BUILDING.md`.

Log says:

> [     7.695][WARN] Failed to * find * [  "allthesky"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "allthesky_complete"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "anutan"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "arabic_al-sufi"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "arabic_arabian_peninsula"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "arabic_indigenous"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "armintxe"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "aztec"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "babylonian_mulapin"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "babylonian_seleucid"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "balinese"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "belarusian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "blackfoot"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "boorong"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "bugis"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "chinese"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "chinese_chenzhuo"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "chinese_manchu"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "chinese_song_dynasty"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "chinese_xianglin"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "egyptian_dendera"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "greek_almagest"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "greek_dante"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "greek_farnese"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "greek_leidenAratea"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "hawaiian_starlines"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "indian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "indian_nakshatras"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "inuit"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "irish"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "japanese_moon_stations"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "kamilaroi"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "khoi-san"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "korean"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "lokono"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "macedonian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "mandar"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "maori"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_chinese"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_hlad"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_iau"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_journey_to_the_west"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_rey"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_st"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "modern_sternenkarten"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "mongolian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "mostimportantconstellations"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "myown"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "navajo"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "norse"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "norse_edda"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "northern_andes"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "orbitsimulator"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "romanian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "ruanui_sky_tahiti_and_society_islands"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "ruelle"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "russian_siberian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "sami"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "samoan"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "sardinian"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "seri"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "testkultur"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "tibetan"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "tikuna"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "tongan"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "tukano"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "tupi"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "vanuatu_netwar"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "wcf_standard"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "xhosa"  ] territory file in sky culture directory
[     7.695][WARN] Failed to * find * [  "zulu"  ] territory file in sky culture directory
[     7.697][WARN] "White Cloud Farm (Standard)" : outdated region key found! Please update the region from 'Southern America' to 'South America'.

mRaetz:
Previously, the existence of skycultures with regions that did not follow the UN Geoscheme naming convention caused Stellarium to crash silently. Now, all unknown regions are grouped under a new category in the list (namely “Other”).

### Description

Fixes #4841 

### Type of change

- [ ] Housekeeping

## Checklist:

### DIFF
--- a/src/gui/ViewDialog.cpp
+++ b/src/gui/ViewDialog.cpp
@@ -1489,7 +1489,11 @@ void ViewDialog::populateLists()
 	// find the earliest beginTime of all cultures (needed in initSkyCultureTime)
 	// ---> evaluate it here so we don't need to iterate over all cultures multiple times
 	int globalBeginTime = QDateTime::currentDateTime().date().year();
+#if (QT_VERSION>=QT_VERSION_CHECK(6,0,0))
 	QMultiMapIterator<QString, QString> cultureRegionIt(cultureRegionMap);
+#else
+	QMapIterator<QString, QString> cultureRegionIt(cultureRegionMap);
+#endif
 	cultureRegionIt.toBack();
 	while (cultureRegionIt.hasPrevious())
 	{


src/gui/ViewDialog.cpp
```diff
@@ -1479,6 +1481,10 @@ void ViewDialog::populateLists()
 	{
 		l->addItem(new SeparatorListWidgetItem(region));
 	}
+	if (occuringRegions.size() > occuringRegionsCounter)
+	{
+		l->addItem(new SeparatorListWidgetItem("Other"));
```
10110111: Hmm, is this string user-visible? If so, it should be translatable. Also, the region names seem to not be translatable while they should, at least I have translated regions in the Location dialog but not in the SC exploration section of the View dialog.

mRaetz: Yes it will be visible in the culturesListWidget.
I'm not sure exactly how this is handled. Is it enough to mark the list of regions in ViewDialog.hpp with “q(...)”? Or do the individual regions need to be marked and translated in this way?

10110111: All the visible instances of the strings must appear in `q_()` (except of course for those in the `.ui` file where they should just be translatable, which is the default).
Also, this kind of region name seems to be used as a key in a map, which might break when the user switches languages. This needs to be double-checked after the strings become translatable.

alex-w: <s>The term "Unknown" is better for this category IMHO</s>

P.S. I've read the bug report - please ignore this comment

alex-w:
Note - I see the region named "Middle Africa" in the list of regions in SCM plugin - where is from come?

mRaetz:
As with “South America,” I took “Middle Africa” from the UN documentation. Since there is no skyculture for this region yet, it seems to have gone unnoticed so far. 
{4805ADA0-9372-4135-B2CA-D0020BD0ED2A}


I compared the current UserGuide on the Stellarium website (April 25) with the currently implemented list and found that there are more cases like this. (Antarctica and northern asia are missing. Antarctica is my fault, idk how i omitted that. Northern Asia is some kind of a special case because it is mentioned as an unofficial subregion (officially this region is part of eastern europe). But i can change that as well.)

gzotti:
> Note - I see the region named "Middle Africa" in the list of regions in SCM plugin - where is from come?

It seems to be the same update that changed Southern America to South America: The English names according to UNO M49. I will change the name of current "Central Africa" in the SUG 9.2.1.

mRaetz:
I assume the same warning message should be displayed for Middle/Central Africa as for South/Southern America?

alex-w:
> I assume the same warning message should be displayed for Middle/Central Africa as for South/Southern America?

Apparently yes

P.S. I’ve fixed translation regions in View dialog - please update your branch

mRaetz:
I just tested to see if everything is working properly and noticed that “Middle Africa” is the only region that hasn't been translated. Am I correct in assuming that this is because this region name didn't exist before and therefore doesn't have a translation yet? (And that the problem will “solve itself” in the future?)

alex-w:
> I just tested to see if everything is working properly and noticed that “Middle Africa” is the only region that hasn't been translated. Am I correct in assuming that this is because this region name didn't exist before and therefore doesn't have a translation yet? (And that the problem will “solve itself” in the future?)

Yes and I hope nobody uses or used Central Africa region in own private sky culture

gzotti:
But we are still using different names. We have Australasia which M49 lists as "Australia and New Zealand". There may be other cases (did not check everything, and Regions.hpp was so far not included in the source list.) Can we fix this in 26.1 and maybe write in the SUG "Scheme M49 names as of March 2026"? Or preferably (as we do NOT follow the nesting, "loosely following the M49 scheme". In this case we can keep the shorter term Australasia and even keep Central Africa which sounds infinitely more familiar. 

AND: We had defined a region "Northern Asia" for the Asian parts of the Russian federation. It is IMO ridiculous to place Siberian tribal SCs into Eastern Europe.

mRaetz:
Im on your side regarding the "Northern Asia" region. I think it was stated somewhere that russia as a whole is only part of eastern europe because of statistical convinience (or something like that).
I can revert the change i made for "Central Africa" in skyCultureMgr (warning like South America) if everyone agrees.

(I prefer the short "Australasia" as well)

alex-w:
Hmm… In this case I should revert my fix from commit https://github.com/Stellarium/stellarium/commit/0760a37d3ef8d35327026f70eb0c426c39e9f0e5 and we should use region names as we defined them in `data/regions-geoscheme.tab` file. Plus some notes in the SUG for it.

gzotti:
Ah, there we even define constants. Should these be harmonized in Region.hpp for something?

So, let us consolidate:
- we keep Central Africa (deviating from M49 spelling)
- we use South America (spelling from M49) but tolerate Southern with warning.
- we use Northern Asia (not in M49) for the Asian part of RF. 
- we use "Australasia" instead of M49 "Austalia and New Zealand"

alex-w:
+ "Other region" to all possible fictional SCs

mRaetz:
Oh yeah, should "Other" be an option in the regionComboBox when creating a new skyculture in SCM?

alex-w:
> Oh yeah, should "Other" be an option in the regionComboBox when creating a new skyculture in SCM?

Yes, of course

alex-w:
Hmm... Should I add "Other" as region 25 (or 00) in regions-geoscheme.tab file also?

gzotti:
> Oh yeah, should "Other" be an option in the regionComboBox when creating a new skyculture in SCM?

Yes. If there is popup help, you can write "'Other' indicates 'not on Earth', e.g. fiction. You can edit the file manually later and specify something else."

gzotti:
> Hmm... Should I add "Other" as region 25 in regions-geoscheme.tab file also?

Good question. What is the actual use of the file? Presumably planetary features, but else?

alex-w:
> Good question. What is the actual use of the file? Presumably planetary features, but else?

It uses for creating a list of translatable names + list of regions in Location dialog. I think no need this "extra region" to adding here. For the future we already have defined regions on extraterrestrial planets.

alex-w:
@mRaetz please see #4844 also and please resolve conflict

gzotti:
Right, I see no application for this until somebody wants to fully create a Tolkien world or so. I won't be the one - have not even read the books.

alex-w:
> Right, I see no application for this until somebody wants to fully create a Tolkien world or so. I won't be the one - have not even read the books.

Probably better way will be processing a possible empty region field as "Unknown region"...

mRaetz:
> @mRaetz please see https://github.com/Stellarium/stellarium/issues/4844 also and please resolve conflict

Does that refer to something other than the error here?

gzotti:
Unknown is always an error, which the program can of course survive. An empty region field must trigger a warning, and the SC can be sorted to Other.

alex-w:
> > @mRaetz please see #4844 also and please resolve conflict
> 
> Does that refer to something other than the error here?

Error #4844 touches GUI of SCM plugin only - at least I see no other places for similar problem.

plugins/SkyCultureMaker/src/types/Region.hpp
```diff
@@ -95,7 +100,8 @@ const std::map<RegionType, Region> REGIONS = {
 	{RegionType::AUSTRALASIA, Region("Australasia", "")},
 	{RegionType::MELANESIA, Region("Melanesia", "")},
 	{RegionType::MICRONESIA, Region("Micronesia", "")},
-	{RegionType::POLYNESIA, Region("Polynesia", "")}
+	{RegionType::POLYNESIA, Region("Polynesia", "")},
+	{RegionType::OTHER, Region("Other", "Use this region for skycultures not located on our earth (e.g. ficitonal earth). You can edit the file manually later and specify the region as you like.")}
```
gzotti: typo: ->"fictional earth"

gzotti:
While testing the map interface: Should double-clicking on the Scandinavia polygon linked to Norse-Edda select and load the Norse-Edda skyculture? Currently it selects Modern. 

Hmm, I see:

[  2676.613][INFO] couldn't find any polygon with name [ "Norse (Edda)" ]!

To create screenshots, I dynamically had switched to English. Maybe the map logic does not yet survive a language switch?

Going back to German seems to have solved it. So, it may work for a single user language, but it's something that must be addressed in 26.2 at the latest.

mRaetz:
> Maybe the map logic does not yet survive a language switch?

Yeah, looks like it. I wasn't aware that dynamic language switching existed. 
In theory, the problem should be solvable by either iterating through all the polygons to update their respective IDs and tooltips, or by removing all the polygons from the map and reloading them.
(I will look into it)

10110111:
Or you could use the English names as keys in the map, and only display the translated ones in the GUI.

mRaetz:
Yes. But the tooltip of each polygon displays the translated skyculture name. Either we loop over all polygon and change it manually or give each polygon a connection to a languageChanged signal.

The simplest solution would probably be to repopulate the map using an existing function (`reloadMap`), similar to how the culturesListWidget is repopulated after translation.

alex-w:
Hmm... How to use territory data - or, better, how to test it?

gzotti:
Unfortunately we only have one territory.geojson file in the Norse-Edda SC. I don't see anyone creating 50 polygons today, but hope we can deliver those by 26.3. The program now survives language switch. 

Not sure, the tabs are now at the bottom of the window. Was that intended? It may take more time for previous users to find that. @alex-w @10110111 opinions?

Atque:
> Unfortunately we only have one territory.geojson file in the Norse-Edda SC. I don't see anyone creating 50 polygons today, but hope we can deliver those by 26.3. The program now survives language switch.
> 
> Not sure, the tabs are now at the bottom of the window. Was that intended? It may take more time for previous users to find that. @alex-w @10110111 opinions?

I don't understand; is the region supposed to be zoomed into when selecting the sky culture? Because if so, I can't get it to work on Norse Edda.

mRaetz:
> I can't get it to work on Norse Edda.

When selecting a skyculture in the list the map is _supposed_ to automatically zoom to the bounding box of the polygons of the respective culture.
It sounds like it never worked for you? It is absolutely possible that there are bugs regarding this because all of my tests / usage only happened on Windows 11.

Atque:
I'm on Ubuntu if that matters. I can try Win11 later today, and MacOS too.

gzotti:
When I select the Norse(Edda) in the list, the polygons become visible, map zooms out first, then zooms into the SC region. Nice effect IMO, thanks! What is tricky is the date range. Before clicking on the name, the polygons may be invisible as they only show up when the time slider below the map is within 800-1400.

alex-w:
I moved subtabs to bottom for consistency of other parts of the GUI

alex-w:
> When I select the Norse(Edda) in the list, the polygons become visible, map zooms out first, then zooms into the SC region. 

What is your language on the tests? I see no effect when Russian language is enabled.

gzotti:
I can test on Ubuntu/WSL whether the zoom effect works. 

> What is your language on the tests? I see no effect when Russian language is enabled.

German. And it survived a language switch to English now.

mRaetz:
> What is your language on the tests? I see no effect when Russian language is enabled.

Which OS are you using? I just tested some languages and it worked every time (on Windows 11).

gzotti:
No problems found on WSL/Ubuntu 24.04. (Qt6.4), so it's OK for me now. (Click on SC in list shows polygons, or preselect date and click on polygon selects SC.)  A macOS test would be fine, of course.

A merge conflict is indicated here, but apparently squash&merge may work.

mRaetz:
_(I don't think that's the problem here, but to clarify:)_
The map is supposed to only automatically zoom in if the corresponding item in the list hasn't been selected before.

alex-w:
> Which OS are you using? I just tested some languages and it worked every time (on Windows 11).

Linux

gzotti:
After asking Google's AI for a GeoJSON (which worked, but with quite simple borders), I found open and public-domain  country boundaries defined in GeoJSON at https://datahub.io/core/geo-countries. It's one big file and showing national borders only, not cultural ones, which actually goes against the usual non-national-borders policy. I have however extracted Romania, Belarus, Mongolia. Not sure about the dates, I used 1945 for Romania and 1991 for the others. Of course the cultures are older. Shall I commit those files to master just so we have a few more polygons, and should we add more from there? (It's moderately easy to copy whole features from there. Some entries use multipolygons, though (e.g. Indonesia, and other archipels of course: Haawaii [needs targeted extraction from USA], Samoa, Tonga, ...), this has not been implemented yet. (->26.2? The GeoJSON actually would also support point features to mark regional capitals or other important places.)

10110111:
> When I select the Norse(Edda) in the list, the polygons become visible, map zooms out first, then zooms into the SC region.

I also don't get any such effect on Ubuntu 22.04. Besides, grepping for `territory.json` doesn't yield any results except these:
```
~/src/stellarium $ grep 'territory\.json' . -rI
./plugins/SkyCultureMaker/src/gui/ScmSkyCultureExportDialog.cpp:        // Export the locations (polygons) of the sky culture to the territory.json file
./po/stellarium/fr.po:msgstr "Échec à l'ouverture en écriture du fichier territory.json."
```
Is the code already there in this PR?

10110111:
Ah, the file has `.geojson` suffix, so confusing. Then the French translation is wrong :)
But anyway, I don't have the effect mentioned, something is missing. Tried in locales `ru`, `en`, `de` with the same results.

gzotti:
Only few days ago I discovered the single territory.json in existence. A short check of the self-defined JSON convinced me (and the rest here) to switch over to official GeoJSON. But there are no translatable items inside. The French is just already outdated.

gzotti:
Trying on my old system, U22.04. Apt cannot find package qt6-svg-dev, therefore cmake fails. @10110111 how did you install it? Or are you on Qt5 there?

alex-w:
> Trying on my old system, U22.04. Apt cannot find package qt6-svg-dev, therefore cmake fails. @10110111 how did you install it? Or are you on Qt5 there?

`sudo apt install libqt6svg6-dev`

gzotti:
Was that renamed? IIRC on U24.04 the names listed in our BUILDING.md are correct.

alex-w:
> Was that renamed? IIRC on U24.04 the names listed in our BUILDING.md are correct.

Yes, names of qt6 packages was changed after releasing qt64

gzotti:
Uh, indeed, it seems on U22.04 we have no SVG display. :-( I see no message in the logfile either.

gzotti:
@alex-w you specified "Linux". That's a wide range of options. IIRC you are on Ubuntu 24. Do you see issues? At the stage of one (or, if you want them, four) geojson visible or not in 26.1, we can still accept it when only old systems show no SVG. Else, based on finding Qt<6.4, the map could be suppressed in 26.2, but this and re-arranging the GUI tabs at runtime is a tedious undertaking for a shrinking user group.

alex-w:
> @alex-w you specified "Linux". That's a wide range of options. IIRC you are on Ubuntu 24. 

Ubuntu 22.04 is still on my home desktop :)

> Do you see issues? At the stage of one (or, if you want them, four) geojson visible or not in 26.1, we can still accept it when only old systems show no SVG. Else, based on finding Qt<6.4, the map could be suppressed in 26.2, but this and re-arranging the GUI tabs at runtime is a tedious undertaking for a shrinking user group.

I see the map and I can zoom it - nothing else.

gzotti:
If U22 is still relevant (actually it is, my old laptop also cannot go further thanks to NVidia abandoning the 5xx), and a fix with older versions of its SVG library cannot be found, maybe the best option then is to hide and ignore the map and display list and text as before on one page. It takes some GUI juggling, though. Is it crucial as of 26.1?

alex-w:
Ideally the list of sky cultures should be on both subtabs IMHO

10110111:
> If U22 is still relevant (actually it is, my old laptop also cannot go further thanks to NVidia abandoning the 5xx), and a fix with older versions of its SVG library cannot be found, maybe the best option then is to hide and ignore the map and display list and text as before on one page.

Is the issue really with Ubuntu's SVG library? I'm using U22 with Qt 6.9.3 built from sources and seeing the same issue.

10110111:
Ah, the culprit is right here:
```
const QString filePath = StelFileMgr::findFile("skyCultures/" + currentCulture + "/territory.geojson");
```
`skyCultures` is not the name of our `skycultures` directory. It passes on Windows, because its filesystems are case insensitive, but fails on Unices.

alex-w:
This branch is crashing in macOS at ViewDialog::populateLists() :-(

gzotti:
>  fails on Unices.

Interesting that the plan9 filesystem in WSL does it (tricking me on U24.4). But great you have found it!

gzotti:
> Ideally the list of sky cultures should be on both subtabs IMHO

I have, months ago, recommended that the one tab should maximize the space for the description text. I still think so.  If we get a flood of complaints, we can change later.

10110111:
> If we get a flood of complaints, we can change later.

OK, I'm complaining preemptively :P

The text tab is the first subtab seen, and if there's no list, it will lead to confusion. I don't think it's a good idea to remove the list from the first subtab. And the second one just naturally requires it, so the list should stay in both.

mRaetz:
> This branch is crashing in macOS at ViewDialog::populateLists() :-(

Oh no. Its hard to test / pinpoint the problem from my perspective but i guess its either about line 1465 and following or when inserting the individual skycultures into the list (if cond on line 1503 and following).

> OK, I'm complaining preemptively :P

Should the list be displayed inside the widget (as it is currently in the mapTab) or outside / next to the tabWidget (see example image)?

{A0B1A177-1489-49C9-9DDB-8F71AF5DF1C0}

alex-w:
> Oh no. Its hard to test / pinpoint the problem from my perspective but i guess its either about line 1465 and following or when inserting the individual skycultures into the list (if cond on line 1503 and following).

I can check it in next few hours, but… why I see code for processing “Other” region as a special case?

10110111:
I guess logically it belongs outside the tabs, since it's common to both.

What we could do to satisfy both requirements of having the SC list visible and having as much space for text as possible is to put these two items into the sides of a splitter, so that the user could choose how much space is taken by one and the other (the default being the same as now, and the state saved to the config as all the other window-related sizes). This would also fix the issue of constant clipping of text in the SC list.

alex-w:
> Should the list be displayed inside the widget (as it is currently in the mapTab) or outside / next to the tabWidget (see example image)?
> 
> {A0B1A177-1489-49C9-9DDB-8F71AF5DF1C0}

Yes, proposed on image solution is good for me

10110111:
Also, I'd like to have something done with the settings, because they take quite a big chunk of the space for text. Maybe also put them into a (horizontal) splitter? And also we could make it scrollable, so that the horizontal size of the window wouldn't be forced to be so big.

Screenshot_2026-03-30_10-06-57

alex-w:
> Also, I'd like to have something done with the settings, because they take quite a big chunk of the space for text. Maybe also put them into a (horizontal) splitter? And also we could make it scrollable, so that the horizontal size of the window wouldn't be forced to be so big.
> 
> Screenshot_2026-03-30_10-06-57

I disagree here

alex-w:
Aargh... I forget about Qt5 specifics :(

gzotti:
> Also, I'd like to have something done with the settings, because they take quite a big chunk of the space for text.

Well, given that we all see that a lot of space is used for the settings, and that we have already 2 tabs, why not have 3 tabs then? Description, settings, locations, all showing the SC list on the left. Or have the settings as click-foldaway so that the description area can be maximized for a better overview. (This is something that we of course could have discussed e.g. in February... Maybe consider for 26.2, and only fix remaining bugs for now)

alex-w:
This is may be strange, but I want to merge this pull request ASAP for preparing RC3 package...

gzotti:
Which is why I said, only fix real bugs now, and please let's discuss big GUI layout changes between releases. If you both want to see the SC list on both tabs, OK for now. When RC3 is out, I need to adapt wording and recreate the screenshots for the SUG.

10110111:
Let's postpone all the GUI tweaks and just make things work for the upcoming release.

10110111:
Not related to this PR, but related to the regions and some recent changes: there's a very suspicious line in `ViewDialog.cpp` in `master`, which may need fixing before the RC3:
```
	for (auto cultureRegionIt = std::prev(cultureRegionMap.cend()), end = std::prev(cultureRegionMap.cbegin()); cultureRegionIt != end; cultureRegionIt--)
```
The application of `std::prev` to a `begin` iterator smells like UB.

10110111:
Ah, I see it's been removed in this PR. Sorry for the noise.

alex-w:
This line was caused crashes on macOS, so, I refactor these block.

gzotti:
Now after merge I see it again: Just after opening the map page, I cannot zoom in with mouse wheel, although I can shift the map around. Only selecting Norse-Edda, the map switches years and then zooms in to Scandinavia. @mRaetz any idea?

Oh wow, after second start, it works. Is there a config.ini key involved?

mRaetz:
> Is there a config.ini key involved?

Not that I know of. I can only imagine that there's some kind of issue with the QGraphicsView functions being used.
(e.g. `transform()`, `scale()`, `mapToScene()`, `viewport()` etc.)

EDIT: i think i found the problem (zoom limit in `scaleView`). I'll test my hypothesis first and report back after.

It might be too late now, but there's one thing I'd like to point out: 
using `Qt::MatchContains` instead of `Qt::MatchExactly` when finding / inserting the items into the list might cause some skycultures in certain languages to not be added correctly. (introduced in commit [5e81cd0](https://github.com/Stellarium/stellarium/commit/5e81cd0628a4f6152e0d10f12fae01d31054cf78))

example: french vs. german (french list is wrong)

Image

10110111:
I still think that the use of localized names as keys is quite unwise and should be improved on in later releases.

mRaetz:
The zoom issue originates from a part I wrote at the beginning of the development process (with pretty naive logic). It’s very easy to fix (just change a threshold), but I made a few more changes to make it work better. (#4846)

PR #4813, opened 2026-03-11
Gamepad Controller Support for Stellarium Remote Control

kutaibaa-akraa:
### Technical Report: Gamepad Controller Module for Stellarium Remote Control Plugin

---

**Note: This report has been revised for scientific accuracy, specifically updated to reflect the pivotal architectural shift from a static SDL database approach to the standards-based W3C Gamepad API.**

This report details the comprehensive gamepad support integrated into the Stellarium Remote Control plugin. The module represents a fundamental architectural shift from a model relying on an external database to one that is built entirely on the browser's native web standards, resulting in a solution that is far more robust, transparent, and maintainable.

#### 1. The Core Architectural Pivot: From a Static Database to the Browser's Dynamic Standard
The single most critical decision in this project was the deliberate move away from the **SDL Game Controller Database** in favor of the **W3C Gamepad API**.

*   **The Problem and the Solution:** The initial, well-intentioned approach attempted to use a community-sourced database (designed for native, compiled applications) within a web browser. This was inherently fragile, unmaintainable, and incorrectly predicted how different operating systems would enumerate controller axes. The correct path was to leverage the `navigator.getGamepads()` API. The browser vendors (Google, Mozilla, Microsoft) have already solved the complex problem of low-level hardware interaction and provide a standardized, unified data stream to JavaScript. By trusting this stream, the module became instantly compatible with any controller recognized by the host operating system, with zero manual configuration.
*   **Real-World Proof:** The validity of this new approach was spectacularly proven during the review process. A user reported that the right joystick's vertical axis wasn't working on Linux. The old code contained a hardcoded assumption (`axes.rightY = 4;`) that was incorrect for his setup. With the new "Live Input" diagnostic panel, the raw browser data immediately showed that the correct axis was `3`. This eliminated guesswork and allowed for an instant, precise fix. This was the moment the new philosophy's superiority became undeniable.

#### 2. Module Capabilities and Features (Based on the New Architecture)
The module is built from several integrated subsystems:

| Feature | Technical Description | User Benefit |
| :--- | :--- | :--- |
| **Automatic Device Identification** | The `gpcontroller.js` module parses the `id` string provided by the browser to extract Vendor ID and Product ID, matching them against an internal list of common controllers (Sony, Microsoft, Nintendo). | Instead of a cryptic technical ID, the UI displays a user-friendly name like "Sony PlayStation 4 Controller (v1)", making device selection and profile management clear. |
[...]

PR #4849, opened 2026-04-01
Restore NO_GUI compilation

gzotti:
### Description

For several years now (really since 2010?) the non-Widget-GUI compilation was not maintained. While not useful in general setups as sole installation in a system, maybe as purely pre-configured RemoteControlled or scripted setups it may still have its value.

Fixes # (issue)

### Type of change

- [x] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: Geforce (but irrelevant)
## Checklist:


gzotti:
Shouldn't we rename the STELLARIUM_GUI_MODE options from (Standard/None) to (QWidgets/None)? Just in case someone writes a new "Standard"...

10110111:
> Shouldn't we rename the STELLARIUM_GUI_MODE options from (Standard/None) to (QWidgets/None)? Just in case someone writes a new "Standard"...

Well, someone could write a new one using QWidgets too. If we get a new GUI that becomes the new standard, it can just be called Modern or something.

gzotti:
Hmm, replacing QWidgets with QWidgets is nothing I expect to happen... but OK, it needs *anything else* to happen. Just that I need to write this name into several CMakefiles, so an earlier decision saves later re-edits.

gzotti:
Also this mode needs the QWidget library because of the ProgressBar. For now it should be done. Purpose is a (slightly generalized) specialized installation that may become more useful with #4813. Docs are in BUILDING.md.

gzotti:
@alex-w IMO this should go in before we evaluate the very interesting RemoteControl GUI changes. Any comments or reservations?

gzotti:
@alex-w  Any objections here?

10110111:
Should we maybe add a CI instance to make sure this doesn't break unnoticed later?

alex-w:
> Should we maybe add a CI instance to make sure this doesn't break unnoticed later?

I've added a CI instance and I see troubles in plugins :(

alex-w (changes requested):
Need fix the compilation with `-DSTELLARIUM_GUI_MODE=None` option

gzotti:
Thanks for setting this up. Must see what fails here. 
Given that this still needs QWidgets, maybe the QTextWidgets for InfoText should also be re-activated later. (we wrote text directly to screen in earlier times, with that panel as fallback when text rendering was broken on some systems.)

gzotti:
Funny observations. It seems MSVC ignores missing includes, while gcc aborts. Seems to work finally. Squash into one commit or keep separate (rebasing the last fixes into the appropriate earlier commits)?

gzotti:
Any further comments?

alex-w (approved):

alex-w:
Probably squash & merge will be better way here

Issue #4400, opened 2025-07-01
AstroCalc ephemeris; label only the first day of month or year


Atque:
In many ephemerides visualized on the web, only the first day of each month, or perhaps only the first of each year for long period objects, are drawn. This also helps in reducing clutter.

Look at this ephemeris of 6 Hebe from a Swedish website covering the whole of 2025:
![Image](https://github.com/user-attachments/assets/be3d9b65-e34c-4beb-9382-41aa1f269621)

Here, only months are labled. It could be drawn as 1/1, 1/2, 1/3 etc (which I'd prefer). or simply the Roman number of month. For years, simply the year would suffice, e.g. Pluto positions from 1992 to 2025, labled one per year.

gzotti:
May be personal preference, but I know an almanac with regular days intervals. Easier for interpolation, and visually also more consistent. This may be added as one of two display modes, but it's not my high priority.

Atque:
> May be personal preference, but I know an almanac with regular days intervals. Easier for interpolation, and visually also more consistent. This may be added as one of two display modes, but it's not my high priority.

Interesting, I'm not aware of such calendar. Is it readable for laymen? This FR is simply for visualizing long-term ephemerides without too much clutter.

gzotti:
If you read German: https://web.astroverein.at/beobachten/almanach
The system now has configurable regular intervals. Adding "first of month" now may require a loop duplication with calendar handling. Yet more code to maintain... Is it worth? Our curve also does not show the brightness change, there are just too many possibilities what different programs can show.

Issue #4137, opened 2025-02-16
astrocal ephemeris: solar altitude step


axd1967:
**Is your feature request related to a problem? Please describe.**
I'd like to know how an SSB moves at specific depressions of the Sun, rather than at fixed time steps/deltas.

E.g I'd like to know the horizontal coordinates of a SSB over successive days for a solar altitude of e.g. -9 degrees.

**Describe the solution you'd like**
Include a choice that uses the current solar twilight finder altitude as a stepper.

**Describe alternatives you've considered**
None

**Additional context**

Atque:
Such feature would be very useful for planning to observe Mercury.

Atque:
Might also be extended to "minutes after sunset/before sunrise".

Issue #2829, opened 2022-11-05
"Adaptive" ephemeris markers


Atque:
I'm not sure about the terminology here, but I sometimes find the ephemeris markers too dense AND too sparse at the same time when showing an ephemeris, because they are currently only showing a fixed interval. What I'm wishing for, is for some kind of "adaptive" markers. See the screenshot below, of comet C/1990 Levy.

![stellarium-149](https://user-images.githubusercontent.com/37028368/200116156-d4becc0e-137a-4278-a040-536475d2ec03.png)

You can see how the markers are too dense near the standstill/retrograde, and too sparse when the comet "gains speed" later in the year. The current solution makes it hard to create and visualize ephemerides for certain bodies.

The website [in-the-sky.org](https://in-the-sky.org/findercharts3.php?id=P5) has ephemeris plotting that really shows what I wish for, see the screenshot below for Mercury:

![mercury_ephem](https://user-images.githubusercontent.com/37028368/200116256-d313065a-316d-4858-a32a-01aa592f1f86.PNG)
Note how the labels show different time intervals, as not to collide with other labels.

It would also be suitable if the markers were adaptive to zooming as well. I.e more density when zooming in.

gzotti:
I only know (printed) ephemerides with fixed steps. This immediately shows where the object runs slower or faster. Adapting label density could be useful.

Atque:
Yes, label density is what I mean, not ephemeris steps. So you can calculate eg Mercurys positions from 2022-01-01 to 2022-12-31 with one-day steps, but showing labels as not to make them collide with one another.

PR #4850, opened 2026-04-02
Ephemeris visualization improvements

### LINKED ISSUE #1315: A few requests for ephemeris visualization (AstroCalc)

Atque:
This multi-fix introduces a few new features to the ephemeris tab of the AstroCalc:

1. Letting users label ephemeris markers in any combination of year, month, days, hours, minutes and seconds by corresponding checkboxes
2. Introduce anti-clutter for both markers and labels, with configurable pixel limit
3. Letting users label only the first of each month and having the label written as three-letter-abbreviation of the month's name
4. The most complex feature: introduce a new time step "Sun at altitude", which lets users plot ephemerides of objects at a specific solar altitude, either in evenings (after noon) or mornings (before noon). It should work even for Arctic and Antarctic locations where the Sun doesn't meet conditions specified. In such cases, no ephemeris marker is drawn for that date. (Note that horizontal coordinates need to be checked for this feature to work as intended!)

Fixes #1315 (partially), #2829, #4137 and #4400

### Screenshots:
The extended ephemeris extra settings window (with ephemeris in the background labeling only the first of each month):
stellarium-034

Venus at solar altitude = -6 during this spring's evening from my location (also showing the new time step "Sun at altitude"):
stellarium-036

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Ubuntu 24.04.4 LTS
Qt 6.4.2.

## Checklist:


gzotti:
YAY! Another long-standing plan of mine, esp. the second picture. 
In the GUI, you could exchange the checkbox by a Combo to switch Morning/Evening. A bit more work, but prettier IMHO.

Atque:
Not a bad idea, I'll see if I can add that.

Atque:
I've added a new time step, which is very specific but I find it interesting; opposition of planets. This is available for the five outer planets (Mars -> Neptune).

stellarium-038

The screenshot shows Saturn oppositions over 29 years. For it to make sense, you need to select the planet of interest both in the ephemeris list AND the combobox "Opposition of planet".

Atque:
@alex-w @gzotti @10110111 Have you given this a look?

gzotti:
Give us a weekend! :-) (We have daytime obligations.)

alex-w:
I'll test it tonight (no time for testing in this day at work :( )

alex-w (approved):

alex-w:
This is maybe strange, but options for new time steps are not intuitive, please add notes for both into SUG

Atque:
Yeah I thought adding a note on them might be helpful. I'll do it tonight.

I guess I should update the screenshots as well? Or does it suffice with a new one?

guide/ch_interface.tex
```diff
@@ -1348,6 +1376,20 @@ \subsection{The Ephemeris Tab}
 In consequence, e.g.\ magnitudes may be affected by atmospheric absorption and may show unexpected values. 
 Remember to switch off atmosphere etc.\ to create extinction free geocentric mean positions as found in almanachs.
 
+Two time steps \newFeature{26.2} require a bit of care to display well. The \emph{Sun at altitude} displays
+an object's position when the Sun is at a specific altitude relative to the horizon (only values ≤ 0° are valid).
+The value can be changed in the \emph{Extra options} window (Fig.~\ref{fig:gui:AstroCalc:Ephemeris:SunAtAltitude}).
+Using it as intented requires the ephemeris to be drawn in horizontal coordinates!
```
10110111: `s/intented/intended/` I guess?

guide/ch_interface.tex
```diff
@@ -1348,6 +1376,20 @@ \subsection{The Ephemeris Tab}
...
+
+The other time step that requires some care is the \emph{Planet at opposition}. This is useful for displaying where
+and when a planet will be at its opposition for a given year in relation to background stars. For it to make sense, 
+the planet in question needs to be selected both in the ephemeris list and in the \emph{Extra options} window 
+(under the title "Opposition of planet", Fig.~\ref{fig:gui:AstroCalc:Ephemeris:PlanetOpposition}).
```
10110111: TeX quotes (two backticks to open and two apostrophes to close) should be used instead of the real quotes: ``` ``Opposition of planet''```

src/core/modules/SolarSystem.cpp
```diff
@@ -2220,6 +2283,29 @@ void SolarSystem::fillEphemerisDates()
...
+		{
+			int y, mo, d;
+			StelUtils::getDateFromJulianDay(AstroCalcDialog::EphemerisList[i].objDate + shift, &y, &mo, &d);
+			// JD of the 1st of this month at 0h local
+			double firstJD;
+			StelUtils::getJDFromDate(&firstJD, y, mo, 1, 0, 0, 0.0);
+			const double dist = std::abs(AstroCalcDialog::EphemerisList[i].objDate + shift - firstJD);
+			const int key = y * 100 + mo;
+			if (!closest.contains(key) || dist < closest[key].second)
+				closest[key] = qMakePair(i, dist);
```
10110111, line 2303: `qMakePair` and `QPair` in general are obsolete. In Qt6 `QPair` is just a typedef for `std::pair`.

gzotti: And in Qt5?

guide/ch_interface.tex
Atque: Right, fixed.

guide/ch_interface.tex
Atque: Ah, I wasn't aware of that. Fixed.

src/core/modules/SolarSystem.cpp
Atque: I see. Do you think they will cause compatibility issues?

gzotti: We have numerous occurrences of QPair which has always worked. I cannot say whether writing std::pair is better, it is certainly not prettier. But qMakePair is indeed officially obsolete (https://doc.qt.io/qt-6/qpair-obsolete.html)

guide/ch_interface.tex
```diff
@@ -1334,6 +1334,18 @@ \subsection{The Ephemeris Tab}
 	\label{fig:gui:AstroCalc:Ephemeris:Asteroids}
 \end{figure}
 
+\begin{figure}[p]
+	\centering\includegraphics[width=0.75\textwidth]{astrocalc_dialog_ephemeris_sun_at_altitude.png}
```
gzotti: You can set both images to 0.85 textwidth to reduce whitespace.

guide/ch_interface.tex
```diff
@@ -1348,6 +1360,19 @@ \subsection{The Ephemeris Tab}
 In consequence, e.g.\ magnitudes may be affected by atmospheric absorption and may show unexpected values. 
 Remember to switch off atmosphere etc.\ to create extinction free geocentric mean positions as found in almanachs.
 
+Two time steps \newFeature{26.2} require a bit of care to display well. The \emph{Sun at altitude} displays
+an object's position when the Sun is at a specific altitude relative to the horizon (only values ≤ 0° are valid).
```
gzotti: instead of "values ≤ 0°"  please write classic LaTeX as in "values $ \le 0^\circ$"

gzotti (approved):
Other than my little typesetting remarks, I did not see an issue now. Thanks for this!

10110111:
The "Label date components (when smart dates is off)" creates too much cognitive load. The more natural way would be using a radiobutton, like so:

Date format
⦿ ☑ Year  ☑ Month  ☑ Day  ☑ Hour  ☑ Min  ☑ Sec
⦾ Smart format

10110111:
Why is "Sun at altitude" limited to be ≤0°? Some objects (e.g. Venus) are observable even with the Sun above the horizon.

Atque:
> Why is "Sun at altitude" limited to be ≤0°? Some objects (e.g. Venus) are observable even with the Sun above the horizon.

Sure but it is barely visible then. I never saw a use for positive solar altitudes.

10110111:
> I never saw a use for positive solar altitudes.

I saw it many times. Of course, it's not as trivially visible as in twilight, you need to know where to look. But once you look in the right direction, it's obviously there, visible effortlessly.

In fact, I have a photo I made exactly 6 years ago, on 2020-04-10 at 13:31:34 UTC+3 from Saint Petersburg, Russia. At that point in time the Sun is 37.9° above the horizon.

Atque:
I've also seen Venus when the Sun is quite high in the sky, one time even without actively looking for it. It will be possible later this year when Venus reaches one of its greater brightnesses, about -4.8. But I still don't see any real use for anyone to use it? @alex-w @gzotti what's your take on this?

gzotti:
I have seen Venus in daylight as well (IIRC down to -3.6 (Müller model) in earlier years, haven't tried lately), but the question here is whether "Venus loop for fixed positive Solar altitude" is of any value. The fixed twilight altitudes are critical in revealing ever dimmer objects, whereas for sky brightness at positive altitudes there is no big difference in its effect to object visibility between, say, +30° and +2°  Solar altitude.  As observer I would even ask for an "object at particular azimuth" ephemeris instead when I have a window of observation between high buildings, for example. Or I just have the celestial positions (loop in the eq.grid) and drag the sky so the loop is visible, then I can read the time.   So, for me the zero limit is fully OK and reasonable.

10110111:
> So, for me the zero limit is fully OK and reasonable.

IMO the limit, while being reasonable per your explanation, is useless. It doesn't help avoid mistakes, and simply hinders exploration. Unless, of course, there's some special math that has to be implemented to lift it, in which case I'm OK with keeping it.

gzotti:
Not even sure about special math, there is just no reasonable application I can see, and setting positive altitude rather confuses than helps exploration (what kind of?). I am going so far as to recommend a manually set config.ini entry for these limits to override the "sane" default by the curious minded 0.02% of users. (List in Appendix D.1 but not elsewhere.)
However, what i did not check, @Atque what if you set -10° Solar altitude but in your summer the Sun never goes that deep? will you show object at time when Sun is deepest in due North?

Atque:
No those steps will be simply skipped.

10110111:
Well, a config entry for this is an overkill. Anyway, I won't insist on
this. If we get a real user with this request, we can reconsider it.

Atque:
> The "Label date components (when smart dates is off)" creates too much cognitive load. The more natural way would be using a radiobutton, like so:
> 
> Date format ⦿ ☑ Year ☑ Month ☑ Day ☑ Hour ☑ Min ☑ Sec ⦾ Smart format

Fixed. Now the `Extra options` window looks like this:
radiobuttons

10110111:
> Now the Extra options window looks like this

Yeah, this looks better, but the horizontal padding of the groupbox makes it seem that the radio buttons are subordinate to the "Label each N steps" checkbox. Same for the two other groupboxes. Should maybe the groupboxes all be at the top?

Atque:
Either that or all at the bottom IMHO.

10110111:
Or/and maybe add a special rule for them to zero out their left/right margins in `data/gui/normalStyle.css` (in the `* AstroCalc` section).

Atque:
All at bottom looks like this:
ephemerisgroupbottom

Opinions? @alex-w @gzotti

alex-w:
Please set `layout*Margins` to 0 for `dateFormatVLayout`, `sunAtAltVLayout`, and `oppositionLayout` layouts

gzotti:
The option "Skip drawing unlabeled markers" (please update wording, while we are at it) is now the only sub-suboption. Is the hierarchy really different from "Use small markers..." and the others?

Atque:
Current solution:
updated-window

gzotti:
The wording of "Sun at altitude" box should be rephrased: 

**Settings for "Sun at altitude" time steps**
Solar altitude:...
Twilight...

gzotti:
Thanks. The title line should also be modified IMHO to not repeat phrases.

Atque:
So "Solar altitude (for "Sun at altitude" time step)"?

gzotti:
No, as said above, 

> Settings for "Sun at altitude" time steps

Atque:
Oh God, where are my eyes...

gzotti:
This happens... :-)

On third and forth view, also to @10110111, @alex-w should we call it more accurately (and everywhere!): "Sun at constant altitude"?

(Of course, `Settings for "Sun at constant altitude" time steps` then)

Atque:
Wouldn't "Sun at altitude" be descriptive enough?

gzotti:
There is clearly something missing. "Sun at altitude -7°" (with dynamically adjusted altitude, whatever was set) would be optimal, but maybe "sun at constant altitude" is easier to implement/remember/translate.

alex-w:
Wait! I have a question: we have dialog for custom time step (it can be activated via selecting specific time step) - why two additional time steps are not using same way for their settings? I didn’t checked latest changes in the code, but new time steps are using for visualization only or for computation too?

alex-w:
OK, I see in the code that new time steps are using for computation ephemeris, so, why the settings for it are defined in visualization properties dialog?

Atque:
The visualizations dialog is simply an "extra" dialog, and I thought all new features would be best placed there. Are you suggesting we move "Sun at altitude" and "Opposition of planet" to the same dialog as "Custom time steps"?

I would actually prefer we move the custom time step to "extra options" and drop the old dialog completely.

alex-w:
> The visualizations dialog is simply an "extra" dialog, and I thought all new features would be best placed there. Are you suggesting we move "Sun at altitude" and "Opposition of planet" to the same dialog as "Custom time steps"?
> 
> I would actually prefer we move the custom time step to "extra options" and drop the old way completely.

Dialog for customization of visualization is have not same meaning as dialog for customization the time steps for computation of ephemeris. The latest one can be modified for giving custom options in depend of selected time step.

Atque:
Okay, yes, I can see a point in this. I'll try that and see how it looks.

Atque:
Does this look good?
stellarium-007

alex-w:
You can manipulate visibility of elements in dependencies of selected time step... ;)

P.S. Newly added blocks by the fact has same hierarchy level as an existed one

gzotti:
The uppermost group in the right dialog also should have its appearance aligned with the rest of the dialog.

Atque:
Please try the latest commit. Now only the relevant time settings show up when time step is selected from the list.

customstepsdialog

alex-w:
Good, but headers of newly added blocks need to rephrase now IMHO. What about simple "Sun at altitude" and "Opposition of planet"? @gzotti @10110111 ?

About positive altitude for the Sun - what about daylight observations of the Sun?

Atque:
Hm, yes, solar observation might be interesting. "What days of the year does the Sun reach 10 degrees above the horizon?" and such.

I've raised the limit to +90 degrees and changed the labels of the blocks. Shall I push it?

alex-w:
> Hm, yes, solar observation might be interesting. "What days of the year does the Sun reach 10 degrees above the horizon?" and such.
> 
> I've raised the limit to +90 degrees and changed the labels of the blocks. Shall I push it?

Yes, but do not forget about the User Guide

gzotti:
It should still be "Sun at **fixed** altitude" or "Sun at **constant** altitude" or "Sun at **preset** altitude" (current favourite) or "Sun at **given** altitude" or "Sun at **configured** altitude" (just a but too long) or "Sun at **selected** altitude".  

I still see no real use for "observing situation at preset positive Solar altitudes", but would like to hear of practical applications in the near future. When setting 90 degrees and going to the equator, will it ever find a solution? Or fail when the Sun reaches only 89.7° North or South around equinox?

alex-w:
+1 for _Sun at preset altitude_ as header

Atque:
> Yes, but do not forget about the User Guide

I don't have more time right now, I'll update in the evening.



About the label, I'll fix that too. And about the limits: I'd say let the users discover and explore it themselves, like Ruslan mentioned last week. A quick look at the equator for this year returns no results above approximately 88.2 degrees. The tool was never meant to be used like this though.

gzotti:
> The tool was never meant to be used like this though.

That's the point. User guidance depends on what the tool was designed for. When the tool is extended and meaningful limits lifted, also the labels `"Twilight": Morning/Evening` should be changed. 

Technically, it could be `"Hemisphere": Eastern/Western`

but this looks weird. 

It could even be dynamic, dependent on preset altitude <0. 
Maybe
```
"Sun in sky": Eastern/Western
"Twilight": Morning/Evening 

```

Atque:
We could improve this in as many ways as possibly imaginable, but I really think we should leave such details for future improvements. Should the limit be lowered to 0 again? That was my intention the whole time.

gzotti:
For me 0 is fine, and @10110111 said he doesn't insist on positive values either. If somebody wants positive solar altitudes and can provide a meaningful use case (more than just mathematical possibility), we can change and introduce dynamic labels later. They just should fit together. Alternatively, make this limit silently configurable for the inquisitive mind, with 0 default, documented in the Appendix. Then labels can keep describing the intended twilight case. (GIGO override option)

Atque:
I'll change back to 0 then.

gzotti:
Oh, @alex-w wanted daylight observations of the sun above. What for, seriously?

Atque:
All right I hope we're good to go now.

gzotti:
For me it's fine. -- I just discovered a new issue, but will report separately. (#4874)

axd1967:
@Atque  nice!

PR #4673, opened 2025-11-27
Jpl horizons algorithm delta t

ssimard2504:
This pull request addresses a single issue:

Update the JPL Horizons Delta-T algorithm to be in-line with JPL Horizons application. The revised algorithm is based on initial guidelines generously provided by Jon Giorgini (JPL) and from information available within the JPL Horizons app user manual:

Jon Giorgini
Senior Analyst
Solar System Dynamics Group (392R)
Navigation and Mission Design Section
Jet Propulsion Laboratory
Pasadena, CA USA
Jon.D.Giorgini@jpl.nasa.gov

source : JPL Horizons app user manual: https://ssd.jpl.nasa.gov/horizons/manual.html

### Description

Changes required to three (3) .cpp files: StelCore.cpp, StelUtils.cpp, StelGuiItems.cpp
Changes required to two (2)  .hpp files: EphemWrapper.hpp, StelUtils.hpp, 
Changes required to one test case: testDeltaT.cpp

As documented in issue #4670, the motivation is essentially to resolve unrealistic deltaT values (i.e.. 0.0) outside the time range (3000BC to 1620) of the previous algorithm and to improve accuracy and concordance with JPL Horizons app.

This change requires access to a routine which can compute leap seconds from 1962-10-20 to present. Software from the ERFA library was integrated within Stellarium and build by CMake as a static external library. It still remains to the Stellarium team to decide if this integration is legally valid.

### Screenshots (if appropriate): 
None

### Type of change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update
- [ ] Housekeeping

NOTE Stellarium User Guide needs to be revised to clarify the need for the use of new JPL DE* files and how to properly use Stellarium and interpret Delta-T results.

### How Has This Been Tested?
Section relevant to JPL Horizons algorithm within testDeltaT.cpp has been revised. Reference data were extracted from JPL Horizons app and compared with values obtained by Stellarium. Differences between Stellarium and JPL Horizons are as follow:

from 9999BC to 1962-01-20 : les than +/- 0.1 seconds differences
from 1962-01-20 to 2650-01-25 :  less than +/- 10 microseconds differences
from 2650-01-25 to 9999AD : less than +/- 2 milliseconds differences

[...]

PR #4863, opened 2026-04-08
Extend precession test with a second epoch one second later

IsseiHasegawa:
Extend precession test with a second epoch (JD + 1s)

The paper table only gives vectors for one JD; repeat the PQXY vs Capitaine matrix comparison for JD + 1s without duplicating tabulated constants.

Made-with: Cursor

### Description
- Extends `TestPrecession::testPrecessionAnglesVondrak()` in `src/tests/testPrecession.cpp`.
- After the existing checks that use the paper’s tabulated vectors for the reference JD, the same PQXY vs Capitaine construction is repeated for `JD + 1 second`.
- No new tabulated constants are added; this only strengthens consistency of the two parametrizations on a nearby epoch.
- No new dependencies.

Fixes # (none — add issue number if you opened one)

N/A

### Type of change
- [x] Housekeeping

### How Has This Been Tested?
- Built the `testPrecession` target and ran:
  - `ctest --test-dir build-test --output-on-failure -R testPrecession`
- Result: `testPrecession` passed.

**Test Configuration**:
* Operating system: macOS (darwin 26.x)
* Graphics Card: N/A (unit test only)

## Checklist:

### DIFF
--- a/src/tests/testPrecession.cpp
+++ b/src/tests/testPrecession.cpp
@@ -116,5 +116,33 @@ void TestPrecession::testPrecessionAnglesVondrak()
 	// according to Fig12 in the paper, a few arcseconds of difference between PQXY and Capitaine parametrisation are allowed.
 	QVERIFY2((angleCap-angleRef)*3600.0<6.0, QString("Angle between rotation matrices too different!").toUtf8());
 
-	//TODO: Add more dates and verify this angle difference is limited to what we can see in Fig.12
+	// Same construction one second later: no tabulated paper vectors here, only matrix/angle agreement.
+	const double JulianDay2 = JulianDay + 1.0 / 86400.0;
+	getPrecessionAnglesVondrak(JulianDay2, &epsilon_A, &chi_A, &omega_A, &psi_A);
+	getPrecessionAnglesVondrakPQXYe(-1234.567, &P_A, &Q_A, &X_A, &Y_A, &epsilon_A);
+	getPrecessionAnglesVondrakPQXYe(JulianDay2, &P_A, &Q_A, &X_A, &Y_A, &epsilon_A);
+	Z = sqrt(qMax(1.0 - P_A * P_A - Q_A * Q_A, 0.0));
+	W = X_A * X_A + Y_A * Y_A;
+	VEC2 = -Q_A * C - Z * S;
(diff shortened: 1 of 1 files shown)


src/tests/testPrecession.cpp
```diff
@@ -116,5 +112,33 @@ void TestPrecession::testPrecessionAnglesVondrak()
 	// according to Fig12 in the paper, a few arcseconds of difference between PQXY and Capitaine parametrisation are allowed.
 	QVERIFY2((angleCap-angleRef)*3600.0<6.0, QString("Angle between rotation matrices too different!").toUtf8());
 
-	//TODO: Add more dates and verify this angle difference is limited to what we can see in Fig.12
```
gzotti, line 119: Adding the same dates plus one second IMO does not qualify to remove the intent of the TODO message. Adding a few centuries might do it, but we have no "official" reference values.

src/tests/testPrecession.cpp
```diff
@@ -17,10 +17,6 @@
  * Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
  */
 
-#include <QObject>
```
gzotti, line 20: These includes were probably used by code now only commented away, and will be needed when taking up development in case any doubts in the precession model appear.

gzotti:
TBH I don't see the point here.

gzotti:
@alex-w @10110111 any comments? Do you think this is useful?

10110111:
This +1s delta makes no sense to me. What does 1 second do to precession? Is the change even above noise level? How does it validate the calculation?

As for the comment on addition of extra data, we could try using JPL Horizons to get some vectors for an Earth-surface-based observer on the two poles, subtract the resulting vectors and thus derive the orientation of the axis at any point in time.

gzotti:
Yes, the +1s makes no sense to me either. #4673 deals with possible accuracy extensions. JPL Horizons uses a different precession model, so mas differences can be expected. But the unit tests are mostly here to ensure some change does not inadvertently break the current models.  If we ever extend into JPL's Olson precession, we must write new tests. IMO we can close this.

gzotti:
4 months without any statement why this would be useful. Created with some AI engine increasing human maintenance burden. Thanks, but no thanks.

PR #4867, opened 2026-04-11
Fix nebula names in search results

### LINKED ISSUE #4858: Some Messier objects don't show as icons
### Expected Behaviour

Messier objects are displayed with icons ("hints)". 

### Actual Behaviour

Searching for M3, the globular is listed as galaxy (???), is selected/centered with blue DSO frame, but has no icon. Likewise, M2, M9 and M10 (found as OCl, displayed as GC without icon), ... 

Other odd observations that may be related:

M4 is found as "star" in the F3 dialog, but then displayed correctly as GC with icon. 
M5 is found as "open star cluster" in the F3 dialog, but then displayed correctly as GC with icon. 
M7 is found as "galaxy" in the F3 dialog, but then displayed correctly as OCl with icon. 

### System
* Stellarium version: 26.1
* Graphics Card: Geforce

### Logfile 

maybe relevant:
[     5.240][INFO] Loading DSO data ...
[     5.287][INFO] [...] Stellarium DSO Catalog, version 3.23 (standard edition)
[     5.603][INFO] Loaded 94899 DSO records
[     5.604][INFO] Loading DSO name data ...
[     5.849][INFO] Loaded 1359 / 1388 DSO name records successfully
[     5.849][WARN] No position data for 29 objects: PGC 2798, PGC 2907, PGC 23521, PGC 28759, PGC 29167, PGC 34658, PGC 42102, PGC 44750, PGC 54559, PGC 59117, PGC 59858, PGC 59953, PGC 69018, PGC 69877, PGC 73957, PGC 95597, PGC 100170, PGC 100170, PGC 119230, PGC 1000714, PGC 1436754, PGC 1694462, PGC 2822840, PGC 2826829, PGC 3098124, SNRG 116.9+00.1, SNRG 194.7+00.4, SNRG 201.1+08.3, SNRG 205.0+00.5
[     5.849][INFO] Loading DSO outline data ...
[     5.858][INFO] Loaded 100 DSO outline records successfully
[     5.858][INFO] Loading DSO discovery data ...
[     5.860][INFO] Loaded 183 / 183 DSO discovery records successfully

10110111:
This appears to be a mistake that had existed even before cc1935b39d38527b766fe1a55182ee00ecd21f14. The search was matching partial names, e.g. "M3" search term matched M31. But it returned the search term itself as the result, which doesn't look like the intended output, since the search term is not what was found.

But since the aforementioned commit the results returned include the actual object pointer, which now conflicts with the partial name, since "M3" is also an exact designation of a completely different object.

This patch fixes the behavior by returning the actual name, rather than the search prefix, which now also lets the user finally get the partial-name search results like M31, M34, etc., which were not found even before the problematic commit.

Fixes #4858.

### DIFF
--- a/src/core/modules/NebulaMgr.cpp
+++ b/src/core/modules/NebulaMgr.cpp
@@ -2206,7 +2206,7 @@ QVector<QPair<QString,StelObjectP>> NebulaMgr::listMatchingObjects(const QString
 			QString constws = constw.mid(0, objUpper.size());
 			if (constws.toUpper()==objUpper)
 			{
-				result.append({constws, StelObjectP(n)});
+				result.append({constw, StelObjectP(n)});
 				continue;	// Prevent adding both forms for name
 			}
 			constw = QString("M %1").arg(n->M_nb);
@@ -2227,7 +2227,7 @@ QVector<QPair<QString,StelObjectP>> NebulaMgr::listMatchingObjects(const QString
 			QString constws2 = QString("Melotte%1").arg(n->Mel_nb).mid(0, objUpper.size());
 			if (constws.toUpper()==objUpper || constws2.toUpper()==objUpper)
 			{
-				result.append({constws, StelObjectP(n)});
+				result.append({constw, StelObjectP(n)});
 				continue;	// Prevent adding both forms for name
 			}
 			constw = QString("Mel %1").arg(n->Mel_nb);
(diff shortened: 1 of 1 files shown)


alex-w (approved):
Thanks for the fix!

gzotti (approved):
Thank you, it works!

PR #4878, opened 2026-04-16
Add ability to let SSO's have multiple epochs

### LINKED ISSUE #453: Very wrong positions of minor planets
_Original report by Torsten Merkel:_ https://bugs.launchpad.net/stellarium/+bug/1291005

I tried to observe three of the minor planets during last two weeks (Pallas, Vesta, Ceres) and had problems to find their positions at the night sky using Stellarium 0.12.4 version. With Pallas and Vesta I succeeded at last, but not really satisfied when comparing real constellation with constellation shown in Stellarium. But with Ceres I failed completely last night. At the position where she should have to be seen very clearly there was nothing than black space. Comparing your calculated ephemerides with an annual guide I am using since a couple of years with great success, I realized that Ceres was shown at a position at least

5 DAYS TOO LATE !

Similarly did that happen to Pallas and Vesta positions, but they differred 'only' about two days in movement (too late also), and
Pallas' orbit line is not really what you will observe in the sky, it moves a bit more to the right (smaller values of rectascension I estimate, maybe smaller values of declination too).

Of course this experience was some sort of frustrating for me as I trusted in calculated positions of such well-known objects in the sky even if Stellarium is an amateur developped program with no cost. Now I am irritated a bit whether or
not there is to be trusted in the shown positions of other minor objects.

Maybe there is an unsufficient input of data for ephemerides of minor planets in summary (no disturbing calculation?
wrong equinox or without adjustment to nowadays?). I only can guess, but I dont know. It would be fine if you find
the reason for this problem and it can be solved.

Thank you for listening to my description!

Best regards,

Torsten Merkel

Atque:
### Description
This branch gives asteroids the ability to have more than one epoch of orbital elements loaded, which is useful for studying past or future events with accurate positional data, without having to load a new `ssystem_minor.ini` every few years of simulation. I've created a Python script that queries JPL Horizons for the desired epochs, after which a JSON file is generated. It should be placed in the Stellarium user data folder an be named `asteroid_elements.json`.

If no such file is found, Stellarium falls back on the values given in `ssystem_minor.ini` and no behavioral changes occur. Note that the asteroid in question **must** be present in the `ssystem_minor.ini`. I.e., the Python script cannot be used to add asteroids, only their extended orbital data.

Only power users will likely be interested in the script, but it's provided in `util/fetch_asteroid_elements.py` of this branch in case someone wants to try it out for themself. The prepared example file `asteroid_elements.json` provided in the `data` folder should be moved to the Stellarium user data folder in order to be found. This file covers Ceres, Pallas, Juno and Vesta during 1-year intervals from 1800 to 2100. It is 582 kB in size and generating it takes about 25 seconds. Fetching the first 500 asteroids for a 100-year period, with 6-month interval epochs, yields a file of about 52 MB and takes about 20 minutes generate, for comparison.

Fixes #453 

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Compare Vesta and Ceres positions in Stellarium with photos from the early 80s and 00s (found online), and against the [JPL Horizons System](https://ssd.jpl.nasa.gov/horizons/app.html#/) (make sure aberration and nutation are switched off when comparing coordinates!).

## Checklist:


alex-w (changes requested):
Please move script into util directory

alex-w:
Please check CodeFactor report - it’s look something strange

gzotti:
Sounds terrific! Not sure where, but this also needs reflection in the User Guide. Either chapter scripting or (rather) in the Appendix D.2.4, pointed to by a link in 19.5. It is an advanced topic and has been asked for a few times (and "bad positions" have been reported several times), but with guidance other interested users might fight through it!

Atque:
Should we add a file that's included in the default installer, and if so, what asteroids and epochs should it cover?

gzotti:
Well, we are exploding anyhow. New Moon model, skycultures, translations, ....  Probably 2MB for the 5 which you mentioned from 1800 onwards won't hurt, and with clear instructions on what to do to create files for 2025-2040 (or so), probably advanced users can create the files they actually need. I am less sure about performance if somebody needs 100.000 objects for a century. Make sure the user never has to copy into the program folder, only the user data folder. 

Now it's just a bit weird that these things are separate Python commands, while the SSE plugin which should be in charge of managing orbital elements gets no attention :-(

Atque:
> Well, we are exploding anyhow. New Moon model, skycultures, translations, .... Probably 2MB for the 5 which you mentioned from 1800 onwards won't hurt, and with clear instructions on what to do to create files for 2025-2040 (or so), probably advanced users can create the files they actually need. I am less sure about performance if somebody needs 100.000 objects for a century. Make sure the user never has to copy into the program folder, only the user data folder.
> 
> Now it's just a bit weird that these things are separate Python commands, while the SSE plugin which should be in charge of managing orbital elements gets no attention :-(

Exciting times indeed. I'll do further accuracy tests tonight. I was under the impression that a 6-month interval was necessary for keeping close apparitions of Eros accurate, but I hadn't fixed the AstroCalc ephemeris bug yet, so it showed the incorrect path in the sky. Maybe 1-year interval epochs is perfectly enough, which would cut the file size in half.

For me, the four first asteroids is fine. I was annoyed a few years ago when I couldn't even simulate the circumstances under which Ceres was discovered.

Any asteroid in a default `asteroid_ephemeris.json` must also be included in the default `ssystem_minor.ini`, but I think the default one already covers the best ones.

gzotti:
Yes, I never had fitting elements to check any old photos in some 1920s books showing a minor planet. Good that elements can now be retrieved, and this path is very promising and welcome for me. 

Ans yes, 6-month intervals may be meaningful. As usual, it's also a question which degree of accuracy you aim at. 

Now I wonder: it is well known that Galileo observed Neptune among Jupiter's moons, and Uranus was included in an early-18ct star map (Flamsteed's 34 Tau), are there other pre-1800s "lost stars" that can be identified as minor planets?

Atque:
> Now I wonder: it is well known that Galileo observed Neptune among Jupiter's moons, and Uranus was included in an early-18ct star map (Flamsteed's 34 Tau), are there other pre-1800s "lost stars" that can be identified as minor planets?

Not that I'm aware of, although I wouldn't be surprised if someone had watched Vesta during peak brightness (+5.3) prior to 1800.

alex-w:
Well, the next big step is implementing fetcher, parser and manager data from JPL HORISONS via Solar System Editor plugin apparently...

gzotti:
Yes, this is one thing. Then also an element editor for a selected object (e.g. Comet absolute magnitudes are often wrong from MPC, users have to manually edit ssystem_minor.ini.) I think I have seen commented-out remains of such functionality, but had no time to restore that.

Atque:
> Well, the next big step is implementing fetcher, parser and manager data from JPL HORISONS via Solar System Editor plugin apparently...

I think that's overambitious. This tool is meant to be used very sparsely, perhaps only once. Integrating it to the SSE is overkill IMHO.

We could compare this to the SMS atmosphere model; Stellarium provides a precomputed set of orbital data that will satisfy most users, while power users and researchers will need to consult the associated tool (via command line) to fetch data of interest.

Atque:
Okay, so a file with Ceres, Pallas, Juno and Vesta with 300 year span (1800 - 2100) and the 496 consecutive following asteroids in a 100 year span between 1950 and 2050, with an epoch time of 1 year, yields a file of 25 MB.

Accuracy for 1-year interval seems to be surprisingly good, even for close approaches (Eros in February 1975 is about 16 seconds of arc from JPL's coordinates). Of course, extreme ones like Apophis for years exceeding 2029 will give uncertain results no matter what data is used.

gzotti:
Thanks for more examples. You might give such examples in the user guide chapter. I think creating the elements for N>4 can still be left as exercise for the curious.
I have not checked yet, do you update the epoch, so that the validity test never fails?

Atque:
I've simply bypassed it by setting it to `orbitGood = 0.0` for asteroids with multiple epochs to avoid the annoying warning. Might be wise to keep the message visible outside the interval though.

> I think creating the elements for N>4 can still be left as exercise for the curious.

Should we cover them for the whole JPL coverage (1600 to 2500)?

gzotti:
No, really, most users would be happy about 2020-2040, 1800-2100 should cover the asteroid age and most users' lifetimes. Who needs more can make it now, and this is great.

data/asteroid_elements.json
```diff
@@ -0,0 +1,14744 @@
+{
```
gzotti, line 1: The file name and terminology entries should be changed: "ephemeris" is the day-to-day list of positions. This list here provides Keplerian orbital elements to compute such ephemeris. So, I suggest 
- asteroid_elements.json as filename
- format: stellarium_asteroid_elements

likewise, rename generator to fetch_asteroid_elements.py

data/asteroid_elements.json
```diff
@@ -0,0 +1,14744 @@
...
+    2487703.5284280935,
+    2488070.0
+  ],
+  "asteroids": [
+    {
+      "number": 1,
+      "name": "1 Ceres (A801 AA)",
+      "absolute_magnitude": 3.35,
+      "slope_parameter": 0.12,
+      "epochs": [
```
gzotti: renaming to "elements" would be clearer.

data/asteroid_ephemeris.json
```diff
@@ -0,0 +1,14744 @@
+{
+  "format": "stellarium_asteroid_ephemeris",
+  "version": 1,
+  "generator": "fetch_asteroid_ephemeris.py (Horizons)",
+  "generated": "2026-04-17T22:01:02.006884+00:00",
+  "center_year": 1950.0,
+  "span_years": 300.0,
+  "epoch_jdes": [
```
gzotti: with the elements renamed (below), this key could be just "epochs" or "epochs_jde" (avoid "jdes").

data/asteroid_elements.json
```diff
@@ -0,0 +1,14744 @@
...
+          "ascending_node": 173.1262263037249,
+          "arg_of_pericenter": 310.0607632470913,
+          "mean_anomaly": 185.9792778440233,
+          "mean_motion": 0.2136579530506651,
+          "time_at_pericenter": 2456634.984473383,
+          "pericenter_distance": 2.130301362880638
+        },
+        {
+          "epoch_jde": 2456186.9732,
+
```
gzotti, line 14442: There should be an extra newline at end of file (github indicates the red minus).

10110111:
What is the role of the `epoch_jdes` array? It seems to simply repeat the JDEs present in the `epoch_jde` entries of the `asteroids`, except the latter are rounded to 4 digits after decimal point, while the latter seem to be given in full precision.

gzotti:
Yes, looks like a preview of what's available. Errm, and yes precision should likely be full in both cases.

src/core/modules/MinorPlanet.cpp
```diff
@@ -88,7 +90,15 @@ MinorPlanet::MinorPlanet(const QString& englishName,
...
 MinorPlanet::~MinorPlanet()
 {
-	//Do nothing for the moment
+	// If we installed an interpolated orbit, make sure Planet::orbitPtr no
+	// longer dangles before we delete the object it points to.
+	if (activeEpochOrbit)
+	{
+		if (orbitPtr == activeEpochOrbit)
+			orbitPtr = nullptr;
+		delete activeEpochOrbit;
```
10110111: Is there any reason for manual memory management instead of using a smart pointer? I think `std::unique_ptr` is in order here.

src/core/modules/MinorPlanet.cpp
```diff
@@ -336,3 +346,135 @@ QString MinorPlanet::renderIAUDesignationinHtml(const QString &plainTextName)
...
+			    std::fabs(epochElements[i].epochJDE - epochElements[i-1].epochJDE));
+		if (std::fabs(jde - lastEpochUpdateJDE) < minSpacing * 0.25)
+			return;
+	}
+
+	lastEpochUpdateJDE = jde;
+	const int n = epochElements.size();
+
+	// Find the snapshot whose epoch is nearest to jde.
+	// We do a simple linear scan; with at most ~20 epochs this is trivial.
```
10110111: At most 20 epochs? Don't we have 300 entries in the default `asteroid_ephemeris.json`?

10110111:
> Yes, looks like a preview of what's available.

I don't think it's useful. It takes space in bytes and in scrolling screens, while the exact same information (up to the precision peculiarity) can be obtained by

```bash
grep '"epoch_jde":' data/asteroid_ephemeris.json | sort -u
```

10110111:
The repository contains this big JSON file, which doesn't get installed by CMake, and thus not used. If we are to have it in the repo, it should be

 1. installed, and
 2. automatically used from the installation location, being overridden by the one in the user's data if present.

Atque:
All right, I think I've fixed all your review comments.

src/core/modules/MinorPlanet.cpp
Atque: Sorry, that was from an earlier attempt when I thought 20 epochs would be enough to cover centuries...

src/core/modules/SolarSystem.cpp
```diff
@@ -4607,3 +4637,122 @@ void SolarSystem::enableSurvey(const HipsSurveyP& colors, const HipsSurveyP& nor
 
 	pl->setSurvey(colors, normals, horizons);
 }
+
+// Extended asteroid elements loader
+
+bool SolarSystem::loadExtendedAsteroidEphemeris(const QString& filePath)
```
gzotti: Again please rename to discern day-to-day positions aka ephemeris from the _elements_ needed to compute them.

Atque: It's painfully obvious that I am no astronomer. :)

gzotti: For that you're doing fine. Really useful stuff so far! We're just picky to ourselves and others so that hopefully nobody is confused later. (I am sure there is still a lot to fix...)

Atque: Thanks a lot for those words. I'm really excited about this tool myself (solving a really old "bug" I reported in back in 2017 I think), so I'm eager to get it as good as possible.

Atque:
~~Linux AMD64 Qt6 fails for some reason.~~

Never mind it ran successfully now.

Atque:
Any further opinions on this?

gzotti:
I just wanted to "replay" some early 19th-ct observations. However:

[     5.913][INFO] File ssystem_minor.ini is loaded successfully...
[     5.925][INFO] ExtendedElements: loaded epoch tables for 0 asteroids from "D:\\StelDev\\GIT\\stellarium\\data\\asteroid_elements.json" ( 0 not matched in ssystem_minor.ini)

The file is 583kB. I did not check the file parsing so far. Maybe still an old name element somewhere?

Atque:
Renaming should have applied everywhere. Did you build the latest commit?

Oh, did you copy the asteroid_elements.json to the user folder?

gzotti:
No, should I have?

Now copied file, restarted, same issue. I am at commit 600c7059e7acdaa , your "New generated..." from April 20.

Atque:
This is weird. I also have the same problem now, but I hadn't when I pushed the commit... :(

Atque:
I think I found it. Try to rebuild now. And you shouldn't have to copy anything to the user/data folder.

gzotti:
Indeed now I can illustrate my reading of Schröter's 1805 report of 1802 observations ...  interesting, Ceres and Pallas were assumed and observed much larger, even with atmosphere. Coordinates for the discovery report of "Juno Georgia" are just slightly off, but this may be a measurement error as well. But interesting: those observers use RA in degrees and arcminutes etc.! Even some partition higher than degrees, but the scan is bad:

grafik
This is the discovery RA (AR=Ascensio Recta) of Sept.1, 1804 22:12 mean time (at Lilienthal obs, near Bremen). Should be 9°24'0'', but how do I read the 0² ? Were these collections of 60° units?  -- we need yet another setting :-) 
Another much enjoyable time sink!

gzotti (approved):
Fine for me now!

src/core/modules/SolarSystem.cpp
```diff
@@ -765,6 +769,21 @@ void SolarSystem::loadPlanets()
...
+
+	// Load extended multi-epoch asteroid ephemeris if available.
+	// findFileInAllPaths checks the user data directory first, then the
+	// installation directory, so a user-provided file overrides the bundled one.
+	// If no file is found at all, all asteroids fall back to their standard
+	// single-epoch KeplerOrbit from ssystem_minor.ini without any change in
+	// behaviour, and the Solar System Editor plugin continues to work normally.
+	const QStringList ephemFiles = StelFileMgr::findFileInAllPaths(
+	    "data/asteroid_elements.json",
+	    static_cast<StelFileMgr::Flags>(StelFileMgr::File));
```
10110111: The explicit cast shouldn't be necessary.

10110111:
> Should be 9°24'0'', but how do I read the 0² ?

Looks more like 2°24'0" to me. The two "2" glyphs do look similar. Besides, I don't think the superscripted glyph in 0² is actually a "2": it looks different from the non-superscripted ones. Would be strange if they used a different font for it. This might be a "τ" or a "z" or something else entirely.

BTW, how do you convert the information from Stellarium to the epoch used in this report?

src/core/modules/SolarSystem.cpp
Atque: Right, solved.

gzotti:
> > Should be 9°24'0'', but how do I read the 0² ?
> 
> Looks more like 2°24'0" to me. The two "2" glyphs do look similar. 

They are indeed similar, but the modern computation gives 9, and also a copperplate resembles what we can draw with an ephemeris map. If the "super-degree" shaves off multiples of 7 (which I don't assume), we'd have "whatever o² means" plus 2°24'. 

> Besides, I don't think the superscripted glyph in 0² is actually a "2": it looks different from the non-superscripted ones. Would be strange if they used a different font for it. This might be a "τ" or a "z" or something else entirely.

This is my question. My writing "0²" is in lack of the correct sign. I have no time to investigate now, but it is not uncommon to find other conventions in other eras. Compare with Zodiacal signs plus degree within. I can only assume multiples of 60 are prepended.
 
> BTW, how do you convert the information from Stellarium to the epoch used in this report?

The measured coordinates are "of date" and within arcseconds with Stellarium's result.

alex-w (approved):

10110111 (approved):

gzotti:
The "o²" may indeed be a "o<sup>z</sup>", z for "Zeichen" (sign), sections of 30° (or double-hours) as was customary in ecliptical longitudes. I have yet to find more examples for this use, though.

Issue #4657, opened 2025-11-19
The SCM plugin should add the functionality of naming celestial bodies


sunshuwei:
Obviously, the target audience of the SkyCultureMaker plugin may not necessarily need to understand programming languages, so the entire SkyCulture construction work should be completed within the graphical interface. The SkyCultureMaker plugin should have the functionality to name stars, planets, and deep-sky objects. In the future, it may also be necessary to add the functionality of creating asterisms.

xLPMG:
I understand that this request is to add what is described in the SUG 9.2.7 to the SCM. 
@alex-w @gzotti  Feel free to assign me to this issue and I will work on this. 
The plan is to finally write the (overdue..) SUG section for the SCM after this is done.

> In the future, it may also be necessary to add the functionality of creating asterisms.

I think this should be a separate issue :)

PR #4879, opened 2026-04-17
SCM: add byname to constellation information and adjust format to SC editor

xLPMG:
While going through the SUG, I noticed that the rather new attribute "byname" was not yet available in the SCM. This PR adds this attribute next to the others like native name or pronounciation.

To simplify the code, i have grouped the different name attributes into a single struct. Furthermore, I removed the "references" object from ScmConstellation: it wasnt even possible to set or get this attribute and the SUG does not mention such an attribute for _common_name_ of a constellation either. I suppose this was simply a mistake. (The _common_names_ attribute of the whole SC has a references attribute, not each constellation)

The other UI changes might be debatable: I noticed that the inputs for the constellation information were different than the sky culture information. Such as using an asterisk for marking optional features or showing hints in the text fields. I have added this in the constellation editor to match the sky culture editor. I can revert this if wanted.

constellation-editor

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Tested manually on MacOS 26.4.1

## Checklist:


xLPMG:
Any recommendations about the usage of std::opt for the common name attributes? It seems fine like this but unnecessarily complex... after all we just use `.value_or("")` and the `connect` functions to the QLine elements also become more complex. I suppose now would be a good time to remove the opts and simply use strings and test if they're empty or not :)

10110111:
> Any recommendations about the usage of std::opt for the common name attributes? It seems fine like this but unnecessarily complex... after all we just use .value_or("") and the connect functions to the QLine elements also become more complex. I suppose now would be a good time to remove the opts and simply use strings and test if they're empty or not :)

Is there ever a situation where an empty string being present must be distinguished from no string being present? Like `"byname": ""` or `"english": ""`. There do exist some such entries in the current SCs, but I think these are mistakes that should be corrected.

xLPMG:
> > Any recommendations about the usage of std::opt for the common name attributes? It seems fine like this but unnecessarily complex... after all we just use .value_or("") and the connect functions to the QLine elements also become more complex. I suppose now would be a good time to remove the opts and simply use strings and test if they're empty or not :)
> 
> Is there ever a situation where an empty string being present must be distinguished from no string being present? Like `"byname": ""` or `"english": ""`. There do exist some such entries in the current SCs, but I think these are mistakes that should be corrected.

I think there is no reason to distinguish them. If the SCM is ever able to import sky cultures, then we have two options
1. interpret an empty name as such and write it back to the json if the user doesn't change the value
2. interpret it as a mistake and don't write it back to the json to correct it

To me 2. makes more sense. The sky culture can't be saved when `english` is empty and all others are optional, so I don't see how an empty string could have a specific meaning

gzotti:
Not sure: Can there be situations where English ("english" tag is always needed, empty string should be reported as error) needs no byname, but others may have it (or vice versa, english byname gets translated to the empty string)? Then an empty byname may be useful. (I have not seen enough feedback about these developments so far. Translators could request bynames for their languages/cultures.)
But coming back to the original question, I was not even aware of std::opt and stuff. If an author does not provide e.g. IPA, there should not be an empty IPA string but just no IPA group. Same goes with native, transliteration, pronounce, ... When importing (reading) the SC, any empty strings need not be made into zero-length strings explicitly but may remain empty QString objects.  (If it makes a difference anywhere.)

xLPMG:
Looking at the code (then at least in the current state) it makes no sense to use opt:
```
if (constellationCommonName.byname->isEmpty())
{
constellationCommonName.byname = std::nullopt;
}
```
The user can not enter an empty name, as it gets turned into a nullopt anyway. So at least for saving sky cultures, empty strings are not allowed and are treated the same as if the user didnt input anything

xLPMG:
As said above, with the current code it made no sense to use opts. Furthermore, its not allowed anymore to enter an whitespace-only (blank) english name. It is a required field and putting a blank string violates this in my opinion.

For the other information, whitespace-only strings are still allowed since they are optional. I'm not sure if those should be checked too. My thought was that since they are optional, the user has the freedom to choose any value...

10110111:
> For the other information, whitespace-only strings are still allowed since they are optional.

They should be allowed in the sense of not denying submission, but then trimmed (`QString::trimmed`), and if empty afterwards, handled as not specified. Actually, I think all the names should be trimmed before saving. Leading or trailing whitespace has happened a few times in the SCs included in the source tree, and it was always a mistake.

plugins/SkyCultureMaker/src/types/ConstellationCommonName.hpp
```diff
@@ -0,0 +1,88 @@
...
+#ifndef SCM_CONSTELLATION_COMMON_NAME_HPP
+#define SCM_CONSTELLATION_COMMON_NAME_HPP
+
+#include <QJsonObject>
+#include <QString>
+
+namespace scm
+{
+
+struct ConstellationCommonName
```
gzotti: This struct replicates much of struct StelObject::CulturalName. It may make sense to extend that with the extras from here and avoid too much duplication.

xLPMG: Yes that makes sense here. I have changed it to extend CulturalName. I have also renamed it to ScmCulturalName and added the functionality of exporting references and the "visible" status. This is in preparation for #4657 which I will try to finish next. Just FYI: after this PR and #4657, I will try to tackle the SUG documentation for the SCM plugin.

gzotti (approved):
No further objection, thanks!

PR #4894, opened 2026-05-03
Update ephemeris for Himalia (Jupiter VI), fixes #4708

schenlap:
### Description

Data from JPL Horizons:
https://ssd.jpl.nasa.gov/api/horizons.api?format=text& COMMAND='506'&
EPHEM_TYPE='ELEMENTS'&
CENTER='500@599'&
REF_PLANE='ECLIPTIC'&
OUT_UNITS='KM-S'&
TLIST='2451545.0'

This change improves Himalaya's position compared to the JPL data. Unfortunately, the error is not eliminated by averaging the data. Please provide an assessment of whether this is the best way to obtain new data.

Fixes #4708 

### Screenshots (if appropriate):
before at 2026-5-03 00:00:10 (diff is 0° 53'):
Screenshot_20260503_133210

after at 2026-5-03 00:00:10 (diff is =0° 13'):
Screenshot_20260503_133608

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [x] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: <Linux>
* Graphics Card: Nvidia

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4908, opened 2026-05-07
add util to fetch new orbit data elemets from JPL for ssystem_major, refs #4894

schenlap:
This is a continuation of pr #4894 , currently in draft state.

### Description

I see this now fulfilled:

 - [x]  Update data with REF_PLANE=BODY,
 - [x] check if ephemeris for 2026 with new data equals (at least almost) the JPL Horizon ephemeris, and if it is,
 - [x] write the process in the SUG, subsubsection in Appendix D2.3. 
 - [x] Then maintainers can routinely update orbital elements of all minor moons every 1-2 years (or even automated per release checklist).
 - [ ] I would also like to see a description of the orbital model in the SUG. Do we know where the ascending node of the orbit against the planet equator is counted from? (It can hopefully be derived from the newer rotation model.)

That script should go into utils, and the process described in SUG and/or MAINTAINER_BUSINESS.md

## Checklist:


schenlap:
with 
_python3 compare_ssystem_major_JPL.py ../data/ssystem_major.ini  himalia --max-error 30 --step 14_
it is possible to compare all moons with JPL stellarium (currently 26.1).
result for location vienna:
[ssystem_compare_26.1.pdf](https://github.com/user-attachments/files/27573236/ssystem_compare_26.1.pdf)

I get small errors for moons of  jupiter, saturn, uranus but I have big (nearly the same error) for neptun, mars an pluto moons. I have to investigate further.

image

gzotti:
Interesting work. I see some systematic parallels in 0.35' yearly amplitude, but then also in much larger parallel curves. Any light time effects? And clearly there is some error in the Moon evaluation, this should be within arcseconds or less, not up to 160° degrees off in monthly period. Is this a static Moon position?
Note that the large moons have their own models (not simple Kepler orbits) with which we can simulate mutual eclipses as observed by a few advanced amateurs, so this should qualify as "accurate". For the rest, yes, more to investigate.

alex-w:
JPL coordinates are topocentric or geocentric? Deimos and Phobos use special theories of orbits, but probably his out of valid range now.

schenlap:
I found my error. It was a radian to degree convertion that was unnecessary (claude.ai added it).
There is still an error that @gzotti mentioned. It looks the same for neptun, jupiter and saturn. Same amplitude, same frequency.

image

Now a few condidates are highlighted that clearly need correction with new data. I would suggest all with max error > 2' (maybe 0.35' came from the planet position itself).
image

> JPL coordinates are topocentric or geocentric?

coord@399 + COORD_TYPE=GEODETIC + SITE_COORD = Topocentric (surface point)
I think there is a light missmatch in observer altitude, which is MSL in stellarium und WGT in JPL. In vienna the diff is around 46m. Adding this does not change the error. It is probably negligible.

schenlap:
when setting 
`flag_light_travel_time                    = false` in config.ini the error is 0.086' instead of  0.218'. I think this would be good enough.

But according to JPL Manual it should be true.
`Positions or values (like RA and DEC) which take into account factors that appear to change the target position with respect to the background coordinate system: light-time, the deflection of light due to large or nearby masses, and stellar aberration.
`

image

gzotti:
Appendix F in the User Guide shows typical errors in past and present times. I cannot say where the 0.35' come from, it's larger than what our list . To exclude any topo efect, switch off topographic positions in Stellarium and Horizon. The Neptune graph looks like it has a monthly wobble on top of an annual wobble. Both are weird, but is the JPL position geocentric or (Earth-Moon-) barycentric? What about nutation? aberration? (isn't that annually 22''~0.35'?!)

worachate001:
I think the error of ~0.35' in J2000.0 coordinates may be the aberration that should not be included in that coordinate system, but technical reason makes it unable to exclude in Stellarium. Because of that, better comparison should be using:
* geocentric coordinates (Tools > uncheck 'Topocentric coordinates')
* apparent coordinates (equinox of date)
* TT timescale (not UTC) (Time > Time correction > 'Without correction')
* DE440 Ephemeris (Main > Ephemeris settings > Use DE440)

Editing some values in the python code generated reasonable and much better results.

Moon

Neptune

The short and sharp peaks of error seem to occur around the time of solar conjunction, so it may be the effect of solar gravitational deflection.

gzotti:
Yes, the 0.35' should be aberration (which you can easily switch off in Stellarium). 
For the moon, I added a minuscule positional correction to adjust for "center of figure" (visible) vs. "center of gravity" (ephemeris result). This might be the "mean" effect visible in the Moon graph. The two other wobbles seek explanation, but as we are discussing in another thread, may stem from differences in precession/nutation models used by the two programs. This gets visible in the sub-arcsecond domain that we did not aim for so far, so should not concern us greatly for now.  The light bending effect is interesting, but as said, we just did not go there. 
Has anyone checked the actual implementation of the "generic moon orbit" code which uses all those angular arguments for Kepler orbits? Does JPL provide the right elements for the implemented model? Or does it need another frame transformation? I did not check our generic-moon code against e.g. anything in the Explanatory Supplement (2013) which gives several solutions for particular moons (and could not have checked an error by observing with my non-existing 1m backyard scope...).

worachate001:
> Yes, the 0.35' should be aberration (which you can easily switch off in Stellarium).

Indeed. You can still use J2000 coordinates, but disable the annual aberration in Stellarium. It gives near-zero errors for planets.

schenlap:
wow, that's for the tests, that look good. 

- geocentric coordinates (Tools > uncheck 'Topocentric coordinates')
- apparent coordinates (equinox of date)
- TT timescale (not UTC) (Time > Time correction > 'Without correction')
can not be changed from API (or I did not find it). I could only add aberration disable. But that is ok for me. I will no work on the manual and test new data for every object above 2' max error.

schenlap:
> Editing some values in the python code generated reasonable and much better results.

what? Can you commit your changes in my PR?

gzotti:
You can select deltaT model in the time settings.

Am 14. Mai 2026 10:14:29 MESZ schrieb schenlap ***@***.***>:

schenlap:
> You can select deltaT model in the time settings.

Yes, I am aware of this. I wanted to automate it via the API, but I couldn't find a way.

gzotti:
core.getDeltaTAlgorithm(), core.setDeltaTAlgorithm("WithoutCorrection")

schenlap:
I think the codeFactor errors are fase positives. The only thing I could change would be the extra white spaces in
```
    params = {
        "format":      "json",
        "MAKE_EPHEM":  "'YES'",
        "COMMAND":     f"'{cmd}'",
        "EPHEM_TYPE":  "'OBSERVER'",
        "CENTER":      "'coord@399'",      # geodetic surface observer on Earth
        "COORD_TYPE":  "'GEODETIC'",
        "SITE_COORD":  f"'{lon:.6f},{lat:.6f},{alt_km:.4f}'",
        "START_TIME":  f"'{jd_to_date_str(jd_start)}'",
        "STOP_TIME":   f"'{jd_to_date_str(jd_stop)}'",
        "STEP_SIZE":   f"'{step_days} d'",
        "QUANTITIES":  "'1'",              # RA/Dec (ICRF)
        "ANG_FORMAT":  "'DEG'",
        "CSV_FORMAT":  "'YES'",
        "CAL_FORMAT":  "'BOTH'",
        "TIME_DIGITS": "'MINUTES'",
        "EXTRA_PREC":  "'YES'",
    }
```
but here the whitespaces make it better readable. Please add a note if i should change this.



@worachate001 
> Editing some values in the python code generated reasonable and much better results.

i am only waiting for feedback here, than my PR wold be ready for review.

worachate001:
I just found an issue in new orbital elements. Look at Elara (Jupiter's moon), orbit_Ascendingnode should be updated to the new epoch, but it's not.

schenlap:
I debugged it. It was an issue with lower-case writing of **n**ode. Fix will follow soon

schenlap:
Errors before this PR (with DE440 installed and enabled):

image

Only the red ones are changed (with error greater than 2')

This is the result after this PR:

image

 I can add the complete pdf report before and after this PR if needed.

schenlap:
>  I would also like to see a description of the orbital model in the SUG. Do we know where the ascending node of the orbit against the planet equator is counted from? (It can hopefully be derived from the newer rotation model.)

This should probably be handled in a follow-up by someone more familiar with this area.

gzotti:
An error of one degree for the Moon is definitely wrong data. You can also see there is no before/after change. Are you sure to have topographic correction switched off? Also Mercury, 1/4 arcminute sounds like its topographic parallax. For the rest, whatever exceeds single-digit arcminutes I guess this indicates a principal issue (misunderstanding) with the geometrical orbit model, even if (probably) mean longitude is updated and therefore the respective moon is at least "on the right side of the planet" for now. But as said, the orbital model is legacy code, unfortunately not documented by the original creator. Somebody should look into the topic with the question "given orbital elements from JPL, how to compute planet moon positions". Thinking it over and starting from scratch.

EDIT: We can of course accept the new data that makes the error apparently less severe, even if the orbital model remains wrong. But please try to remove the systematic error (topocentric parallax) in your investigation - just observe from the geocenter.

schenlap:
> An error of one degree for the Moon is definitely wrong data. You can also see there is no before/after change.

I did only change objects with orbit data, not special function like `coord_func=lunar_special`. Our moon ist not changed.

My Settings changed from default:
image


image


image

Is there anything else I should change?
I am still not able to get as good results as https://github.com/Stellarium/stellarium/pull/4908#issuecomment-4437831213 but don't know why.

gzotti:
Did you also set geocentric observer in JPL Horizons?

schenlap:
I changed everything to geocentric. Now moon result is 1 arcmin.

image

alex-w:
I think remarks (`# source: https://arxiv.org/pdf/1803.01907.pdf`) for some objects should be removed to avoid puzzling in future

alex-w (approved):

util/compare_ssystem_major_JPL.py
```diff
@@ -0,0 +1,792 @@
...
+    """
+    cmd = OBJECT_IDS[name.lower()]
+
+    params = {
+        "format":      "json",
+        "MAKE_EPHEM":  "'YES'",
+        "COMMAND":     f"'{cmd}'",
+        "EPHEM_TYPE":  "'OBSERVER'",
+        "CENTER":      "'500@399'",      # geocentric earth
+        "START_TIME":  f"'{jd_to_date_str(jd_start)}'",
```
worachate001:
```suggestion
        "START_TIME":  f"'{jd_to_date_str(jd_start)} TT'",
```

alex-w:
@schenlap please update python code according to PEP8 guide.

schenlap:
> @schenlap please update python code according to PEP8 guide.

done

alex-w (approved):
Thanks!

guide/app_ssystem_ini.tex
```diff
@@ -722,6 +722,117 @@ \subsubsection{How to manually look up a body}
 \noindent The displayed trajectory should start at Voyager~2's encounter with Neptune on August 18, 1989 (JD 2447757.5). 
 The value of \texttt{orbit\_good=} ($2460554.5-2447757.5 = $) \texttt{12797} fulfills this idea. 
 
+\subsubsection{How to check and update moons orbital elements}
+\label{sec:ssystem.ini:MoonsOrbitalElements}
+
```
gzotti, line 728: If you want, you can sign as this section's author with 
\sectionauthor*{Your Real Name}

(Also for the Multiple Epoch section.)

guide/app_ssystem_ini.tex
```diff
@@ -722,6 +722,117 @@ \subsubsection{How to manually look up a body}
...
+Moons orbiting planets are modelled in Stellarium using Keplerian orbits. While this
```
gzotti: The paragraph should start with:

If not handled by dedicated orbital functions, moons...

guide/app_ssystem_ini.tex
```diff
@@ -722,6 +722,117 @@ \subsubsection{How to manually look up a body}
...
+approach is computationally efficient, it does not account for gravitational perturbations
+caused by other planets and the Sun. Over longer time spans these perturbations cause the
+predicted positions to drift noticeably from the true positions, so the orbital elements
+need to be refreshed periodically.
+
+Two Python scripts are provided to support this process. Both scripts require the path to
+\texttt{ssystem\_major.ini} as their first argument.
```
gzotti: To avoid the bad line wrap, write "require as their first argument the path..."

util/compare_ssystem_major_JPL.py
```diff
@@ -0,0 +1,893 @@
...
+            f"{datetime.now(timezone.utc).strftime('%Y-%m-%d %H:%M UTC')}",
+            ha="center", va="center", fontsize=11, color="#444444",
+        )
+        pdf.savefig(fig_t, bbox_inches="tight")
+        plt.close(fig_t)
+
+        # One page per object
+        for name, (dates, errors) in all_results.items():
+            fig, ax = plt.subplots(figsize=(11, 5))
+            make_plot(ax, name, dates, errors)
```
gzotti: It would be nice if the plot title also could also show IAU moon number (from ssystem_maior.ini:[objName]iau_moon_number). Some minor moon names may not ring the bell...

gzotti:
I wonder about the remaining error between major planet positions. Esp. a 2-month period with Saturn error. Maybe as discussed elsewhere just spurious differences with precession models. Else this is a very nice verification contribution. And I am pretty sure the principal orbit geometry for "Keplerian" moons is flawed. But this is for another PR.

gzotti:
> I think remarks (`# source: https://arxiv.org/pdf/1803.01907.pdf`) for some objects should be removed to avoid puzzling in future

I think these comments refer to data entries like B-V, and should therefore stay. But probably another comment or otherwise unused data entry like "orbit_updated=fetch_ssystem_major.py on 2026-06-09" could be added. If this comment/key is found and value starts with fetch_ssystem...", update it as well.

guide/app_ssystem_ini.tex
```diff
@@ -722,6 +722,117 @@ \subsubsection{How to manually look up a body}
...
+  \item Run \texttt{compare\_ssystem\_major\_JPL.py} again for the same object to verify
+        that the new elements reduce the error. Comparing the before and after plots
+        provides a clear picture of the improvement.
+  \item Repeat as needed for additional objects.
+\end{enumerate}
+
+
 \subsection{Multiple Epochs for Minor Planets}
 \label{sec:ssystem.ini:MultiEpoch}
 \indexterm{asteroid!multi-epoch ephemeris}
```
gzotti: Although not exactly in this PR, please also fix those entries: The actual index entries are in rectangular brackets, visible term in braces (can be empty as in here.)

\indexterm[asteroid!multi-epoch ephemeris]{}
\indexterm[JPL Horizons!multi-epoch script]{}

schenlap:
> But probably another comment or otherwise unused data entry like "orbit_updated=fetch_ssystem_major.py on 2026-06-09" could be added. If this comment/key is found and value starts with fetch_ssystem...", update it as well.

this is redundant to orbit_epoch. Do you really what to add a extra date? maybe we should add a note to epoch?
`orbit_Epoch=2461172.5 # fetch_ssystem_major.py on 2026-05-15`

gzotti:
"Real" comments in an ini file are in danger of "vanishing" when written by programs that ignore comments when reading the ini, which is why "unused entries" may be more durable. You are right, epoch alone may be enough, but needs decoding. I don't insist on that, just suggest. @alex-w? @10110111 ?

alex-w:
> > I think remarks (`# source: https://arxiv.org/pdf/1803.01907.pdf`) for some objects should be removed to avoid puzzling in future
> 
> I think these comments refer to data entries like B-V, and should therefore stay. But probably another comment or otherwise unused data entry like "orbit_updated=fetch_ssystem_major.py on 2026-06-09" could be added. If this comment/key is found and value starts with fetch_ssystem...", update it as well.

OK, but comments like `orbit_Epoch=2461172.5 # fetch_ssystem_major.py on 2026-05-15` is enough for me (this file is not updated by robots almost)

gzotti:
Then I hope whichever robot will eventually update will still not mess with documentary comments.

alex-w:
I approve it

gzotti (approved):
OK for me now, thanks!

PR #4912, opened 2026-05-11
SCM: add common_names tab (#4657)

### LINKED ISSUE #4657: The SCM plugin should add the functionality of naming celestial bodies

xLPMG:
This PR adds a "Common Names" tab to the sky culture editor, which lets the user insert cultural names of stars, plansts and DSOs

Fixes #4657 

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Created different sky cultures on my macbook.

## Checklist:

### DIFF
--- a/plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
+++ b/plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
@@ -1242,9 +1242,8 @@ void ScmSkyCultureDialog::cnAddNew()
 	if (!cnCheckObjectExists(key)) return;
 
 	cnEntries.append({key, name});
-	const int newRow = cnEntries.size() - 1;
 	cnRefreshTable();
-	ui->cnEntriesTable->selectRow(newRow);
+	cnClearForm();
 }
 
 void ScmSkyCultureDialog::cnSaveChanges()
--- a/plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
+++ b/plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
@@ -292,7 +292,7 @@ void ScmSkyCultureDialog::createDialogContent()
 	ui->cnVisibleLbl->setEnabled(false);
 	ui->cnRemoveEntryBtn->setEnabled(false);
 	connect(ui->cnObjectTypeCB, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
-	        &ScmSkyCultureDialog::cnUpdateVisibleField);
+	        [this](int i) { cnUpdateVisibleField(static_cast<CnObjectType>(i)); });
 	connect(ui->cnRemoveEntryBtn, &QPushButton::clicked, this, &ScmSkyCultureDialog::cnRemoveEntry);
 	connect(ui->cnUseSelectionBtn, &QPushButton::clicked, this, &ScmSkyCultureDialog::cnUseSelectedObject);
 	connect(ui->cnAddNewBtn, &QPushButton::clicked, this, &ScmSkyCultureDialog::cnAddNew);
@@ -303,7 +303,8 @@ void ScmSkyCultureDialog::createDialogContent()
 	{
 		if (cnEditingRow >= 0) ui->cnSaveChangesBtn->setEnabled(true);
 	};
-	connect(ui->cnObjectTypeCB, QOverload<int>::of(&QComboBox::currentIndexChanged), this, onFormEdited);
+	connect(ui->cnObjectTypeCB, QOverload<int>::of(&QComboBox::currentIndexChanged), this,
+	        [onFormEdited](int) { onFormEdited(); });
 	connect(ui->cnObjectIdLE, &QLineEdit::textChanged, this, onFormEdited);
 	connect(ui->cnEnglishLE, &QLineEdit::textChanged, this, onFormEdited);
 	connect(ui->cnNativeLE, &QLineEdit::textChanged, this, onFormEdited);
@@ -315,7 +316,7 @@ void ScmSkyCultureDialog::createDialogContent()
 	connect(ui->cnVisibleCB, QOverload<int>::of(&QComboBox::currentIndexChanged), this, onFormEdited);
 	connect(ui->cnEntriesTable, &QTableWidget::itemSelectionChanged, this,
 	        &ScmSkyCultureDialog::cnOnTableSelectionChanged);
-	cnUpdateVisibleField(0); // initial placeholder text
+	cnUpdateVisibleField(CnObjectType::Star); // initial placeholder text
 }
 
 void ScmSkyCultureDialog::handleFontChanged()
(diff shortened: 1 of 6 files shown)


xLPMG:
I'm not 100% sure what the most intuitive workflow is. Currently possible:
1. To create a new name: Fill data --> Press "Add Entry" -> repeat
2. To edit a name: click in list on common name, then press "Load selected" to load data or simply fill from scratch -> Press "Save Entry" to override or "Add Entry" to add as a new name

plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
```diff
@@ -914,3 +950,224 @@ void ScmSkyCultureDialog::cancelAddPolygon()
...
+	ui->cnEntriesTable->setRowCount(0);
+	for (const auto &pair : cnEntries)
+	{
+		const int row = ui->cnEntriesTable->rowCount();
+		ui->cnEntriesTable->insertRow(row);
+
+		QString displayKey = pair.first;
+		switch (pair.second.special)
+		{
+		case StelObject::CulturalNameSpecial::Morning: displayKey += QLatin1String(" (morning)"); break;
```
alex-w: The words "morning" and "evening" are not translatable here. Is it correct?

xLPMG: It would make sense to make this translatable :)

alex-w: Please add the context for these strings to avoid troubles for translators

alex-w (approved):
Thank you very much!

10110111:
A button to add the currently selected object would be useful. Otherwise one has to manually type in the HIP id.

Besides, the current behavior of _Add Entry_ → _The "English" field is required_ is inconvenient. The "Load selected" is also quite unintuitive.

What I'd expect is:

1. Add an entry (the currently selected one or empty, depending on what button was clicked).
2. Fill in all the necessary fields, including the identifier (which would also be mandatory along with the English name).
3. Maybe even the type combobox could be one of the necessary fields, rather than a global control.
4. Have it all updated in the table on the fly or have a button to Apply/Accept/Save new contents of input fields.
5. If a copy is really necessary, add a button to Copy.
6. On finishing with the common names, check that
  - Mandatory fields have been filled in
  - There are no duplicate objects in the list

xLPMG:
> A button to add the currently selected object would be useful. Otherwise one has to manually type in the HIP id.
> 
> Besides, the current behavior of _Add Entry_ → _The "English" field is required_ is inconvenient. The "Load selected" is also quite unintuitive.
> 
> What I'd expect is:
> 
> 1. Add an entry (the currently selected one or empty, depending on what button was clicked).
> 2. Fill in all the necessary fields, including the identifier (which would also be mandatory along with the English name).
> 3. Maybe even the type combobox could be one of the necessary fields, rather than a global control.
> 4. Have it all updated in the table on the fly or have a button to Apply/Accept/Save new contents of input fields.
> 5. If a copy is really necessary, add a button to Copy.
> 6. On finishing with the common names, check that
> 
> * Mandatory fields have been filled in
> * There are no duplicate objects in the list

sounds good. this made me realize that I totally forgot about Gaia IDs as well..
I have implemented the checks for mandatory fields and duplicates directly per common name (add new or edit). I think thats more user-friendly than potentially being bombarded with issues after creating many table entries.

Lastly, I'm still unsure where to place the buttons. At the bottom like now or between table and fields?

gzotti:
The background of the Common Names list is too bright IMHO. All other backgrounds are in "dark mode", why is this bright?

Else for me it seems to work. We can keep buttons at bottom, and if users react negatively, be ready to further improve.

xLPMG:
Screenshot 2026-05-22 at 16 49 02
I guess this is not how it looks for you?

Edit: @gzotti I have added the QTableWidget to the style css and set the background color there. Please try again and see if this fixes it. If there is anything else that looks different than the picture above, please let me know or post a screenshot of how it looks for you. Thanks!

gzotti:
Now it looks OK, thanks.

gzotti (approved):
Looks good for me now, thanks.

10110111:
I added LDN 1436 from selection, then switched the combobox to _Planet_, and pressed _Save Changes_. And suddenly I got an object with `NAME LDN 1436`, which of course doesn't exist.

Should the objects be checked maybe?

plugins/SkyCultureMaker/src/gui/ScmSkyCultureDialog.cpp
```diff
@@ -914,3 +968,326 @@ void ScmSkyCultureDialog::cancelAddPolygon()
 {
 	hideAddPolygon();
 }
+
+void ScmSkyCultureDialog::cnUpdateVisibleField(int typeIndex)
+{
+	// typeIndex: 0=Star, 1=Planet, 2=DSO
```
10110111: Why not use an enum for these named numbers?

xLPMG: i guess it was easier for me in the moment to directly use the checkbox indices but I agree that an enum mapping makes this more readable

xLPMG:
> I added LDN 1436 from selection, then switched the combobox to _Planet_, and pressed _Save Changes_. And suddenly I got an object with `NAME LDN 1436`, which of course doesn't exist.
> 
> Should the objects be checked maybe?

Good idea, I didn't even think about this. I added a small warning dialog with the options to cancel saving, save anyway and to also turn off the warning for the current session. I didn't make this a permanent config parameter on purpose, because I dont want the user to be able to turn it off and forget about it.

10110111:
A UX issue for me is the following sequence (I'm just following my intuition here as a new user used to other programs):

 * Select a star
 * Press _Add New_
 * See an error message. Try to figure out how this star can have existed if there is no such entry. Realize that you should have used a button on the opposite side of the window to import the selection.
 * OK, _Use Selected Object_.
 * Add a name
 * Save Changes
 * Argh, I've overwritten my previous entry!

Now, why does this happen?

First, "Add New" does sound like "add the one that doesn't exist yet", with picking the selection as the default. In fact, I think we don't even need the "Use Selected Object", because the worst that could happen if "Add New" does this is that we'll just have to replace the id filled by default. And here I mean that "Add New" will create a new entry and make it current, and only complain that it doesn't have an English name if we try to apply the changes (or switch tabs). When switching tabs, the error dialog may offer to discard the entry and proceed or continue editing.

And second, "Save Changes" gives too few clues as to whether it would create a new entry to save the new data, or it would update the currently selected one. I'd relabel it to something like "Apply to current entry" (feel free to rephrase to a more laconic version, it just needs to be unambiguous).

---

It may be useful to find an example of an unrelated program that has a list with editable entries and see how it handles this. Maybe it has a good UI worthy of imitation.

10110111:
An unrelated issue (not for this PR) with this dialog is that I can't resize it by putting the cursor near the border (at an invisible border). Instead I have to use a square handle in the corner. This makes the window different from all the other Stellarium windows.

And also, for some reason, when I try to move the window by dragging by title bar, it jumps back and forth, with only the average position reflecting the direction of dragging. Again, this differs from the behavior of all the other Stellarium windows.

This is not to be fixed in this PR, just something to keep in mind for future improvements (or shall I file an issue report for this?).

alex-w:
> This is not to be fixed in this PR, just something to keep in mind for future improvements (or shall I file an issue report for this?).

Please fill separate bug report for tracking

gzotti:
This is caused by both this plugin and OnlineQueries using the dialog class that uses separate windows, so both can be moved to another screen. In 2022 I did not manage to bring the scale (EDIT: resize) function to the edge, so I added the tiny square, but I dont object changing this if you know wat to do.

xLPMG:
> First, "Add New" does sound like "add the one that doesn't exist yet", with picking the selection as the default. In fact, I think we don't even need the "Use Selected Object", because the worst that could happen if "Add New" does this is that we'll just have to replace the id filled by default. And here I mean that "Add New" will create a new entry and make it current, and only complain that it doesn't have an English name if we try to apply the changes (or switch tabs). When switching tabs, the error dialog may offer to discard the entry and proceed or continue editing.

I think "Use Selected Object" is a really useful button, as it not only tells the user that selections can be used in this dialog, it also allows the user to edit/replace data. Furthermore, I'm really no fan of the "switching tabs" thing, as it kind of forces the user to stay in a tab or finish/discard their progress. It feels like an unnecessary constraint.

> And second, "Save Changes" gives too few clues as to whether it would create a new entry to save the new data, or it would update the currently selected one. I'd relabel it to something like "Apply to current entry" (feel free to rephrase to a more laconic version, it just needs to be unambiguous).

Fair point, I have renamed the buttons to "Save as new entry", "Update selected entry" and "remove selected entry". Furthermore, the "Copy Data From Selected Object" button is now placed directly above the input fields. Its easier to find and I hope this conveys the feeling that it only fills the fields instead of adding some new entry.
Also, after "Save as new entry", the form is cleared and no entry is selected in the table. Therefore you can't accidentally override an existing entry anymore after having added one.

In general, I think your intuition is that the table at the top is like a "working area" to you where you add partial results and edit them etc. However, the idea was that the table area represents the current, validated information that will end up in the json. The line edits below are the "working area" where data can be entered and then either used to be added as a new entry in the table or to update an existing one (with direct validation).

> It may be useful to find an example of an unrelated program that has a list with editable entries and see how it handles this. Maybe it has a good UI worthy of imitation.

Sadly I couldn't find such an example. All apps I know either use 1) an extra dialog that opens for inputting data with "Save" and "Cancel" or 2) an interactive table where all entries are editable and then one save button for the whole table (which is impractical here, we have too many fields).

What do you think about these changes? Does it still feel unintuitive or error-prone?

10110111 (approved):
Yeah, now it's much more understandable. Might be a good idea to select the English name after filling (otherwise the user might think that it's part of the combined id), or maybe not even fill it.

But I'm not sure. @alex-w, @gzotti?

Generally though, this LGTM.

gzotti:
What is the "Visible" entry doing now? I tried to add an entry, but "Visible" seems disabled. 
Ah - it's for planets only. I don't see tooltips in this window. Should there be tooltips? Then it should explain "Planets only". Alternatively, should the "Visible" and its dropdown just be hidden when Type&ne;Planet?

xLPMG:
> What is the "Visible" entry doing now? I tried to add an entry, but "Visible" seems disabled. Ah - it's for planets only. I don't see tooltips in this window. Should there be tooltips? Then it should explain "Planets only". Alternatively, should the "Visible" and its dropdown just be hidden when Type≠Planet?

I'm not a big fan of tooltips because I think the UI itself should be intuitive enough. For longer explanations, tooltips can of course be very useful though.

Screenshot 2026-06-01 at 18 22 22

What if we directly write "Planets only" in the UI? does it feel out of place?

If any of you have a recommendation and prefer hiding it or a tooltip, then of course I can implement it that way. I'm sure you know your userbase much better than I do ;)

gzotti:
Yes, like this it's ok for me. But maybe rather `Visible (planets only):` Hiding full lines may also feel too surprising for some. @10110111 , @alex-w  ?

10110111:
> But maybe rather Visible (planets only): Hiding full lines may also feel too surprising for some.

Agreed.

alex-w:
> But maybe rather `Visible (planets only):` Hiding full lines may also feel too surprising for some.

It’s ok for me

xLPMG:
done!

PR #4915, opened 2026-05-14
Fix tracked object jumps while zooming with viewport offset

### LINKED ISSUE #4887: Display glitches in 26.1
Yay for new versions! 😊 

I’ve discovered a display glitch in 26.1. Zooming the view under certain conditions produces a bumping effect rather than the expected smooth zoom.

Conditions:
An object must be selected and centred
Vertical viewport offset ≠ 0 (bigger is better to notice the effect)

Appears when using repeated presses of page up/down to zoom in/out. Holding the key down or using the mouse wheel to zoom doesn’t show the effect.

When page up/down key is pressed, the starfield experiences a small “bump” up and down in addition to the expected zoom. The bump seems proportional to the offset – 1° produces no visible bump, 50° produces a big one (I usually use -20° and the bump is visible).

See this video for a demo (the glitch appears at 21 seconds): [https://cleanshot.com/share/W0WsdnwG](https://cleanshot.com/share/W0WsdnwG)
The video has 6 sections, each with a few presses of page up/down to zoom in/out:
00:00 No object selected / Vertical viewport offset 0° (no bumping visible)
00:06 Object is selected / Vertical viewport offset 0° (no bumping visible)
00:11 No object selected / Vertical viewport offset 20° (no bumping visible)
00:17 Object is selected / Vertical viewport offset 20° (no bumping visible)
00:21 Object is centred / Object is selected / Vertical viewport offset 20° (bumping visible)
00:29 Object is centred / Object is selected / Vertical viewport offset 50° (more bumping visible)

My system:
MacBook Pro (14-inch, 2023)
Chip Apple M2 Pro
macOS 26.4.1
Stellarium v26.1

_Originally posted by @Monque27 in https://github.com/Stellarium/stellarium/discussions/4854#discussioncomment-16639907_

Symmetricity:
## AI Disclosure

This pull request was generated with AI assistance and reviewed locally before submission.

## Summary

This fixes a viewport-offset tracking issue where a selected/tracked object could visibly jump during manual zoom when a non-zero vertical viewport offset was active.

The change reorders an existing `updateVisionVector(deltaTime)` call so tracking compensation runs after manual FOV changes for the frame. No extra per-frame work is added.

Fixes #4887.

## Pre-Fix Behavior

With tracking enabled and vertical viewport offset set, manual zoom could render a frame using the new FOV while the tracking aim still reflected the previous FOV. The selected object visibly shifted during zoom.

In the local scripted repro, the tracked object moved by up to 61 px during repeated zoom pulses with a 20% vertical viewport offset.

## Post-Fix Behavior

Tracking compensation uses the FOV that will be rendered for the frame. The selected object remains stable during manual zoom with the same viewport offset.

In the same local scripted repro, the tracked object stayed within normal pixel-rounding variance during zoom.

## Performance Impact

None expected. This only reorders an existing per-frame update; it does not add another call, loop, allocation, projection calculation, or render step.

## Visual Evidence

Before/after video:

https://github.com/Symmetricity/stellarium/releases/download/pr-4915-evidence/target1-viewport-offset-before-after.mp4

Preview:

before-after preview

## Reproduction

Also ran a local scripted Xvfb repro that:

1. selects and tracks Vega,
2. applies a 20% vertical viewport offset,
3. performs repeated manual zoom-in/zoom-out pulses,
4. logs the selected object's screen position during each pulse.

Pre-fix: max observed jump during zoom was 61 px.

Post-fix: movement stayed within pixel-rounding variance.

The local repro script is not included in this PR because it is a standalone GUI/Xvfb automation script rather than a project test. I can share it if useful.

## Testing

Ran locally:

```bash
cmake --build <build-dir> --parallel 2
ctest --test-dir <build-dir> --output-on-failure
git diff --check
```

### DIFF
--- a/src/core/StelMovementMgr.cpp
+++ b/src/core/StelMovementMgr.cpp
@@ -1027,8 +1027,6 @@ void StelMovementMgr::setFOVDeg(float fov)
 // Increment/decrement smoothly the vision field and position
 void StelMovementMgr::updateMotion(double deltaTime)
 {
-	updateVisionVector(deltaTime);
-
 	const StelProjectorP proj = core->getProjection(StelCore::FrameJ2000);
 	// the more it is zoomed, the lower the moving speed is (in angle)
 	double depl=keyMoveSpeed*deltaTime*1000*currentFov;
@@ -1083,11 +1081,13 @@ void StelMovementMgr::updateMotion(double deltaTime)
 		deltaAlt = rateY*M_PI_180*deltaTime;
 	}
 
+	updateVisionVector(deltaTime);
 	panView(deltaAz, deltaAlt);
 	updateAutoZoom(deltaTime);
 }
 
-// called at begin of updateMotion()
+// Called after manual FOV changes so viewport-offset tracking uses the FOV
+// that will be rendered for this frame.
 void StelMovementMgr::updateVisionVector(double deltaTime)
 {
 	// Specialized setups cannot use this functionality!


10110111 (approved):
Seems to make sense.

The only state changed in `updateMotion` and used in `updateVisionVector` is `currentFov` (and only for the problematic offset), and it's the one that needs to be up-to-date (or it will fail to match the calculations on the next frame in the absense of user input), so it makes sense to first update it, and then call the method.

The projector `proj` is only used for `deltaZoom`, which uses the `currentFov` supplied, and this member isn't changed in `updateVisionVector` AFAICT.

alex-w (approved):
Thanks!

Issue #3731, opened 2024-04-30
Display eclipse map directly within the program


Atque:
As per #3719, we can export a raster image of the shadow path of the Moon during a solar eclipse. It would be nice if this could be shown directly within the program, possibly as a tab in the "Eclipses" section of the AstroCalc.

10110111:
I initially thought this could go into the _Location_ dialog, so that one could use the eclipse chart as a guide for choosing where to go.

But OTOH, in this case the user may want to also automatically adjust time, so that by default switching to the new location would go to maximal eclipse at that geographical point. So maybe indeed it should be somewhere in the AstroCalc dialog—or in a dialog that could be opened by some button in AstroCalc, like "Show on a map".

gzotti:
Given screen space restrictions, a separate panel may be best. The Location dialog is already very complicated. This new panel may also show the optimal location, or set location/(optionally!) best time on clicking the map. Some functionality will still have to be taken from LocationDialog. Subclass?

Issue #3525, opened 2023-11-26
No User Guide chapter for Lens Distortion plugin


gzotti:
### Expected Behaviour

All plugins are described in the User Guide

### Actual Behaviour

The new Lens Distortion Estimator Plugin is not described in the User Guide. I assume it is of interest to astrophotographers or just photographers who need a way to find lens parameters for other applications (which? Name a few examples. ). 

Please describe the plugin and what it can be used for (practical examples) in the User Guide. 1-2 images, 2-3 pages. 

On a different idea, once a match has been accomplished, can this be used for an entry in nebulae\...\textures.json? It may require exporting the undistorted image. Or to find angle/coordinate parameters when creating a SkyImage? Or to link a piece of sky art to three stars? (To use as constellation art in skycultures.)

10110111:
> can this be used for an entry in nebulae...\textures.json? It may require exporting the undistorted image.

Yes, may be a good use. Do you have an example of a distorted input image?

> Or to find angle/coordinate parameters when creating a SkyImage?

Do you mean the parameters for image placement? If so, yes, these can also be added to the generator.

> Or to link a piece of sky art to three stars? (To use as constellation art in skycultures.)

Not sure why you need this plugin for such a task. Doesn't it simply require you knowing pixel position of the stars?

As for distortion, the constellation art images seem to be quite fuzzy regarding the position of the stars. Is there any example where you'd need a precise placement of an image with respect to the stars?

gzotti:
> > can this be used for an entry in nebulae...\textures.json? It may require exporting the undistorted image.
> 
> Yes, may be a good use. Do you have an example of a distorted input image?

You can use any self-made DSO image. I can PM you some. Probably lens distortion is less of an issue to telescope images, but image placement in J2000 coordinates would be: find image corner coordinates. 

 
> > Or to find angle/coordinate parameters when creating a SkyImage?
> 
> Do you mean the parameters for image placement? If so, yes, these can also be added to the generator.

Yes, there are arguments for either position&angle, or four corner coordinates. When showing some large image insert in a planetarium show, the four corners would be fine to avoid large distortion on the dome. In this use case, output should be alt-azimuthal coordinates, but J2000 coordinates should also be available for "real" sky insertions. 
 
> > Or to link a piece of sky art to three stars? (To use as constellation art in skycultures.)
> 
> Not sure why you need this plugin for such a task. Doesn't it simply require you knowing pixel position of the stars?

[...]

PR #4924, opened 2026-05-20
New plugin: Visibility map

### LINKED ISSUE #945: show earth view in map projection
I want to see where an object is in the local zenith.

A checkbox in the projection section 
image

that merges the Earth photo
image

so that I can see where e.g. the Sun is on a given moment (here on a star globe, lowest prio layer):

image

https://www.gaisma.com/en/location/gent.html
image

**Additional context**

### LINKED ISSUE #3268: Calculate time for observing an object
Currently, there is a function that returns the time when a star rises above the horizon, respectively 0.00001° above the horizon.
Although such stars or objects can't be observed, we believe to observe an object it should at least be around 10-15° above the horizon, depending on the altitude/light pollution where the observer is. Also, we'd like to count in the sun's position (should be at least 6° below the horizon, so it's dark enough to observe.)

Any idea how we could alter the Vec4d StelObject::getRTSTime to do the above? 

Would changing this 
`ho = altitude*M_PI_180;`

to `ho = 6*M_PI_180;` (6 degrees) be suitable to calculate an object above 6 degrees?

Also, we'd have to check if the sun is settled, which means it'd be quite a lot of iterations. Any clue on how to solve it?

Atque:
### Description
This new plugin gives users the ability to study visibility maps across several fields. It consists of four main tabs.

The first tab _Twilight zones_ shows a map of the Earth with the three twilight stage limits (civil, nautical and astronomical) drawn across the map with appropriate shades and colored border lines. It also adds symbols for where the Sun and the Moon are at zenith.

The second tab _Rise/set isolines_ shows times for sunrise/sunset and moonrise/moonset as isolines across the map. Added are also the start and end times for civil, nautical and astronomical twilight. Areas from which the Sun does not set or rise are shaded blue (polar night) or yellow (polar day). Areas where the Moon doesn't rise or set are also shaded. Users are able to select UTC offset through GUI.

The third tab _Object visibility_ lets users select a star or a DSO, and have a few lines and symbols drawn across the map: latitude of zenith transit, latitude of circumpolar limit, latitude of "good" visibility (customizable, defaulting to +5 degrees above horizon at culmination) and latitude limit of visibility. This feature was heavily inspired by the _The 49 brightest stars in the night sky – when and where can we see them?_ on https://astro-geo-gis.com/ found [here](https://astro-geo-gis.com/the-49-brightest-stars-in-the-night-sky-when-and-where-can-we-see-them/).

A bonus for this tab, is the "best visibility calendar". It shows the meridian transit of the star compared to observer latitude. For northern observers such as yours truly, this is helpful when trying to determine when e.g. Antares is best visible in the bright spring sky.

The fourth tab _Daylight length_ lets the user see how much time the Sun is above the horizon on a given date. It also contains the time the Sun is within civil, nautical and astronomical twilight. Areas from which the Sun never rises or sets are shaded blue (polar night) and yellow (polar day).

Twilight zones:
bild

Rise/set isolines:
bild

Object visibility map:
bild

Object visibility "best visibility calendar":
bild

Daylight length tab:
bild

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Fedora 44 KDE Plasma
* Qt version: 6.11.1

## Checklist:


alex-w (changes requested):
Please update `po/stellarium/POTFILES.in` file also and fix compilation with Qt5.

P.S. Great plugin, thanks!

gzotti:
Very impressive, thank you!

Observations/suggestions:
- twilight tab: sync with Stellarium. I expect changing the time in Stellarium's time panel would change twilights. Nope. Should be connected with StelCore::timeSyncOccurred()
- Tab "Rise/set isolines": Narrow the window. Lines are mirrored top/bottom. Maybe just distort the view to whatever window size there is?
- Maybe add a flag/checkbox to "set location on click into map". 
- Tab "Object visibility": When displaying twilight map, the two right buttons "Current location"/"World" seem to be not doing anything and could be disabled/hidden. They are useful in the world map mode, of course. 
- In this tab, when selecting a new object, the map should auto-update. (connect to StelObjectMgr::selectedObjectChanged())
- In this tab, when selecting a planet, no new maps are generated. The map for the year is surely more complicated, but could be interesting. If that is not planned, some message ("words of solace") should be displayed over an empty map. 
- Tab "Daylight length": Maybe also link to current Stellarium time. (Bidirectional)
- Finally, please add an "Export map" button to each tab, with size selection, file type etc (or use file type from the screenshot settings, or filename system panel extension dropdown).
- And of course, section 13.5 in the User Guide...

General question: should we combine this with the current location panel and its old low-res map?

alex-w:
@Atque what about license and source of the map?

alex-w:
> latitude of "good" visibility (+5 degrees above horizon at culmination) 

Why 5 degrees? Standard altitude for good visibility is 18 degrees (for photometry of course)

P.S. What about adding line for 18 degrees too for astrophotographers?

Atque:
> Tab "Rise/set isolines": Narrow the window. Lines are mirrored top/bottom. Maybe just distort the view to whatever window size there is?

You mean the lines are mirrored between northern and southern hemisphere? They should be that way. Although I noticed that time zone does not update when selecting a new location. It reloads upon changing day.

> In this tab, when selecting a new object, the map should auto-update. (connect to StelObjectMgr::selectedObjectChanged())

I would want to have that optional in that case.

> In this tab, when selecting a planet, no new maps are generated. The map for the year is surely more complicated, but could be interesting. If that is not planned, some message ("words of solace") should be displayed over an empty map.

Solar system objects are excluded for this reason. Yes, a message to inform users might be sane.

gzotti:
> > latitude of "good" visibility (+5 degrees above horizon at culmination)
> 
> Why 5 degrees? Standard altitude for good visibility is 18 degrees (for photometry of course)
> 
> P.S. What about adding line for 18 degrees too for astrophotographers?

Obvious solution: make it configurable.

Atque:
> @Atque what about license and source of the map?

It comes from the [Natural Earth website](https://www.naturalearthdata.com/downloads/10m-raster-data/10m-natural-earth-1/), scaled and converted to JPG by me. I don't know if we can use it freely though.


> > latitude of "good" visibility (+5 degrees above horizon at culmination)
> 
> Why 5 degrees? Standard altitude for good visibility is 18 degrees (for photometry of course)

That's directly from the Astro-Geo-GIS source mentioned in the PR description.

alex-w:
> > @Atque what about license and source of the map?
> 
> It comes from the [Natural Earth website](https://www.naturalearthdata.com/downloads/10m-raster-data/10m-natural-earth-1/), scaled and converted to JPG by me. I don't know if we can use it freely though.

These maps are under public domain - we can use it! Well, the `CREDITS.md` file should be updated too.

alex-w:
> General question: should we combine this with the current location panel and its old low-res map?

Are you want to use small copy of new map for Location dialog as replacement of the current map? The new map is under public domain license and we can freely use it within planetarium.

gzotti:
> > General question: should we combine this with the current location panel and its old low-res map?
> 
> Are you want to use small copy of new map for Location dialog as replacement of the current map? The new map is under public domain license and we can freely use it within planetarium.

No, I mean combining all interface elements from the Location dialog onto another tab of this plugin and then replace Location Dialog by this plugin (which would become a main feature) in order to have a Location dialog with better map where users can scale up to hit location from map more clearly, and then have twilight zones on the map. Having twilight zones as from the ancient X11 sunclock was a very old feature request. And #3731 can probably also combined.

gzotti:
> > Tab "Rise/set isolines": Narrow the window. Lines are mirrored top/bottom. Maybe just distort the view to whatever window size there is?
> 
> You mean the lines are mirrored between northern and southern hemisphere? They should be that way. Although I noticed that time zone does not update when selecting a new location. It reloads upon changing day.

No. Narrow the window until it is higher than wide. Then observe that above the north rim and below the south rim of the map area there are mirrored copies of the lines. Highly decorative, but meaningless. As remedy, consider distorting the earth map to always fill the window in any geometry, or only show the map and suppress everything on the overlaid drawing that sticks out of the map area.

Atque:
Ah yes now I see it. Really ugly, I must fix that.

alex-w:
> > Are you want to use small copy of new map for Location dialog as replacement of the current map? The new map is under public domain license and we can freely use it within planetarium.
> 
> No, I mean combining all interface elements from the Location dialog onto another tab of this plugin and then replace Location Dialog by this plugin (which would become a main feature) in order to have a Location dialog with better map where users can scale up to hit location from map more clearly, and then have twilight zones on the map. 

This is very bad idea - plugin can be optionally loaded, Location dialog is part of main program. Many features of this plugin shouldn’t be in Location dialog. I disagree with replacing the Location dialog by the plugin.

> Having twilight zones as from the ancient X11 sunclock was a very old feature request.

This plugin is solving it.

> And #3731 can probably also combined.

As an extension of AstroCalc/Eclipses tool - yes, of course.

gzotti:
> > No, I mean combining all interface elements from the Location dialog onto another tab of this plugin and then replace Location Dialog by this plugin (which would become a main feature) in order to have a Location dialog with better map where users can scale up to hit location from map more clearly, and then have twilight zones on the map.
> 
> This is very bad idea - plugin can be optionally loaded, Location dialog is part of main program. Many features of this plugin shouldn’t be in Location dialog. I disagree with replacing the Location dialog by the plugin.

Why not? It would be "the dialog with earth map". Tab1=Location Panel replacement with current functionality but zoomable map. Tab2..n whatever new feature this plugin delivers. As of now, there are separate "two dialogs with different earth maps".

> > Having twilight zones as from the ancient X11 sunclock was a very old feature request.
> 
> This plugin is solving it.

Which is why I mentioned it. 
 
> > And #3731 can probably also combined.
> 
> As an extension of AstroCalc/Eclipses tool - yes, of course.

But probably only when it's not a plugin...

10110111:
Why do we need one more Earth map? Can't we just reuse the Blue Marble we already have in the Location dialog (see `data/gui/miscWorldMap.jpg`)? Or replace the existing one with this one, if we must (though the old one has less compression artifacts, so I'd prefer to keep it).

Atque:
I thought we needed a brighter one for this plugin.

Also, we could use a PNG for this, but it's 11 MB, and I believe that'd be too large?

10110111:
> Also, we could use a PNG for this, but it's 11 MB, and I believe that'd be too large?

Yeah, I think it is. What we could do is reuse this map for the Location dialog, but bringing it to the resolution of the current map (or even higher, like 8k×4k), and saving with a better JPEG quality (I suppose the original source is much higher resolution, right?). Then we could remove the old 1.6M file.

This would also be an improvement for #4925, in which currently contrast is a bit lacking in the sea areas.

P.S.: It would also make sense to move the Location dialog's map to `textures` dir, so that it's commonly accessible.
P.P.S.: The current map is also used as the background for exported eclipse maps, need to check how they would look with the brighter one.

10110111:
Some notes:

 * The Visibility Map dialog has a strange gray border and title, unlike all the other dialogs
 * The lines should use antialiasing. The isolines seem to already have it, but the twilight borders don't.
 * The vertical and horizontal grid lines might look better without antialiasing (they are never aliased, but their position between pixels sometimes makes their width twice the "nice" one).
 * Rise/set isolines' labels are clipped when GUI font size is larger than default
 * All labels and symbols should be scaled with GUI font (see `StelApp::getScreenScale`, but you need to define `StelApp::getGuiScale` in a similar way — or maybe just use `StelApp::getInstance().guiFontSizeRatio()` if HighDPI scaling is automatically applied by Qt).
 * I haven't seen the SVG map anywhere in the GUI — how is it used?

10110111:
> General question: should we combine this with the current location panel and its old low-res map?

Its current map is higher-res than that of this plugin :)

gzotti:
A high quality JPG may be better than a PNG, size-wise. What I like here is the zoom option, a functionality I would very much welcome in the location panel, kept separate or not. Just guessing, but probably the cities in the zoomed-in map are from the svg.

10110111:
> What I like here is the zoom option, a functionality I would very much welcome in the location panel, kept separate or not.

Agreed, I'm planning to implement it after #4925.

Atque:
> I haven't seen the SVG map anywhere in the GUI — how is it used?

Sorry, that was a leftover from when I thought SVG would work better. It's deleted now.

Atque:
Should we go for an 8k res Earth texture? IIRC, older graphics cards struggle with such large image files.

10110111:
> Should we go for an 8k res Earth texture? IIRC, older graphics cards struggle with such large image files.

Not sure how QPainter handles this, but `StelTexture` downsamples textures if they are out of OpenGL limits.

10110111:
Besides, current Location dialog already has a texture larger than 4k×2k, and nobody seems to have complained.

10110111:
> Should we go for an 8k res Earth texture?

As for this question, I guess you could first try to see how much size it would take — when compressed to a quality where even a very detailed visual study wouldn't notice any artifacts.

Atque:
Ah, I see. Well I'll try and see how it looks. Any idea how large of a texture we could accept?

Oh, and shall I go ahead and place it in the `textures` directory?

alex-w:
> The Visibility Map dialog has a strange gray border and title, unlike all the other dialogs

+1

@Atque, is it possible to unificate the look of this plugin with other plugins?

alex-w:
> Oh, and shall I go ahead and place it in the `textures` directory?

When it used in plugin only - no.

Atque:
> > The Visibility Map dialog has a strange gray border and title, unlike all the other dialogs
> 
> +1
> 
> @Atque, is it possible to unificate the look of this plugin with other plugins?

I'm not sure I understand what you mean. Is this the grey border you're referring to?

bild

alex-w:
> I'm not sure I understand what you mean. Is this the grey border you're referring to?

Please open any other plugin and check the look & feel

Atque:
An 8k JPG texture of the Earth is 7.3 MB in size and looks "okay" when zoomed in max. Looks grainy when zoomed out however, I'm no good at image compression. :(

Atque:
> > I'm not sure I understand what you mean. Is this the grey border you're referring to?
> 
> Please open any other plugin and check the look & feel

Ah, right. I see it now.

Atque:
Hm, my plugin doesn't really solve any of those. I could add a night texture of Earth, but I'm not sure were to get one for free. Plus blending it in twilight could be tricky (when are lights lit?).

alex-w:
> Hm, my plugin doesn't really solve any of those. I could add a night texture of Earth, but I'm not sure were to get one for free. Plus blending it in twilight could be tricky (when are lights lit?).

Partially - without night lights from cities, but now users can see day and night parts of the planet.

gzotti:
The smallest SBC will struggle with 8k, even just loading and auto-reducing. We lost RPi3 already.  ANGLE is OpenGL ES, probably only 2k-safe, but here the auto-reducing should work. I can test these platforms not before June, though.

10110111:
> Looks grainy when zoomed out however

Well, this is likely because of the way you're zooming out: possibly, by nearest-neighbour downsampling. If you do it like `MapWidget` does, you should get better results.

> The smallest SBC will struggle with 8k, even just loading and auto-reducing.

OK, what about 5400×2700? Our `miscWorldMap.jpg` already has this size.

Actually, never mind. We can unify these textures after this PR gets merged.

10110111:
> I'm not sure I understand what you mean. Is this the grey border you're referring to?

Yes, and in my case (Ubuntu 22.04) the title bar is also grey.

Atque:
> > I'm not sure I understand what you mean. Is this the grey border you're referring to?
> 
> 
> 
> Yes, and in my case (Ubuntu 22.04) the title bar is also grey.

Could you try the latest commit? Should be fixed now.

10110111:
> Could you try the latest commit? Should be fixed now.

Yeah, this is fixed now.

Some more notes:

 * I see you've added antialiasing to some lines, but for some reason they (particularly isolines and twilight borders) still look jagged, although with the blur typical for antialiasing. Do you see this too? Is it maybe due to rounding of coordinates (like using `QPoint` instead of `QPointF`)? See the screenshot:

Screenshot_2026-05-22_13-17-00

 * Opening the dialog for the first time after Stellarium launch takes 2 seconds on my laptop (Ubuntu 22.04 with Intel UHD Graphics 620 / Intel Core i5-8265U). This is too long. If the computations take so much time, they may better be done in a separate thread, or, if it requires being in the main thread, then a limited number of steps per frame.

 * Zooming/panning the isolines map happens at 3.4 FPS instead of the base 28 FPS when idle. At the very least, you could pan a ready-made image instead of recalculating it every time. And maybe keep the already computed data to only be rendered on zooming.

 * Switching language doesn't update labels
 * Object visibility lines sometimes don't reach all the way to the western side of the map (see the screenshot below). Can be reproduced by panning east-west. Same for daylight length lines on the eastern side.

Screenshot_2026-05-22_12-59-25

Atque:
> * Opening the dialog for the first time after Stellarium launch takes 2 seconds on my laptop (Ubuntu 22.04 with Intel UHD Graphics 620 / Intel Core i5-8265U). This is too long. If the computations take so much time, they may better be done in a separate thread, or, if it requires being in the main thread, then a limited number of steps per frame.

I think that's because of the 8k texture. It takes a second or two on my system as well.

> * Zooming/panning the isolines map happens at 3.4 FPS instead of the base 28 FPS when idle. At the very least, you could pan a ready-made image instead of recalculating it every time. And maybe keep the already computed data to only be rendered on zooming.

Huh, really? I know I had that problem in earlier attempts (not pushed to GitHub), but that should be fixed now (earlier, time had to be stopped in the planetarium in order for the whole plugin to work smoothly). I have no weaker system to test on.

Atque:
Okay, the language issue should be fixed. I hope the lines look better now too.

10110111:
The lines look similarly strange (there does seem to be some improvement, but still not good enough, especially the thin isolines).

But more importantly, now the performance of panning of the isolines map has dropped to 1.8 FPS. I've tried going back one commit just to make sure, and indeed, at 3876d9b26a197c8e6163f40c30046757c75c4c63 I have 3.4 FPS.

10110111:
Widening the window to several screens size reveals the issue with the lines in more detail, see the screenshot below. Most of the lines in this screenshot look OK, but one of them is strangely zigzagged. It does look like rounding of coordinates even more to me, although this time to some grid coarser than pixels.

Screenshot_2026-05-23_10-09-22

10110111:
A few issues in the visibility chart, see the screenshot below.

 * Clipped labels (I marked them in red)
 * Missing label on the left vertical scale? At least the scale differs from that on the right side, so needs a similar label explaining what it is.

Screenshot_2026-05-23_10-20-52

alex-w:
> A few issues in the visibility chart, see the screenshot below.
> 
> * Clipped labels (I marked them in red)
> * Missing label on the left vertical scale? At least the scale differs from that on the right side, so needs a similar label explaining what it is.
> 
> Screenshot_2026-05-23_10-20-52

See also strange position for checkbox "Click map to set location"

alex-w (changes requested):
Please fix compilation with Qt5

alex-w:
First tab has wrong work on macOS

stellarium-007

10110111:
> I think that's because of the 8k texture. It takes a second or two on my system as well.

Well this is unacceptable then. Let's revert to the previous version of the texture for now, I'll try to see later (after this PR gets merged) how to improve on this.

Atque:
I think I've bitten off a little more than I can chew. I'm working on a rewrite, and I'd like to present a new plugin (starting with the "Object visibility" tab) which I can extend with more functionality in the future.

I'll close this PR but leave the branch in case someone finds it useful.

gzotti:
What a pity, giving up the Marathon at 40km. :-(

Atque:
I just want to re-run it with new tactics ;)

I think a segmented approach is better.

axd1967:
Thank you @Atque .

As a side note, hopefully this will lead the effort for updating the feeble Mercator map that Stellarium users are forced to use.
image  (see https://github.com/Stellarium/stellarium/issues/945#issuecomment-4511717074)

Hopefully one day Stellarium will provide the users with an additional (and scrollable and zoomable...) projection: an **orthographic** map of Earth. (Example: https://www.timeanddate.com/eclipse/globe/2026-august-12)

May I ask the team how checkboxes such as  "This change requires a documentation update" 

image

are being dealt with? Is there a separate backlog of UG updates ? Or does someone have to somehow remember to update the UG? IS there automation that tracks this?

Related:
- https://github.com/Stellarium/stellarium/issues/3525
-

PR #4925, opened 2026-05-21
Location dialog: show location filter circle

10110111:
Currently it's quite unintuitive that when one clicks in the map, the location list gets filtered. This patch gives some feedback on this, darkening the area that is filtered out and showing an outline around the remaining area.

### Default look

stellarium-022

### After a click

stellarium-021

### DIFF
--- a/src/core/StelLocation.cpp
+++ b/src/core/StelLocation.cpp
@@ -229,6 +229,18 @@ StelLocation StelLocation::createFromLine(const QString& rawline)
 	return loc;
 }
 
+Vec3d StelLocation::pointAtDistanceDegrees(const double centerLon, const double centerLat, const double distDeg, const double bearing)
+{
+	Vec3d centerPoint;
+	StelUtils::spheToRect(centerLon * M_PI_180, centerLat * M_PI_180, centerPoint);
+	const Mat3d R = Mat4d::rotation(centerPoint, -bearing * M_PI_180).upper3x3();
+
+	Vec3d pointNorthOfCenter;
+	StelUtils::spheToRect(centerLon * M_PI_180, (centerLat + distDeg) * M_PI_180, pointNorthOfCenter);
+
+	return R * pointNorthOfCenter;
+}
+
 // Compute great-circle distance between two locations
 float StelLocation::distanceDegrees(const float long1, const float lat1, const float long2, const float lat2)
 {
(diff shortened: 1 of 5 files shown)


src/gui/MapWidget.hpp
```diff
@@ -40,6 +41,8 @@ class MapWidget : public QWidget
 	void setMarkerPos(double longitude, double latitude);
 	//! allow hiding the location arrow (if sitting on an observer)
 	void setMarkerVisible(bool visible);
+	//! Set the search circle to mark the locations available
```
gzotti, line 44: for clarity, please add that args are degrees (not radians)

gzotti:
good idea! I have not tested but what happens on the edges? Do you wrap the circle?

10110111:
> but what happens on the edges? Do you wrap the circle?

Oh yes, the method that CodeFactor barks at is the implementation of the wrapping :)

In fact, it's not a circle generally on the equirectangular map:

Screenshot_2026-05-21_16-36-37

src/core/StelLocation.cpp
```diff
@@ -229,6 +229,18 @@ StelLocation StelLocation::createFromLine(const QString& rawline)
...
 }
 
+Vec3d StelLocation::pointAtDistanceDegrees(const double centerLon, const double centerLat, const double distDeg, const double bearing)
+{
+	Vec3d centerPoint;
+	StelUtils::spheToRect(centerLon * M_PI_180, centerLat * M_PI_180, centerPoint);
+	const Mat3d R = Mat4d::rotation(centerPoint, -bearing * M_PI_180).upper3x3();
+
+	Vec3d pointNorthOfCenter;
+	StelUtils::spheToRect(centerLon * M_PI_180, (centerLat + distDeg) * M_PI_180, pointNorthOfCenter);
```
gzotti, line 239: what if centerLat==87 and distDeg=5?

10110111: This method will just work as intended: the point will go through the pole (the conversion is a rotation by latitude followed by a rotation by longitude). The issues are to be expected at the rendering stage, but this is also handled:

Screenshot_2026-05-21_16-48-22

gzotti (approved):
Fine for me!

alex-w (approved):
Great!

PR #3091, opened 2023-03-09
Planes plugin reactivation

gzotti:
### Description

Recently, we had a discussion about tracking airplane traffic, similar to satellites (https://groups.google.com/g/stellarium/c/F4Z7lTI9C6M). Alexander recalled that in 2013, a plugin to track planes based on the ADS-B system was under development by @bi0ha2ard. Unfortunately it was never merged, and the current developers have no ADS-B hardware to test or update anything in this plugin. 

I only rebased this 10-year old code on top of current master (2023-03-08) and added enough updates so that it compiles with Qt5.15.2 and Qt6.4 and then added some obvious updates, e.g. use the StelProperty system, and harmonized the color selection. 

It may be possible to extend it based on e.g. data from FlightRadar24, by somebody more familiar with airplane tracking. However I cannot say currently what has to be changed. My time is too limited to continue on that, but maybe somebody else wants to take over?

This is a branch in the main repository. A merge to master is obviously only useful if the plugin demonstrably does something useful. If "conflict" warnings appear, they are usually solved quickly by a rebase from master at the onset of further work.  

Fixes # (issue)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] This change requires a documentation update

### How Has This Been Tested?

**Test Configuration**:
* Qt version: 5.15.2, 6.4
* Operating system: Windows 11
* Graphics Card: (Geforce, but irrelevant)

## Checklist:

- [ ] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #4926, opened 2026-05-21
Restore and modernize the Planes plugin with live ADS-B providers

kamilzaras95-cmyk:
### Description

  This PR restores the built-in `Planes` plugin on current Stellarium `master` and modernizes it into a working live aircraft plugin.

  The original `Planes` work and the later revival attempt in #3091 provided the upstream basis for this plugin. This PR brings it back as a native Stellarium object module and replaces the old offline /
  receiver-oriented approach with live online ADS-B providers.

  Main changes:
  - restore `plugins/Planes` as a built-in static plugin on current `master`
  - integrate aircraft as native `StelObject` / `StelObjectModule` objects
  - fetch live aircraft data around the current observer location
  - support two providers:
    - `adsb.fi`
    - `airplanes.live`
  - draw aircraft directly in Stellarium without external overlay widgets
  - support selection, search, object info, and the standard Stellarium selection pointer
  - add a working configuration dialog and toolbar button
  - add safe refresh limits and provider-specific radius limits
  - keep live requests disabled until the user explicitly enables the plugin

  User-facing behavior:
  - aircraft can be shown/hidden from the plugin button or configuration dialog
  - labels can be enabled
  - label content can be switched between flight number and aircraft model
  - refresh intervals below 15 seconds are blocked to reduce the risk of provider rate limits
  - provider websites are linked from the configuration panel
  - selected aircraft use the standard Stellarium selection marker

  Technical notes:
  - the plugin uses dead reckoning between refreshes based on track, ground speed, and vertical rate
  - selected aircraft instances are preserved across updates by `icao24`, so the selection marker keeps following the aircraft after refreshes
  - provider-specific response differences are handled internally (`aircraft` vs `ac` payload keys)
  - the plugin is disabled by default in `default_cfg.ini`

  This PR focuses on a usable upstream MVP and does not try to solve every future enhancement in one go.

  Related:
  - Revived earlier work: #3091
  - Discussion context: https://github.com/Stellarium/stellarium/discussions/4909

  Attribution:
  This plugin builds on the earlier `Planes` work and revival effort, with the current live ADS-B integration and maintenance update by Kamil Zaraś (astronow.pl).

  ### Screenshots:

  - Planes toolbar button and configuration dialog
  image

  - Aircraft labels and selection marker in the sky
  image

  - Aircraft info panel
  image

  ### Type of change
  - [x] Housekeeping

  ### How Has This Been Tested?

  Manual local testing on current Stellarium `master` build with the built-in `Planes` plugin enabled.

  Tested areas:
  - plugin loads correctly at runtime
  - toolbar button works
  - configuration dialog opens without crashes
  - live provider switching works (`adsb.fi` / `airplanes.live`)
  - aircraft rendering works
  - labels work
  - selection and standard Stellarium pointer work
  - selection persists across refreshes
  - dead reckoning updates aircraft position between refreshes
  - refresh interval lower bound is enforced
  - provider-specific radius limits are enforced
  - provider links open correctly from the configuration dialog

  **Test Configuration**:
  * Operating system: Windows 10 Pro 22H2 (build 22621.4387)
  * CPU: AMD Ryzen 7 3700X (8 cores / 16 threads)
  * RAM: 32 GB
  * Graphics Card: NVIDIA GeForce RTX 4070

  ## Checklist:


plugins/Planes/src/gui/PlanesDialog.cpp
```diff
@@ -0,0 +1,241 @@
+#include "gui/PlanesDialog.hpp"
```
gzotti, line 20: This needs the usual copyright disclaimer found in any of our hpp/cpp source files. Whatever is based on the old Planes plugin should bear the original author's name first please. If heavily modified, of course you can add a second line for you.

kamilzaras95-cmyk: Added the standard Stellarium source headers and kept the original author first, with my line added second where the files were heavily modified.

I also cleaned up one unused include while touching the files.

alex-w (approved):
Thanks for the plugin!

It's good for me as first implementation and public testing feature, but IMHO informational box need to be redesigned and in future getting the ADS-B data also from connected real device will be very helpful.

kamilzaras95-cmyk:
@alex-w thanks!

I’m all for it. I wanted to do something I could get done quickly—that’s why I chose the API. 
I plan to buy or borrow the equipment I need to develop the next feature.
As soon as I have the necessary equipment, I’ll get to work on the redesign and implementation ;

Atque:
Very nice plugin! I think vertical rate could also be given in ft/min.

kamilzaras95-cmyk:
@Atque I will fix this. I hope today

gzotti:
Looks good with finally live data! Current wind means I have planes only out of sight, though, so cannot compare in this hour.  The original Planes plugin also had track lines. Do you have plans to revive them?

Related, when I change the time in Stellarium, the original implementation moved the planes. Not sure what kind of data the online retrieval delivers or whether tracks can be constructed on which the planes then would be shown moving. Currently the planes seem to be plotted where they are "now" (system time). It then may be better to suppress plotting if not in realtime mode. 

The color of the GUI panel title bar seems to differ from other plugin panels. Please try to harmonize with other plugins.

Atque:
This plugin could probably use the same selection marker as the Satellites plugin.

kamilzaras95-cmyk:
I pushed a follow-up update that addresses the recent feedback:

  - @Atque the aircraft info panel now shows vertical rate in both `m/s` and `ft/min`
  - @Atque the selection marker now uses the same pointer style as the `Satellites` plugin
  - @gzotti the configuration dialog title bar is now wired the same way as other plugin dialogs, so its appearance matches the standard plugin panels
  - live aircraft are now shown only in real-time mode; outside real-time mode they are hidden and live updates are suppressed
  - the plugin status now reports this in the configuration panel

  Regarding track lines and time changes:

I agree that the old Planes behavior was useful there. For the current live ADS-B implementation, the data source is   fundamentally “live now”, not true historical trajectory data. 



  My current plan is to treat track lines and limited backward time playback as a follow-up improvement:
  - keep a local recent history buffer per aircraft while the plugin is running
  - use that for drawing recent track lines
  - allow limited backward playback only within that locally recorded history window
  - keep aircraft hidden again when going into future simulation time, or outside the retained local history range

  That would provide a more honest time behavior without pretending to have full historical backend data.

alex-w:
@kamilzaras95-cmyk please update `po/stellarium/POTFILES.in` file also

alex-w:
I want to refactor infobox after merge to harmonize the data with other objects

PR #4950, opened 2026-06-04
Prevent occasional race condition in SolarSystem::computePositions()

gzotti:
I was already pretty sure where to investigate: Moons needing their planets' positions, after all, access of which may collide with writing them in another thread. This solution was proposed by Claude Code: Segment solar system into "buckets" (systemPlanetsByLevel: Sun; Planets/Minorbodies/Comets; Moons; [theoretical moons of moons]; ...)), and process the buckets in sequence. This should guarantee correct execution sequence.

In addition, it turned out some moon systems were not as independent as previously thought and needed to be made threadsafe. Now each moon has its own data set and as long as not the same computation is run in parallel for the same moon, all should be OK. Given their even higher importance with the bucket-wise processing, I have sorted the commits before the actual threading magic.

The fourth commit is as-is from Claude Code (including extensive comments discussing the changes), but needs testing, and yes, some comments can be even shortened.

### Description

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant

## Checklist:

### DIFF
--- a/src/core/planetsephems/tass17.c
+++ b/src/core/planetsephems/tass17.c
@@ -3057,7 +3057,7 @@ static void CalcLon(double t,double lon[7])
 	}
 }
 
-static void CalcTass17Elem(double t,const double lon[7],int body,double elem[6])
+static void CalcTass17Elem(const double t, const double lon[7],const int body, double elem[6])
 {
 	const struct Tass17MultiTerm *tmt_begin,*tmt;
 	int i;
@@ -3143,24 +3143,23 @@ const double TASS17toJ2000[9] = {
 */
 
 #define TASS17_DIM (8*6)
-static double t_0 = -1e100;
-static double t_1 = -1e100;
-static double t_2 = -1e100;
-static double tass17_elem_0[TASS17_DIM];
-static double tass17_elem_1[TASS17_DIM];
-static double tass17_elem_2[TASS17_DIM];
+static double t_0[TASS17_NMOONS] = {-1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100};
+static double t_1[TASS17_NMOONS] = {-1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100};
+static double t_2[TASS17_NMOONS] = {-1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100};
+static double tass17_elem_0[TASS17_NMOONS][TASS17_DIM];
+static double tass17_elem_1[TASS17_NMOONS][TASS17_DIM];
+static double tass17_elem_2[TASS17_NMOONS][TASS17_DIM];
 /* 1 day: */
 #define DELTA_T 1.0
 
-static double tass17_jd0 = -1e100;
-static double tass17_elem[TASS17_DIM];
+static double tass17_jd0[TASS17_NMOONS] = {-1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100};
+static double tass17_elem[TASS17_NMOONS][TASS17_DIM];
 
 void CalcAllTass17Elem(const double t,double elem[TASS17_DIM], void *user)
 {
-	int body;
 	double lon[8];
 	CalcLon(t,lon);
-	for (body=0;body<=7;body++) CalcTass17Elem(t,lon,body,elem+(body*6));
+	for (int body=0;body<=7;body++) CalcTass17Elem(t,lon,body,elem+(body*6));
 }
 
 void GetTass17Coor(double jd,int body,double *xyz, double *xyzdot)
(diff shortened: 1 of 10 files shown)


10110111:
This method is becoming awfully long. It needs to be split into separate ones, one per algorithm. It should be cheap, given that you're using Claude for this PR :P

gzotti:
Well, actually a colleague with Claude access offered his assistance. But it's an interesting teaser into utilizing advanced AI assistants. I also got a "problem discussion and update plan" for StelMovementMgr which (I think) shows pre-Qt code and is a minefield of side effects in deep need of cleanup/rewrite. 

Not sure why splitting the 4 cases into 4 methods would really be helpful. Case 2 is default, others can be configured in config.ini.

Actually we could decide on the best working of 2/3 and delete the other. TBH I have not fully ingested your solution 3 but always used case 2 (or actually stopped using extra threads soon after hangs appeared: Have not used my 43.000 objects asteroid belt since). I think case 1 is still bugged and could be removed. Case 0 should be left for historical reasons (contains the basic algorithm).

gzotti:
It still isn't done. Something bad has crept in again...

10110111:
> It still isn't done.

Do you mean the model is still working on a new commit?

gzotti:
No, I have no access to that until Monday. I am seeing a hang and scratching my own head again over it.

gzotti:
Interestingly the same code is running for 3 hours, ~11000 SS objects and 3-7 extra threads now without issue on another PC. Cannot reproduce the problem here as seen from Earth and SSObserver. Sad to see the Trojans apparently dissolving over centuries as orbital elements are not updated. And it becomes ever clearer that not the SS computation is the bottleneck but the Marker rendering...

gzotti:
Another test, 8 hours and no issue. Maybe that one hang had other reasons? Please test with a few extra threads.

gzotti:
Ran overnight successfully. Now gave another try with case3. got stuck after 2 hours, but finally in TASS17. (Saturn moons.) This is not really nice to read... It seems each moon needs longitudes of the others, so here may be an issue when there is only a single target array for some results.

gzotti:
I also had some hangs during startup. They resolved to tass17 and marssat which indeed were not thread safe. Maybe the separation into correct bins even exacerbated the situation when the few moons fought for computing. All planetephems should now be thread safe. Since the last changes, startup always succeeded, a few ten times. Is there a trick to test in a CI cycle: launch a program, check that it comes up, exit after 5 seconds, repeat 500x?

gzotti:
Observing Saturn and its moons, I see unfortunate short stalls. It looks even smoother without extra threads (default ssystem_minor.ini). :-( 26.1 is smoother but can occasionally freeze. Is it all just too much overhead?

10110111:
> Is there a trick to test in a CI cycle: launch a program, check that it comes up, exit after 5 seconds, repeat 500x?

You can try using Xvfb on a Linux builder. And then use the `timeout` command to kill the process after a given duration.

Whether the program comes up you can check by exit status I think, e.g. consider the following command:

```
LC_ALL=C time timeout 3 sleep 30
```

I use the C locale to avoid unexpected issues like dot vs comma etc. (and this also results in the use of `/usr/bin/time` instead of the Bash builtin).

Now, `time` measures the actual time elapsed: `0:03.00elapsed` in the output. `timeout 3` sends a signal to the process `sleep 30` when time is up. Exit status `124` means that the command timed out (or, well, that the process returned 124). Any other will mean that the process terminated by itself (possibly by crash).

The question is whether 5 seconds is enough for all your initialization. Stellarium starts in about 15 seconds on my machine, for example.

10110111:
> Observing Saturn and its moons, I see unfortunate short stalls. It looks even smoother without extra threads (default ssystem_minor.ini). :-( 26.1 is smoother but can occasionally freeze. Is it all just too much overhead?

Do we actually want to parallelize the main planets + a bunch of natural satellites? Maybe only do this for minor bodies? Then we'll get rid of the parent nesting problem I think (right?), as well as of the need to make satellite ephemerides computation thread safe.

gzotti:
That could be a way, splitting major planets plus moons from minor bodies (comets, asteroids, KBOs, ...), and parallelize only the latter. The thread-safeness for moons was similar to VSOP87, just added a few storage spaces for results. This could stay IMHO.

Not sure what to do for 26.2 now. Hide the thread setting (with flag for devs) and keep extrathreads at 0?

10110111:
> Not sure what to do for 26.2 now. Hide the thread setting (with flag for devs) and keep extrathreads at 0?

Maybe simply disable the spinbox and add a tooltip explaining why (with a reference to the corresponding SUG chapter number).

10110111:
> Maybe simply disable the spinbox and add a tooltip explaining why (with a reference to the corresponding SUG chapter number).

Though I wonder what if the spinbox shows a value≠0. In this case the user may want to be able to switch it back to 0.

gzotti:
Of course I'd initialize to zero.

gzotti:
New hope: I rebased and rearranged the commits to start with the thread-safe moon ephemerides, then applied the first Claude work which split SSO into buckets by parent numbers. Running this commit (4b99d93) looks smooth, and I did not manage to create a startup hang which Claude tried to fix with QThreadpool magic which may have caused the hiccups. Probably the latter is not required after all. 
If this runs overnight, this might be the solution. Just running the 40000 asteroid system again. 10fps is not much but looks awesome. 

EDIT: Seem to work!! 8 hours running, and then just working startup more than 10 times without issues.

10110111:
We have just a few days before the release, maybe we shouldn't cram this into it, and could postpone this PR to 26.3?

gzotti:
It just works better than before. I could now not cause further hangs, and the resolutions were fully traceable: ephemeris conflicts which I had not seen in 2024 (solved) and the occasional race in steps 2 and 3 (solved by Claude).  The second commit from Claude (which actually helped me understand the actual issue around moons now fighting more often for their siblings' data) is no longer included.

The "avoidance fix" so far was setting 0 extra threads. Nothing indicates to me it became even worse when doing so. N.B. with default case 2 only. I have not further investigated your case 3. This can be aimed for in 26.3, we need probably only one, well-working solution.

gzotti:
Another day has passed. @alex-w @10110111 Can you observe any problem in this bugfix?

10110111:
I'll only have time to review this on the weekends.

alex-w (approved):
My tests are passed

src/core/planetsephems/tass17.c
```diff
@@ -3143,24 +3143,23 @@ const double TASS17toJ2000[9] = {
...
+static double tass17_jd0[TASS17_NMOONS] = {-1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100, -1e100};
+static double tass17_elem[TASS17_NMOONS][TASS17_DIM];
 
 void CalcAllTass17Elem(const double t,double elem[TASS17_DIM], void *user)
 {
-	int body;
 	double lon[8];
 	CalcLon(t,lon);
-	for (body=0;body<=7;body++) CalcTass17Elem(t,lon,body,elem+(body*6));
+	for (int body=0;body<=7;body++) CalcTass17Elem(t,lon,body,elem+(body*6));
```
10110111, line 3162: This function is called from `CalcInterpolatedElements` (see line 3182), which is supposed to be called per body, but here we ignore this per-body logic and calculate all the bodies, potentially in parallel threads. So there's still a chance that one and the same body will be computed in parallel, writing to the same set of variables from different threads.

Same problem with `CalcAllMarsSatElem`.

10110111: Ah, actually no, it will write to different arrays. The whole logic is quite ugly though, because `TASS17_DIM` already splits the array into 8 bodies, and we duplicate this into 8 (=`TASS17_NMOONS`) more per-body arrays.

10110111:
I guess this can go in now, but the concurrency of tass17 and marssat should be refactored at some point.

src/core/planetsephems/tass17.c
gzotti: Yes, this was the solution. I agree it's ugly, but did the job.

PR #4966, opened 2026-06-10
Lightweight atmosphere for ANGLE

10110111:
This PR is an attempt to make the LW atmosphere work on ANGLE. Please check that this doesn't break other platforms and, most importantly, that it works as expected on ANGLE. If things go well, we'll be able to remove the GLESv3 vs GLESv2 line from the SUG.

### DIFF
--- a/guide/ch_skylight.tex
+++ b/guide/ch_skylight.tex
@@ -159,7 +159,7 @@ be given in a future version of this User Guide.
 
 \subsection{Intermediate Mode: The Lightweight Skylight Model}
 
-This model \newFeature{26.2} combines a realistic view of the sky as in ShowMySky with light weight of the Preetham model, at the cost of being fixed to simulating a sea level sky (just as Preetham), lacking the realism of eclipse simulation (also as Preetham), and having no parameters to tweak, and not supporting OpenGL ES v2 (only v3 is supported, so it won't work with ANGLE included with Qt5).
+This model \newFeature{26.2} combines a realistic view of the sky as in ShowMySky with light weight of the Preetham model, at the cost of being fixed to simulating a sea level sky (just as Preetham), lacking the realism of eclipse simulation (also as Preetham), and having no parameters to tweak.
 
 The Lightweight model based on a set of meshes generated from a render of ShowMySky, so its look is just as realistic (at sea level and without any eclipse) as that of ShowMySky, except for the solar aureole, which is a bit simplified.
 
--- a/src/core/TextureAverageComputer.cpp
+++ b/src/core/TextureAverageComputer.cpp
@@ -190,7 +190,7 @@ void TextureAverageComputer::checkNeedForWorkaround()
 	for(int n=data.size(); n<width; ++n)
 		data.emplace_back(0,0,0,0);
 
-	GL(gl.glTexImage2D(GL_TEXTURE_2D,0,GL_RGBA8,data.size(),1,0,GL_RGBA,GL_UNSIGNED_BYTE,&data[0][0]));
+	GL(gl.glTexImage2D(GL_TEXTURE_2D,0,GL_RGBA,data.size(),1,0,GL_RGBA,GL_UNSIGNED_BYTE,&data[0][0]));
 	const auto mipmapAverage = getTextureAverageSimple(texture, width, 1);
 
 	const auto sum = std::accumulate(data.begin(), data.end(), Vec4f(0,0,0,0),
--- a/src/core/TextureAverageComputer.cpp
+++ b/src/core/TextureAverageComputer.cpp
@@ -36,12 +36,12 @@ public:
 	}
 	void capture()
 	{
-		GL(gl.glGetIntegerv(GL_DRAW_FRAMEBUFFER_BINDING, &oldFBO));
+		GL(gl.glGetIntegerv(GL_FRAMEBUFFER_BINDING, &oldFBO));
 	}
 	~FBORestorer()
 	{
 		if (oldFBO >= 0)
-			GL(gl.glBindFramebuffer(GL_DRAW_FRAMEBUFFER, oldFBO));
+			GL(gl.glBindFramebuffer(GL_FRAMEBUFFER, oldFBO));
 	}
 };
 }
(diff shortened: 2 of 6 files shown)


10110111:
On Wine I see that it only works in `--angle-d3d9` mode, but not in `--angle-mode`, and it seems that near the twilight the brightness goes down too much, and if it's enabled at startup, then the sky is purely white. But this all may be due to the imperfection of WineD3D, I cannot tell whether it's so. Needs testing on real Windows.

gzotti:
Can you please trigger making a downloadable installer? My 2010 ANGLE dependent PC has no build env. I can hopefully test tonight (in 13 hrs.)

10110111:
> Can you please trigger making a downloadable installer?

OK, I'll push a trigger commit here.

gzotti:
Thanks, will test there later. While compiling on another PC, I see a warning:

[158/255 2.3/sec] Building CXX object src\CMakeFiles\stelMain.dir\core\modules\AtmosphereLightweight.cpp.obj
D:\StelDEV\stellarium\src\core\modules\AtmosphereLightweight.cpp(657): warning C4312: 'reinterpret_cast': conversion from 'const _Ty' to 'void *' of greater size
        with
        [
            _Ty=std::seed_seq::result_type
        ]

Is that clean?

10110111:
> Is that clean?

Well the very need for a `reinterpret_cast` here is due to a deficiency of the API, whereby the function wants a simple integer (offset) while its C signature expects a pointer in this slot. Generally I think this warning is harmless, although I guess I could silence it by tweaking the code a bit.

gzotti:
Works largely OK. Speed is about 13fps vs. Preetham 17fps. Pretty good!

[     0.998][INFO] OpenGL module type: 1
[     0.998][DBG ] StelGLWidget constructor
[     0.998][DBG ] StelGraphicsScene constructor
[     1.047][DBG ] initializeGL(windowWidth = 1280, windowHeight = 1024)
[     1.047][INFO] OpenGL supported version:  "OpenGL ES 2.0 (ANGLE 2.1.99f075dade7c)"
[     1.047][INFO] Current Format:  QSurfaceFormat(version 2.0, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize 24, redBufferSize 8, greenBufferSize 8, blueBufferSize 8, alphaBufferSize 8, stencilBufferSize 8, samples 0, swapBehavior QSurfaceFormat::DefaultSwapBehavior, swapInterval 1, colorSpace QSurfaceFormat::DefaultColorSpace, profile  QSurfaceFormat::NoProfile)
[     1.047][INFO] Initializing StelMainView
[     1.047][INFO] Luminance textures are supported
[     1.047][INFO] Running in Low Graphics Mode
[     1.047][INFO] Maximum texture anisotropy: 16
[     1.047][INFO] Maximum 2D texture size: 8192
[     1.047][INFO] Detected: OpenGL ES 2.0
[     1.047][INFO] Driver version string: OpenGL ES 2.0 (ANGLE 2.1.99f075dade7c)
[     1.047][INFO] GL vendor: Google Inc.
[     1.047][INFO] GL renderer: ANGLE (Intel(R) HD Graphics Direct3D11 vs_4_0 ps_4_0)
[     1.047][INFO] GL Shading Language version: OpenGL ES GLSL ES 1.00 (ANGLE 2.1.99f075dade7c)

During twilight and when the sun rises higher, the darkest area shows banding. With ANGLE there is no dithering. But the general view compared to Preetham is much nicer.

I wanted to attach a screenshot but this is buggy on this PC, at least with this build.
stellarium-001

Screenshot with Windows capture:


Unbenannt

gzotti:
Addition: Screenshot works OK in 26.1.

10110111:
> Addition: Screenshot works OK in 26.1.

Doesn't it work with this PR too when you're using Preetham atmosphere?

Also, does screenshotting (with LW atmosphere) work in non-ANGLE mode on the PCs where it's supported? And does it work with the 26.2-RC1?

10110111:
Generally this seems to be an improvement after all, despite all the shortcomings, since the RC1 currently crashes when LW atmosphere is selected.

10110111:
Just tried it on a real Windows 11, and

 * No crash while RC1 does crash on selecting LW atmosphere;
 * The white sky when starting with LW enabled also reproduced, just as seen before in Wine. (Switching to Preetham and back to LW makes it work.);
 * The screenshots in ANGLE mode are also broken when LW atmosphere is active and enabled, but work in non-ANGLE mode;
 * Average luminance computation also works wrongly, just as it did in Wine.

10110111:
OK I reproduce the broken screenshots on Linux with GLES-based Qt and `MESA_GLES_VERSION_OVERRIDE=2.0` and `MESA_GLSL_VERSION_OVERRIDE=100` environment variables.

gzotti:
"+" Banding reduced
"+" Screenshot works again
"-" White-out (Average luminance computation). now broken also here. This worked yesterday. Can be mitigated a bit by tonemapping, setting screen lum. to 1800 nits or so, but this looks like an error. And: with setting 1800 nits, stars vanish.

10110111:
> White-out (Average luminance computation). now broken also here. This worked yesterday.

Yeah, this is an elusive bug. Doesn't reproduce anywhere except ANGLE. Even on MESA with forced GLESv2 everything is in order, but not on ANGLE.

10110111:
OK, I know what's going on. ANGLE from Qt5 doesn't support `GL_*_vertex_array_object`. None of these extensions (ARB, APPLE, OES) that `QOpenGLVertexArrayObject` supports. So the simple `MESA_GLES_VERSION_OVERRIDE=2.0 MESA_GLSL_VERSION_OVERRIDE=100` is not sufficient to reproduce the problem in a Linux-based development environment, one has to add `MESA_EXTENSION_OVERRIDE=-GL_ARB_vertex_array_object` too.

10110111:
So this should fix the issue. Please check if it does for you.

gzotti:
OK, waiting for the build to become available... what a queue!

10110111:
The queue has two useless entries that will take time to build before the final one builds. You can use the [artifact from my own instance of AppVeyor](https://ci.appveyor.com/project/10110111/stellarium/builds/54204714/artifacts).

gzotti:
Great, thanks! I installed the patch over yesterday's package. And .. tadaa, it works! Screenshots are good. 13fps vs 17.5 for Preetham, that's probably acceptable for the increase in quality. Still, the Preetham solution should remain available (still cheaper on field laptops where energy/battery time counts more than visual atmosphere simulation quality. But if Lightweight turns out to work everywhere during 26.2 worldwide deployment I have no objection to make it default in 26.3.

Issue #2719, opened 2022-10-04
Solar disk lacks limb darkening


10110111:
### Expected Behaviour

Zooming into the Sun should show a solar disk that has smooth reduction of brightness from the center to the rim. Here's a photo of the Sun I took in real life through a solar filter:

![sun-photo](https://user-images.githubusercontent.com/6376882/193814480-116bb4ae-68da-441f-8e42-c82e2421ed9b.png)

The [theoretical brightness profile](https://en.wikipedia.org/wiki/Limb_darkening#Calculation_of_limb_darkening) is obviously not flat:

![brightness](https://user-images.githubusercontent.com/6376882/193817407-4404930b-edd4-4f2d-99d6-3118289a5d7d.png)

### Actual Behaviour

The solar disk is flat (if we disregard the grain texture and the irrelevant "halo" texture), and there's a strange bright rim. Here's a screenshot:

![sun](https://user-images.githubusercontent.com/6376882/193815128-df520d7e-965f-4b0e-9d4c-9f2dfab3c8dc.png)

If we take the RGB profile from the screenshot (~150px away from the center to avoid the "halo"), we'll see this:

![profile](https://user-images.githubusercontent.com/6376882/193815629-3656c98c-548b-427e-920b-09939441881a.png)

Both the bright rim and the flat interior are obvious here.

### Steps to reproduce

1. Select the Sun
2. Center on selected object <kbd>Space</kbd>
3. Disable atmosphere <kbd>A</kbd>
4. Zoom in so that the solar disk fills almost the whole screen.

### System
* Stellarium version: v1.0-13-g6bffbc7c0a
* Operating system: Ubuntu, Windows
* Graphics Card: NVIDIA GeForce GTX 750Ti

gzotti:
Indeed, limb darkening should be easy to add with normal vector consideration. The bright rim may even be an artifact of the corona texture painted after the sun (?) and a slight mismatch. In this case, of course invert drawing order. About Solar Corona, there is another issue about it not being affected by refraction. This would need a different approach to map the Corona into the sky.

10110111:
> The bright rim may even be an artifact of the corona texture painted after the sun (?)

Yes, it appears to be. I noticed that we do have an option `landscape/early_solar_halo`, which defaults to `false` and influences the corona too. Why? When is it useful? I think it should be true unconditionally.

gzotti:
[...]

PR #4973, opened 2026-06-14
Fix Solar Corona refraction

### LINKED ISSUE #2264: Corona texture not adjusted for refraction
Viewing solar eclipses close to the horizon, we see that the corona texture is not affected by atmospheric refraction. See the screenshot below.

stellarium-117

Note how the Moon is squeezed while the corona texture is not, creating an ugly artifact.

gzotti:
### Description

The original implementation of the Solar corona uses screen-space drawing (sprite), which cannot accommodate for differential refraction or map projections.  Solar eclipses at sunset looked somewhat bad.

This branch utilizes a texturized vertex array centered on the sun that handles refraction better. Vertices are adjusted for declination, and the mesh is rotated around its center to adjust for the angle between ecliptic and the equatorial grid so that the corona stays aligned with the ecliptic plane as before (should probably become solar equator, but the difference in appearance is small and corona changeable).

Fixes #2264 

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Check that the corona does encircle the Sun, that no dark inside area is visible but also the prominence should be visible. 

**Test Configuration**:
* Graphics Card: Geforce

## Checklist:

### DIFF
--- a/src/core/StelSkyDrawer.cpp
+++ b/src/core/StelSkyDrawer.cpp
@@ -520,12 +520,14 @@ bool StelSkyDrawer::drawPointSource(StelPainter* sPainter, const Vec3d& v, const
 
 // Draw's the Sun's corona during a solar eclipse on Earth.
 // painter: in FrameJ2000
-// pos: solar position, J2000
+// posJ2000: solar position, J2000
 // radius: angular dimension of the square texture, not the sun. --> NEW: angular radius of the sun, radians
 // color: attenuated sunlight color (extinction may have reddened the sun)
 // alpha: transparency
 // NO MORE: angle is 0 for an equatorial sky view, solar parallactic angle for an azimuthal mount.
-void StelSkyDrawer::drawSunCorona(StelPainter* painter, const Vec3d& pos, double radius, const Vec3f& color, const float alpha)
+// eclipticAngle intersection angle between ecliptic and equatorial coordinates of date (not J2000!).
+// This has to be applied to the image, so that the corona is aligned along the ecliptic.
+void StelSkyDrawer::drawSunCorona(StelPainter* painter, const Vec3d& posJ2000, double radius, const Vec3f& color, const float alpha, const double eclipticAngle)
 {
 	radius *= (512.f/193.f); // Texture size is 1024, solar radius within is 192 or 193. Increase radius to the width/height of actual image, in radians.
 	texSunCorona->bind();
@@ -544,7 +546,7 @@ void StelSkyDrawer::drawSunCorona(StelPainter* painter, const Vec3d& pos, double
 	// Partly copied from Constellation artwork.
 
 	double ra=0, dec=0;
-	StelUtils::rectToSphe(&ra, &dec, pos);
+	StelUtils::rectToSphe(&ra, &dec, posJ2000);
 
 	// Tessellate texture into an equispaced 5x5 field
 	static const int nbPoints=5;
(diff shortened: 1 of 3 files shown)


gzotti:
Looks good now I think. Some cleanup still pending.

src/core/StelSkyDrawer.cpp
```diff
@@ -519,20 +519,124 @@ bool StelSkyDrawer::drawPointSource(StelPainter* sPainter, const Vec3d& v, const
...
+	// Create a 5x5 mesh with texture coordinates to display a texture quad.
+	// Partly copied from Constellation artwork.
+
+	double ra=0, dec=0;
+	StelUtils::rectToSphe(&ra, &dec, posJ2000);
+
+	// Tessellate texture into an equispaced 5x5 field
+	static const int nbPoints=5;
+	static QVector<Vec2f> texCoords((nbPoints-1)*(nbPoints-1)*6);
+	static bool texInitialized=false;
```
10110111: Why not just make these static variables normal data members and initialize them in the constructor? It would be so much cleaner, including getting rid of this `texInitialized` flag.

gzotti: Yes, that's the other option for that. If you prefer...

gzotti:
Another question to @alex-w , although it has been over 10 years since this line was committed. Why do we need the last line here, postDrawPointSource() ?

Another ugliness: A zoomed-in sun should not have the solar "halo" staining the now beautifully textured solar center. Not related to this PR, just another observation.

10110111:
> Another ugliness: A zoomed-in sun should not have the solar "halo" staining the now beautifully textured solar center.

Yes, I complained about this a few years ago, and the solution was to have the halo drawn before the Sun. I don't remember where in the GUI the relevant option is though.

alex-w:
> Another question to @alex-w , although it has been over 10 years since this line was committed. Why do we need the last line here, postDrawPointSource() ?

I don't remember details, but this line should flush the buffer - you can try disable it and check the behavior of drawing SSO.

gzotti:
It's called in the method to draw the corona (which is called only under some tested conditions). This IMHO has nothing to do with the pointsource buffer.

alex-w:
> It's called in the method to draw the corona (which is called only under some tested conditions). This IMHO has nothing to do with the pointsource buffer.

It may be related to old non-trivial hooks - please disable this call and check how app works

src/core/StelSkyDrawer.cpp
gzotti: Done

gzotti:
Can you trigger the debug message? If not, the call can be removed.

gzotti:
> Yes, I complained about this a few years ago, and the solution was to have the halo drawn before the Sun. I don't remember where in the GUI the relevant option is though.

Ah, landscape/early_solar_halo=true. This should probably become default, @alex-w ?

alex-w:
> Ah, landscape/early_solar_halo=true. This should probably become default, @alex-w ?

I'm not very good familiar to this part of code :( @10110111?

10110111:
> I'm not very good familiar to this part of code :( @10110111?

Me neither, I only noticed it [as noted here](https://github.com/Stellarium/stellarium/issues/2719#issuecomment-1399518051), and when I suggested that it should be unconditionally true, @gzotti said this:

> Cannot say. There were some changes in 2015/16 when I prepared using Stellarium in an exhibition. "false" should then be the old behaviour. Some effects look different, and I was probably not convinced if it was really better in all settings. E.g., check halo with a rising sun.

gzotti:
OK, if at that time corona and "halo" were both drawn after the sun, a corona distorted at the horizon would have caused big headaches. Now that corona is drawn before the 3D sphere, there is no bright rim line, and distortion at the horizon also works. So, flipping at least the default seems OK for me, if not even removing the "halo after sphere" option at some later date.

gzotti:
If you are else OK, I'd revert that "Temporary" commit, remove the postDrawPointSource() call which in my observations never came beyond testing `nbPointSources==0`, keep the drawing order toggle with new default and squash&merge this branch.

src/core/StelSkyDrawer.cpp
```diff
@@ -135,6 +137,21 @@ StelSkyDrawer::StelSkyDrawer(StelCore* acore) :
 	}
 	texImgHalo=QImage(StelFileMgr::getInstallationDir()+"/textures/star16x16.png");
 	texImgHaloSpiky=QImage(StelFileMgr::getInstallationDir()+"/textures/star16x16_rays.png");
+
+	// Tessellate texture into an equispaced 5x5 field. Vertices have to be computed per frame.
+	coronaTextureCoords=QVector<Vec2f>((coronaMeshDim-1)*(coronaMeshDim-1)*6); coronaTextureCoords.clear();
```
10110111: This whole line is completely unnecessary. There's no need to do any kind of reservation, since the filling happens only once at startup, so has negligible performance impact.

10110111:
> If you are else OK, I'd revert that "Temporary" commit, remove the postDrawPointSource() call which in my observations never came beyond testing nbPointSources==0

I'd be conservative a week before the release, and not change things not completely understood until the release.

src/core/StelSkyDrawer.cpp
gzotti: probably right.

gzotti:
OK, then I'd revert that "Temporary" commit (or shall I keep it and modify warning to "If you can read this, please send a report!"), keep (but still flag "What for") the postDrawPointSource() call which in my observations never came beyond testing nbPointSources==0, keep the drawing order toggle with old default and squash&merge this branch.

10110111:
> or shall I keep it and modify warning to "If you can read this, please send a report!"

I don't think it's necessary, you can just do the actual removal after the release, and we'll have 3 months to see if something breaks.

gzotti:
@alex-w @10110111 any further comments or objections?

10110111 (approved):
Seems to work from what I have tested.

alex-w (approved):

PR #4957, opened 2026-06-06
SUG: write section for SCM

xLPMG:
Good news! I have finally found time to write a complete guide for the SCM plugin :)

I would appreciate any feedback.

@sushoff

PR #4978, opened 2026-06-15
SUG: improve SCM section

xLPMG:
Implements feedback from #4957


sushoff:
pardon, ... I am always responding to mail - so when GitHub mails me from the old branch, I tend to overlook that it is closed. Thanks for fixing typos etc ... that's why I shared it as comment, so that more eyes can look at it!

alex-w (approved):

PR #4991, opened 2026-06-20
 Add Gaia DR3 level 9/10 star catalog support with compact storage and generation tools

SceneryMC:
### Description

Add a Gaia DR3-based star catalog generation pipeline and the engine changes needed to render the resulting high-level catalogs smoothly.

**Catalog generation (`util/gaiacat/`)**

- `skychart2cat` converts Gaia DR3 `.bin` extracts into Stellarium `.cat` catalogs. SkyChart `.dat` files are also supported as a fallback, but their parallax and proper-motion values have a known precision issue (~10^4–10^5× magnification for near-zero measurements).
- `extract_columns.py` compresses the full Gaia DR3 CSV release into compact `.bin` files (60 B/star) containing only the fields needed for the Star2 format.
- `dedupcat` removes duplicate stars between adjacent magnitude levels.
- `cmpcat` compares two catalogs zone-by-zone for validation.

The pipeline preserves stars missing BP/RP photometry, includes proper motion, parallax and B-V, and supports level 8/9/10 catalogs in the Star2 format.

**Engine changes**

- `SpecialZoneArray::draw()` now returns early when the level's brightest star (`mag_min`) is fainter than the current visibility limit. This avoids paging in huge mmap'd catalogs every frame just to discover nothing is visible, which removes the stuttering seen with the new level 8+ catalogs.
- Optional compact storage for high-level catalogs: when enabled, only the zone-count table and star data are kept, omitting the per-zone `SpecialZoneData` array. Restricted to `level >= 9`. Controlled by `stars/flag_compact_star_storage` (default `false`).
- `StarMgr::searchGaia()` searches all loaded levels using `maxGeodesicGridLevel` instead of the draw-limited `getMaxSearchLevel()`, and Phase 2 uses a 60" search radius around the HEALPix pixel center to catch stars near pixel edges.
- Fixed an integer overflow in `StarMgr::checkAndLoadCatalog()` when computing MD5 checksums for `.cat` files larger than 2 GB.

图片

Galactic center — dense star field loaded smoothly, no stuttering

图片

Star at ~22 mag, found via the search window

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Generated `stars_8` (6.5 GB), `stars_9` (19 GB) and `stars_10` (26.7 GB) Star2 catalogs from Cartes du Ciel `.dat` files and from Gaia DR3 CSV extracts on Windows 10. All catalogs were placed on an HDD. Browsing the galactic center at FOVs from 60° down to 0.01° shows no stuttering; at FOV 0.1° performance is limited by GPU draw-call throughput, not disk I/O.

`testGaiaSearch` was run on levels 8, 9 and 10 with up to 100,000 samples per level. Phase 1 finds the correct zone for ~83–95% of stars; Phase 2 finds the remainder with zero misses.

**Test Configuration**:
* Operating system: Windows 10 64-bit
* Graphics Card: NVIDIA RTX 2070 laptop, driver version 551.86

## Checklist:


SceneryMC:
The `.cat` files needed for testing are being generated. I'll drop a download link in the comments once they're ready, along with the corresponding `starsConfig.json`

10110111:
Wow, the first screenshot looks impressive. Are these stars ultimately all taken from Gaia DR3?

SceneryMC:
> Wow, the first screenshot looks impressive. Are these stars ultimately all taken from Gaia DR3?

Yes. I converted Cartes du Ciel's Gaia DR3 `.dat` files (https://sourceforge.net/projects/skychart/files/catalog/) into Stellarium's `.cat` format, since the original Gaia DR3 CSV archive is over 700 GB and impractical to download. The formulas for deriving Vmag and B-V from G, Gbp, and Grp are the same as those used by Cartes du Ciel. Hope I didn't make a mistake somewhere.

10110111:
> Hope I didn't make a mistake somewhere.

It may be useful to have the script you used for the conversion in this PR (under the `util` directory), both to check correctness and to enable users to make the conversion themselves.

SceneryMC:
Uploaded. That said, the Python script is almost unoptimized — I assumed it would only run once, and I had access to a server with 128 GB RAM and 64 cores (128 threads), so a typical computer would likely struggle to run it. I'm currently rewriting it in C++ so it can complete in an acceptable time on my 16 GB laptop. Once it's done and I've confirmed the output matches the Python version, I'll push it as well.

gzotti:
Looks impressive, indeed. However please consider:
Stellarium runs on 2010 hardware, old field laptops, RaspberryPi 4 (used to run on RPi3 but this is just no longer enough with our current Gaia catalog) etc, so this extension should remain optional. Also, is it OK to gut the catalog from CdC? Adding the tool to self-convert the catalog is important, as are instructions for beginners how to create the big catalog. Similar to the instructions around orbital elements added recently (described in sections D2.3.2 and D2.4 in the Guide in master), I would like to see such chapter describing the process of catalog creation. 
Also note, our catalog separates out binaries with known orbits, and properly handles stars with proper motion fast enough to leave their Zone. I hope this PR likewise takes care of this.

SceneryMC:
Understood.

1. I will keep the `DynamicZoneArray` feature disabled by default in the `config.ini` file to ensure compatibility with low-end hardware.
2. Regarding the data sources: For the Star3 format, the CdC `.dat` files are sufficient, aside from some minor precision issues. However, for the Star2 format, the CdC data lacks the `plx_err` field, which means this data has to be extracted from the original Gaia CSVs.
3. I have generated the new catalogs for levels 8–10 (levels 0–7 remain unchanged). I regenerated level 8 because the original magnitude range (16.75–18.0) seemed somewhat truncated. The new magnitude ranges are as follows:

| Level | Vmag range |
|-------|-----------|
| 8     | 16.75–18.50 |
| 9     | 18.50–20.25 |
| 10    | 20.25–23.00 |

Since the Star2 and Star3 formats lack the `hip` field, the generated deep catalogs cannot explicitly map binary stars. This matches the behavior of the original level 8 catalog, and since stars fainter than 16.75 mag do not have Hipparcos numbers anyway, I assume this is acceptable.

Additionally, I have ported the catalog generation script from Python to C++. The generation process now completes in about an hour on my laptop (16 GB RAM), with memory consumption staying well under 8 GB. I am currently working on adding global zone support for both the C++ and Python scripts.

SceneryMC:
Wait — I just noticed the User Guide mentions:

> Only level 0 to level 2 star catalogs have their global zone populated (i.e. level 3–8 global zones are empty).

Since my current C++ and Python converters are targeting levels 8 to 10 (stars fainter than 16.75 mag), this means global zone handling isn't actually needed for them. So the scripts are basically already complete for this specific task.

However, if you'd prefer to have a more generalized tool for the project, I could try to extend the converters to support global zones and binary stars (HIP IDs), making it possible to generate level 3–7 catalogs as well in the future. (I would prefer to leave levels 0–2 to the existing pipeline. Safely handling the complex blending with the Hipparcos catalog is quite challenging for me right now.)

gzotti:
Maybe it's enough to keep the Global Zone to levels 0...2 then. Contemporary observers will mostly care for the time until 2050 where I don't expect big changes, and if a star mag22 goes absent, I won't really note it. The necessity for special handling is mostly about brighter stars going AWOL in antiquity.

SceneryMC:
Got it. 

Also, while reviewing the User Guide, I noticed that the description in Section C.2.1 regarding coordinates still refers to the old method. 

> Individual star records do not contain full positional coordinates, instead they contain coordinates relative to the central position of the zone they occupy. Thus, when parsing star catalogues, it is necessary to know about the zone model to be able to extract positional data.

Since the new Gaia DR3 format uses absolute RA/Dec in mas or 0.1 arcsec, this paragraph seems to be a leftover from the pre-25.1 era. I’ll go ahead and correct this description while adding the documentation for the catalog creation process.

SceneryMC:
I’ve made some progress on the catalog generation and uncovered some interesting technical insights I'd like to share:

### Catalog generation

1. **BP/RP-missing stars preserved**: Unlike the existing level 7+ catalogs, my scripts retain stars that lack BP and RP (B-V defaults to 0, G magnitude used directly as V). Notably, the ADQL query in the reference notebook at [henrysky/stellarium_star_catalogs](https://github.com/henrysky/stellarium_star_catalogs) includes a clause that explicitly selects stars without BP/RP:

   ```sql
   WHERE (x.bp_rp IS NOT NULL AND x.v_mag > 15.5 AND x.v_mag <= 16.75)
      OR (x.bp_rp IS NULL AND x.phot_g_mean_mag > 15.5 AND x.phot_g_mean_mag <= 16.75)
   ```

   This suggests bp/rp-missing stars should be included — but the downstream processing drops them anyway. It is unclear whether this is intended behavior.

2. **SkyChart `.dat` parallax/PM accuracy**: Compared to Gaia CSV raw data, about 80% of stars whose parallax, pmra, or pmdec is expressed in `e-n` scientific notation (absolute value < 1e-3) appear magnified by exactly 10^n in the `.dat` files. Example: `Gaia DR3 6603941581186299520` — CdC shows parallax 4.222 mas, but the ADQL result is 0.00004221 mas, exactly a 1e5× difference. Roughly 2 million out of 1.8 billion CdC stars are affected. RA, Dec, and V magnitude are unaffected. I haven't found the root cause in CdC's data processing scripts, and the ~80% rather than 100% rate is puzzling. The old SkyChart-to-cat converter is retained as a fallback.

3. **Cross-validation against existing level 7 catalog**: After downloading the full Gaia DR3 CSV and converting with my pipeline, comparison against the existing `stars_7` catalog shows:
   - 900k+ additional stars (those missing BP/RP, dropped in the existing catalog)
   - Only 414 stars placed in a different (and probably adjacent) zone
   - Apart from the 900k+ extra stars, the total star count differs by just 1 (a single star allocated to level 6 or 8 instead of 7)
   - The discrepancies are consistent with C++ vs Python rounding differences. Otherwise the results match exactly.
   - B-V was not compared. The existing catalogs generate B-V from Gaia DR3 XP continuous spectra using `GaiaXPy` synthetic photometry (as described in the [reference pipeline's README](https://github.com/henrysky/stellarium_star_catalogs)), which requires processing ~15 TB of spectral data.

### `DynamicZoneArray` — and an unexpected finding

The real reason `DynamicZoneArray` eliminated stuttering was not the async I/O itself. `SpecialZoneArray`'s `draw()` blocks on `mmap` to retrieve the entire zone data, then checks the first star's visibility. I had inadvertently added a pre-check in `DynamicZoneArray`: before reading the zone from disk, verify whether the level's brightest star (`mag_min`) is visible; if not, return immediately without touching I/O.

This simple optimization is extremely effective. I've since refactored the code (CRTP `ZoneArrayImpl`) to reuse `SpecialZoneArray`'s logic with minimal duplication — a side effect of sharing the draw path was that `SpecialZoneArray` also gained the pre-check — and its stuttering disappeared as well. From the perspective of eliminating UI lag, `DynamicZoneArray` is no longer strictly necessary. I'm keeping it, however, because it avoids `ZoneData` — whose size grows exponentially with level, reaching ~320 MB at level 10. `DynamicZoneArray` saves that memory.

---

Please let me know if you have any feedback or questions regarding these finding.

gzotti:
Thanks for your efforts. I hope @henrysky can comment on the catalog differences, and you find a solution!

10110111:
> From the perspective of eliminating UI lag, `DynamicZoneArray` is no longer strictly necessary. I'm keeping it, however, because it avoids `ZoneData` — whose size grows exponentially with level, reaching ~320 MB at level 10.

Now in what cases will it be used? For catalogs with level>N for some N?

src/core/modules/StarMgr.cpp
```diff
@@ -500,9 +500,9 @@ bool StarMgr::checkAndLoadCatalog(const QVariantMap& catDesc, const bool load)
 			else
 			{
 #if QT_VERSION >= QT_VERSION_CHECK(6, 4, 0)
-				md5Hash.addData(QByteArrayView(reinterpret_cast<const char*>(cat), static_cast<int>(cat_sz)));
+				md5Hash.addData(QByteArrayView(reinterpret_cast<const char*>(cat), static_cast<qsizetype>(cat_sz)));
```
10110111, line 509: These explicit casts of integers aren't even necessary. Why were they here? @gzotti?

gzotti: I am afraid I'd have to study the full git history for that answer. I can only guess after almost 3 years that either Qt6.4 refused to compile or at least a syntax change to use QByteArrayView was suggested by the CLang checker in QtCreator, and keeping the use of int in the length specification just worked without issue. Nowadays qsizetype (presumably uint64 in most cases?)  is to be preferred, like size_t instead of int elsewhere.

SceneryMC:
> Now in what cases will it be used? For catalogs with level>N for some N?

After thinking it through, I ended up removing `DynamicZoneArray` entirely. The memory savings are now implemented independently in `SpecialZoneArray`, and the FPS/memory usage numbers barely changed in my tests, so `DynamicZoneArray` turned out to be unnecessary.

As a result, the scope of this PR shifted: it's now mainly about the star catalog generation scripts, the early-return optimization, and a handful of fixes that become more noticeable at level 9/10. I've rewritten the PR title and description to reflect this.

The remaining main work includes updating guide.pdf and adding support for the star3 format to the generation scripts. The PR is close to done.

henrysky:
> This suggests bp/rp-missing stars should be included — but the downstream processing drops them anyway. It is unclear whether this is intended behavior.

This is the intended behaviour to grab the data (the we potentially need downstream) first from Gaia Archive and filter/process them later. I agree that we should consider NOT dropping them just because lacking of colour information. Filling `B-V=0` by default would not be ideal, something like `B-V=1` would be better. The rationale is those dim stars are most likely low mass K/M dwarfs.

I am not familiar with SkyChart so I cannot comment on their catalogs. Is there any reason why we now have to convert SkyChart data file instead of grabbing directly from the Gaia Archive? I think it would be better to grab the data from Gaia directly instead of from SkyChart (and I assume our licenses allow it?).

Another consideration is Gaia DR4 release date (Dec 2, 2026) is less than 150 days from now. Should we wait a few months for DR4? Gaia DR3 was released on Jun 13, 2022 but Skychart only added Gaia DR3 on January 17, 2024 (at least from what I can see, correct me if I am wrong). Does that mean we have to wait Skychart to add DR4 before we can update that Stellarium catalogs?

Thanks for this PR and it looks great! Let me know if you need anything from me.

gzotti:
I don't follow closely enough: Are Gaia IDs stable from DR3 to DR4? Skycultures recently started to use "DR3 xxx" IDs. Else I'd also prefer to not reprocess CdC data (adding a dependency) but work from the original ESA data if there is not a good reason.
Should the missing B-V be marked with a special value that "could" be recognized as "unknown" for the InfoString? _Displaying_ it with the right color for B-V=1 sounds reasonable.

SceneryMC:
Yes, I also think using original ESA data is better. In fact, my latest local catalog is generated this way and no longer depends on CdC's `.dat` files at all. The reason I used CdC's `.dat` before was that storage space was insufficient and I happened to have CdC's `dat` files, which was a preliminary attempt; later I managed to get enough storage space.

Should I remove the CdC `dat` processing functionality from the current catalog generation script?

Additionally, I modified the catalog generation script and Stellarium to support missing B-V:
1. The generated` Star2/Star3` format .cat files map missing B-V to the maximum positive value of the field. For `Star2` this is 32.767 mag, and for `Star3` this is 5+ mag; neither can be reached by a real star.
2. When Stellarium `draw()` reads the maximum positive field value, it replaces it with `stars/missing_bv_value` from the `.ini` file, defaulting to 1.0.
3. For `InfoString`, the maximum positive field value is displayed as `--`.

henrysky:
> I don't follow closely enough: Are Gaia IDs stable from DR3 to DR4? Skycultures recently started to use "DR3 xxx" IDs.

I would expect majority of Gaia IDs to be stable from DR3 to DR4.

> Should I remove the CdC dat processing functionality from the current catalog generation script?

If you are not using it, then I think you can remove the CdC .dat processing script as none of us is using it.

Issue #4983, opened 2026-06-18
I believe that Stellarium should avoid using controversial border maps as much as possible to prevent unnecessary trouble.


sunshuwei:
I believe that Stellarium should avoid using controversial border maps as much as possible to prevent unnecessary trouble. Astronomy software should remain as detached from political disputes as possible. For historical Sky Cultures, I think historical maps should be used whenever possible. Using modern maps can cause unnecessary complications when territorial disputes are involved (for instance, being banned from distribution in certain countries due to border issues).
For example, China and India have territorial disputes. If Arunachal Pradesh is mapped into India according to current maps, China will be dissatisfied; if Aksai Chin is mapped into China, India will be dissatisfied. Just as it is highly reasonable for the Korean Sky Culture to encompass both North and South Korea, by the same logic, the Chinese cultural area should include both the PRC (Mainland China, Hong Kong, Macau) and the ROC (Taiwan)—similar to how the "China" entry on the German Wikipedia ( https://de.wikipedia.org/wiki/China) is handled. Alternatively, this could be managed by using historical maps from different eras.
Additionally, regarding the Tibetan Sky Culture, the distribution of the Tibetan ethnic group is not limited to the Tibet Autonomous Region; Tibetans are also distributed across Qinghai, northwestern Yunnan, western Sichuan, southern Gansu, as well as Bhutan, Sikkim, and Ladakh.

<img width="1714" height="927" alt="Image" src="https://github.com/user-attachments/assets/8f5afe19-3488-4fd4-bca3-c02b248cf641" />

10110111:
Indeed, the use of political borders goes contrary to #1844.

Guanjin0562:
I notice that many Western sky cultures seem to have drawn on historical political boundaries (e.g., the Roman Empire). Should Chinese sky culture perhaps follow a similar approach?

In fact, there is some justification for using historical territories—traditional Chinese skycultures was not confined to the core Chinese cultural regions, but also extended to the frontier areas of several dynasties. For instance, a Han‐period star chart has been unearthed in Xinjiang, and Qing‐era star maps survive in Inner Mongolia. During the Yuan dynasty, Guo Shoujing conducted astronomical observations at the four extreme points of the Mongol–Yuan Empire; the northernmost station was located near latitude 64° N on the Yenisei River (as determined by the altitude of the North Pole above the horizon).

[...]

PR #4994, opened 2026-06-21
Added territory.geojson for Chinese SCs

Guanjin0562:
Added `territory.geojson` files for several Chinese historical dynasties. I am not a historian—these boundaries were drawn based on historical maps published in China, and I do not guarantee their accuracy.

We should avoid all modern territorial disputes. For this reason, I set the end date for the maps of China, Korea, and Japan to the early 20th century, so as not to display modern borders. Until a better solution is found, I ask that no one arbitrarily change these end dates to the present. In addition, I have temporarily removed the modern Indian territories, as they are disputed; these should instead be left for others to map historically, without showing modern boundaries.

In China, using controversial modern maps can lead to serious consequences such as website bans, which are irreversible. Although we do not care about politics, we must avoid such controversies.

My work may be a bit rushed because the deadline is too close, and I hope to resolve similar conflicts as soon as possible to avoid problems with the new version.


Guanjin0562:
Some Chinese sky cultures should not add to map, such as the "single" type of Xianglin and modern(Journey_to_the_West). Modern(Chinese) cannot be added either, as it will directly involve modern boundary.

#4983

gzotti:
Thanks a lot! 

The polygons should show where the connected skyculture was followed. A missing map invites addition. The Xianglin polygon could at least show the location where the map comes from. Journey to the West could show the area where this story is commonplace and known to "everybody". This may include current borders or raw lines through today's countries as needed. But if today's politics is as stupid as it is, do what you think is best. 

You can leave a remark/comment in the territory file "do not change, politically sensitive" or so. Just utilize an otherwise unused key like "caution".

sushoff:
> Some Chinese sky cultures should not add to map, such as the "single" type of Xianglin and modern(Journey_to_the_West). Modern(Chinese) cannot be added either, as it will directly involve modern boundary.
> 
> #4983

actually, we tried to avoid that for JTW but if you think it is better not to show it, that may be fine ,t oo

sushoff:
> Thanks a lot!
( ...)
> You can leave a remark/comment in the territory file "do not change, politically sensitive" or so. Just utilize an otherwise unused key like "caution".

GREAT idea!

Guanjin0562:
Added comments in geojson and also added map polygons for Xianglin SC. But two modern SCs in China cannot add maps anymore.

skycultures/tibetan/territory.geojson
```diff
@@ -4,18 +4,21 @@
   "features": [
    {
     "type": "Feature",
+    "comment": "This includes the traditional Tibetan cultural area: all of China's Tibet Autonomous Region, parts of Qinghai, Xinjiang, Sichuan, Yunnan, and Gansu, as well as the whole of Bhutan and Ladakh in India. Due to complex territorial disputes in the Himalayas, Kashmir, and elsewhere, areas after 1911 are temporarily not shown.",
     "properties": {
-	"name": "Tibet/China", 
+	"name": "Qinghai-Tibet_Plateau(青藏高原)", 
 	"beginTime": 600,
-	"endTime": 9146,
+	"endTime": 1911,
```
gzotti, line 11: which is a pity, as those observations/interviews were from 2019 (~today), when Tibetan authorities had at least (again?) power over their traditional calendar. So, the SC and territory files are now disjunct.

sunshuwei: As an ethnic group, the transnational distribution of Tibetans extending into modern times shouldn't involve any political issues. If @Guanjin0562 has concerns, I think it should be fine if @gzotti extracts the data and submits it instead.

Guanjin0562: If you think it is necessary, could you please make the changes yourself afterwards? I set the end temes for several SC regions to the early 20th century to indicate that "I wish to avoid any possible controversy." Although the extent of the Tibetan area is not problematic per se, the fact that its cultural region spans multiple countries still concerns me. If you believe that the Tibetan, Manchu, Korean, etc., are not an issue, please go ahead and change their dates to 9146 yourself—it is still important to distinguish who made the revisions. Just keep China's borders and don't extend beyond 1911.

gzotti: My first polygon for Tibet was made from a published polygon set of the regions of China. I am well aware the Tibetan culture area has different borders and that political censors will disagree likely with whatever cultural area we would propose, but undisputed today's borders should be OK for them. I did not check these latest files. You mean I can just change back the end year number?

Guanjin0562:
Regarding the map polygons for Xianglin SC: according to expert opinion, it is not reasonable to delineate its circulation/distribution range, because we cannot verify its precise locations or extent of dissemination (for example, we know it also circulated in Japan, but obviously Japan should not be included). In the interest of rigor, and in the absence of a unified standard for what constitutes "circulation," we have decided to delete it.
At the same time (1634), the Chinese official authorities also published a star catalog. This catalog should be taken as the representative of China's official SC for the period 1634–1756, and we should create a new SC for it and add corresponding map polygons.

skycultures/tibetan/territory.geojson
sunshuwei: Administrative_Division_Tibet_(PRC_claimed) svg
China and India have disputes over tens of thousands of square kilometers of pale yellow land on the map, and even went to war in the 1960s.

Guanjin0562: Yes, you can keep the boundaries of the cultural district and adjust the year yourself.

Guanjin0562: Actually, I am very interested in Tibetan SC. I have studied Tibetan alphabet and also tried to collect information. I still haven't found any star chart, but I have discovered the spelling and meanings of the twenty-eight mansions in Tibetan, which seem to have different names in different sources, similar to the Latin alphabet names you provided.

image

gzotti: It's time we make an obfuscated world map with rectangular tiles for Eastern Asia that show nothing in detail, just to please politicians. 😵‍💫

gzotti: Maybe you can improve my lunar station identification for a later release. The topology of the actual asterisms must be identical to what's visible on the observing platform, but maybe 2-3 are in reality totally different from what I found.

sushoff: > It's time we make an obfuscated world map with rectangular tiles for Eastern Asia that show nothing in detail, just to please politicians. 😵‍💫

I should have had this idea a year ago, when I discussed the map with with the SCE developer and foresaw this issue (exactly with China's borders)

sushoff:
> Regarding the map polygons for Xianglin SC: according to expert opinion, it is not reasonable to delineate its circulation/distribution range, because we cannot verify its precise locations or extent of dissemination (for example, we know it also circulated in Japan, but obviously Japan should not be included). In the interest of rigor, and in the absence of a unified standard for what constitutes "circulation," we have decided to delete it. At the same time (1634), the Chinese official authorities also published a star catalog. This catalog should be taken as the representative of China's official SC for the period 1634–1756, and we should create a new SC for it and add corresponding map polygons.

yes - that is the general issue with books and written cultures: these pieces are used beyond their region of creation and their time. For Arabic Al-Sufi and the Almagest (written in Africa, later used in Europe for more than a millennium), I think, we now agreed that we will only indicate the area of origin. How far it spread ... well ... 

Keep in mind that these polygons are only an indicator for the user to show where it stems from. The Stellarium star simulation can display the names and constellations at all times and places on Earth. :-)

gzotti:
> For Arabic Al-Sufi and the Almagest (written in Africa, later used in Europe for more than a millennium), I think, we now agreed that we will only indicate the area of origin. How far it spread ... well ...

Almagest: I'd say starting in the Roman Empire 2nd ct AD.  Derivatives where applicable: Sufi: Islamic world of that time. Later successors: "Christian Europe", as far as literature from antiquity was circulated. Distribution was of course not down to the peasant, but scholarly or monastic circles.

sushoff:
> > For Arabic Al-Sufi and the Almagest (written in Africa, later used in Europe for more than a millennium), I think, we now agreed that we will only indicate the area of origin. How far it spread ... well ...
> 
> Almagest: I'd say starting in the Roman Empire 2nd ct AD. Derivatives where applicable: Sufi: Islamic world of that time. Later successors: "Christian Europe", as far as literature from antiquity was circulated. Distribution was of course not down to the peasant, but scholarly or monastic circles.

that's exactly what I did. I take the Roman Empire in the borders from 20 to 395 CE as origin, but I leave the rest undefined. In the end, Stellarium's function is to simulate the sky and the Sky Culture Explorer is only meant to give an idea where this image culture *stems from*. If we wanted to also display where it was additionally used/ varied/ adopted... this would become a research+visualisation project on its own. Interesting, no doubt, but beyond our current capacity (and scope).

gzotti (approved):
OK from my side, thanks for various important considerations.

gzotti:
I changed the Tibetan territory file now, and modified the description with an explicit remark: "This map shows a cultural region and does not show or imply political borders."

PR #5003, opened 2026-06-23
Protect the whole jpl_pleph with the mutex

### LINKED ISSUE #4993: Flickering planets + labels when additional threads are used
Using the latest master https://github.com/Stellarium/stellarium/commit/b9a0a6c4f153f552a35c84a01990a0bde0e40134 on Fedora 44, I see some flickering when using more than 0 threads. See the video below.

https://github.com/user-attachments/assets/9cdadb58-b6ad-49d7-b4af-6d2658b1779c

Possibly introduced by #4950.

Fedora 44, KDE 6.7.0
Qt 6.11.1
Nvidia GeForce RTX 4070 Ti Super, driver version 595.80

10110111:
While `jpl_state()` is not thread-safe, the use of the whole `jpl_eph_data` isn't thread-safe either, and it's accessed by `jpl_ephem()`.

Although `jpl_ephem()` is called from `GetDe440Coor`, `GetDe441Coor`, etc., these functions aren't used concurrently, so there's no issue with protecting `jpl_ephem()` itself by a global JPL mutex.

The `jpl_state()` itself is only ever used from `jpl_ephem()`, so it doesn't really need any additional protection.

Fixes #4993

### DIFF
--- a/src/core/planetsephems/jpleph.cpp
+++ b/src/core/planetsephems/jpleph.cpp
@@ -151,6 +151,8 @@ double DLL_FUNC jpl_get_double(const void *ephem, const int value)
 int DLL_FUNC jpl_pleph(void *ephem, const double et, const int ntarg,
                       const int ncent, double rrd[], const int calc_velocity)
 {
+    QMutexLocker locker(&mutex);
+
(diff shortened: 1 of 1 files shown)


alex-w (approved):

gzotti (approved):
Right, thanks!!

Issue #4979, opened 2026-06-16
Improvement Suggestions for SCM & SCE


sushoff:
The Sky Culture Maker (SCM) and the Sky Culture Explorer (SCE) were 2025 student projects at the University of Jena. The results were tested before release, but there are known issues or wishes for future improvements that only turned out after more tests & practical usage. Here we collect items. 

- [ ] SCE geo map was designed to show only selected data (enough orientation for the user, but not overload the software with nonastronomical content). Thus, some islands, rivers, lakes ... might be missing. Here we collect cases where we would need them (e.g. if their SC is added). 

-  Anutan 
- 

- [ ] SCM we need to add the feature to edit an existing SC
- [ ] SCM

youlaafifah:
> The Sky Culture Maker (SCM) and the Sky Culture Explorer (SCE) were 2025 student projects at the University of Jena. The results were tested before release, but there are known issues or wishes for future improvements that only turned out after more tests & practical usage. Here we collect items.
> 
> * * [ ] 
>     [ ]  SCE geo map was designed to show only selected data (enough orientation for the user, but not overload the software with nonastronomical content). Thus, some islands, rivers, lakes ... might be missing. Here we collect cases where we would need them (e.g. if their SC is added).* Anutan* * [ ] 
>     [ ]  SCM we need to add the feature to edit an existing SC* [ ] 
>     [ ]  SCM

In adding geopolygon of Maori, anutan, balinese, and tongan, some small islands are missing:

- Chatham islands for maori migration;
- No anutan island because it's so tiny;
- Nusa lembongan and Nusa Penida islands for Balinese migration;
- missing islands between Neiafu and Tongatapu for tongan migration.
- many islands shape from SCM different with GIS data when I export to online map like mapOG  and google earth.
- In some cultures, I have set the start time to a value less than 0 on the time layer for example, -900 for 900 BC, but the slider and the minimum year for the final display only go up to 0 and cannot go below that.

xLPMG:
If nobody has started working on it yet, I'd have a look at the SC edit feature :)

gzotti:
The SCM can now edit SCs. Does adding SC territories (#5062) close this?

PR #5019, opened 2026-07-01
Add sky culture loader to SCM

xLPMG:
This PR extends the SCM with the functionality of editing existing sky cultures by loading them.
It fixes part of #4979

### Type of change

- [ ] Housekeeping

### Testing

I tested this on MacOS 26.5.1. Please test this on Windows and Linux.

### Checklist:


xLPMG:
@gzotti @10110111 @alex-w 

The SCM needs to parse the index.json in order to load the sky culture into the SCM. I decided not to reinvent the wheel and simply use the `read` method from `Constellation.cpp`, BUT I had to make it public for that. Now my question to you: is this acceptable? Do you have any other idea? At last, I can always simply copy the function over, but then we have two places that need to be adjusted if anything about the index.json changes.

xLPMG:
I encountered one more issue. From the Constellation.cpp, I tried to read the native name and pronounciation with `Constellation.getNameNative();` and `Constellation.getNamePronounce`. But these are the same value! Looking into `Constellation.hpp`, we can see: `QString getNamePronounce() const override {return (culturalName.pronounceI18n.isEmpty() ? culturalName.native : culturalName.pronounceI18n);}`

It checks for `pronounceI18n` and returns the native name if it doesnt exist. But in the `read` function, `pronounceI18n` is never set. Only the `pronounce` field is being read: `culturalName.pronounce = names["pronounce"].toString().trimmed();`.

So my question is: why doesnt `getNamePronounce` return `culturalName.pronounce`? Why is the `pronounce` value thats used in index.json files ignored by the reader?

gzotti:
I think you could make SCM friend of Constellation/ConstellationMgr, and if necessary add a getCulturalName method that returns the raw CulturalName struct. I had to modify various getters because some name parts like pronounce are optional, but there must be something for display.

10110111:
> I decided not to reinvent the wheel and simply use the read method from Constellation.cpp, BUT I had to make it public for that. Now my question to you: is this acceptable? Do you have any other idea?

I don't see any reason to protect `class Constellation` from being created in other places than its friend `class ConstellationMgr`. This `private` declaration dates back to 2006 (c1f8fdfcd8c1d1189b4fafc66b11f93912601b2e), and the commit message doesn't explain why it has to be so.

So I guess the constructor & destructor can be made public.

10110111:
> I think you could make SCM friend of Constellation/ConstellationMgr

No, this breaks modularity. No core part of Stellarium should ever know anything about plugins' implementations, especially at the level of befriending them.

10110111:
> But in the read function, `pronounceI18n` is never set.

Generally it's supposed to be set by `ConstellationMgr::updateI18n` I think.

xLPMG:
Thank you for the fast and useful feedback!

> if necessary add a getCulturalName method that returns the raw CulturalName struct

Yes, that is a better way than to copy the individual fields. I totally forgot that ScmCulturalName inherits from StelObject::CulturalName.

plugins/SkyCultureMaker/src/ScmSCLoader.hpp
```diff
@@ -0,0 +1,110 @@
...
+		Authors,
+		About,
+		Acknowledgements,
+		License
+	};
+
+	static inline Section sectionOf(const QString &heading)
+	{
+		const QString h = heading.toLower().trimmed();
+		if (h == "introduction") return Section::Introduction;
```
gzotti: You should use L1S("introduction") etc (defined in StelUtils.hpp)  for comparing static string views.

gzotti:
Seems to work. However: After loading a SC from the program directory, the SC loads, but the main program view dialog's SC tab still shows the 'Modern' SC loaded. Probably just a missing connect with a notification SkycultureChanged or such? Editing then seems to have worked.

EDIT: Tiny addition: If user has no skycultures subdir in the USER_DATA_DIR, consider creating it if necessary. I know, it can be created on the spot in the file explorer.  But in this otherwise overprotective plugin, maybe even this should be automatic.

xLPMG:
> Seems to work. However: After loading a SC from the program directory, the SC loads, but the main program view dialog's SC tab still shows the 'Modern' SC loaded. Probably just a missing connect with a notification SkycultureChanged or such? Editing then seems to have worked.

I'm not sure thats what we want. The SCM draws its own constellations anyway, why would we want to set it in the ViewDialog? What I've done now is: when loading with the SCM, the sky culture (like modern) that was shown in stellarium via ViewDialog is hidden while the SCM is active. After the SCM is closed, this is restored and the sky culture is shown again. When a sky culture is exported via SCM, it is set as the active one.

What i have also fixed now:
- a skycultures folder is created automatically if not existent
- exporting a sky culture that already existed was not possible. eg "load -> edit -> overwrite" was not possible. I've added a dialog for this asking the user if they want to overwrite the existing SC
- the message dialogs looked weird on MacOS.  this was bc we supplied a "parent" widget instead of always using "&StelMainView::getInstance()". this is fixed now too.

plugins/SkyCultureMaker/src/SkyCultureMaker.cpp
```diff
@@ -141,6 +142,38 @@ void SkyCultureMaker::setActionToggle(const QString &id, bool toggle)
 	}
 }
 
+// Core constellation-display actions hidden while the editor draws its own sky culture.
+static const QStringList coreConstellationActions = {
+	"actionShow_Constellation_Lines",
+	"actionShow_Constellation_Art",
+	"actionShow_Constellation_Labels",
+	"actionShow_Constellation_Boundaries",
+};
```
gzotti, line 155: This should then also affect the Asterism actions. .-)

xLPMG:
Ah one more TODO which I will adress in a future PR:

Writing a new SC is a step by step process, roughly
-> save description file
-> save illustrations 
-> save index.json

and if any of these steps fail, the process is aborted and all files are deleted, so no partial or broken SC is written.

The overwrite mechanic is currently:

1) Check if SC exists, if yes and user wants to overwrite, delete original data
2) Start SC export

If the SC export then fails and aborts, the original data is lost (because it was deleted in step 1).

I would improve it in a way that when the user wants to overwrite an existing SC, the new SC is first created as a temp folder and only if the creation succeeds, the old SC is deleted and replaced by the newly exported SC. So a failed export does not delete the existing SC.

gzotti:
Works for me in a brief test. @alex-w , @10110111 ?  One small improvement may be to move the sky to center on an existing constellation that the willing editor double-clicks in the list. (Don't select constellation, just move the sky.)

alex-w (changes requested):
Obvious issues: 
- `QObject::tr()` -> `q_()`
- Please add `ScmSCLoader.cpp` into `po/stellarium/POTFILES.in` file

alex-w (approved):

alex-w:
Thanks for quick fixes

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
```diff
@@ -381,16 +395,16 @@ QPolygonF ScmGeoLocGraphicsView::convertViewToWGS84(const QPolygonF &viewCoordin
...
 		result.append(QPointF(xInMeter, yInMeter));
 	}
 
 	// convert map coordinates to Lat/Lon coordinates (EPSG: 4326)
 	for (auto &point : result)
 	{
-		qreal xInLon= (point.x() * 180.0) / 20037508.3427892439067363739014;
-		qreal yInLat = (std::atan(std::exp(((point.y() * 180.0) / 20037508.3427892439067363739014) * (M_PI / 180.0))) * (360.0 / M_PI)) - 90.0;
+		qreal xInLon = (point.x() * 180.0) / MERCATOR_HALF_WIDTH_PROJ;
+		qreal yInLat = (std::atan(std::exp(((point.y() * 180.0) / MERCATOR_HALF_WIDTH_PROJ) * (M_PI / 180.0))) * (360.0 / M_PI)) - 90.0;
```
10110111: So many nested parentheses here. Could've used `M_PI_180` and `M_180_PI` from `StelUtils.hpp` to avoid some of them (and to get `360/M_PI` you won't need any parentheses when multiplying `M_180_PI` by 2, since multiplication by 2 is exact).

And the parentheses in expressions like `(a * b) / c` and `(a * b) - c` are simply useless, since multiplication would be done first anyway.

All this said, this expression could be simplified to
```
std::atan(std::exp(point.y() * 180.0 / MERCATOR_HALF_WIDTH_PROJ * M_PI_180)) * 2 * M_180_PI - 90.0
```

plugins/SkyCultureMaker/src/types/License.hpp
```diff
@@ -53,7 +53,9 @@ enum class LicenseType
 	CC_BY,
 	CC_BY_NC,
 	CC_BY_ND,
-	CC_BY_NC_ND
+	CC_BY_NC_ND,
+	CC_BY_SA,
+	GNU_GPL_V2
```
10110111: Since you need to extend this from time to time, please add a comma after the last item. Then a subsequent addition will just be a `+` line in the diff.

10110111:
The `Constellations` section appears to be appended into the SC `Description` field ("Culture Description"). Why? Don't we have a special field for this in the Constellation Editor?

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
xLPMG: Hm I don't agree on the parenthesis part.. just because the result stays the same, doesn't make them simply useless. To me they added a lot of readability. But thats very subjective of course, so I've changed it.

xLPMG:
> The `Constellations` section appears to be appended into the SC `Description` field ("Culture Description"). Why? Don't we have a special field for this in the Constellation Editor?

Fixed, they're being extracted now. I also thought it would be good to notify the user about which headings the tool couldnt recognize instead of silently appending to the sky culture description. A good example to test this is the Aztec SC for example.

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
```diff
@@ -395,16 +396,16 @@ QPolygonF ScmGeoLocGraphicsView::convertViewToWGS84(const QPolygonF &viewCoordin
...
 		result.append(QPointF(xInMeter, yInMeter));
 	}
 
 	// convert map coordinates to Lat/Lon coordinates (EPSG: 4326)
 	for (auto &point : result)
 	{
-		qreal xInLon = (point.x() * 180.0) / MERCATOR_HALF_WIDTH_PROJ;
-		qreal yInLat = (std::atan(std::exp(((point.y() * 180.0) / MERCATOR_HALF_WIDTH_PROJ) * (M_PI / 180.0))) * (360.0 / M_PI)) - 90.0;
+		qreal xInLon = point.x() * 180.0 / MERCATOR_HALF_WIDTH_PROJ;
+		qreal yInLat = std::atan(std::exp(point.y() * 180.0 / MERCATOR_HALF_WIDTH_PROJ * M_PI_180)) * 2 * M_180_PI - 90.0;
```
gzotti: This is actually:

qreal yInLat = std::atan(std::exp(point.y()  * M_PI / MERCATOR_HALF_WIDTH_PROJ)) * 2 * M_180_PI - 90.0;

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
gzotti: I agree in parts about readability. In some cases parentheses may however prevent precomputation of constants. E.g. consider  

qreal xInLon= (point.x() * 180.0) / 20037508.3427892439067363739014;

This requires two computations at runtime. 

The same expressed as:

qreal xInLon= point.x() * (180.0 / 20037508.3427892439067363739014);

can process the division at compile time.

xLPMG:
> One small improvement may be to move the sky to center on an existing constellation that the willing editor double-clicks in the list. (Don't select constellation, just move the sky.)

Good idea :) In a new PR or just add it here?

gzotti:
If it's easy, just add it here while fixing the (hopefully) last review inputs. If you have no time to look into this now, keep it for later. We should try to have this merged ASAP, then fix the begin/endyear mess (#5070), then add the (cleaned-up) territory polygons (#5062).

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
xLPMG: Ah right, that makes sense

xLPMG:
> If it's easy, just add it here while fixing the (hopefully) last review inputs. If you have no time to look into this now, keep it for later. We should try to have this merged ASAP, then fix the begin/endyear mess (#5070), then add the (cleaned-up) territory polygons (#5062).

I've had the fix ready already but wasn't sure if i want to bloat this PR more. But it wasn't much.

plugins/SkyCultureMaker/src/gui/ScmGeoLocGraphicsView.cpp
10110111: > The same expressed as:

> qreal xInLon= point.x() * (180.0 / 20037508.3427892439067363739014);

> can process the division at compile time.

Right, but so would
```
qreal xInLon = 180 / 20037508.3427892439067363739014 * point.x();
```
I did suggest the use of `M_PI_180` and `M_180_PI` though, which does use compile-time-computed constants here.

And of course, there's no real issue with this particular example, since there's only one set of parentheses without nesting. But trying to match `(((` to the closing parentheses is a bit too much for a human when it can be avoided.

gzotti: Right. In any case, keep the constants visible, and don't precombine constants manually with a pocket calculator into something incomprehensible. Algorithmic shortening out of "*180/180" or re-sorting factors to allow the compiler to optimize constants should be done, though. A few brackets for added clarity won't bother me much, but yes, that's also a matter of taste.

10110111:
> I also thought it would be good to notify the user about which headings the tool couldnt recognize instead of silently appending to the sky culture description. A good example to test this is the Aztec SC for example.

In the Aztec SC the extra headings are level-3+ headings, so they must be put under the corresponding level-2 section, not simply into _Description_. Particularly, the "Thanks to" and "Conclusions" are subsections of the _Authors_ section, so should be retained there (even though "Conclusions" is out of place there, the SCM can't analyze logic, so should preserve the structure the best it can).

10110111:
I got a crash "corrupted doubly-linked list" in the following scenario.

1. Load Aztec SC (from the source tree)
2. Open _The Market_ constellation to edit
3. Move the reference stars on the third tab (I moved all three)
4. (Maybe switch through the tabs 3-2-1-2-3, I'm not sure if it matters)
5. Press Save

The crash doesn't happen reliably when Stellarium runs freely, but on the next run under Valgrind I did get the following error:
```
==1409977== Invalid write of size 1
==1409977==    at 0x12CA4D4: ScmConstellationDialog::close() (ScmConstellationDialog.cpp:160)
==1409977==    by 0x12CC2DF: ScmConstellationDialog::saveConstellation() (ScmConstellationDialog.cpp:500)
==1409977==    by 0x6FE950A: void doActivate<false>(QObject*, int, void**) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Core.so.6.9.3)
==1409977==    by 0x52BE585: QAbstractButton::clicked(bool) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BE81D: QAbstractButtonPrivate::emitClicked() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BFF05: QAbstractButtonPrivate::click() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52C0132: QAbstractButton::mouseReleaseEvent(QMouseEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x51F389D: QWidget::event(QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x51916AA: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x519AC97: QApplication::notify(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x6F85FB1: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Core.so.6.9.3)
==1409977==    by 0x5199871: QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool, bool) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==  Address 0x43de8c40 is 0 bytes inside a block of size 664 free'd
==1409977==    at 0x484BB6F: operator delete(void*, unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1409977==    by 0x131ADBF: operator() (unique_ptr.h:85)
==1409977==    by 0x131ADBF: reset (unique_ptr.h:182)
==1409977==    by 0x131ADBF: operator= (unique_ptr.h:167)
==1409977==    by 0x131ADBF: operator= (unique_ptr.h:212)
==1409977==    by 0x131ADBF: operator= (unique_ptr.h:371)
==1409977==    by 0x131ADBF: __remove_if<__gnu_cxx::__normal_iterator<std::unique_ptr<scm::ScmConstellation>*, std::vector<std::unique_ptr<scm::ScmConstellation> > >, __gnu_cxx::__ops::_Iter_pred<scm::ScmSkyCulture::removeConstellation(const QString&)::<lambda(const std::unique_ptr<scm::ScmConstellation>&)> > > (stl_algo.h:824)
==1409977==    by 0x131ADBF: remove_if<__gnu_cxx::__normal_iterator<std::unique_ptr<scm::ScmConstellation>*, std::vector<std::unique_ptr<scm::ScmConstellation> > >, scm::ScmSkyCulture::removeConstellation(const QString&)::<lambda(const std::unique_ptr<scm::ScmConstellation>&)> > (stl_algo.h:894)
==1409977==    by 0x131ADBF: scm::ScmSkyCulture::removeConstellation(QString const&) (ScmSkyCulture.cpp:66)
==1409977==    by 0x12CBA58: ScmConstellationDialog::saveConstellation() (ScmConstellationDialog.cpp:486)
==1409977==    by 0x6FE950A: void doActivate<false>(QObject*, int, void**) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Core.so.6.9.3)
==1409977==    by 0x52BE585: QAbstractButton::clicked(bool) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BE81D: QAbstractButtonPrivate::emitClicked() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BFF05: QAbstractButtonPrivate::click() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52C0132: QAbstractButton::mouseReleaseEvent(QMouseEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x51F389D: QWidget::event(QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x51916AA: QApplicationPrivate::notify_helper(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x519AC97: QApplication::notify(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x6F85FB1: QCoreApplication::notifyInternal2(QObject*, QEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Core.so.6.9.3)
==1409977==  Block was alloc'd at
==1409977==    at 0x4849013: operator new(unsigned long) (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==1409977==    by 0x1319CC4: make_unique<scm::ScmConstellation, const QString&, const std::vector<scm::ConstellationLine, std::allocator<scm::ConstellationLine> >&, bool const&> (unique_ptr.h:962)
==1409977==    by 0x1319CC4: scm::ScmSkyCulture::addConstellation(QString const&, std::vector<scm::ConstellationLine, std::allocator<scm::ConstellationLine> > const&, bool) (ScmSkyCulture.cpp:55)
==1409977==    by 0x130F1E4: ScmSCLoader::parseIndexJsonConstellations(QJsonArray const&, QDir const&, scm::ScmSkyCulture*) (ScmSCLoader.cpp:193)
==1409977==    by 0x1310599: ScmSCLoader::parseIndexJson(QDir const&, scm::ScmSkyCulture*, QString*) (ScmSCLoader.cpp:116)
==1409977==    by 0x1312015: ScmSCLoader::loadFromDirectory(QDir const&, QString*, QList<QString>*) (ScmSCLoader.cpp:57)
==1409977==    by 0x1312664: ScmSCLoader::selectAndLoad(QWidget*, QString const&, QString*, QList<QString>*) (ScmSCLoader.cpp:76)
==1409977==    by 0x12F72D1: ScmStartDialog::startScmEditingProcess() (ScmStartDialog.cpp:137)
==1409977==    by 0x6FE950A: void doActivate<false>(QObject*, int, void**) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Core.so.6.9.3)
==1409977==    by 0x52BE585: QAbstractButton::clicked(bool) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BE81D: QAbstractButtonPrivate::emitClicked() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52BFF05: QAbstractButtonPrivate::click() (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
==1409977==    by 0x52C0132: QAbstractButton::mouseReleaseEvent(QMouseEvent*) (in /home/ruslan/opt/qt6.9.3/lib/libQt6Widgets.so.6.9.3)
```

10110111:
So opening seems OK now. But I think the "open, edit, save" sequence needs some testing.

10110111:
Also, in this branch I get the following warning in the log:

```
[    82.276][WARN] QObject::connect(StelMovementMgr, NomenclatureMgr): invalid nullptr parameter
```

Maybe it's just because something has been fixed in `master`, so it may be worth merging `master` into this branch once again.

gzotti:
Yes, that was caused by bad initialisation sequence. But actually years ago we learned that 'git rebase master' is cleaner.

10110111:
Sure rebase is cleaner, but given that we are going to squash this PR
anyway, there's not much difference.

xLPMG:
issue should be fixed now :)

PR #5020, opened 2026-07-01
Update messier_marathon script

martignoni:
### Description
Update messier_marathon script:

- Add missing descriptions.
- Save and restore initial state.
- Delete a unused variable.
- Rename a variable.

### Type of change

- [x] Housekeeping

## Checklist:

### DIFF
--- a/scripts/messier_marathon.ssc
+++ b/scripts/messier_marathon.ssc
@@ -6,8 +6,9 @@
 //          Alexander Wolf (alex-w)
 //          Nicolas Martignoni (martignoni)
 // Description: Messier Marathon on March 20, 2020. The data for the script are taken from
-//              GaoXing's observation log: http://www.xjltp.com/2_huodong/2_030301.htm (now defunct).
+//              GaoXing's observation log (now defunct): http://www.xjltp.com/2_huodong/2_030301.htm.
 //              Missing descriptions added by Nicolas Martignoni.
+
 // Change Log:
 // 2019-09-23   ultrapre@github.com
 //     * First version
@@ -193,7 +194,7 @@ var PasteText = [tr("Difficulty is 0"),
 		 tr("Too big, scattered, not easy to verify"), 
 		 tr("Bigger than M6"), 
 		 tr("Easy to locate and verify"),
-		 tr("Medium size, easy to verify"), tr("Small, difficult"),
+		 tr("Small, not easy to verify"), tr("Small, difficult"),
 		 tr("Faint, check the Y shape of the star pattern"),
 		 tr("Brighter than imagined, can be detected with 60 binos"), 
 		 tr("Darker than M97, long strips, look hard"), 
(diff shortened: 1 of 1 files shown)


scripts/messier_marathon.ssc
```diff
@@ -1,14 +1,30 @@
 //
 // Name: Messier Marathon
-// Author: ultrapre@github.com
-// Version: 1.2
+// Version: 1.3
 // License: Public Domain
-// Description: Messier Marathon on March 20, 2020. The data for the script are taken from GaoXing's observation log: http://www.xjltp.com/2_huodong/2_030301.htm
+// Authors: ultrapre@github.com (Original author)
```
alex-w: Please use list of author with comma delimiter

scripts/messier_marathon.ssc
```diff
@@ -1,14 +1,30 @@
...
+//          Alexander Wolf (alex-w)
+//          Nicolas Martignoni (martignoni)
+// Description: Messier Marathon on March 20, 2020. The data for the script are taken from
```
alex-w, line 6: I'm not sure for correct extracting text for translations. What about one line text?

scripts/messier_marathon.ssc
```diff
@@ -162,37 +178,30 @@ var PasteText = [tr("Difficulty is 0"),
...
-		 tr("Easy to locate and verify"), "", "", "", 
+		 tr("Easy to locate and verify"),
+		 tr("Small, not easy to verify"), tr("Small, difficult"),
+		 tr("Faint, check the Y shape of the star pattern"),
 		 tr("Brighter than imagined, can be detected with 60 binos"), 
 		 tr("Darker than M97, long strips, look hard"), 
 		 tr("Easy to verify with more than 20 times binos"), 
 		 tr("Difficult to locate, bright, easy to verify"), 
-		 tr("In the morning light, it looks too hard, dark"), ""];
-var showtime = ["19:30","19:40","19:45","19:47","19:47","19:47","19:48",
```
alex-w, line 181: showtime isn't used?

martignoni: No, it is not. I've tested the script without it, and the variable does not seem to be used.

gzotti: Has it been used originally? It seems to be the timeline to observe the objects in a manageable temporal sequence, which may have been the original intention.

martignoni: AFAIK, the script sets the current observation time from `NowTime` array and display this time in a label `nowlabel` (line 267), so there's no need of a second array for the time.

This way, an adequate temporal sequence is indeed set.

scripts/messier_marathon.ssc
martignoni: Done

scripts/messier_marathon.ssc
martignoni: I'm afraid I don't know. Can you point me a reference where I could learn that?

martignoni: If this can help: the description is correctly extracted for displaying in the GUI, see screenshot.
script-mm

gzotti:
I think this needs further changes:

- setFlagLandscapeSetsLocation(false)  should be used to keep the current observer location and shows a sequence you'd be able to follow. Else you only get the marathon run from Guereins, France, and all the logic that comes later that adjusts for latitude is meaningless. What happens could probably be also explained in a short introductory paragraph.
- That very logic that adjusts for latitude applies time shifts to the sequence of observations. These should be applied before the first object is shown. I.e., apply all time corrections, sort list by time, then start displaying. Maybe it would make things easier to handle if the three 110-entry-lists would be joined into a single list of 110 triplets [timestamp, object, description]. 
- Also, the displayed descriptions are often cut off. Is there a label width cutoff, or other limitation? (Probably related to a hiDPI screen!)

martignoni:
@gzotti I agree that the current script is not optimal (but for this PR I made minimal changes to it, for adding the missing translations).

I've adapted this script for the needs of my local observatory. It's more maintainable than the current one. If you think it's better, it's OK for me if you want to use it in Stellarium instead (see attachment).

In fact my version sets the landscape (and the associated location).

[ependes-messier-marathon.ssc.txt](https://github.com/user-attachments/files/29611950/ependes-messier-marathon.ssc.txt)

PR #5022, opened 2026-07-02
Add option to reposition toolbars in other corners

### LINKED ISSUE #685: Change toolbar position
I would like to have the option to change the toolbar position. I'm teaching Stellarium in a classroom and the students in the back can not see the toolbar in the lower part of the screen.
I am currectly using the Windows 10 magnifier to show the toolbar at the top of the screen.

Thanks in advance.

### LINKED ISSUE #4783: Change Toolbar Position
I recall seeing a planetarium show where the presenter used Stellarium and a spheric mirror projection system, where the horizontal toolbar was at the top of the laptop screen (and hence not visible to the audience in the dome). (The accompanying status info – Location, FOV, Date & Time, etc – was not showing.) However I cannot see how to achieve this in the config.ini file, and it’s not mentioned in the User Guide. Please can you advise.

xLPMG:
I came across issue #4783 and thought "why not?". I decided to give it a try and got this working version:

The "Tools" tab in the Configuration dialog has two checkboxes which let the user move the toolbars to the top, to the right or both. This makes it possible to position the toolbars in any of the four corners, which should be enough customization.

Fixes #685 #4783

Screenshot 2026-07-03 at 00 51 49

Screenshot 2026-07-03 at 00 52 26

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Manually on MacOS 26.5.1

## Checklist:


src/gui/StelGuiItems.hpp
```diff
@@ -363,7 +371,7 @@ class StelBarsFrame : public QGraphicsPathItem
 	public:
 		StelBarsFrame(QGraphicsItem* parent);
 		//! defines a line around the two button bars
-		void updatePath(BottomStelBar* bottom, LeftStelBar* left);
+		void updatePath(BottomStelBar* bottom, LeftStelBar* left, bool barOnTop=false, bool barOnRight=false);
```
10110111: I don't think the default parameter values would help in any use case, will they? It seems they will only override user settings if a function call omits them.

gzotti:
Not sure if we really need that. OTOH it was requested twice in 7 years, so OK. Just that the names become increasingly weird. BottomStelBar sounds like a design decision from 2002 (?). Should we think about renaming a few GUI classes, e.g.:
- DialogMenuBar for what used to be on the left vertical.
- ToolMenuBar for what used to be on the bottom horizontal.

Have you tested how this works together with Oculars plugin GUI (top-right corner), and with the Calendars screen output (bottom-right corner)? What about InfoText output, is that shifted down? Does it move up when the 'BottomStelBar' (now at top) retracts?

The checkmarks in the config GUI might look better as Comboboxes: Dialog buttons [Left Edge/Right Edge], Tool Buttons [Bottom Edge/Top Edge]

xLPMG:
> Not sure if we really need that. OTOH it was requested twice in 7 years, so OK. Just that the names become increasingly weird. BottomStelBar sounds like a design decision from 2002 (?). Should we think about renaming a few GUI classes, e.g.:
> 
> * DialogMenuBar for what used to be on the left vertical.
> * ToolMenuBar for what used to be on the bottom horizontal.
> 
> Have you tested how this works together with Oculars plugin GUI (top-right corner), and with the Calendars screen output (bottom-right corner)? What about InfoText output, is that shifted down? Does it move up when the 'BottomStelBar' (now at top) retracts?
> 
> The checkmarks in the config GUI might look better as Comboboxes: Dialog buttons [Left Edge/Right Edge], Tool Buttons [Bottom Edge/Top Edge]

Yes the names are a little weird now, but I wanted to keep the changes minimal. I interpreted it like "this tool bar was named BottomStelBar because that where it lives by default and it keeps this name, even if the user decides to optionally show it at the top instead". If you are fine with more code changes, then renaming won't be a problem.

It fully overlaps the Oculars plugin, but I do not see this as an issue. If the user wants to use the Oculars GUI, then they shouldn't put the toolbar in the same corner if they want to use it correctly. They can always turn off the Oculars GUI or choose any of the three other corners for the toolbars.

I tested the Calendar plugin now and only the vertical bar slightly overlaps, but as far as i can see it, the vertical bar also overlaps the info text even in its default position (bottom left) on smaller screens.

With the latest commit, the info panel shifts down and retracts when the horizontal bar is at the top. It also moves to the right if the vertical bar is at the top.

As for the comboboxes: sure, changed this in the latest commit.

gzotti:
OK, renamings can be made in a separate commit later. I tried it, and it does work quite well. However, when vertical bar is right, the mouse-over context help (tooltip) is not visible. I even like bottom-right as it avoids the InfoText collision. But the collision with the Oculars plugin button bar should be handled in Oculars. Not sure @alex-w @10110111, just enable a warning in the Oculars config GUI that main GUI should be moved to any other corner? Or can the button panel be moved out of the corner easily? The actual collision looks bad and will cause troubles.

Combobox is better, but side-by-side please, not stacked, to make better use of screen space. Or actually, as we in fact have only four settings, a single combobox in the upper area (Planetarium Options, bottom-right I see space for it) and a single enum property could be enough: Menu Bars: [0=bottom-left, 1=bottom-right, 2=top-left, 3=top-right].

alex-w:
I see issue for pixel distance between text and borders on toolbar (please see text when toolbar on bottom and on top).

xLPMG:
I do like your suggestions of DialogMenuBar, ToolMenuBar and calling them "menu bars" together, but the term "toolbars" is already used throughout the project to refer to both of them, so I'd rather just stay consistent with the current code for now. (or rename everything consistenly but thats extra effort and complexity)

alex-w:
> OK, renamings can be made in a separate commit later. I tried it, and it does work quite well. However, when vertical bar is right, the mouse-over context help (tooltip) is not visible. I even like bottom-right as it avoids the InfoText collision. But the collision with the Oculars plugin button bar should be handled in Oculars. Not sure @alex-w @10110111, just enable a warning in the Oculars config GUI that main GUI should be moved to any other corner? Or can the button panel be moved out of the corner easily? The actual collision looks bad and will cause troubles.

Oculars GUI panel created a bit hackish and I'm not sure that it can be moved easy... In any wat currently top-right variant give ugly look.

> Combobox is better, but side-by-side please, not stacked, to make better use of screen space. Or actually, as we in fact have only four settings, a single combobox in the upper area (Planetarium Options, bottom-right I see space for it) and a single enum property could be enough: Menu Bars: [0=bottom-left, 1=bottom-right, 2=top-left, 3=top-right].

Hmm... Are you remember for hidden font settings (available for advanced users)? :) But yes, the new options should be more compact.

gzotti:
> Hmm... Are you remember for hidden font settings (available for advanced users)? :) But yes, the new options should be more compact.

There is empty space right of "Screen button scale". The potentially hidden font area is above.

gzotti:
We could make drawing the Oculars GUI depend on toolbar corner not being top-right. It's not a setting you change all the time. If you use the plugin, just use another corner, or the Oculars GUI toolbar is not available.

gzotti:
Here is a first attempt to move the Oculars panel out of the way.

```
diff --git a/plugins/Oculars/src/gui/OcularsGuiPanel.cpp b/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
index 609928f6ac..deb9c63166 100644
--- a/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
+++ b/plugins/Oculars/src/gui/OcularsGuiPanel.cpp
@@ -19,7 +19,9 @@ Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.

 #include "Oculars.hpp"
 #include "OcularsGuiPanel.hpp"
+#include "SkyGui.hpp"
 #include "StelApp.hpp"
+#include "StelGui.hpp"
 #include "StelGuiItems.hpp"
 #include "StelTranslator.hpp"
 #include "StelActionMgr.hpp"
@@ -240,6 +242,10 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
        //Night mode
        connect(&stelApp, SIGNAL(colorSchemeChanged(const QString&)), this, SLOT(setColorScheme(const QString&)));
        setColorScheme(stelApp.getCurrentStelStyle());
+
+
+       StelGui *gui=dynamic_cast<StelGui*>(StelApp::getInstance().getGui());
+       connect(gui, SIGNAL(toolbarCornerChanged(int)), this, SLOT(updatePosition()));
 }

 OcularsGuiPanel::~OcularsGuiPanel()
@@ -311,6 +317,7 @@ void OcularsGuiPanel::foldGui()

 void OcularsGuiPanel::updatePosition()
 {
+       const double cornerRadius = 12.0;
        updateGeometry();
        /*qDebug() << "Widget:" << size()
                << "Buttonbar:" << buttonBar->size()
@@ -320,11 +327,17 @@ void OcularsGuiPanel::updatePosition()
                << "Layout" << mainLayout->geometry();*/
        qreal xPos = parentWidget->size().width() - size().width();
        qreal yPos = 0;
+       const int buttonBarCorner=StelApp::getInstance().getStelPropertyManager()
+               ->getStelPropertyValue("StelGui.toolbarCorner").toInt(); // (0=BL, 1=BR, 2=TL, 3=TR)
+       if (buttonBarCorner ==3)
+       {
+               yPos += (cornerRadius + static_cast<SkyGui*>(parentWidget)->getLeftBarWidth());
+               xPos -= (cornerRadius + static_cast<SkyGui*>(parentWidget)->getBottomBarHeight());
+       }
        setPos(xPos, yPos);

        //Update border/shading
        QPainterPath newBorderPath;
-       double cornerRadius = 12.0;
        QPointF verticalBorderStart = geometry().topLeft() + QPointF(-0.5,0.5);
        QPointF horizontalBorderEnd = geometry().bottomRight() + QPointF(-0.5,0.5);
        QPointF cornerArcStart(verticalBorderStart.x(),
@@ -334,6 +347,7 @@ void OcularsGuiPanel::updatePosition()
        newBorderPath.arcTo(cornerArcStart.x(), cornerArcStart.y(), cornerRadius, cornerRadius, 180, 90);
        newBorderPath.lineTo(horizontalBorderEnd);
        newBorderPath.lineTo(horizontalBorderEnd.x(), verticalBorderStart.y());
+       newBorderPath.closeSubpath();
        borderPath->setPath(newBorderPath);
 }

```

(@10110111 I don't know why, but I cannot apply the new connect syntax here. --> EDIT: Solved)

10110111:
> I don't know why, but I cannot apply the new connect syntax here.

Really?.. This compiles perfectly well for me, both in Qt 5 and 6:

```cpp
    StelGui *gui=dynamic_cast<StelGui*>(StelApp::getInstance().getGui());
    connect(gui, &StelGui::toolbarCornerChanged, this, &OcularsGuiPanel::updatePosition);
```

gzotti:
Whatever my system did not like. No idea, I had tried like the 500+ (?) connections I changed last week...  yes now the same works.

gzotti:
Any further work or comments here?

gzotti:
@alex-w, @10110111 any comments?

alex-w:
I don't like movement the bottom toolbar on the top, because it introducing the messy in the GUI, but looks like someone want it. IMHO, the PR may be accepted.

10110111:
> Here is a first attempt to move the Oculars panel out of the way.

Isn't this (or a variation of it) still waiting to be pushed to this PR?

gzotti:
Yes, I had expected @xLPMG to use/modify/push the patch. It's a working solution, may still further improved though. The patch avoids collision of the Oculars panel, but most users would just not use top-right main menu plus oculars menu. I just want to progress small things. We have 1 month for the longer PRs.

xLPMG:
> Yes, I had expected @xLPMG to use/modify/push the patch. It's a working solution, may still further improved though. The patch avoids collision of the Oculars panel, but most users would just not use top-right main menu plus oculars menu. I just want to progress small things. We have 1 month for the longer PRs.

Hi, sorry, I was caught up in exams the past weeks but I'm free now. I'll take a look at the patch asap.

xLPMG:
I've tried to solve it dynamically so the oculars panel moves together with the toolbars but this didn't work: the panel was too close to the toolbars so when hovering over an oculars panel button, it triggered the toolbar which then expanded and moved the oculars panel, making it impossible to actually click something.

I've applied the static patch and only switched getLeftBarWidth and getBottomBarHeight.

Personally I'm not a fan of this because it now looks somewhat out of place but I suspect 99% of users wont use this configuration (top right toolbar and oculars). Remember that the original feature request in #685 was not just simply a preference but a specific case where students couldnt see the bottom of the screen from the back of the room. I think for this, such a solution is totally acceptable :)

gzotti:
> I've applied the static patch and only switched getLeftBarWidth and getBottomBarHeight. 

Oopsie :-) (It was 33°C on that day, but good you have fixed this...)

But now I see another issue, the tiny box/triangle icons in the corner keys that control auto-slide seem to have vanished.

xLPMG:
Screenshot 2026-08-18 at 18 26 26

I still have them. The issue came after the last commit for you?

gzotti:
I think so, but must check again. Later tonight...

plugins/Oculars/src/gui/OcularsGuiPanel.cpp
```diff
@@ -240,6 +242,10 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
 	//Night mode
 	connect(&stelApp, SIGNAL(colorSchemeChanged(const QString&)), this, SLOT(setColorScheme(const QString&)));
 	setColorScheme(stelApp.getCurrentStelStyle());
+
```
gzotti, line 246: To clean up a bit, one blank line should be enough.

src/gui/ConfigurationDialog.cpp
```diff
@@ -528,6 +534,29 @@ void ConfigurationDialog::updateDateTimeDisplayFormat()
...
+{
+	QComboBox* cb = ui->toolbarCornerComboBox;
+	const int cur = cb->count() > 0 ? cb->currentIndex()
+	                                : StelApp::getInstance()
+	                                          .getStelPropertyManager()
+	                                          ->getStelPropertyValue("StelGui.toolbarCorner")
+	                                          .toInt();
+	cb->blockSignals(true);
+	cb->clear();
+	cb->addItem(q_("Bottom-left"));
```
gzotti: To make life easier for translators, I'd add context:

```
cb->addItem(qc_("Bottom-left", "GUI alignment"));
cb->addItem(qc_("Bottom-right", "GUI alignment"));
cb->addItem(qc_("Top-left", "GUI alignment"));
cb->addItem(qc_("Top-right", "GUI alignment"));
```

gzotti:
Don't know what happened before, now the icons returned. After cleanup it's OK for me now!

gzotti (approved):
Good to go!

10110111:
So what do we do with the overlap? I get this currently:

stellarium-020

10110111:
Also, shouldn't the background of the panel icons be inverted vertically when the panel is at the top? Otherwise the rounded corners are near the edge while they were supposed to be on the farther side.

stellarium-021

gzotti:
> So what do we do with the overlap? I get this currently:
> stellarium-020

Ah, with top-left on a too narrow screen this is what you get. In this case, bottom left/right is better.

gzotti:
> Also, shouldn't the background of the panel icons be inverted vertically when the panel is at the top? Otherwise the rounded corners are near the edge while they were supposed to be on the farther side.


Indeed, they have asymmetrical round corners!

xLPMG:
i didnt even notice the rounded corners before.. fixed now. as for the oculars panel: I went the "correct" route here and its moved down dynamically as soon as the rect overlaps. the same behaviour now also applies to the calendars panel.

It is computationally more expensive but as far as I understand it, this is only done when the position needs to be updated, so only during window resizes and when the toolbar is moved.

10110111:
Top-right doesn't behave right:

stellarium-020

gzotti:
Currently it is not intended to move automatically, just avoid collision with slid-out bars. It is usable in this config, but may not be pretty. Most users who need Oculars will likely use one of the 3 other corners, or keep the bars locked visible.

10110111:
Ah, I didn't realize that the horizontal shift is just to give way for the vertical panel. Could we maybe then just move the Oculars panel to the left in this case, so that it at least was attached to a side, rather than hang in the air?

(I wonder though how its motion is even implemented, given that it's in a plugin; need to check the code...)

gzotti:
There is no motion implemented for now. :-) Really, users of Oculars likely select another setting to avoid this hanging panel.

10110111:
The buttons are shifted too much down in the top-* orientation, see for comparison:

stellarium-022
stellarium-021

10110111:
> There is no motion implemented for now.

I mean the Oculars panel does get shifted to fit the main panels. This is what I'm wondering about.

plugins/Oculars/src/gui/OcularsGuiPanel.cpp
```diff
@@ -240,6 +242,9 @@ OcularsGuiPanel::OcularsGuiPanel(Oculars* plugin,
...
+	StelGui *gui=dynamic_cast<StelGui*>(StelApp::getInstance().getGui());
+	connect(gui, SIGNAL(toolbarCornerChanged(int)), this, SLOT(updatePosition()));
```
10110111: Is the old-style connect required here and in other places of this diff? If not, please use the modern function-pointer-based overload.

New code shouldn't use string-based connect unless absolutely necessary.

10110111:
Other than the connect style, the code looks OK to me.

10110111:
Though, the buttons on the vertical panel could be shifted farther away from its top/bottom edge, so that their placement is the same as in current `master` by default.

plugins/Oculars/src/gui/OcularsGuiPanel.cpp
xLPMG: it was easy enough to fix so i did it for the whole file.

xLPMG:
> Though, the buttons on the vertical panel could be shifted farther away from its top/bottom edge, so that their placement is the same as in current `master` by default.

this was trickier.. I suppose the gap on master originally was more like a bug. The "gap" isn't actually padding, it's an invisible helpLabel (the ones that appear next to the buttons on hover). because it's a child item, it got iterated before the actual buttons in the layout loop and reserved a slot, therefore shifting them down and creating the gap. this also resulted in the problem you mentioned earlier where the buttons were shifted too far down when the bar was at the top. 

I fixed this originally by removing the invisible help label from the layout to make the top/bottom orientation more consistent but that fully removed the gap. I've added it back now by starting the vertical bar buttons at positive or negative one button gap (9.5 * buttonSizeRatio) instead of at yPos=0.

10110111:
>  I suppose the gap on master originally was more like a bug. The "gap" isn't actually padding, it's an invisible helpLabel

Hmm, OK. Let it be as you've done then.

10110111 (approved):

alex-w (approved):

PR #5029, opened 2026-07-06
TelescopeControl: fix LX200 protocol incompatibilities

rumengb:
The changes are well explained in the comments.

### Description
The current implementation misses some dialects of the LX200 protocol 

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:

## Checklist:

### DIFF
--- a/plugins/TelescopeControl/src/Lx200/Lx200Command.cpp
+++ b/plugins/TelescopeControl/src/Lx200/Lx200Command.cpp
@@ -366,71 +366,40 @@ int Lx200CommandSyncSelected::readAnswerFromBuffer(const char *&buff,
 		p++;
 	}
 
-	switch (first_byte)
+	// Unlike ":MS#" (goto), the Meade ":CM#" (sync) command does not return a
+	// numeric status. On success it replies with a '#'-terminated string - the
+	// name of the object the telescope was synchronised to (e.g. "M31#"), a
+	// message such as "Coordinates matched.#", or simply "OK#" as sent by
+	// INDIGO's LX200 emulation. Accept any '#'-terminated response as success
+	// and consume it. A bare '0' (used by some controllers) is still accepted
+	// as a single-byte answer.
+	if (first_byte == '0')
 	{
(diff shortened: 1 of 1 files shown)


alex-w (approved):
The code looks OK, but I can test it on hardware

plugins/TelescopeControl/src/Lx200/Lx200Command.cpp
```diff
@@ -366,71 +366,40 @@ int Lx200CommandSyncSelected::readAnswerFromBuffer(const char *&buff,
...
+	}
 
-		default:
-			#ifdef DEBUG4
-			*log_file << Now()
-				  << "Lx200CommandSyncSelected::readAnswerFromBuffer: "
-				     "sync returns something weird"
-				  << StelUtils::getEndLineChar();
-			#endif
+	for (;;p++)
```
10110111, line 387: What is the actual change in the logic here? It seems that you've effectively only removed the `buf++;` from the case when `first_byte` is `'1'` or `'2'` (assuming the debug macros are not defined). Is there anything else that was incompatible with something in the handling of this command?

rumengb:
Sorry I am on the phone. And I can see just a small snippet. But I think
you are taking about the part where different controllers reply differently
to the command. Some reply with 0 other with OK# some give object name
before # others even reply with whole sentences before #. This change  is
supposed to make it very permissive. As it was rejecting valid responses.
As far as I remember I explained it well in the comment above.

plugins/TelescopeControl/src/Lx200/Lx200Command.cpp
10110111: OK, I was under the impression that `1` and `2` were special answers that needed some special processing, and your change removed this. Now I see that actually they were supposed to start a `#`-terminated string, and with your change every `#`-terminated response is made valid.

10110111 (approved):

PR #5030, opened 2026-07-06
TelescopeControl: add TCP transport for LX200 protocol

rumengb:
This PR adds new feature - ability to connect to LX200 mounts over TCP/IP

### Description
Lately there are LX200 mounts that can be accessed over WIFI (TCP/IP) like ML Astro SAL33, ZWO AM3/5/7 and basically all OnStep nased mounts. So adding this functionality is a nice addition to Stellarium. Also indigo_mount_agent offers LX200 translation layer over TCP and all indigo mounts can be accessed via the common LX200 commands. TCP or Serial connection can be chosen when the telescope is created or modified. The current serial transport is unchanged TCP/IP is added as an alternative. Plugin help page is updated accordingly.

Fixes # (issue)

Screenshot From 2026-07-04 14-05-57

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

Tested with OnStepX mount and INDIGO lx200 server.
Slew/sync/abort/ connect disconnect all work as expected.

**Test Configuration**:
* Operating system: Ububtu 
* Graphics Card: AMD Radeon

## Checklist:

### DIFF
--- a/plugins/TelescopeControl/src/Lx200/Lx200Connection.cpp
+++ b/plugins/TelescopeControl/src/Lx200/Lx200Connection.cpp
@@ -7,6 +7,8 @@ It also contains sample server classes (dummy, Meade LX200).
 Author and Copyright of this file and of the stellarium telescope library:
 Johannes Gajdosik, 2006
 
+TCP transport support added by Rumen Bogdanovski, 2026.
+
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU General Public License
 as published by the Free Software Foundation; either version 2
@@ -31,7 +33,8 @@ Foundation, Inc., 51 Franklin Street, Suite 500, Boston, MA  02110-1335, USA.
 #include <cmath>
 
 
-Lx200Connection::Lx200Connection(Server &server, const char *serial_device) : SerialPort(server, serial_device)
+template <class Transport>
+void Lx200ConnectionImpl<Transport>::initTiming(void)
 {
 	time_between_commands = 0;
 	next_send_time = GetNow();
@@ -42,16 +45,17 @@ Lx200Connection::Lx200Connection(Server &server, const char *serial_device) : Se
 //! Resets the connection.
 //! Removes all commands in the queue without executing them and
 //! cleans both the read and the write buffers.
-void Lx200Connection::resetCommunication(void)
+template <class Transport>
+void Lx200ConnectionImpl<Transport>::resetCommunication(void)
 {
 	while (!command_list.empty())
 	{
 		delete command_list.front();
 		command_list.pop_front();
 	}
-	
-	read_buff_end = read_buff;
-	write_buff_end = write_buff;
+
+	this->read_buff_end = this->read_buff;
+	this->write_buff_end = this->write_buff;
 	#ifdef DEBUG4
 	*log_file << Now() << "Lx200Connection::resetCommunication" << StelUtils::getEndLineChar();
 	#endif
@@ -60,13 +64,14 @@ void Lx200Connection::resetCommunication(void)
 	next_send_time = GetNow() + 10000000;
 	read_timeout_endtime = 0x7FFFFFFFFFFFFFFFLL;
 	goto_commands_queued = 0;
-	static_cast<TelescopeClientDirectLx200&>(server).communicationResetReceived();
+	static_cast<TelescopeClientDirectLx200&>(this->server).communicationResetReceived();
 }
 
 //! Commands the telescope to slew to the given right ascension and declination.
 //! Converts the coordinates and queues the necessary sequence of Lx200Command
 //! objects to perform the slew.
-void Lx200Connection::sendGoto(unsigned int ra_int, int dec_int)
+template <class Transport>
+void Lx200ConnectionImpl<Transport>::sendGoto(unsigned int ra_int, int dec_int)
 {
 	if (goto_commands_queued <= 1)
 	{
@@ -84,10 +89,10 @@ void Lx200Connection::sendGoto(unsigned int ra_int, int dec_int)
 		int ra = static_cast<int>(std::floor(0.5 + ra_int * (86400.0/4294967296.0)));
 		if (ra >= 86400)
 			ra -= 86400;
-		sendCommand(new Lx200CommandStopSlew(server));
-		sendCommand(new Lx200CommandSetSelectedRa(server, ra));
-		sendCommand(new Lx200CommandSetSelectedDec(server, dec));
-		sendCommand(new Lx200CommandGotoSelected(server));
+		sendCommand(new Lx200CommandStopSlew(this->server));
+		sendCommand(new Lx200CommandSetSelectedRa(this->server, ra));
+		sendCommand(new Lx200CommandSetSelectedDec(this->server, dec));
+		sendCommand(new Lx200CommandGotoSelected(this->server));
 		goto_commands_queued++;
 	}
 	else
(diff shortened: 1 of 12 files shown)


gzotti:
You can and should just keep working on a PR and add further commits (occasionally rebasing/squashing intermediate commits), not close and PR the next one minutes later. It will keep discussions available and makes tracking progress easier (actually possible).

rumengb:
> You can and should just keep working on a PR and add further commits (occasionally rebasing/squashing intermediate commits), not close and PR the next one minutes later. It will keep discussions available and makes tracking progress easier (actually possible).

Sorry about that! There was not too much going on there anyways. 
What do you want me to do about that?

gzotti:
Just keep working in this PR now, and be patient, we have daytime jobs. When after one of us had already checked like 10 of 12 files you close and start over, that invested review time is lost.  As said in the previous one, I cannot really test it without hardware and deeper code insight, but you need to update the User Guide as well. This is the instruction reference users need to rely upon.  @10110111 detected your attempt to "fix" the */° issue in the first of these PRs. After half a year of receiving several complaints I had to revert what had seemed to be closer to the specs to the current state. No repetition of that necessary. 

But we'd have need for a thorough plugin review. Most of the code is from before 2010, made by previous team members with a previous generation of GOTO devices. Meanwhile GOTO protocols have changed, there are azimuthally mounted GOTOs (IMO fully inconceivable 20 years ago...), auto-plate-solving, etc., this was all not accounted for when the Stellarium telescope communication protocol was devised.  also, the current draw() queries some isAlive which sends a request over serial or TCP for connected status. Once per frame. I think this should be loosened a bit, like once per second, and the connected status cached. Without hardware, I won't be able to test the consequences of such change, though. This was reported when a native Alpaca client was contacted by the ASCOM module and caused what looked like too much IP traffic.  So, the necessity of this query is dubious and it may in fact slow down the program, and in total, we need a native Alpaca module (branch for that exists, but is fragmentary.). If you are interested...

rumengb:
I am not sure I have time to work on this, i also have a daytime job and I am a core developer of INDIGO platform which already takes almost all of my free time. My primary motivation for these PRs is to make Stelarrium talk seamlessly to INDIGO. As it does with Sky Safary and other planetarium programs as they have this feature already - lx200 over TCP. This is why we added this protocol adapter to indigo. It saves a lot of trouble. Stellarium does not have it. I discussed this with Alex years ago, but nothing happened no one took the initiative to make it so here am I after many user complaints that it does not work with INDI adapter for many reasons. However I have fixed it with the PR you already accepted, but I wanted to do this in more standard way. Rather than adding indigo dependency and indigo mount support to Stellarium which is even harder to justify ;) and This TCP layer is something all other planetariums have and many mounts have. For me this is no brainier.

What comes to Alpaca we have alpaca adapter in indigo. While developing it we figured out how inappropriate this approach is for control of astronomical equipment. It is full of race conditions by design. It is extremely ineffective. for a single poll of the mount coordinates you have to open connection, request some url with some tokens, receive the response and then close the connection. Imagine doing this for mount, focuser position, camera cooler, etc... 10-15 connects/disconnects per second with at least 1kb per connection. Compared to lx200 over TCP approach: one connection and several bytes per second - a lot faster, cleaner and simpler. So i am not a fan of alpaca.

gzotti:
Ah, I was unaware of your involvement in INDIGO. But yes, such changes best come from people who need it and who know what they are doing...

rumengb:
> @10110111 detected your attempt to "fix" the */° issue in the first of these PRs. After half a year of receiving several complaints I had to revert what had seemed to be closer to the specs to the current state.

This sound like i wanted to commit some crime :) I did it based on my experience  and i had no intention to break Stellarium - In INDIGO we support over 10 dialects of LX200 and we have thousands of users both on Mac and Linux using them and we did not have a single complaint about this. But if you say it happens  -maybe we do not support this particular type of mount. any way... I removed it.. indigo accepts anything as a separator in this case. So let us put this to rest.

rumengb:
> Ah, I was unaware of your involvement in INDIGO. But yes, such changes best come from people who need it and who know what they are doing...

It is questionable if I understand what I am doing in Stellarium :-), but this looked the most straight forward approach to achieve the goal... you had all the pieces in place I just had to glue them together with minimal intrusion...

gzotti:
> This sound like i wanted to commit some crime :)

No, that was not implied. Just memory of a bad attempt to fix something that was not broken, only weird.  Changing back to "weird" stopped the complaints.

10110111:
> But if you say it happens -maybe we do not support this particular type of mount.

I think the reality here is that the mounts don't care what they are fed at this position (INDIGO apparently passes `*`, Stellarium `°`, Stellarium Mobile `:`), while what they reply is checked rigorously by Stellarium. And their reply normally contains `°`. I suppose we could change what we pass here, but I wouldn't dare do this to prevent regressions. So far this has worked well, judging by the amount of issue reports since the revert (none).

10110111:
I've checked this with an LX200 emulator on Ubuntu 22.04, seems to work. Will have to also check on Windows.

10110111:
A few strange things I noticed:

 * Connection type is said as "local, Stellarium", which doesn't generally make much sense for a TCP connection;
 * The label in the radio button under "Telescope controlled by" reads "Stellarium, directly through a serial port", which is not quite correct now.

rumengb:
> A few strange things I noticed:
> 
>     * Connection type is said as "local, Stellarium", which doesn't generally make much sense for a TCP connection;
> 
>     * The label in the radio button under "Telescope controlled by" reads "Stellarium, directly through a serial port", which is not quite correct now.

Sorry for the oversight. Is it ok now?

10110111 (approved):
It works in the Windows version run in Wine, as well as natively on Ubuntu 22.04. Also, I've tested in Ubuntu 22.04 with a virtual serial port (#4787), and it also works as it used to, so I haven't found any regressions.

From my side this looks OK. Thanks for implementing this!

rumengb:
> It works in the Windows version run in Wine, as well as natively on Ubuntu 22.04. Also, I've tested in Ubuntu 22.04 with a virtual serial port (#4787), and it also works as it used to, so I haven't found any regressions.
> 
> From my side this looks OK. Thanks for implementing this!

Thank you! Please review the protocol fixes PR. I made Stellarium parsing more permissive, so that it can work with wider variety of LX200 dialects. Including INDIGO :)

gzotti:
Why is this merged?

rumengb:
> Why is this merged?
> 
>     * [x]  This change requires a documentation update
> 
>     * [ ]  I have updated the respective chapter in the Stellarium User Guide

Is it my responsibility to update the docs? The in app help is updated accordingly... Any way I will see what I can do in the lunch break.

gzotti:
Yes, as specified several times. This is why the checklist entry reminds you. The User Guide resides in the project sources so that developers can immediately tell the users what they have done. If I never connected a TCP LX200 and pressed any necessary buttons found on that device, I cannot write about it. I can help a bit with formatting if needed, not with such technical content around optional hardware. Users may have questions, look into the Guide and see outdated information. They will ask the maintainers, and not PM you.  We will have to say "don't know". Such confusion can be prevented if developers care for users and describe their stuff.

rumengb:
> Yes, as specified several times. This is why the checklist entry reminds you. The User Guide resides in the project sources so that developers can immediately tell the users what they have done. If I never connected a TCP LX200 and pressed any necessary buttons found on that device, I cannot write about it. I can help a bit with formatting if needed, not with such technical content around optional hardware. Users may have questions, look into the Guide and see outdated information. They will ask the maintainers, and not PM you. We will have to say "don't know". Such confusion can be prevented if developers care for users and describe their stuff.

please see - TelescopeControl: Lx200 over TCP documentation update #5035

PR #5031, opened 2026-07-06
fix: use regionKey() for sky culture region matching to avoid CJK substring issues

### LINKED ISSUE #5008: Bug: Indian sky cultures incorrectly categorized in CJK locales
In Chinese, Japanese, and Korean (CJK) locales, Indian sky cultures (e.g., Indian , Indian Nakshatras ) were incorrectly categorized under "South-eastern Asia" (东南亚) instead of "Southern Asia" (南亚). This issue did not occur in English or other Latin-script languages.

### Expected Behaviour

Indian sky cultures should appear under "南亚" (Southern Asia) in CJK locales.

### Actual Behaviour

Indian sky cultures appear under "东南亚" (South-eastern Asia) in CJK locales.

### Root Cause
The code used Qt::MatchContains to find the correct region separator in the sky culture list:
`l->findItems(q_(cultureRegionIt.value()), Qt::MatchContains)` (src\gui\ViewDialog.cpp:1551)
In CJK languages, this creates a substring matching problem:

| English | Chinese | Japanese |
|-------|--------|--------|
| Southern Asia | 南亚 | 南アジア |
| South-eastern Asia | 东南亚 | 東南アジア |

When searching for "南亚" (Southern Asia) with MatchContains , Qt also matched "东南亚" (South-eastern Asia) because the latter contains the former as a substring. This caused all Indian sky cultures to be placed under the wrong region separator.

Additionally, SeparatorListWidgetItem wraps text with decorative characters ( "⸻ " + text + " ⸻" ), so a naive MatchExactly on the raw region name would fail.

### Possible Solutions

The original code:

src\gui\ViewDialog.cpp:1551-1552
```
QString itemName = (l->findItems(q_(cultureRegionIt.value()), Qt::MatchContains).empty()) ? q_("Other") : q_(cultureRegionIt.value());
l->insertItem(l->row(l->findItems(itemName, Qt::MatchContains).first()) + 1, item);
```

replace as:
```
QString translatedRegion = q_(cultureRegionIt.value());
QString separatorText = "⸻ " + translatedRegion + " ⸻";
QList<QListWidgetItem*> foundItems = l->findItems(separatorText, Qt::MatchExactly);
if (foundItems.empty())
{
    foundItems = l->findItems("⸻ " + q_("Other") + " ⸻", Qt::MatchExactly);
}
if (!foundItems.empty())
{
    l->insertItem(l->row(foundItems.first()) + 1, item);
}
else
{
    // Fallback: append to end if no separator found
    l->addItem(item);
}
```

I have checked whether this modification affects other components. The SkyCultureMaker plugin should not be affected because it uses a completely different region handling mechanism (ComboBox with enum values stored as item data, rather than text-based searching with `findItems()` ). However, I am uncertain about potential impacts on other areas of the codebase and would appreciate your review or suggestions for a better fix if available.

Because it's too late for my time zone, I didn't personally submit the pull request.

mmustafasenoglu:
## What this PR does

Fixes incorrect categorization of Indian sky cultures in CJK locales (Chinese, Japanese, Korean).

## Why

In CJK locales, `Qt::MatchContains` creates a substring matching problem:
- "Southern Asia" → "南亚"
- "South-eastern Asia" → "东南亚"

When searching for "南亚" (Southern Asia), Qt also matches "东南亚" (South-eastern Asia) because the latter contains the former as a substring. This causes all Indian sky cultures to be placed under the wrong region separator.

## Solution

Instead of matching against translated strings (which is unreliable), this PR stores the original (untranslated) region key in `SeparatorListWidgetItem` and uses it for matching.

### Changes:
1. **SeparatorListWidgetItem.hpp/cpp**: Added `regionKey` parameter to constructor and `m_regionKey` member with getter
2. **ViewDialog.cpp**: 
   - Pass region key when creating separator items
   - Use `regionKey()` for matching instead of translated text
   - Added fallback to "Other" separator if region not found

This approach is more robust than `MatchExactly` on decorated text because it does not depend on translation strings at all.

## Testing

The fix can be verified by:
1. Switching to a CJK locale (Chinese/Japanese/Korean)
2. Opening Settings → Sky Cultures
3. Verifying that Indian sky cultures appear under "Southern Asia" (南亚/南アジア) instead of "South-eastern Asia" (东南亚/東南アジア)

Fixes #5008

### DIFF
--- a/src/gui/SeparatorListWidgetItem.cpp
+++ b/src/gui/SeparatorListWidgetItem.cpp
@@ -14,13 +14,14 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
 #include "SeparatorListWidgetItem.hpp"
 
-SeparatorListWidgetItem::SeparatorListWidgetItem(const QString &text, QListWidget *parent)
+SeparatorListWidgetItem::SeparatorListWidgetItem(const QString &text, const QString &regionKey, QListWidget *parent)
 	: QListWidgetItem(parent, 1318)
+	, m_regionKey(regionKey)
 {
 	setFlags(Qt::NoItemFlags);
 	setTextAlignment(Qt::AlignCenter);
--- a/src/gui/SeparatorListWidgetItem.hpp
+++ b/src/gui/SeparatorListWidgetItem.hpp
@@ -14,7 +14,7 @@
  * GNU General Public License for more details.
  *
  * You should have received a copy of the GNU General Public License
- * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ * along with this program; if not, see <http://www.gnu.org/licenses/>.
  */
 
 #ifndef SEPARATORLISTWIDGETITEM_HPP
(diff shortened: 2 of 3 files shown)


alex-w (approved):
Thanks!

PR #5041, opened 2026-07-13
Add new tab 'Twilight limits' to Object Visibility plugin

Atque:
This branch adds a new tab to the Object Visibility plug-in; Twilight limits. It shows lines of latitude for the twilight stage limits during summer and winter solstices. For example, if someone is interested in showing the line of limit of the civil twilight in northern hemisphere summer solstice, the line is drawn on the map. Equator, tropic lines and polar circles are also drawn, as are limits of maximum solar altitude during winter solstice (think Longyearbyen, Svalbard during winter solstice with Sun reaching maximum -11.6 degrees). These lines are dashed.

We calculate according to the epoch of the planetarium, meaning we can follow e.g. the polar circle movement across millennia.

Another improvement to the plug-in, is location labels, called from Stellarium's own location database. We filter by population and avoid clutter when zooming out. This applies to both tabs.

Code was generated by AI (Codex). If this is interesting and worth merging, please test it thoroughly. Suggestions for improvement most welcome!

### Screenshots:
bild

bild

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Fedora 44
Qt 6.11.1

## Checklist:

### DIFF
--- a/plugins/ObjectVisibility/src/ObjectVisibility.cpp
+++ b/plugins/ObjectVisibility/src/ObjectVisibility.cpp
@@ -60,7 +60,9 @@ StelPluginInfo ObjectVisibilityStelPluginInterface::getPluginInfo() const
 	info.description = N_("Shows on a planet map where a selected star or "
 	                      "deep-sky object is visible.  Supports observation "
 	                      "from Earth, the Moon, the eight planets, Pluto, "
-	                      "and the four Galilean moons.  Based on the "
+	                      "and the four Galilean moons.  Also shows "
+	                      "Earth-only solstice twilight latitude limits "
+	                      "computed from obliquity of date.  Based on the "
 	                      "geometric criteria of the Astro-Geo-GIS article "
 	                      "\"The 49 brightest stars in the night sky – when "
 	                      "and where can we see them?\"");
--- a/plugins/ObjectVisibility/src/ObjectVisibility.hpp
+++ b/plugins/ObjectVisibility/src/ObjectVisibility.hpp
@@ -57,6 +57,12 @@ Given a star with declination @f$\delta@f$ at the current epoch
 The plugin does **not** handle planets, the Sun, the Moon, asteroids,
 comets or artificial satellites: their motion across the sky is too
 fast for a single static snapshot to be useful.
+
+The Twilight limits tab is Earth-only. It draws the equator, tropics,
+polar circles, and solstice latitudes where the Sun's lowest or highest
+altitude is -6°, -12°, or -18°. These latitudes are computed from
+Earth's obliquity of date, so they remain valid for historical and
+future epochs supported by Stellarium.
 @}
 */
 
--- a/plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
+++ b/plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
@@ -28,6 +28,7 @@
 #include "StelFileMgr.hpp"
 #include "StelGui.hpp"
 #include "StelLocation.hpp"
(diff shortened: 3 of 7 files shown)


10110111:
1. Don't we want to make place labels also available in the Location dialog? This feature would deserve a separate PR.
2. The whole window doesn't fit in my FullHD screen at 100% scaling and 19px fonts (both GUI and screen). In `master` it does.
3. I think the polar circle lines should be more different from the twilight lines, otherwise all these different colors at the same thickness and solidity just mix up and hinder comprehension.
4. Also, the black equator is hardly visible. Maybe make it white/gray?

alex-w:
UI should be compact - currently it very wide. The population filter for locations given wrong results.

Atque:
> 1. Don't we want to make place labels also available in the Location dialog? This feature would deserve a separate PR.
> 
>     2. The whole window doesn't fit in my FullHD screen at 100% scaling and 19px fonts (both GUI and screen). In `master` it does.
> 
>     3. I think the polar circle lines should be more different from the twilight lines, otherwise all these different colors at the same thickness and solidity just mix up and hinder comprehension.
> 
>     4. Also, the black equator is hardly visible. Maybe make it white/gray?

1. Hm, yes indeed. Do you think it's portable from this plug-in, or should I drop it and wait for a PR for the Location dialog?
2. Ah, I'll look into it.
3. Indeed we could alter its appearance in some way.
4. Yes, I also noticed this. It needs a better color.

> UI should be compact - currently it very wide. The population filter for locations given wrong results.

Population filter gives wrong results in what way? We filter also by zooming; try to zoom in if a town is not visible immediately.

Atque:
Suggestion for new UI layout (not pushed yet):

bild

alex-w:
> > The population filter for locations given wrong results.
> 
> Population filter gives wrong results in what way? We filter also by zooming; try to zoom in if a town is not visible immediately.

A quick example - my city has population over 600 thousands (in our database 620000+), but it isn't visible on the map when I selected filter for 500 or 100 thousands.

Atque:
I can't reproduce. Is it visible when you uncheck "Near lines"?

alex-w:
> I can't reproduce. Is it visible when you uncheck "Near lines"?

Ouch... you're right - I missed this option :(

alex-w (approved):

gzotti (approved):
Looks good!

10110111:
I think there should be less space between the legend lines and the corresponding text (about one or two widths of `W` letter). Otherwise e.g. "-18° astronomical" for the Lowest Sun is closer to the line on its right than that on its left, which is wrong.

Atque:
Ah yes, good input. Looks a bit confusing, I should fix that.

Atque:
After latest commit:
bild

10110111:
Yeah, looks better now.

There are a few issues with the Places feature though.

 * "Places" and "Min. population" labels are not aligned well on my Ubuntu 22.04 with 19px fonts. It may be better to simply use the label of the checkbox to read "Places, min. population:", followed immediately by the combobox.
 * Some labels blink as I'm panning the map. Particularly Norilsk (in Russia) disappears too soon if it's moved closer to the top of the image (after a single wheel click to zoom the map), but reappears if the map is then panned left-right, only to disappear again on continued motion.

Atque:
It's already merged so I can't push any new changes, but I could look into these while working on the third tab. The blinking I guess is expected (compare anti-clutter of the ephemeris markers).

10110111:
> The blinking I guess is expected (compare anti-clutter of the ephemeris markers).

No, the Norilsk label has no close neighbours, at least on the Twilight Limits tab, and yet it blinks.

PR #5047, opened 2026-07-20
Add new tab "Twilight map" to Object Visibility

Atque:
This introduces the third tab in the Object Visibility plug-in. It's fairly simple; a map of the Earth with current twilight limits visualized by lines and color shading. We calculate the upper limb of the Sun for the terminator on Earth. On other bodies, we simply use the terminator (no twilight zones, naturally) and calculate from the center of the solar disk. Where the Sun (or Moon, but only from Earth) appears at zenith at the given time, we draw a small symbol.

We update the map every 0.25 seconds, which seems fast enough. This causes some apparent stuttering when running through time and date, but I think that's negligible.

A new button to sync the map between the tabs was added. This is helpful when wanting to compare different tabs for particular areas of the map.

Also, the bug not solved in #5041 with flickering city labels has hopefully been properly solved. At least I can't reproduce it.

Code was generated by Codex (supervised by me ;) ).

### Screenshots:
Twilight limits on Earth:
bild

Terminator on the Moon:
Skärmbild_20260720_012953

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Fedora 44
Qt 6.11.1.

## Checklist:


10110111:
1. Single scroll up to zoom in
2. Pan to the left to see the east part of the map
3. Result on the screenshot

stellarium-022

Atque:
Ouch, that looks bad. Weirdly, right-panning does not cause this.

Atque:
Issue should be fixed now.

10110111:
~1. Why are there no place labels in the sunlit part (see the screenshot below)?~
2. Places<sub>Min. population:</sub> misalignment is still present. Besides, even clicking it as a single label would be more convenient, so once again: please make it a single checkbox label: "Places, min. population:".
3. The filling between the contours is aliased (you can see it in the previous screenshot without the lines, but with the lines it's also visible).
4. On zooming, the filling gets upsampled instead of being redrawn at a new resolution, resulting in a blurry pixelization (again, visible in the previous screenshot).

stellarium-022

alex-w:
> 1. Why are there no place labels in the sunlit part (see the screenshot below)?

Please try uncheck "Near lines"

alex-w:
The naming of legend elements for twilight should be synced between 2 new tabs to reduce the work of translators.

10110111:
> Please try uncheck "Near lines"

Oh! You're right. OK, this isn't an issue then.

Atque:
> 2. PlacesMin. population: misalignment is still present. Besides, even clicking it as a single label would be more convenient, so once again: please make it a single checkbox label: "Places, min. population:". 
>3. The filling between the contours is aliased (you can see it in the previous screenshot without the lines, but with the lines it's also visible). 
>4. On zooming, the filling gets upsampled instead of being redrawn at a new resolution, resulting in a blurry pixelization (again, visible in the previous screenshot).

2. That should be easy.
3. Shouldn't it be?
4. Ah yes, I've noticed that too. I'll see if I can make the res increase on zooming.

10110111:
> Shouldn't it be?

Of course not, it looks ugly, and antialiasing should be just a matter of enabling the corresponding `QPainter` flag. Though, it might just be visible because of the upsampling on zooming.

alex-w (approved):
OK for me, but please reflect the changes in User Guide.

Atque:
Yes, I was planning for a total of 5 tabs, and to write the SUG chapter when it's completed.

t4saha:
It would be nice to have the overlay in the location window map.

Atque:
> It would be nice to have the overlay in the location window map.

Yes, but I don't feel confident enough to touch too much of the main program, especially since I'm using AI. But if anyone would like to port it to the Location dialog after merge, I would welcome it.

Atque:
@gzotti @10110111 Any further opinions on this? I don't want to continue on other tabs before this one's merged.

gzotti:
Actually I would not laugh about writing docs later. These should come as features grow, as integral part of development.

Atque:
> Actually I would not laugh about writing docs later. These should come as features grow, as integral part of development.

Okay, fair enough. Where does it fit, though? 13.13, right after Calendar plugin?

gzotti:
Rather after Time Navigator. It is an UI plugin with more general utility.

10110111:
The filling looks weird in the following screenshots (see the corners of the "smooth rectangle"). I guess it may be acceptable that it gets simply upsampled instead of getting redrawn when zooming, but at the smallest scale it should look perfectly, which it doesn't.

stellarium-020

stellarium-021

gzotti:
In the legend, the values 0/-6/-12/-18 should be right-justified. It may require a separate column, but would look better.

10110111:
> In the legend, the values 0/-6/-12/-18 should be right-justified.

Really? Have you seen this anywhere? Normally the legend symbols are placed adjacent to their meaning (with a small gap of course).

What I'd do here instead is extend the lines to the width of the `night` label, so that the symbol column has a consistent width.

gzotti:
grafik
grafikAnd yes, of course, the legend lines should extend to the same width in the left legend column. 


Another issue:
grafik
grafik

Working on my scaled-display (150%), I notice that the wrap-around map (which I actually find good!) does not show until you pan the map further to the left. Also, I can move (pan) the map upward until 1/3 of the map height is gray at the bottom. Downward motion is not possible.

10110111:
> Working on my scaled-display (150%), I notice that the wrap-around map (which I actually find good!) does not show until you pan the map further to the left. Also, I can move (pan) the map upward until 1/3 of the map height is gray at the bottom. Downward motion is not possible.

This must be a bug in the `MapWidget` component itself, unrelated to this PR. Can you confirm it in the Location Dialog?

gzotti:
Huh? I never saw I could pan and zoom this! Yes, indeed, same issue.

10110111:
>  I never saw I could pan and zoom this!

Since #4927.

10110111:
Filling still looks wrong in the corners, see the screenshots below. Also, the legend lines aren't long enough with 100% scale and 19px fonts (also visible in the screenshots).

stellarium-024

stellarium-025

gzotti:
I meant the numbers should be right-aligned. I.e., one column has the lines, the next has numbers (right'-aligned), the third has the descriptive label (left-aligned).

Atque:
> I meant the numbers should be right-aligned. I.e., one column has the lines, the next has numbers (right'-aligned), the third has the descriptive label (left-aligned).

Like this?

bild

10110111:
Why do the contours change in two steps when I change month? E.g. I change month 7→8, and the contours shift to the left together with the subsolar and sublunar points, and then shift back and get a bit higher.

Atque:
I can't see anything strange. Going from 7 -> 8 works just fine.

But a guess: Is it the 0.25 second update you're seeing?

10110111:
> Is it the 0.25 second update you're seeing?

Whatever the update period, the intermediate state I get is an artifact. Here's how it looks.

https://github.com/user-attachments/assets/3c8cf40a-2d92-4b8a-9477-975266d3400d

gzotti:
Maybe a 1-second (or even 15-second) interval is enough for the timer (or a fraction of that if on limelapse), but explicit timeChanged() (and locationChanged() if applicable somewhere) signals must be connected to force an immediate update.

Atque:
Okay I see it now (only on X11 session though).

Atque:
I think I've fixed it now. At least I can't see anything weird on X11. Please try again @10110111.

plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
```diff
@@ -202,11 +202,11 @@ void ObjectVisibilityDialog::createDialogContent()
 	connect(core, &StelCore::dateChanged,
 	        this, &ObjectVisibilityDialog::refreshTwilightLimits);
 	connect(core, &StelCore::dateChanged,
-	        this, &ObjectVisibilityDialog::refreshTwilightMap);
+	        this, &ObjectVisibilityDialog::scheduleTwilightMapRefresh);
 	twilightMapTimer = new QTimer(this);
 	twilightMapTimer->setInterval(250);
 	connect(twilightMapTimer, &QTimer::timeout,
-	        this, &ObjectVisibilityDialog::refreshTwilightMap);
+	        this, [this](){ refreshTwilightMap(); });
```
10110111: What's the point of this change?

10110111:
Yes, the behavior on time change has been fixed.

The upsampling artifacts still look ugly:

stellarium-020

plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
Atque: Codex motivates this change by saying:
 > refreshTwilightMap is now overloaded. The lambda disambiguates the no-argument overload for the timer
  > connection; the timer still performs the normal throttled refresh, while forced immediate updates use the
  > bool force overload.

10110111: This is an overkill. It could've simply used `qOverload` to disambiguate the overloaded methods.

Atque: Resolved now.

Atque:
@10110111 @gzotti is this PR okay for you?

alex-w:
@Atque I'm didn't compile latest changes, but I have question - did you synced latest changes in legends between two Twilight tabs?

Atque:
No, but I can't see any issue with the other tabs' legends.

alex-w:
> No, but I can't see any issue with the other tabs' legends.

So, translators will be translate these similar lines twice :)

Atque:
Ah, yes. I'll fix that, sorry.

10110111:
I've redone the icons in `master`, you may want to apply this to your SUG additions.

Atque:
Okay I hope we're good to go now.

plugins/ObjectVisibility/src/gui/ObjectVisibilityMapWidget.cpp
```diff
@@ -419,35 +954,30 @@ void ObjectVisibilityMapWidget::drawPlaceLabels(QPainter& painter) const
...
 	const double logicalHeight = height() * ratio;
+	const double degreesPerPixel =
+		mapWidth > 0.0 ? 360.0 / mapWidth : 360.0;
+	const double nearLineDegrees =
+		std::max(minNearLineDegrees, nearLinePixels * ratio * degreesPerPixel);
 
 	QVector<const PlaceLabel*> candidates;
-	candidates.reserve(static_cast<int>(std::min<qsizetype>(
-		placeLabels.size(), 2048)));
+	candidates.reserve(static_cast<int>(std::min<qint64>(
```
10110111: This bunch of casts doesn't make sense. `QVector::reserve` takes a `qsizetype`, not `int`, and `size()` returns the same. (In Qt5 it's an `int` in both cases, so no need for a cast either.) The innermost cast is not needed at all if you use an explicit template parameter—but that also shouldn't be `qint64` but rather `qsizetype`.

plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
```diff
@@ -508,10 +682,106 @@ void ObjectVisibilityDialog::refreshTwilightLimits()
...
+	{
+		double ra = 0.0;
+		double dec = 0.0;
+		StelUtils::rectToSphe(&ra, &dec,
+		                      body->getEquinoxEquatorialPos(core));
+		double lon = StelUtils::fmodpos(ra * 180.0 / M_PI -
+		                                primeMeridianSiderealDeg, 360.0);
+		if (lon > 180.0)
+			lon -= 360.0;
+		return QPair<double, double>(lon, dec * 180.0 / M_PI);
```
10110111: What is the point in `QPair<double, double>` instead of `QPointF`?

10110111:
I get about 8 FPS when this tab is open compared to 25 FPS when the whole dialog is closed.

Atque:
> I get about 8 FPS when this tab is open compared to 25 FPS when the whole dialog is closed.

I can confirm on Ubuntu/X11 but not on Fedora/Wayland. I'll investigate, thanks for notifying.

Atque:
@10110111 Could you try it now and see if the FPS drop is gone?

plugins/ObjectVisibility/src/gui/ObjectVisibilityMapWidget.cpp
Atque: Right. Fixed.

gzotti:
In the Visibility tab, the legend still looks strange, as the line/dashed line/chain of 5 crosses/chains of 5 triangles are of unequal length. Likewise, on the Twilight tab, the dashed legend line for Polar circles is shorter than the others. 
Not sure if scaling has to do with it: 
I have started Stellarium on a 150% scaling monitor, but shifted on my second, non-scaling. This switch was acknowledged by the message "[   127.147][DBG ] Changing high-DPI scaling factor from 1.5 to 1". Still the lines are too short. Else, the map error persists, but if caused elsewhere, it's probably SEP.

plugins/ObjectVisibility/src/gui/ObjectVisibilityDialog.cpp
Atque: Sorry, an unnecessary choice. I've changed to `QPointF` now.

Atque:
> In the Visibility tab, the legend still looks strange, as the line/dashed line/chain of 5 crosses/chains of 5 triangles are of unequal length. Likewise, on the Twilight tab, the dashed legend line for Polar circles is shorter than the others. Not sure if scaling has to do with it: I have started Stellarium on a 150% scaling monitor, but shifted on my second, non-scaling. This switch was acknowledged by the message "[ 127.147][DBG ] Changing high-DPI scaling factor from 1.5 to 1". Still the lines are too short. Else, the map error persists, but if caused elsewhere, it's probably SEP.

Indeed, the legend is a bit messy still. I'll try to harmonize it a bit. What issues are you seeing with the map?

gzotti:
Those I mentioned earlier ('last week'). The right part is cut off where the texture image ends, and is only reloaded when the map is moved further to the left. But maybe MapWidget has to be fixed for that.

Atque:
> Those I mentioned earlier ('last week'). The right part is cut off where the texture image ends, and is only reloaded when the map is moved further to the left. But maybe MapWidget has to be fixed for that.

Okay, I see it too. Hm, I would not like to touch the main program unless I have to. Especially if it's a deeper problem.

It's present in `master` too.

10110111:
I'll take care of this issue later myself. Please file an issue report to
track it.

10110111:
Will try in a few days. Currently I don't have a convenient access to my
dev env.

10110111:
> Could you try it now and see if the FPS drop is gone?

Nothing changed, still the same drop.

alex-w:
I can confirm drop FPS from 60 to 14-20 on Linux/X11 when Twilight tabs are opened.

Atque:
I can't reproduce FPS drop on X11 Ubuntu, could you please try the latest fix @alex-w, @10110111?

alex-w:
Performance in macOS is OK

Atque:
Did you try Ubuntu too? Or whichever X11 Linux system you have.

10110111:
I've tried on Ubuntu, it doesn't have any serious FPS drop now.

PR #5065, opened 2026-08-03
cmake: add OHOS platform detection module

joinother:
## Summary

Add a small CMake module that exposes `STELLARIUM_PLATFORM_OHOS` when the OpenHarmony toolchain sets `CMAKE_SYSTEM_NAME` to `OHOS`.

This is intentionally limited to platform detection. Follow-up patches will use this flag to make unavailable Qt modules optional, one feature at a time.

## Validation

- Rebased onto current `master` (`a8d81f7fb1`).
- `git diff --check` passes.
- Verified the module with the DevEco CMake 3.28.2 runner.
- A full local configure is not available on this machine because it does not have a desktop Qt SDK installed.

## Scope

No Harmony application code, signing material, generated files, or UI changes are included.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,6 +127,7 @@ ENDIF()
 # Use customized cmake macros
 SET(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 INCLUDE(CPM)
+INCLUDE(StellariumPlatform)
 
 # Show platform info
 MESSAGE(STATUS "Platform: ${CMAKE_SYSTEM} (${CMAKE_SYSTEM_PROCESSOR})")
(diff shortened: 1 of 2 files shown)


alex-w:
By various reasons the minimal cmake is set to version 3.18, but definition  OHOS for CMAKE_SYSTEM_NAME is added in version 3.30, so, the current solution will give always wrong results on Ubuntu 22.04 (standard support until May 2027) for example. 

What about minimal versions of linux for Harmony OS SDK (I didn't download it due requirements of registration)?

10110111:
> definition OHOS for CMAKE_SYSTEM_NAME is added in version 3.30, so, the current solution will give always wrong results on Ubuntu 22.04

This variable will never STREQUAL "OHOS", and Ubuntu isn't Harmony OS, so I don't see any problem with the default-FALSE result.

alex-w:
> > definition OHOS for CMAKE_SYSTEM_NAME is added in version 3.30, so, the current solution will give always wrong results on Ubuntu 22.04
> 
> This variable will never STREQUAL "OHOS", and Ubuntu isn't Harmony OS, so I don't see any problem with the default-FALSE result.

I hope in SDK isn't used cross-compilation :)

alex-w (approved):

gzotti (approved):
OK for me.

PR #5066, opened 2026-08-03
Dome mode for scripting engine

alex-w:
### Description
Extending existing methods and adding new methods into scripting engine for support the dome mode (hello to real planetariums with domes!). The ideas are coming from cases of real using the stellarium in Barnaul planetarium.

### Type of change

- [ ] Housekeeping

## Checklist:

### DIFF
--- a/src/scripting/StelMainScriptAPI.cpp
+++ b/src/scripting/StelMainScriptAPI.cpp
@@ -1422,14 +1422,58 @@ void StelMainScriptAPI::moveToObject(const QString& name, float duration, float 
 	}
 }
 
-void StelMainScriptAPI::moveToSelectedObject(float duration)
+void StelMainScriptAPI::moveToSelectedObject(float duration, float shift)
 {
 	StelObjectMgr* omgr = GETSTELMODULE(StelObjectMgr);
 	if (omgr->getSelectedObject().isEmpty())
 		return;
 
 	StelMovementMgr* mvmgr = GETSTELMODULE(StelMovementMgr);
-	mvmgr->moveToObject(omgr->getSelectedObject()[0], duration); // Object may be without English name
+	StelObjectP obj = omgr->getSelectedObject()[0]; // Object may be without English name
+
+	if (qFuzzyCompare(shift, 0.f))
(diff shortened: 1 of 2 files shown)


10110111:
Could you explain a little what problem this is supposed to solve? And why does the shift direction only specify latitude, but not longitude?

alex-w:
> Could you explain a little what problem this is supposed to solve? And why does the shift direction only specify latitude, but not longitude?

By default these methods centering objects, but in domes the center is located in zenith - this is not very good for observers. I’ve added optional parameters to ability shift these object from zenith to the south through meridian - e.g. the “center” of the sky is shifted to 20-40 degrees from the zenith to comfort observation. This hook is coming from the experience of using Stellarium is Barnaul’s planetarium.

gzotti:
Ah, this implies non-180 fov, but you zoom in somewhat, right?

alex-w:
> Ah, this implies non-180 fov, but you zoom in somewhat, right?

No, FOV is 180-200 degrees, without zooming. But we actively using multipling factors for Solar system bodies for their “zooming” on the dome to better show the phenomenae.

gzotti:
Ah, now I see the effect live. So you shift azimuth to have object in your main view direction (parallel rows of seats in the auditorium?), then just move it off-zenith view. OK for me. Or should the shift better take percentage of FOV so that the focus is always in the same location, regardless of current sky fov? (Of course, can also be left to the show author.)

alex-w:
The shift in degrees is more predictable and configurable for different sizes of domes IMHO. And yes, orientaion of seats was included into account, at least 2 planetariums - Barnaul’s and Novosibirsk’s - has south point in direct of observers view.

gzotti:
Well, that's the point. In Vienna there was the rule, you never rotate the sky in azimuth (even if the digital planetarium now allows it).  The old Zeiss setup had North/South lamps in the perimeter.  Centering and shifting means placing the object's azimuth in view center, effectively moving south away from the main view orientation. But adapting a screen-oriented simulation to a dome-oriented is hard, I know...

ADD: Which is why we added api/main/focus with mode=mark.

PR #5070, opened 2026-08-06
Remove global SC beginTime and endTime

xLPMG:
Removes loading and saving of beginTime and endTime from the index.json of a sky culture.
TODO: remove these values from all index.json files.

I suppose this can be seen as a fix for sky cultures in which the index.json contains different values than the defined territories.

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Manually on MacOS 26.6 with M3 Pro GPU.

## Checklist:


gzotti:
Thanks, very much in favor of this. Hopefully find time to review still later today.

xLPMG:
> very much in favor of this.

me too ;) thats why I immediately wanted to try it.

gzotti:
I just had a flash of concern about SCs without polygons (esp. the "global" ones). But it's not the skyculture, it is the territory polygon that has date entries. Its purpose is to allow showing/hiding polygons when dragging the slider. Or was there ever any technical need for date clamps in SCs? Like shortening the SC list in the menu to those "currently" active (at slider date)? I am afraid this would more confuse than help.

xLPMG:
> I just had a flash of concern about SCs without polygons (esp. the "global" ones). But it's not the skyculture, it is the territory polygon that has date entries. Its purpose is to allow showing/hiding polygons when dragging the slider. Or was there ever any technical need for date clamps in SCs? Like shortening the SC list in the menu to those "currently" active (at slider date)? I am afraid this would more confuse than help.

The global time limits of the SC are only used for the min/max time labels shown for the selected culture, computing the slider range across all cultures and the "apply time filter to list" checkbox. SCs without polygons get the sentinels for "unknownBeginTime" and "presentEndTime", so they dont affect the slider range and are always shown. 
What is your concern?

xLPMG:
The SCM only allows SCs to be created with a minimum date of -6500. Where does this number come from? Does it make sense? Should I replace it with "unknownBeginTime = -50k"?

gzotti:
I was not involved in these number/age definitions. To be on the safe side, probably our "unknown/unspecified begin" should even be earlier, like -100.000. This has then the similar non-literal value 'unspecified' as 9146. 
Or -7353, if you know the secret German message behind this number shown on a 7-segment LCD display. :-) (Hint: rotate 180 degrees). But yes, it should be smaller than -40.000 used for AUS cultures. So I propose -100.000.

alex-w:
Please add context "unknown date" for translatable terms "Unknown"

gzotti:
Is this still in draft, or should we review it now?

src/gui/ViewDialog.cpp
```diff
@@ -1649,9 +1649,13 @@ void ViewDialog::populateLists()
...
 	}
 
-	// find the earliest beginTime of all cultures (needed in initSkyCultureTime)
-	// ---> evaluate it here so we don't need to iterate over all cultures multiple times
-	int globalBeginTime = QDateTime::currentDateTime().date().year();
+	// find the slider limits across all cultures (needed in initSkyCultureTime)
+	// ---> evaluate it here so we don't need to iterate over all cultures multiple times.
+	// The "unknown" begin sentinel and the "present/∞" end sentinel are ignored so that cultures
+	// with no defined time range do not stretch the slider to its extremes.
+	const int currentYear = QDateTime::currentDateTime().date().year();
```
gzotti, line 1659: A long-running installation may run into issues I think. Consider starting in a public exhibition on December 28 and running for 2 weeks 24/7. Then maybe we need a signal that fires at turn of year and re-initializes these values.

xLPMG:
> Is this still in draft, or should we review it now?

Please review. Its still in draft mode so it doesnt get merged yet. I wanted to wait for some feedback and when thats done, delete beginTime & endTime from all index.json files. Or should I rather not do that?

src/gui/ViewDialog.cpp
xLPMG: please check the latest commit. I've added a function handleVisibleChanged that runs when the dialog is opened (changed to visible=true) and updates the year if its outdated. The caveat is that it won't work if you keep the ViewDialog open over new year and never close it, but i this approach to me seems the most simple and straightforward.

gzotti: Right. If the year is only used in context of setting the slider max, in the mentioned context of an exhibition the GUI should not be shown, at least not permanently. If a guide operates this occasionally, this looks reasonable for me now.

gzotti (approved):
Seems OK for me, thanks.

xLPMG:
@gzotti 
Something I just noticed after freshly looking at the code again:

Global endTime is initialized with "int globalEndTime = currentYear;"
And then we try to find the global endTime over all sky cultures:
```		
// Find the latest real end year across all cultures
if (cultureEndTime < StelSkyCulture::presentEndTime && cultureEndTime > globalEndTime)
{
	globalEndTime = cultureEndTime;
}
```

Afaik the SCM can only create polygon data until the "present" time, which will then set it to the year 9146.
However here, when an SC author sets the territory data (without SCM) to something after currentYear and before 9146, for example the year 3000, then the ViewDialog slider will also go up to the year 3000. 
The SCM Dialog slider only goes up to max the currentYear.

I think this is mostly a matter of perspective. It makes sense that the SCM doesn't allow setting future dates at all and consequently forces everything to be max the current year.

Does it make sense for the ViewDialog to show dates between currentYear and 9146 if the author wrote this into the territory file? Or should this also be hard capped at the current year?

gzotti:
I think using current year as max year for the slider is the only reasonable thing. Specifying a SC that ends in 3000 (or whenever >now) may be possible manually, but looks pointless (unless for SciFi authors or, hm, ugly thoughts in ugly times, with endDate in the near future, will look like specified by wannabe genocidaires :-(.

PR #5076, opened 2026-08-14
Render stars with point spread function profile (PSF)

### LINKED ISSUE #2679: Star discs not projected
Good morning,

I am interested in using the cylindrical (equirectangular) projection as an exchange format with other SW (Blender/...). I have noticed that, when activated, this projection (but also other types) applies to the galaxy background, and to the positions of the stars, but not to their discs, which appear circular even at very high latitudes.

I am not sure what is the current use-case for the projections, but for me, it would be useful if the geometry transformations were applied after the rendering of the star discs on the sphere, so as to affect also the shape of the star discs.

Alternatively, for my needs, I could try and follow the recipe given [here](http://paulbourke.net/panorama/stellariumsphere/). But this adds a lot of extra processing if, like in my case, one wants to export footage for high-res videos.

**Additional context**
High latitude portion of the sky obtained with Stellarium 0.22.2. Star discs are circular across the entire map.
Capture
High latitude portion of an equirectangular star map. Stars appear more elongated with increasing latitude.
Capture2

My apologies if I've overlooked something.

Kind regards,

Atque:
This branch brings the ability for users to draw stars through a point spread function profile (PSF), which gives stars a bit of a more realistic look. Code is inspired, but not directly ported, from Celestia (which has a very similar PSF star renderer), and has been generated by AI (Codex).

In the PSF mode, users are able to draw stars with Jacobian projection correction, meaning stars will elongate and bend proportionally as they approach the edge of the screen when FOV is large enough. Note that this mode is highly costly at maximum FOV, especially with e.g. cylinder projection. This mode is enabled by a checkbox in the PSF properties dialog.

(Sort of) Fixes #2679.

### Screenshots:
stellarium-005

With Jacobian projection correction (perspective):
stellarium-006

Cylindrical:
stellarium-007

The properties of PSF stars:
Skärmbild_20260813_224455

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Tested in Fedora 44 and Ubuntu 24.04.4 LTS

## Checklist:


alex-w:
I think new GUI elements in View/Sky dialog may be more compact, like
[ ] Spiky stars [ ] PSF star {Properties}
(See lines above)

Is _PSF star_ right term here, maybe _PSF stars_ is more correct?

Atque:
> I think new GUI elements in View/Sky dialog may be more compact, like [ ] Spiky stars [ ] PSF star {Properties} (See lines above)

Wouldn't that make translations more dense?

> Is _PSF star_ right term here, maybe _PSF stars_ is more correct?

It's already named "PSF stars".
bild

10110111:
Hmm, the code looks needlessly large and complicated for such a seemingly simple feature. It seems to me that what would suffice is just rendering each point object as a quad or a small bunch of quads (like 4×4), where each vertex of the quad would be projected independently, much like currently it is for planets.

Also, "Max irradiance", "Exposure", "Dim clip factor" — what is this all about? Is it an attempt to add yet another channel of tone mapping?

Atque:
> Hmm, the code looks needlessly large and complicated for such a seemingly simple feature. It seems to me that what would suffice is just rendering each point object as a quad or a small bunch of quads (like 4×4), where each vertex of the quad would be projected independently, much like currently it is for planets.
> 
I could try a simpler approach, yes.

> Also, "Max irradiance", "Exposure", "Dim clip factor" — what is this all about? Is it an attempt to add yet another channel of tone mapping?

Celestia has tone mapping abilities, so yes, it likely stems from Celestia's implementation.

alex-w:
> > I think new GUI elements in View/Sky dialog may be more compact, like [ ] Spiky stars [ ] PSF star {Properties} (See lines above)
> 
> Wouldn't that make translations more dense?
> 
> > Is _PSF star_ right term here, maybe _PSF stars_ is more correct?
> 
> It's already named "PSF stars". bild

The right side is empty and it can be use for putting checkbox and button of new feature

Atque:
> The right side is empty and it can be use for putting checkbox and button of new feature

Do you think the whole "PSF stars" + checkbox + Properties should move to the right, or would it suffice with "Properties"?

Atque:
All right, I tried a different approach, but rendering all stars uniformly as tessellated sprites had serious impact on FPS (basically useless in wider FOV's) plus it made the faint stars flicker when panning around, and applying it only to the brightest stars naturally made the fainter ones un-projected near screen edges, which is not what I wanted with the projection correction.

10110111:
I've just tried running your branch, and when I checked "PSF stars", my frame rate dropped to 0.06–0.12 FPS on Ubuntu 22.04 with Intel UHD Graphics 620 (at least when using the orthographic projection). Unchecking it brings the frame rate back to 24 FPS.

Besides, the additional tonemapping is not only useless in principle (we already manage stars blobs' brightnesses elsewhere), but it also makes the picture radically different even near the center of projection by default. See the screenshots below.

#### Original

stellarium-021

#### This PR

stellarium-020

Atque:
> Besides, the additional tonemapping is not only useless in principle (we already manage stars blobs' brightnesses elsewhere), but it also makes the picture radically different even near the center of projection by default. See the screenshots below.

Did you trim the settings under `Properties`?

10110111:
> Did you trim the settings under `Properties`?

No, I just checked the checkbox right after I found it. Didn't even click _Properties_.

Atque:
Ah yes, we might need better default values. Please play around with the settings. I've noticed myself that perspective projection gives the highest FPS (and initially, I wanted to limit this mode to perspective projection).

10110111:
Honestly, I don't really like this approach at all, not least because of the huge amount of code it adds while only helping with the point objects.

What if you instead try to render everything into, say, a cube map, and then sample it when rendering in the current projection similarly to how MilkyWay or ZodiacalLight do (except for the source projection, of course)? When the FoV is smaller than 90° you'd avoid the 5 invisible sides of the cube and render in a simplified way similar to the current one.

This could make the corners of the cube map look suboptimal, but I think it's worth trying.

alex-w:
> > The right side is empty and it can be use for putting checkbox and button of new feature
> 
> Do you think the whole "PSF stars" + checkbox + Properties should move to the right, or would it suffice with "Properties"?

Like this, just compare...

stellarium-008

alex-w:
Interesting, I see no difference for performance in macOS.

Atque:
> Honestly, I don't really like this approach at all, not least because of the huge amount of code it adds while only helping with the point objects.
> 
> What if you instead try to render everything into, say, a cube map, and then sample it when rendering in the current projection similarly to how MilkyWay or ZodiacalLight do (except for the source projection, of course)? When the FoV is smaller than 90° you'd avoid the 5 invisible sides of the cube and render in a simplified way similar to the current one.
> 
> This could make the corners of the cube map look suboptimal, but I think it's worth trying.

I've tried that now, and results are not satisfying. Code might be cleaner and simpler, but the functionality I'm seeking cannot be achieved. So I'll keep the current approach and await further inputs.

10110111:
> the functionality I'm seeking cannot be achieved.

How so? Aren't points stretched? Do you enable texture filtering—with a simple `GL_NEAREST` this may indeed fail to work? What do you actually get (could you make a screencast?)?

Atque:
> How so? Aren't points stretched? Do you enable texture filtering—with a simple `GL_NEAREST` this may indeed fail to work? What do you actually get (could you make a screencast?)?

See the screenshot below. Stars near the cube face center are more prominent than stars further to its edges, and the border between cube faces is uncomfortably obvious (the bright blob is the Moon).

image

Cube face resolution is 2048 px, and my experiment fades between PSF and cubemap between 60 and 90 degree FOV. Already at 2048 px we notice FPS drop, and it doesn't even look that good.

10110111:
Hmm, OK. But in any case, the original implementation also has a huge performance impact. 

And the points don't even look the same in the center of projection, even when zoomed in: they are too dim and blurred. There should be some correspondence between what we have with and without this feature. Particularly, in the center of the projection there should be no difference at all, since it's affected the least by the projection differences.

The additional tone mapping should be removed, instead doing only the stretching of the stretched areas while preserving the shapes and brightnesses of the central ones. If we ever want to have separate brightness controls for point objects, this should be a separate PR. Currently it looks useless and goes against the previous work on balancing what is visible and what isn't depending on illumination.

Atque:
> Hmm, OK. But in any case, the original implementation also has a huge performance impact.

The mode is optional. Or are you experiencing performance loss without Jacobian projection correction?

> And the points don't even look the same in the center of projection, even when zoomed in: they are too dim and blurred. There should be some correspondence between what we have with and without this feature. Particularly, in the center of the projection there should be no difference at all, since it's affected the least by the projection differences.

What, really? I can't see any difference in the center of screen, only closer to the edges.

> The additional tone mapping should be removed, instead doing only the stretching of the stretched areas while preserving the shapes and brightnesses of the central ones. If we ever want to have separate brightness controls for point objects, this should be a separate PR. Currently it looks useless and goes against the previous work on balancing what is visible and what isn't depending on illumination.

All right, I'll see if I can use other criteria. They can't use relative/absolute scale though, we need size, flare decay etc.

10110111:
> The mode is optional. Or are you experiencing performance loss without Jacobian projection correction?

If we do add the projection correction for points, I'd like to use it by default—and even as the only available mode, because we already do this for planet meshes.

Currently the performance drop is inconsistent though. This time I started with stereographic projection, opened the View dialog, and after checking the checkbox I got the huge drop. But when I made a screenshot (by pressing <kbd>Ctrl</kbd>+<kbd>s</kbd>), after that frame rate suddenly became an order of magnitude higher, reaching about 4 FPS.

> What, really? I can't see any difference in the center of screen, only closer to the edges.

Well, maybe you're using some non-default settings in your new dialog, but this is what I get by default:

#### Original
stellarium-021

#### New
stellarium-020

Atque:
> If we do add the projection correction for points, I'd like to use it by default—and even as the only available mode, because we already do this for planet meshes.

Me too, but I know many people won't really care about this, and thus it ought to remain optional, especially if it hits performance.

> Currently the performance drop is inconsistent though. This time I started with stereographic projection, opened the View dialog, and after checking the checkbox I got the huge drop. But when I made a screenshot (by pressing Ctrl+s), after that frame rate suddenly became an order of magnitude higher, reaching about 4 FPS.

That is weird, I haven't experienced this myself. Some projections are more prone to performance dips. Perspective is the one that works the best, I've noticed.


> Well, maybe you're using some non-default settings in your new dialog, but this is what I get by default:

Forgive me, but what exactly is wrong with the new mode in these screenshots? Stars in the center look fine, while stars in the edges are enlarged (as one would expect by the stereographic projection).

10110111:
> what exactly is wrong with the new mode in these screenshots?

Well in this particular pair, if you open the screenshots in separate tabs and switch between them back and forth, you'll notice that the new one lacks lots of stars, and the ones with labels look dimmer. The stars between the Sun and the red "З" have the same size as in the old state (so not affected much by stretching) but are still blurry. This is not how projection distortion should work.

See how it should work on the example of the Moon in the following pair. See how crisp the Moon is, and how different your stars look even taking into account their original sprites' smooth edges.

#### Old

stellarium-020

#### New

stellarium-021

Atque:
> See how it should work on the example of the Moon in the following pair. See how crisp the Moon is, and how different your stars look even taking into account their original sprites' smooth edges.

But wouldn't we want distorted stars to be proportionally blurry? Same brightness stretched over a wider area?

Atque:
All right, I've fixed the UI, and renamed the controls (hopefully it's easier to understand now).

image

Do note that number of stars drawn are still controlled by the absolute/relative scaling!

Askaniy:
Hi there! I'm the original designer of the rendering method, you can find the reference Python implementation here: https://github.com/Askaniy/CelestiaStarRenderer
I would like to be credited. Also, Levin Li done a great job porting it to Celestia, my own branch was less successful.

Atque:
> Hi there! I'm the original designer of the rendering method, you can find the reference Python implementation here: https://github.com/Askaniy/CelestiaStarRenderer I would like to be credited. Also, Levin Li done a great job porting it to Celestia, my own branch was less successful.

Interesting, thank you for sharing! Yes, of course, I'll credit you. 

Do you think your original implementation would be a cleaner base to build this PSF renderer on?

Askaniy:
At first glance, yes, my architecture does seem more straightforward to me. Also, in the original repository, there is an example of a projection (simple cylindrical) that was reimplemented in this PR. There is also a test comparing my simplified PSF with the reference scientific paper. However, I am not a C/C++ developer, so I can't really judge which codebase would be better for Stellarium.

Askaniy:
These issues could help clarify some of the differences between my repo and Celestia's version:
https://github.com/CelestiaProject/Celestia/issues/2559
https://github.com/CelestiaProject/Celestia/issues/2560

10110111:
> But wouldn't we want distorted stars to be proportionally blurry? Same brightness stretched over a wider area?

First of all, the unstretched stars (e.g. in the center of projection) should look as closely to the current state of Stellarium as possible. Not more blurry, not dimmer. These stars are the reference for the look.

Then, generally for all stars, I think we want to get such a projection that, when a patch of it is reprojected to a small-FoV perspective view, it would look like it does in the current state (up to the artifacts of finite resolution, of course). 

I just don't really understand why you think blurring of stars is appropriate. Stretching is not the same as blurring. Or am I mistaking the goal of your PR? Maybe you want to achieve something completely different from what I'm thinking of?

10110111:
Now that I re-read the description in the OP, it seems like your "Projection correction" is actually a secondary part of this feature, while the blurring (the PSF profile) is the main one. Right?

Then I still wonder why you think the blurry stars are more realistic. To the sharp human vision (natural or corrected) real-life stars look like crispy points (of different sizes depending on brightness). They might have tiny rays due to aberrations of the eye optics, but definitely not be blurry generally. They do become blurry on photographs though, but Stellarium usually aims at simulating the way sky appears to the eye, not to a camera. The PSF I get from this PR doesn't match the stars I remember on photos though.

Atque:
> Now that I re-read the description in the OP, it seems like your "Projection correction" is actually a secondary part of this feature, while the blurring (the PSF profile) is the main one. Right?

Yes, the projection correction is not the vital change of this PR (and doesn't affect legacy stars).

> Then I still wonder why you think the blurry stars are more realistic. To the sharp human vision (natural or corrected) real-life stars look like crispy points (of different sizes depending on brightness). They might have tiny rays due to aberrations of the eye optics, but definitely not be blurry generally. They do become blurry on photographs though, but Stellarium usually aims at simulating the way sky appears to the eye, not to a camera. The PSF I get from this PR doesn't match the stars I remember on photos though.

I guess I'm trying to visualize a pseudo-camera-ish mode. Stars in Stellarium have long bothered me. They are ugly and do not look like they do to my naked eye or my camera. This PR aims at launching an optional, more realistic mode. As a bonus, to simulate wide-angle photographs, I also decided to add projection correction. It might not be perfect, but at least it gives an idea of how light sources behave in the edges of a projected screen. Users who find it demanding on their hardware or annoying to the eye, could simply turn it off.

Askaniy:
> They might have tiny rays due to aberrations of the eye optics, but definitely not be blurry generally. They do become blurry on photographs though, but Stellarium usually aims at simulating the way sky appears to the eye, not to a camera.

@10110111 I just processed the real measurements of the human eye PSF, provided by [Spencer et al. 1995](https://dl.acm.org/doi/10.1145/218380.218466), and designed a much more efficient formula for the "bloom"/"flare" mode. Please check out the article and the [ReadMe](https://github.com/Askaniy/CelestiaStarRenderer#celestia-star-renderer). Current star rendering in Stellarium is much better than it was in Celestia, but it's not fully realistic, either.

gzotti:
GUI wise, PSF stars ignore the spiky stars setting. And the details settings seem to re-implement the classical absolute/relative scale settings of the traditional blobs, while still being influenced by them. Maybe some default set of abs/rel (most obvious setting: 1/1) should be frozen for PSF use (or at least, set relative==1 and allow 'more stars' with absolute scale?), so that we don't have so many screws to turn. We would then have 3 modes: classic (old blobs governed by 2 settings abs/rel), spiky (likewise), and PSF (governed by its own settings plus 'abs'). The moon's glare currently is a monstrous burnt-out patch. This had been dimmed down to allow the moon globe rendered with phase even if tiny, so this glare must be dimmed. And the solar glare must be covered as well. I see the 'Bright source cap'. but maybe SSOs should be handled differently?  Note also that the Oculars plugin has a specialized mode to save these detail settings when entering Oculars view (or camera view), and replaces them with its own, because oculars view just may require different settings.  This will have to be extended by PSF settings.   But IMO the projection correction is most interesting, at least in a few projections. Do you intend to extend this idea to DSO icons as well? (separate PR, likely)

On blurry vs. photographic: Wide angle photos often have unblurred but just overexposed bright stars, surrounded by well- but not overexposed dimmer stars of the same apparent disk diameter. It's hard to identify the bright stars in those pictures. Some artists therefore use slight softening filters to increase bright stars.

10110111:
> I just processed the real measurements of the human eye PSF

I don't quite get why one would want to simulate the PSF of the eye to make a realistic presentation of something that the same kind of eye would look at. This way—even if you make the PSF correct both in brightness pattern shape and in angular size—will generally result in seeing a doubly-convolved image: first by the simulated PSF, second by the actual physiology of the eye.

What we should strive to achieve in Stellarium is making it seem that we're looking at point sources of different brightnesses, but taking into account the constraint that the actual screens have limited brightness. This in particular means that the dimmest stars should be 1 px wide when aligned to the pixel grid and 2×2 px in size when we have to antialias their non-integral position. This PR currently makes them too blurry.

It does make sense though to show the tail of the PSF as something that can't appear physiologically due to the limited brightness of the display. But this shouldn't get in the way for the points that we *can* display well, such as the dimmer stars.

Atque:
> I don't quite get why one would want to simulate the PSF of the eye to make a realistic presentation of something that the same kind of eye would look at. This way—even if you make the PSF correct both in brightness pattern shape and in angular size—will generally result in seeing a doubly-convolved image: first by the simulated PSF, second by the actual physiology of the eye.

Unless the user sits in a perfectly dark room and has launched Stellarium on their OLED screen with incredible brightness, then I don't think eye physiology matters too much.

> What we should strive to achieve in Stellarium is making it seem that we're looking at point sources of different brightnesses, but taking into account the constraint that the actual screens have limited brightness. This in particular means that the dimmest stars should be 1 px wide when aligned to the pixel grid and 2×2 px in size when we have to antialias their non-integral position. This PR currently makes them too blurry.

This contradicts my experience of dim stars. Dim stars are, to my eye, not smaller, only dimmer. The ones close to the NELM are washed out, blurry patches that flicker in and out of view. Simulating stars as a single pixel would not represent this too well.

About the blurriness; is it also noticeable when projection correction is turned off? You could play with the settings and get rather sharp stars close to the visual limit.

> It does make sense though to show the tail of the PSF as something that can't appear physiologically due to the limited brightness of the display. But this shouldn't get in the way for the points that we _can_ display well, such as the dimmer stars.

Again, I don't agree that Stellarium currently displays dim (nor bright) stars well. At least this new PSF is a bit more accurate.

Atque:
> GUI wise, PSF stars ignore the spiky stars setting. And the details settings seem to re-implement the classical absolute/relative scale settings of the traditional blobs, while still being influenced by them. Maybe some default set of abs/rel (most obvious setting: 1/1) should be frozen for PSF use (or at least, set relative==1 and allow 'more stars' with absolute scale?), so that we don't have so many screws to turn. We would then have 3 modes: classic (old blobs governed by 2 settings abs/rel), spiky (likewise), and PSF (governed by its own settings plus 'abs'). The moon's glare currently is a monstrous burnt-out patch. This had been dimmed down to allow the moon globe rendered with phase even if tiny, so this glare must be dimmed. And the solar glare must be covered as well. I see the 'Bright source cap'. but maybe SSOs should be handled differently? Note also that the Oculars plugin has a specialized mode to save these detail settings when entering Oculars view (or camera view), and replaces them with its own, because oculars view just may require different settings. This will have to be extended by PSF settings. But IMO the projection correction is most interesting, at least in a few projections. Do you intend to extend this idea to DSO icons as well? (separate PR, likely)

Yes, the Moon is admittedly very difficult to balance. I disagree that SSO's should behave differently. I think the brightness cap covers the brightest objects. Perhaps the Moon needs a special case to help visualize the phases, just like the current blob texture. About the Sun; should it also use PSF? Or are you suggesting a new halo texture?

About DSO markers; no, they are not my focus, but it could be interesting to have them projection corrected too. Yes, another PR in that case.

10110111:
Can the PSF be allowed to be different? This may help with enabling projection stretching for the old `star16x16.png` PSF, as well as having different PSFs for the Moon and the Sun.

src/core/StelSkyDrawer.cpp
```diff
@@ -217,17 +254,104 @@ void StelSkyDrawer::init()
 	starShaderVars.color = starShaderProgram->attributeLocation("color");
 	starShaderVars.texture = starShaderProgram->uniformLocation("tex");
 
+	const char *psfVsrc =
```
10110111: What do these shaders do? I can even delete the whole hunk, and the sky still renders.

10110111:
> I disagree that SSO's should behave differently. I think the brightness cap covers the brightest objects.

No it doesn't. Just try zooming into any planet, e.g. Mercury. You'll get a huge white blob that grows with zoom until it suddenly disappears, revealing a small ball of the planet.

10110111:
Also, there's an issue with the placement of labels: they normally are far away to not touch the halos, but in PSF mode they are too close, see the screenshots below.

#### Old

stellarium-021

#### New

stellarium-020

Atque:
> Can the PSF be allowed to be different? This may help with enabling projection stretching for the old `star16x16.png` PSF, as well as having different PSFs for the Moon and the Sun.

You mean three separate PSF (stars/point sources, Moon and Sun)? And would it be interesting to stretch the 16x16 texture? I thought of this before deciding to go for PSF.

Atque:
> No it doesn't. Just try zooming into any planet, e.g. Mercury. You'll get a huge white blob that grows with zoom until it suddenly disappears, revealing a small ball of the planet.

Hm, ok. We should probably hide it earlier. When I say SSO's shouldn't behave differently, I'm mainly referring to size. Stellarium had (and partly still has) a problem with size of planet halos vs stars'.

10110111:
> You mean three separate PSF (stars/point sources, Moon and Sun)?

Yes.

> And would it be interesting to stretch the 16x16 texture?

Well sure. I'd be interested to preserve the current PSF while having the projection stretching applied to it. Especially given that I don't have the huge performance drop now that I had in the beginning of this PR.

The circular textures `star16x16.png` and `haloLune.png` could be approximated by some polynomials or something, not necessarily remain textures. This would avoid possible pixellation. Not so sure about `star16x16_rays.png` and `halo.png`.

gzotti:
> > No it doesn't. Just try zooming into any planet, e.g. Mercury. You'll get a huge white blob that grows with zoom until it suddenly disappears, revealing a small ball of the planet.
> 
> Hm, ok. We should probably hide it earlier. When I say SSO's shouldn't behave differently, I'm mainly referring to size. Stellarium had (and partly still has) a problem with size of planet halos vs stars'.

Well, because some planets can become extremely bright. At some point the disk becomes ugly. And users complain about the Galilean moons just being big bright blobs and not turning into spheres earlier when viewed in a telescope. And Venus when just not covered by the Moon (Sept.14) being much too dim. (Not a blob size/PSF issue though.)

Atque:
> Well sure. I'd be interested to preserve the current PSF while having the projection stretching applied to it. Especially given that I don't have the huge performance drop now that I had in the beginning of this PR.
> 
> The circular textures `star16x16.png` and `haloLune.png` could be approximated by some polynomials or something, not necessarily remain textures. This would avoid possible pixellation. Not so sure about `star16x16_rays.png` and `halo.png`.

I could try it, yes. Solar halo also likely needs stretching then.


> Well, because some planets can become extremely bright. At some point the disk becomes ugly. And users complain about the Galilean moons just being big bright blobs and not turning into spheres earlier when viewed in a telescope. And Venus when just not covered by the Moon (Sept.14) being much too dim. (Not a blob size/PSF issue though.)

I could try and see if I can at wider FOV then. About Venus; I have thought about making the PSF move to the sunlit part of the crescent, but I'm not sure if it's possible or how good it'd look. That could possibly help a bit.

gzotti:
Actually, once you clearly see the disk/crescent of Venus, there is no need for a PSF blob. But we must do something about the texture rendering. The visible part of Venus is always blindingly bright, the illuminated area of the disk should be rendered just not burned-out. Maybe for a dark-sky environment, surround a tiny crescent by a semi-bright blob, similar to the Moon, to visualize its brightness.

Atque:
Yes, but that's for another PR. For Venus we'd need an atmosphere edge too (Celestia will probably be useful to base that on). And we have the Moon burnout: #3067 

But hiding the PSF rendering when planet takes up enough of FOV is something I'll hopefully add later today. Should fix Jupiter's moons too.

src/core/StelSkyDrawer.cpp
Atque: Sorry, that appears to be a left-over from a previous attempt.

Atque:
Locally I've added bloom for the Moon, making the weak phases stand out more against the sky. Bloom is configurable (0.10 - > 5.00) and disables the PSF drawing for the Moon when enabled. It is disabled in daylight as not to burn out the Moon further.

Moon 17% illuminated at 1x size at 50 degree FOV:
bild

Do you think this would be interesting to add to this PR's code?

10110111:
> Do you think this would be interesting to add to this PR's code?

I think it's better for a separate PR. This one is already quite a bunch of code to review.

Askaniy:
@Atque Looking at the last screenshot, could it be that gamma correction is applied twice on the colors? (talking about the glow, not the overexposed centers)
For example, Betelgeuse looks paler than my calculations. Not gamma-corrected (linear scaled) color of Betelgeuse should be around `#ffb134`, sRGB gamma-corrected around `#ffd97c`, and gamma-corrected twice is `#ffeeb9` (very similar to the screenshot).

Atque:
> @Atque Looking at the last screenshot, could it be that gamma correction is applied twice on the colors? (talking about the glow, not the overexposed centers) For example, Betelgeuse looks paler than my calculations. Not gamma-corrected (linear scaled) color of Betelgeuse should be around `#ffb134`, sRGB gamma-corrected around `#ffd97c`, and gamma-corrected twice is `#ffeeb9` (very similar to the screenshot).

Hm, yes, it seems stars and planets were too pale in coloration. I was wondering why Mars was so white. I hope it looks better now, thanks for notifying! :)

Atque:
@10110111 @gzotti @alex-w Any further comments on this?

alex-w (changes requested):
Please reflect the changes in `po/stellarium/POTFILES.in` file too

alex-w:
Hmm... default properties for PSF:

stellarium-036
stellarium-037

Atque:
@alex-w I take it you think the Moon is too dominant in the sky? The `bright source cap` option sets a roof for the PSF magnitude. If -8.5 is too high, we could set it even lower as default (although 1/1 in abs/rel scale already is unrealistic and I mostly use it for "learning constellations").

10110111:
> I take it you think the Moon is too dominant in the sky?

I think you were striving to make the look more realistic, no? The Moon in the screenshot above looks more like the Sun from a human vision PoV. In reality the lunar glare never saturates the eye receptors, only the lunar disk does (when looking at it with dilated pupils, e.g. when moving the gaze towards the full Moon from some star).

Atque:
> > I take it you think the Moon is too dominant in the sky?
> 
> I think you were striving to make the look more realistic, no? The Moon in the screenshot above looks more like the Sun from a human vision PoV. In reality the lunar glare never saturates the eye receptors, only the lunar disk does (when looking at it with dilated pupils, e.g. when moving the gaze towards the full Moon from some star).

Yes, but the Moon is always a special case. Taking wide-angle photos of the sky will show the Moon as a ridiculously bright blob, and to the naked eye (which has excellent dynamic range) it looks bright and you clearly see the form. I can't see any other solution right now, other than to cut bright moonlight off.

gzotti:
As said above, at least Sun and Moon need special treatment. Unsure if we can cap at mag -4.5 (Venus max) to keep whatever blob is necessary non-ugly. The Moon should have an aureole when the body would be too small to be clearly visible, similar to the current semi-bright blob. But I'd welcome the projection effect.

Atque:
> As said above, at least Sun and Moon need special treatment. Unsure if we can cap at mag -4.5 (Venus max) to keep whatever blob is necessary non-ugly. The Moon should have an aureole when the body would be too small to be clearly visible, similar to the current semi-bright blob. But I'd welcome the projection affect.

I think -4.5 is too low of a cap. I want to cover the brightest supernovae as well.

Atque:
New options to draw Moon added. We're presented with the option to draw PSF with "Moon glare reduction", or as a textured halo (reuses `haloLune.png`).

moon

bild

alex-w:
@Atque please add button to reset PSF properties

Atque:
Great idea, I'll do that.

alex-w:
Interesting, should be better use separate section for psf options in `config.ini` file?

gzotti:
IMO psf_* keys in the stars section are OK.

10110111:
Is it intended that the Moon looks much less prominent that Venus in PSF mode? It loses its internal halo, see the screenshots below.

#### Old

stellarium-022

#### New

stellarium-023

Atque:
Yes, when you use the textured halo you lose a lot of "wow" effect on the Moon.

10110111:
> Yes, when you use the textured halo you lose a lot of "wow" effect on the Moon.

Hmm, but then what's the use of this option? I thought this option would enable the double halo (external thin + internal bright) we have normally.

Atque:
> Hmm, but then what's the use of this option? I thought this option would enable the double halo (external thin + internal bright) we have normally.

The option is for those who want to view the lunar phase in large FOV without having to scale the Moon. Using PSF for the Moon will unavoidably draw it as a bright blob, and thus the phase cannot be distinguished.

gzotti:
But can't you draw a half-bright large blob that leaves phase visible?

Atque:
I'm not sure that would look any better. :/

10110111:
> Using PSF for the Moon will unavoidably draw it as a bright blob, and thus the phase cannot be distinguished.

I'm not so sure of this. You already use the texture as a PSF for the Moon, why not just combine the two textures as is done normally and use *this combination* as the PSF? In current normal (non-PSF) mode the phase is visible despite there being a halo:

stellarium-021

Atque:
> I'm not so sure of this. You already use the texture as a PSF for the Moon, why not just combine the two textures as is done normally and use _this combination_ as the PSF? In current normal (non-PSF) mode the phase is visible despite there being a halo:
> stellarium-021

I was never a big fan of the semi-transparent Moon halo, which was introduced after a harmonization to the planet halos that made the Moon an uncomfortably big of a blob, so to me, that looks like a worse solution than the one I'm currently providing in this branch.

gzotti:
Until we have real HDR sceens, we must work with compromises.

Atque:
> Until we have real HDR sceens, we must work with compromises. 

Don't you think we've reached one with this solution? I think we've got plenty of options now to display the Moon satisfactorily.

I have local code that supports bloom around the Moon but I won't push it into this branch.

10110111:
No, HDR rendering would imply

1. Rendering of everything physical in linear light to a floating-point (or
at least wide enough for the dynamic range) render target,
2. Applying a sensible PSF to everything at once (ideally, by a
convolution),
3. Proper tone mapping – not just for luminance adaptation, but also to
simulate the dynamic range of the human eye on an sRGB screen.

This PR only simulates a PSF for point objects. And generally in current
Stellarium everything is rendered by an ad-hoc sRGB brightness (and often
even hue!) instead of properly rendering in some units of brightness.

gzotti:
In the current build I can have a large, very soft textured halo with (zoomed out) an invisibly small moon. The label 'Moon' is placed somewhere above-right of the halo. The place where the moon should be is visually empty.  I don't want to have the PSF details panel open all the time to fine-tune the view depending on FOV. IMO the large halo texture needs a really bright patch in the center (but not a saturated PSF blob) to look quite similar to the current conspicuous glow.

src/core/StelSkyDrawer.hpp
```diff
@@ -112,9 +123,9 @@ class StelSkyDrawer : public QObject, protected QOpenGLFunctions
 	//! @param checkInScreen whether source in screen should be checked to avoid unnecessary drawing.
 	//! @param twinkleFactor allows height-dependent twinkling. Recommended value: min(1,1-0.9*sin(altitude)). Allowed values [0..1]
 	//! @return true if the source was actually visible and drawn
-	bool drawPointSource(StelPainter* sPainter, const Vec3d& v, const RCMag &rcMag, int bVindex, bool checkInScreen=false, float twinkleFactor=1.0f)
+	bool drawPointSource(StelPainter* sPainter, const Vec3d& v, const RCMag &rcMag, int bVindex, bool checkInScreen=false, float twinkleFactor=1.0f, float appMag=std::numeric_limits<float>::quiet_NaN())
```
10110111, line 126: `appMag` should be documented, particularly what it actually is and whether it has any physical meaning, or is just an ad-hoc manipulated version of the magnitude.

Besides, it seems like something that could be added as one more field to `RCMag` instead of passing it to the drawing functions. Maybe it's a better approach?

src/core/StelSkyDrawer.cpp
```diff
@@ -43,15 +43,43 @@
...
+#include <algorithm>
+#include <cmath>
 
 // The 0.025 corresponds to the maximum eye resolution in degree
 #define EYE_RESOLUTION (0.25f)
 #define MAX_LINEAR_RADIUS 8.f
+// Keep very bright point-source PSF halos under control. This preserves contrast
+// for objects such as Venus and bright supernovae while letting the full Moon
+// stand out without using its full integrated magnitude.
+#define DEFAULT_PSF_BRIGHT_SOURCE_MAG_LIMIT (-8.5f)
```
10110111, line 55: I'd convert these into `constexpr` variables instead of defines. And also it's better to replace `static` (in the functions below and for the variables you'd create for these values) with a single anonymous namespace.

src/core/StelSkyDrawer.cpp
```diff
@@ -163,6 +214,11 @@ StelSkyDrawer::~StelSkyDrawer()
 	
 	delete starShaderProgram;
 	starShaderProgram = Q_NULLPTR;
+	delete psfPointShaderProgram;
```
10110111, line 217: Please use `std::unique_ptr` for new such objects. There's no benefit in using manual memory management in new code.

Askaniy:
> No, HDR rendering would imply
> 
> 1. Rendering of everything physical in linear light to a floating-point (or
> at least wide enough for the dynamic range) render target,
> 2. Applying a sensible PSF to everything at once (ideally, by a
> convolution),
> 3. Proper tone mapping – not just for luminance adaptation, but also to
> simulate the dynamic range of the human eye on an sRGB screen.
> 
> This PR only simulates a PSF for point objects. And generally in current
> Stellarium everything is rendered by an ad-hoc sRGB brightness (and often
> even hue!) instead of properly rendering in some units of brightness.

@10110111 is right, ideally we need a proper convolution for HDR instead of the PSF function hack.

My PSF is designed for SDR. In its transition logic, it uses "overexposure limit" that is true for SDR. And then skips proper PSF calculations inside the "overexposure radius", leaving only "bloom" asymptotics.
HDR do not operate with "overexposure", so this optimization does not make sense and makes central areas to be just clipped (not realistically), or go to infinity at the center following the bloom asymptotic.

PR #5082, opened 2026-08-20
Fix saving/retrieving stored main window size on scaled screens

### LINKED ISSUE #4955: v26.1 Qt6 regression: Application forces fullscreen on launch and ignores saved window geometry on 4K display
There is a window geometry regression introduced in version 26.1 Qt6 compared to version 25.4 Qt6. On version 25.4, the application correctly remembers and obeys the last saved window size upon launch. In version 26.1, the application completely overrides saved configurations and automatically forces itself into a maximized/fullscreen state at startup.

Steps to Reproduce:
1. Install Stellarium v26.1 Qt6 on Windows 11 with a 4K resolution monitor (3840x2160).
2. Press F11 to break out of fullscreen into windowed mode, resize the window frame, and click "Save Settings".
3. Close the application.
4. Relaunch the application. It immediately forces itself back to a maximized/fullscreen layout, ignoring the user dimensions.
*Note: Appending `--windowed` to the execution shortcut targets also fails to bypass this behavior.*

System Environment Info from Log:
- Stellarium Version: 26.1 (Qt6 Build)
- Operating System: Windows 11
- Display Resolution: 3840x2160 (4K UHD)
- GPU: NVIDIA GeForce RTX 3060 Ti
- Working Baseline: Version 25.4 Qt6 operates flawlessly and respects window constraints on this exact hardware environment.

### LINKED ISSUE #5074: Window position will not be saved on a 4K display with Win11 (25H2)
Hello.

On a 4K display with Windows 11 (25H2), when using "window mode", as opposed to "full screen", the window position will not be retained when reopening Stellarium between sessions, even if the "Save Settings" button was pressed in the Configuration Menu. It reopens with the "windowed" Stellarium in an awkward position. It has to be resized manually to be able to view all relevant information, especially the bottom Tool Bar.

I have uninstalled Stellarium from this Windows computer and deleted all the associated files, then installed anew, to no avail. The odd behavior remains.

Please note that this is not an issue with a Windows computer with an HD display, nor is it a problem on MacOS with an iMac 27" (Intel, 2020) that has a 5K display. This is strictly with Windows and a 4K display, as far as my situation is concerned.

I've searched for reported problems of this issue, but did not find one.

This is with the latest official Stellarium, 26.2, and with the latest weekly build, as well as earlier versions of Stellarium.

Regards.

Martin

Image
Image

gzotti:
A problem at loading window size has been reported for 26.1 and 26.2 that shows only on hiDPI screens and when using the old-fashioned manual "store settings".

### Description

Fixes #5074 (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11 24H2
* Graphics Card: Geforce 2070

## Checklist:

### DIFF
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1624,21 +1624,14 @@ void StelMainView::moveEvent(QMoveEvent * event)
 	//const QPoint &pos=event->pos(); // top-left of drawing area
 	const QPoint &pos=frameGeometry().topLeft(); // including frame!
 
-	// We use the glWidget instead of the event, as we want the screen that shows most of the widget.
-	QWindow* win = glWidget->windowHandle();
-	if(win)
-	{
-		stelApp->setDevicePixelsPerPixel(win->devicePixelRatio());
-	}
-	else
-		qWarning() << "Cannot get window handle (to set device pixel ratio). (uncritical)";
-
 	if (StelApp::isInitialized())
 	{
-		qDebug() << "saving position with devicePixelRatio=" << devicePixelRatio();
+		const qreal dpp = devicePixelRatio();
+		stelApp->setDevicePixelsPerPixel(dpp);
+		qDebug() << "saving position with devicePixelRatio =" << dpp;
 
-		StelApp::immediateSave("video/screen_x", int(std::lround(pos.x()*devicePixelRatio())));
-		StelApp::immediateSave("video/screen_y", int(std::lround(pos.y()*devicePixelRatio())));
+		StelApp::immediateSave("video/screen_x", int(std::lround(pos.x()*dpp)));
+		StelApp::immediateSave("video/screen_y", int(std::lround(pos.y()*dpp)));
 	}
 }
 
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -727,7 +727,6 @@ void StelMainView::resizeEvent(QResizeEvent* event)
 			guiItem->setGeometry(QRectF(0.0,0.0,sz.width(),sz.height()));
 		if (StelApp::isInitialized()  && !isFullScreen())
 		{
-			qDebug() << "saving size with devicePixelRatio=" << devicePixelRatio();
 			StelApp::immediateSave("video/screen_w", int(std::lround(sz.width() *devicePixelRatio())));
 			StelApp::immediateSave("video/screen_h", int(std::lround(sz.height()*devicePixelRatio())));
 		}
@@ -1621,14 +1620,12 @@ bool StelMainView::needsMaxFPS() const
 
 void StelMainView::moveEvent(QMoveEvent * event)
 {
-	//const QPoint &pos=event->pos(); // top-left of drawing area
-	const QPoint &pos=frameGeometry().topLeft(); // including frame!
+	const QPoint &pos=frameGeometry().topLeft(); // including frame! (Original event->pos() returns top-left of drawing area)
 
 	if (StelApp::isInitialized())
 	{
 		const qreal dpp = devicePixelRatio();
 		stelApp->setDevicePixelsPerPixel(dpp);
-		qDebug() << "saving position with devicePixelRatio =" << dpp;
 
 		StelApp::immediateSave("video/screen_x", int(std::lround(pos.x()*dpp)));
 		StelApp::immediateSave("video/screen_y", int(std::lround(pos.y()*dpp)));
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -718,6 +718,16 @@ StelMainView::StelMainView(QSettings* settings)
 
 void StelMainView::resizeEvent(QResizeEvent* event)
 {
+	qreal devicePixelRatio=1.0;
+	// We use the glWidget instead of the event, as we want the screen that shows most of the widget.
+	QWindow* win = glWidget->windowHandle();
+	if(win)
+	{
+		devicePixelRatio=win->devicePixelRatio();
+	}
+	else
+		qCritical() << "resizeEvent: CANNOT GET WINDOW HANDLE! --> CANNOT SET devicePixelRatio";
+
 	if(scene())
 	{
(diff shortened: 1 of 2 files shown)


gzotti:
I tried to use code from StelMainView::moveEvent(QMoveEvent * event) to get devicePixelRatio, however glWidget->windowHandle(); returns null. Not sure if this has any other consequences elsewhere or why Florian implemented it like that so many years ago. I can use StelMainView::getInstance().devicePixelRatio(). But should I just remove the code that retrieves null, or does it retrieve non-null on Linux? @10110111 Please run this first commit on Linux and move/resize the main window. The logfile should show CRIT warnings.

gzotti:
Apart from removing the qDebugs and QWarnings, IMO this should work. The net effect over 25.4 is that the stored window position is now not shifted further down (as it used to be since 2013).

10110111:
Interesting. I get null pointer from `windowHandle()` all the time, even after all the initialization has long finished. I suppose we'll never get a valid handle on Qt6. The actual effect of this is that `stelApp->setDevicePixelsPerPixel(...)` call never happens, which is not good. But I don't really understand why the pixel ratio has to be queried from `QWindow` rather than `QWidget`. Maybe it's also a relic of old versions of Qt?

The following patch makes things more sensible I think:
```diff
diff --git a/src/StelMainView.cpp b/src/StelMainView.cpp
index 3cf2ec7d58..6bc3b6a645 100644
--- a/src/StelMainView.cpp
+++ b/src/StelMainView.cpp
@@ -1624,21 +1624,15 @@ void StelMainView::moveEvent(QMoveEvent * event)
     //const QPoint &pos=event->pos(); // top-left of drawing area
     const QPoint &pos=frameGeometry().topLeft(); // including frame!
 
-    // We use the glWidget instead of the event, as we want the screen that shows most of the widget.
-    QWindow* win = glWidget->windowHandle();
-    if(win)
-    {
-        stelApp->setDevicePixelsPerPixel(win->devicePixelRatio());
-    }
-    else
-        qWarning() << "Cannot get window handle (to set device pixel ratio). (uncritical)";
-
     if (StelApp::isInitialized())
     {
-        qDebug() << "saving position with devicePixelRatio=" << devicePixelRatio();
+        // We use the glWidget instead of the event, as we want the screen that shows most of the widget.
+        const auto dpp = glWidget->devicePixelRatioF();
+        stelApp->setDevicePixelsPerPixel(dpp);
+        qDebug() << "saving position with devicePixelRatio =" << dpp;
 
-        StelApp::immediateSave("video/screen_x", int(std::lround(pos.x()*devicePixelRatio())));
-        StelApp::immediateSave("video/screen_y", int(std::lround(pos.y()*devicePixelRatio())));
+        StelApp::immediateSave("video/screen_x", int(std::lround(pos.x()*dpp)));
+        StelApp::immediateSave("video/screen_y", int(std::lround(pos.y()*dpp)));
     }
 }
 
```

Maybe we could even get rid of the use of `glWidget` and simply call the method of `this` as you did here for `immediateSave`.

gzotti:
Old code probably, yes. maybe from the older QOpenGLWidget.  But would that mean we need to test again on Qt5? Or is it safe to just avoid glWidget stuff? 

I have one weird observation: Double-clicking a window that was shifted to the top (video/screen_y=0) will make it 'maximized'. On a scaled (1.5) screen (at least) this makes the window frame narrower and records '-11' as video/screen_y. Don't know if this is another peculiarity of Windows or does the Linux version behave the same?

10110111:
> But would that mean we need to test again on Qt5? Or is it safe to just avoid glWidget stuff?

I think we should test, though I don't expect any issues here.

> Double-clicking a window that was shifted to the top (video/screen_y=0) will make it 'maximized'.

It's expected (if you mean double-clicking the title bar, not anywhere in the window).

> On a scaled (1.5) screen (at least) this makes the window frame narrower and records '-11' as video/screen_y.

Well IIRC, on Windows XP this was the default behavior of maximized windows: they were stretched to a size a big bigger than the screen and thus had a smaller title bar, and this caused some issue for Wine's simulation of this behavior, I don't remember what exactly was the problem.

> Don't know if this is another peculiarity of Windows or does the Linux version behave the same?

On Linux this generally depends on the window manager and the windowing system. On my system with X11, XFCE4 panel and `xfce4-windowck-plugin` I have no title bar when the window is maximized, but the panel is on top of the screen, so the real y=36 is saved. When the window isn't maximized, its true top-left corner is saved, including window decorations into the window rect.

gzotti:
No windowHandle with Qt5 either. OK, that part goes away. However, despite scaling 125%, in Qt5, devicePixelRatio=1, in Qt6 I have correct 1.25.  I have no problem not hunting to fix the Qt5 branch (needed on old PCs without scaling screens...)

10110111:
> However, despite scaling 125%, in Qt5, devicePixelRatio=1, in Qt6 I have correct 1.25.

It's OK, Qt5 doesn't really support fractional scaling.

src/StelMainView.cpp
```diff
@@ -727,8 +727,9 @@ void StelMainView::resizeEvent(QResizeEvent* event)
 			guiItem->setGeometry(QRectF(0.0,0.0,sz.width(),sz.height()));
 		if (StelApp::isInitialized()  && !isFullScreen())
 		{
-			StelApp::immediateSave("video/screen_w", sz.width());
-			StelApp::immediateSave("video/screen_h", sz.height());
+			qDebug() << "saving size with devicePixelRatio=" << devicePixelRatio();
```
10110111: I'd rephrase this, because `immediateSave` doesn't necessarily save anything. Maye something like "Calling immediateSave for size with devicePixelRatio =".

Same for the position.

10110111:
Other than the log messages, this LGTM.

src/StelMainView.cpp
gzotti: Actually the messages were here only for debugging and are scheduled for removal before merge.

Issue #5080, opened 2026-08-20
Support standalone dynamic plugins in official MSVC Windows builds


qlifee:
### Problem

The official Stellarium 26.2 Qt6 Windows x86_64 package appears unable to
support standalone third-party C++ plugins built with the matching MSVC
toolchain.

I am developing VisibilityContours as a standalone dynamic plugin. It builds
and loads successfully against an unmodified native Linux Stellarium 26.2
installation. The equivalent Windows plugin needs to resolve Stellarium API
symbols such as `StelModule`, `StelPainter`, `StelCore`, `StelApp`,
`SolarSystem`, and `Planet`.

### Tested package

- Stellarium 26.2 official Qt6 Windows x86_64 installer
- Qt 6.9.3
- MSVC 2022
- GitHub Actions `windows-2022`

### Findings

1. The official installer contains no `stellarium.lib`, `stelMain.lib`, or
   `stelMain.dll`.
2. `dumpbin /exports stellarium.exe` reports no exported Stellarium API
   symbols, including all six API families listed above.
3. Therefore an import library cannot be generated from the installed
   executable, and a standalone plugin cannot link against the official host.
4. Stellarium 26.2 sets `GENERATE_STELMAINLIB=0` for MSVC, with the source
   comment that splitting the main binary into a DLL for dynamic plugins does
   not seem to work with MSVC:
   https://github.com/Stellarium/stellarium/blob/v26.2/CMakeLists.txt#L525-L532

### Reproducer and diagnostics

- VisibilityContours PR with the Windows probe:
  https://github.com/qlifee/VisibilityContours/pull/2
- Successful diagnostic run against the official installer:
  https://github.com/qlifee/VisibilityContours/actions/runs/32342301401

The workflow downloads the exact official installer and source tag, installs
the package silently, inventories DLL/import-library files, and inspects the
`stellarium.exe` export table with MSVC `dumpbin`.

### Questions

1. Is there an endorsed way to build standalone third-party plugins for the
   official MSVC Windows package that I have missed?
2. If there is currently no supported path, would the project accept work to
   make a shared/exported `stelMain` library and matching import library work
   under MSVC?
3. If so, what API export and ABI policy would the maintainers prefer for
   external plugins?

The goal is for users to install only the plugin DLL under their normal
Stellarium user modules directory, without patching or rebuilding Stellarium.

gzotti:
[...]

PR #5085, opened 2026-08-25
Enable external Windows plugins with Qt 6 and MSVC

qlifee:
### Description

Addresses the technical foundation requested in #5080.

This change restores an opt-in shared stelMain build for Qt 6 with MSVC and packages the small supported surface needed by standalone dynamic plug-ins. It provides:

- stelMain.dll for the Stellarium runtime and installer
- stelMain.lib for external MSVC linking
- a relocatable companion SDK archive with three minimal headers
- a Stellarium CMake package and imported Stellarium::stelMain target
- machine-readable package, compiler, Qt, architecture, configuration, and runtime metadata
- SHA-256 manifests and archive checksum
- developer documentation for building and installing a plug-in
- an end-to-end Windows test with a source-isolated external plug-in

The SDK deliberately supports only implementing StelModule and StelPluginInterface. Other DLL exports remain internal.

The documented ABI policy is conservative: a binary plug-in must match the exact Stellarium package build, architecture, MSVC toolset, Qt version, Release configuration, and dynamic MSVC runtime. ABI compatibility across Stellarium releases or separately rebuilt packages is not guaranteed.

GENERATE_STELMAINLIB remains off by default under MSVC. An official Windows release job must opt in and publish the companion SDK archive before this becomes an end-user feature.

### Maintainer decisions requested

- Whether official Qt 6/MSVC package builds should enable GENERATE_STELMAINLIB by default
- Whether the companion SDK ZIP should be published beside the installer
- Whether WINDOWS_EXPORT_ALL_SYMBOLS is acceptable as an interim launcher implementation or should be replaced by a fully explicit internal export list
- Whether the initial three-header SDK surface is the preferred first supported API boundary

### Screenshots

Not applicable; this is build, packaging, documentation, and integration-test infrastructure.

### Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Housekeeping

### How Has This Been Tested?

GitHub Actions proof for the current rebased commit: https://github.com/qlifee/stellarium/actions/runs/32826433410

Test configuration:

- Operating system: Windows Server 2022
- Compiler: Visual Studio 2022, MSVC v143, x64
- Qt: 6.9.3
- Configuration: Release with the dynamic MSVC runtime
- Graphics card: not applicable; the integration test uses the low-graphics startup path

The workflow:

1. Builds Stellarium as stellarium.exe plus stelMain.dll.
2. Builds the SDK ZIP and validates its checksum, exact contents, and compatibility metadata.
3. Extracts the ZIP beneath a relocated path containing spaces.
4. Copies the test plug-in source outside the Stellarium source tree.
5. Configures and builds that plug-in only through the installed Stellarium CMake package.
6. Loads it with an independent Qt plug-in loader and checks its interface and module version.
7. Installs it in a temporary Stellarium user directory, launches Stellarium, and verifies that initialization writes a sentinel.

The default Linux and non-MSVC behavior remains unchanged. The focused Windows job completed successfully at current commit f3ee1c6 after a conflict-free rebase over data and translation-only upstream commits.

## Checklist

- [x] My code follows the project code style
- [x] I have commented hard-to-understand areas
- [x] I have made corresponding changes to the plug-in developer documentation
- [ ] I have updated the Stellarium User Guide (not applicable to this developer SDK)
- [x] My changes generate no known new warnings
- [x] I have added tests that prove the feature works
- [ ] Existing unit tests pass locally (the focused proof disables the unrelated test suite)
- [x] No dependent downstream changes are required for this draft


alex-w:
> Whether official Qt 6/MSVC package builds should enable GENERATE_STELMAINLIB by default

Yes, the official Windows/Qt6 (at least x86_64 - I’m not sure for arm64 architecture) package builds should enable GENERATE_STELMAINLIB by default. We’re use AppVeyor for producing and signing an official packages - so, probably some changes in `.appveyor.yml` file is needed.

gzotti:
If it works on x64, I can test on Arm. If the solution are MSVC switches or other configuration issues, they hopefully also apply there.

IMHO only support for Qt6 is needed here in case Qt5 causes any issues.

qlifee:
Thanks, @alex-w and @gzotti — this gives me a clear scope for the next revision. I will keep the general MSVC option opt-in, but enable GENERATE_STELMAINLIB in the official AppVeyor Win64/Qt6 package path. Qt5/MSVC will remain unchanged.

I will also make that job build and validate the matching SDK ZIP and checksum. One release-pipeline question: the existing SignPath webhook receives the AppVeyor job artifact list. Should the SDK ZIP and checksum be exposed through that same signed-release job, or kept outside the signing deployment as ordinary AppVeyor artifacts?

The implementation is architecture-aware, but only x64 has been tested end to end. Once x64 is green, I can prepare an ARM64 compile/package artifact and instructions for native testing. I will not claim ARM64 runtime support until that succeeds.

alex-w:
> I will also make that job build and validate the matching SDK ZIP and checksum. One release-pipeline question: the existing SignPath webhook receives the AppVeyor job artifact list. Should the SDK ZIP and checksum be exposed through that same signed-release job, or kept outside the signing deployment as ordinary AppVeyor artifacts?

A good question. Is SDK really need for end users?

gzotti:
Well the whole point is to have the ability to build dynamic plugins for end users who just don't want to recompile everything. I have never done this, and the solution needs clear  instructions to build our example https://github.com/Stellarium/stellarium-dynamic-plugin using QtCreator. From there, users should find their way. What would be the consequences if the SDK is not signed? (I have not yet tried anything in here.)

qlifee:
Thanks, @alex-w and @gzotti. By SDK users I mean plug-in developers, not ordinary Stellarium users. Someone installing Stellarium or an already-built plug-in only needs the official Stellarium package to provide the matching `stelMain.dll`. A developer building a standalone plug-in needs the matching headers, import library, CMake metadata, and build information as an optional companion download.

Leaving that SDK ZIP unsigned would not prevent compiling or loading plug-ins. The headers and import library do not use Windows Authenticode, and the ZIP's SHA-256 checksum detects corruption but does not independently authenticate the publisher. The SDK copy of `stelMain.dll` is only for isolated build/load tests; normal use relies on the copy supplied by the matching Stellarium installation.

I would avoid adding the ZIP blindly to the existing SignPath deployment. AppVeyor sends the uploaded artifact list to the webhook, while signing a DLL nested in the ZIP would change both its internal manifest and the ZIP checksum. My conservative proposal is to keep the current SignPath installer flow unchanged, expose the SDK ZIP and checksum only as ordinary artifacts from non-signing development/PR builds for testing, and decide official release authentication separately if a dedicated SignPath artifact configuration is wanted. Would that separation work for the project?

I agree that building the official `stellarium-dynamic-plugin` example with Qt Creator should be an acceptance criterion. I inspected it: it currently assumes Qt5/Linux and a complete Stellarium source tree through `STELROOT`; its old dialog also uses the removed `BarFrame` class. Building it unchanged would expand this first SDK by roughly 33 internal dependency headers. Instead, I propose a separate modernization of that example to Qt6, `find_package(Stellarium CONFIG REQUIRED)`, `Stellarium::stelMain`, and a normal Qt dialog, with complete Qt Creator/MSVC Release instructions. That keeps this initial SDK surface small while providing a visible example developers can actually follow.

One CI note: AppVeyor did not report a compilation or test failure for the latest revision. Build 54626175 was cancelled while still queued, before any matrix job started, and its record was then removed. All GitHub checks, including the Windows plug-in SDK workflow, passed. If convenient, an AppVeyor project maintainer could rerun the current PR head when capacity permits; otherwise the next legitimate revision will trigger a fresh build.

gzotti:
OK, for now I confirm with GENERATE_STELMANLIB the program builds and works as before on Arm64 (Win11). There is a stelMain.dll and a tiny stellarium.exe. I don't build installers or such, just run the program from within the build directory.

Now, given it never worked, I have never built a dynamic plugin and have long forgotten what I had tried. What to do next? 

EDIT: @qlifee I just saw you have developed what reads like an interesting plugin. Does that come with build instructions for 'advanced users', i.e. those who know what directories are, what to do with ZIP files, can operate a mouse, download and install programs and follow 'click here' instructions?  If a so-far regular user who has installed the regular build (must be tested with a beta installer!) wants to have it, what to do: install MSVC2022/26 and Qt6, set envvars, open QtCreator, locate a CMakeLists file and ... (click on build, wait for completion, copy dll to USER_DIR/modules, ..., restart Stellarium, activate plugin, ...). And given that such dynamic plugins are targeted at users (top 20%?), these instructions really go into the User Guide, chapter "12.3: Dynamic Plugins". (Don't worry about formatting if you are unfamiliar with LaTeX.) This should assume Stellarium has been installed from the installer package (which then includes your SDK). 

Another issue arises for those with several builds using the same user data directory. If the DLLs need to have same Qt version and Stellarium release version number, is there a way to still have multiple plugin DLLs in that folder? E.g. DLLs or subdirs with release number and Qt version in the names?

PR #4956, opened 2026-06-06
Comprehensive Remote Control Enhancement

kutaibaa-akraa:
This pull request introduces a major overhaul of the Stellarium Remote Control plugin, adding a professional‑grade script development environment, interactive API tools, and significant enhancements to the Sky Culture.

#### 1. Script Editor with CodeMirror (`scripteditor/`)

A fully featured script editor has been added, providing a modern coding experience directly in the browser.

**Key capabilities:**
- **Syntax highlighting** for Stellarium Script (`.ssc`) – recognizes built‑in objects, functions, keywords, and comments.
- **Intelligent autocomplete** – triggered by `Ctrl+Space` or typing a dot (`.`) after a namespace. Suggests all Stellarium API functions with real‑time filtering.
- **Case‑sensitive function validation** – valid functions appear in **green** (bold), invalid ones in **red** (wavy underline), case errors in **orange** (wavy underline).
- **Auto‑validation toggle** – off by default; users can enable real‑time validation as they type.
- **Revalidate button** – performs a full document validation, reporting the number of valid/invalid functions and case errors.
- **Parameter hints** – when you type `(`, a popup shows the function’s parameters, types, and descriptions.
- **Function detail popup** – click any function in the Quick Reference to see signature, description, parameters, return type, and an example. You can insert the example at the cursor or at a specific line.
- **Example collector** – loads all examples from the module JSON files (e.g., `ConstellationMgr.json`, `MilkyWay.json`) and inserts them into the editor, optionally as commented code.
- **Quick Reference panel** – displays all API functions grouped by module, with a search filter.
- **Execution history** – stores the last 20 executed scripts; click to reload.
- **File operations** – open/save `.ssc` files.
- **Line counter** – shows the current line count in the toolbar.
- **Copy to clipboard** – copies the entire editor content with a success notification.
- **Keyboard shortcuts help** – a comprehensive dialog listing all editor shortcuts (run, run selection, find, replace, go to line, indent, line operations, etc.).
Script_Editor

### 2. Code Generator (`codeGenerator.js`)

An interactive wizard that produces ready‑to‑use code for all Remote Control API operations.
Code_Generator_2

**Supported output formats:**
- **cURL** – command‑line requests with dynamic base URL resolution.
- **Stellarium Script (.ssc)** – directly usable in Stellarium.
- **Python** – using the `requests` library.
- **JavaScript** – using the Fetch API.

**Features:**
- Step‑by‑step process: Category → Action → Parameters → Generate.
[...]

PR #5086, opened 2026-08-26
RemoteControl: Add OpenAPI (SwaggerUI)

gzotti:
### Description

This adds a simple interactive explorative documentation/execution UI to the RemoteControl web interface.

Call with http://localhost:8090/docs/index.html

Based on swagger-ui-5.32.14. 

Setting up and the original stelOpenAPI.json was originally outlined by Google Gemini, but the JSON will likely grow in further steps, as the AI missed a few important parameters. 

This is a separate directory and should in no way conflict or question the much larger improvement, #4956

Fixes (complaints found elsewhere that the RC API is so badly documented.)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: (irrelevant)

## Checklist:


alex-w:
Please fix missing records in cmake file for *.json files

kutaibaa-akraa:
@gzotti,

This is impressive and a great step forward.

I have a few questions to better understand your implementation and next steps:  

- Did you use any automated tooling to generate the `stelOpenAPI.json` (e.g., from Doxygen comments or the source code), or was it manually written/curated?  I'm curious about the workflow.  

- Integration with my `apiExplorer.js`: As you know, I built a manual API explorer in `apiExplorer.js`. I completely agree that it should be deprecated or replaced with a solution based on OpenAPI/Swagger UI. 

What's your plan for the next phase?

gzotti:
> @gzotti,
> 
> This is impressive and a great step forward.

Thanks. actually I have seen such UI elsewhere and investigated a bit how to do that. It turned out everything is here already, only the yaml (or JSON) API description file was missing. 
 
> I have a few questions to better understand your implementation and next steps:
> 
>     * Did you use any automated tooling to generate the `stelOpenAPI.json` (e.g., from Doxygen comments or the source code), or was it manually written/curated?  I'm curious about the workflow.

As written above I asked Google's AI for a first shot (see first commits; it's really a good and easy start, although it had some errors), but when asked for a full file it only pointed me to the known Doxygen-created page. So now I am filling in the rest. It turned out that yaml allowed multi-paragraph documentation much more easily than JSON, so I switched format. Almost done, but -- famous last words -- I see an error :-P

It would be great to have the yaml generated from the *Service.cpp files automatically, but no. And there are certainly small errors like unknown/unspecified HTTP return values. But these are minor and rather harmless issues. 

>     * Integration with my `apiExplorer.js`: As you know, I built a manual API explorer in `apiExplorer.js`. I completely agree that it should be deprecated or replaced with a solution based on OpenAPI/Swagger UI.

I cannot estimate the amount of work you put into your explorer. Which is why I am a bit scared of "inheriting" it later for maintenance. Your extension around the scripting generator/editor is much more, but if a manually written API explorer can be replaced by the standardized Swagger OpenAPI, I think it would be easier to maintain. So if the script generator does not depend on your API explorer, and is otherwise not better, we can/should remove it.
 
> What's your plan for the next phase?

Actually this SwaggerUI is just a quick improvement for using the RC plugin API online. Indeed it used to be needlessly cumbersome to develop simple scripts like CURL command lines. And this Swagger UI just shows the right parameters and then even the actual CURL call.  And it is all in just one subdir of the RC webroot! One customized yaml that was manually written based on the Doxygen-generated page, and else just a standard JS package. 

If we agree on the following: merge of this ASAP, then you can adapt your branch, remove your API explorer if possible (I hope it is), and we hope to then have the script editor integrated for 26.3.

gzotti:
Not sure why, the /main/view POST of coordinates does not work yet. Else looks pretty good.

10110111:
> As written above I asked Google's AI for a first shot (see first commits; it's really a good and easy start, although it had some errors), but when asked for a full file it only pointed me to the known Doxygen-created page. So now I am filling in the rest.

You can try asking an LLM to check your results for mismatches with the actual C++ code. I've tried doing this with Claude AI (free plan) and the current state of this PR, with the following prompt (the code I pasted is skipped here for brevity):
```
Below are some C++/Qt implementations of Stellarium's RemoteControl API HTTP GET and POST handlers. Some services may implement only GET or only POST, or both. After the implementations I'm posting their description in OpenAPI format. Your task is to check whether the description matches the implementations: find mistakes in the description, missing entries in the docs, extra entries whose corresponding implementations I haven't given. If you find something that needs more context to decide if there's a mistake (e.g. a definition of a class, etc.), tell me.

Here are the implementations:
// ------------ start ----------------

void LocationSearchService::get(const QByteArray& operation, const APIParameters &parameters, APIServiceResponse &response)
<...>
}

// -------------- end ---------------------

Now here's the OpenAPI YAML:

openapi: 3.0.3
info:
  title: Stellarium RemoteControl API
  version: 1.1.4
<...>
```

And it found some things: https://claude.ai/share/168de3c7-3d4b-4025-b3c8-68a1cc071a1d, which even seem to make sense.

10110111:
Oh, missed the last two commits. The above link only discusses what was before them.

gzotti:
Good finds, thanks. Can Claude find what to change in the yaml to make the input fields in the /main/view POST function work? I will stop for ~10 hrs now.

10110111:
> Can Claude find what to change in the yaml to make the input fields in the /main/view POST function work?

Yes. Here's its reply for the current version of the YAML with my request to additionally comment on this issue.

---

## Fix needed for `/main/view` POST to actually work

The `j2000`, `jNow`, and `altAz` fields are declared as:

```yaml
j2000:
  type: array
  items:
    type: number
  maxItems: 3
  minItems: 3
```

But the implementation does **not** parse form-array-style parameters (`j2000[]=1&j2000[]=2...` or repeated `j2000=1&j2000=2&j2000=3`, which is what `type: array` in an `application/x-www-form-urlencoded` body normally implies). Instead it does:

```cpp
QByteArray j2000 = parameters.value("j2000");
QJsonDocument doc = QJsonDocument::fromJson(j2000);
QJsonArray arr = doc.array();
```

This means the single string value of the `j2000` parameter must itself be a JSON-encoded array literal, e.g. `j2000=%5B1.0%2C2.0%2C3.0%5D` (URL-encoded `[1.0,2.0,3.0]`). If a client follows the current schema and serializes it as a real array parameter, the server will fail to parse it (empty/invalid JSON array, size ≠ 3, silently ignored/error).

**Fix:** change `j2000`, `jNow`, and `altAz` to:

```yaml
j2000:
  type: string
  description: JSON-encoded array of 3 numbers [x, y, z] in the J2000 frame, e.g. "[0.1,0.2,0.97]".
```

(same for `jNow` and `altAz`).

Note this exact same bug pattern also affects `/main/focus`'s `position` field, which is parsed identically via `QJsonDocument::fromJson(parameters.value("position"))` — it should likewise be `type: string`, not `type: array`, for the same reason.

## Other mistakes / issues found in this version

**Improvements confirmed** (fixed from the prior version): `/main/status` location field is now `region` (correct), `propId` description now correctly references `propertyChanges`, the fabricated "synchronization array filters" text is gone, `/main/fov`'s fabricated smooth-zoom text is gone, `/view/skyculturedescription/` path no longer has the stray space, and Location/Script/SimbadService/StelAction/StelProperty `list`/`info`/`status`/`direct`/`stop` endpoints are now documented.

**New/remaining issues:**

1. **`/objects/info` responses block is malformed YAML/OpenAPI:**
```yaml
content:
  application/json:
    schema:
      type: object
  HTML/text:
```
`HTML/text:` isn't a valid media type and has no schema — should be `text/html:` with `schema: {type: string}`.

2. **`/view/skyculturedescription/` example text still copy-pasted from landscape:** "An example: landscapedescription/image.png returns image.png from the current skyculture directory" — should read `skyculturedescription/image.png`.

3. **`/locationsearch/nearby` description is copy-pasted from `/locationsearch/search`:** it says "Searches the term in the list of predefined locations... returns a JSON string array of the results," but this operation has no `term` parameter at all — it does a geographic proximity search via `pickLocationsNearby(planet, longitude, latitude, radius)`. The description needs to actually describe the nearby/radius search.

4. **`/locationsearch/nearby` parameters use `value:`** (e.g. `value: Earth`, `value: 3`) — not a valid OpenAPI parameter field; should be `example:`.

5. **`/location/planetlist` description cites `SolarSystem::getAllPlanetEnglishNames`**, but the code calls `ssys->getAllPlanets()` and iterates `PlanetP` objects — the cited method name doesn't match what's shown. (Flagging — can't fully confirm without `SolarSystem`'s class definition, but it doesn't match the given implementation.)

6. **`/scripts/info`:**
   - JSON format example omits the `version` field, which the code does insert (`obj.insert("version", scriptMgr->getVersion(scriptId).trimmed())`) — missing entry.
   - `html` query parameter is marked `required: true`, but the code only checks `parameters.contains("html")` as an optional flag (`if(parameters.contains("html"))`) — should be `required: false`.
   - Response description "Script pipeline initialization successful" is copy-pasted from `/scripts/run` and doesn't describe what this endpoint actually returns (script metadata JSON or HTML).

7. **`/scripts/status`** response description is also the copy-pasted "Script pipeline initialization successful" — doesn't fit a status-query endpoint.

8. **`/scripts/direct`**: `useIncludes` is typed as `type: bool` — invalid OpenAPI type; must be `type: boolean`.

9. **`/stelproperty/list`** JSON format description omits two fields the code actually inserts: `canNotify` and `isWritable` (`item.insert("canNotify", ...)`, `item.insert("isWritable", ...)`) — missing entries.

10. **`/stelaction/do`** still doesn't document the special-case behavior for `actionQuit_Global` / `actionShow_Remote_Control`: these trigger asynchronously and the response body becomes `"error: Stellarium or remote control closing!"` instead of the normal true/false/ok — this remains undocumented.

11. Minor: the top-level `description` bullet list has a dangling "and" — `**Search catalogs** and` immediately followed by a new bullet `**Manipulate the field of view**...` — reads as a leftover fragment from reordering the list.

gzotti:
Yes, specifying the V3D arrays as string was my fallback solution. I had just hoped these nicely separated x/y/z inputs would be possible.

gzotti:
OK, I hope I am done here. Some issues came from doc omissions in the doxygen doc which should also be fixed. Or should we point from Doxygen to the running SwaggerUI?

10110111:
> Or should we point from Doxygen to the running SwaggerUI?

No, Doxygen docs are simply available on our website, while SwaggerUI is a runtime feature of a Stellarium instance. Or do I misunderstand this?

alex-w:
The current fixes should be reflected in doxygen docs too for consistency between static docs (generated by doxygen for website) and dynamic docs in RC plugin.

gzotti:
You are right in terms of completeness to have everything in Doxygen. OTOH, now it's double-documented, and I guess most users would now explore the API interactively.  So, we either fix or remove the page from Doxygen.

alex-w:
Please fix page in doxygen :) OpenAPI is good addition, but not complete documentation

10110111:
ChatGPT found the following issue:

---

`/main/view` (GET)
The implementation accepts
```
ref=on
ref=off
anything else -> auto
```
The docs describe this, but say:

> "On"

with capital O.

Implementation is lowercase only.

`/main/view` (POST)
Exactly the same issue.

---

And indeed, when atmosphere is disabled, these two calls give different outputs:

 * http://localhost:8090/api/main/view?coord=jNow&ref=on
 * http://localhost:8090/api/main/view?coord=jNow&ref=On

10110111:
In fact, there's a bunch of things reported, you can try giving my prompt to different AIs. ChatGPT is completely free to use, it can even reply from an Incognito window without logging in. Claude requires a login, but has a free plan. I'm attaching the prompt as a file for those who want to run the checks. (If you want to edit/update the C++ reference code or the YAML in this prompt, search for `-- start` and `-- end` — they delimit the C++ code.)

[openapi-check-prompt.txt](https://github.com/user-attachments/files/31555114/openapi-check-prompt.txt)

10110111:
One thing that Claude doesn't find but ChatGPT does is this:

---

`/stelaction/list`
The documented JSON example is malformed.

It shows
```
<group> : [
    <actionName> : {
```
but the implementation returns
```
"group": [
    {
        "id": "...",
        ...
    },
    {
        ...
    }
]
```
The array contains objects.

There is no `<actionName>:` key.

gzotti:
Alright, ChatGPT reads docs thoroughly. Are our location altitudes metres above sea level (geoid) or actually above WGS84 ellipsoid (used in Satellites plugin)? Or the IAU1976 ellipsoid mentioned by Meeus in Astrn.Algorithms? 20 years ago it did not matter much. But when discussing clear specifications, we should add it.

10110111:
> Are our location altitudes metres above sea level (geoid) or actually above WGS84 ellipsoid (used in Satellites plugin)?

Looks like it's supposed to be WGS84 since 35687bde5f085f27033d28a2898cbbc14b43c3d0. Though, it's not quite so, because WGS84 semi-major axis is 6378137.0 m, while `StelUtils.hpp` has
```cpp
// Equatorial radius of the Earth in km (WGS-84)
#define EARTH_RADIUS 6378.1366
```
Maybe we shouldn't try to use the IAU2009 radius if we follow WGS84.

---

BTW, the `__f` is also not the same as in WGS84, see `stdsat.h`:
```
static const double __f=3.352779E-3;
```
This differs from 1/298.257223563 = 3.35281066474748071985e-3 for WGS84.

gzotti:
OK, these are details in the Satellites plugin. Unfortunately some deviations from the original sources (~2008?) are undocumented and should be reviewed. As I understand the original formulation used WGS-72. Is using WGS-84 an official update?
What about our location catalog, base_locations.txt? @alex-w ? What is the source of the altitudes? (Altitude corrections are important for (mostly) lunar parallax, i.e. solar eclipses, stellar occultations or such.) IAU, satellite operators/trackers and geodesy have different standards, we should decide for one and keep that. (Or add proj4 library and allow everything...)

To finalize this PR, I can just omit 'WGS84' and leave it open for now. Else I think the PR is ready to squash.

alex-w:
> What about our location catalog, base_locations.txt? @alex-w ? What is the source of the altitudes? (Altitude corrections are important for (mostly) lunar parallax, i.e. solar eclipses, stellar occultations or such.) IAU, satellite operators/trackers and geodesy have different standards, we should decide for one and keep that. (Or add proj4 library and allow everything...)

All properties of Earth's locations was obtain here - https://www.geonames.org

gzotti:
OK, that means if geonames elevation data come from gtopo30 (https://d9-wret.s3.us-west-2.amazonaws.com/assets/palladium/production/s3fs-public/atoms/files/GTOPO30_Readme.pdf) they are metres above sea level. Lon/Lat coordinates are GPS compatible WGS84, so we should probably adjust all our earth ellipsoid data and algorithms to WGS84, strictly, and stop using IAU1976 or other shapes. Not in this PR, but soon, like for 26.4. And then we need at least a geoid correction to find true WGS84 distance from the geocenter. (Proj4 or some restricted set.)

alex-w (approved):
Thanks!

alex-w:
Hmm… Should be link to API doc section in standard pages?

gzotti:
OK, I can add a sentence and a link to localhost:8090/docs/index.html to Doxygen RC plugin docs.

gzotti:
@alex-w the `QStringList StelLocationMgr::getRegionNames(const QString& planet) const ` will deliver regions of Mars if called with 'xbcmarsiantic' or 'xbcmarvenusiantic' will deliver Venus regions. Is that intended, or should the call not be a bit more restrictive, like matching full names only? OK for caseInsensitive.

plugins/RemoteControl/doc/remoteControlApi.doxygen
```diff
@@ -23,7 +23,13 @@
...
 The \ref remoteControl "RemoteControl plugin" provides an HTTP-based interface to Stellarium, implemented on the server-side through implementations of AbstractAPIService.
 The APIController maintains the list of registered services, and dispatches HTTP requests to the right service.
-The API is accessible under the server path `/api/`. For example, if you have the server running on the default port of 8090,
+The API is accessible under the server path `/api/`. 
+
+The API of a running RemoteControl plugin webserver instance of Stellarium can be explored interactively 
+using a <a href="https://swagger.io/open-source/swagger-ui/">Swagger UI</a> 
+at <a href="http://localhost:8090/docs/index.html"> http://localhost:8090/docs/index.html</a>.
+
+To actually use the API in s web-based application, for example, if you have the server running on the default port of 8090,
```
10110111, line 32: _"in s web-based"_ — did you mean "in **a** web-based"?

gzotti: Uhm, sure.

alex-w:
> @alex-w the `QStringList StelLocationMgr::getRegionNames(const QString& planet) const ` will deliver regions of Mars if called with 'xbcmarsiantic' or 'xbcmarvenusiantic' will deliver Venus regions. Is that intended, or should the call not be a bit more restrictive, like matching full names only? OK for caseInsensitive.

No, minor planets has regions too, but you may call names of minor planet with and without designation.

plugins/RemoteControl/webroot/docs/stelOpenAPI.yaml
```diff
@@ -0,0 +1,1083 @@
...
+    post:
+      tags:
+        - Script Service
+      summary: Stops running script.
+      description: Stops the execution of a running script.
+      requestBody:
+        required: false
+      responses:
+        '200':
+          description: Script pipeline initialization successful
```
10110111, line 979: Maybe "Script stopping requested successfully" would make more sense here.

plugins/RemoteControl/webroot/docs/stelOpenAPI.yaml
```diff
@@ -0,0 +1,1083 @@
...
+                id:
+                  type: string
+                  description: System toggle shortcut name (e.g., 'actionShow_Atmosphere').
+      responses:
+        '200':
+          description: |
+            Shortcut method fired (ok, or true/false results for toggle actions).
+            
+            When actions 'actionQuit_Global' or 'actionShow_Remote_Control' are given, 
+            these trigger asynchronously and the response body becomes "error: Stellarium or remote control closing!" instead
```
10110111, line 1031: `error: invalid id parameter` response isn't documented.

10110111:
Also these two are found by Claude:

* `/main/view` **POST** has no error responses documented. The code can return "invalid j2000/jNow/altAz format, use JSON array of 3 doubles" or "requires at least one of az,alt,j2000,jNow parameters", none of which appear under responses.
* `/main/focus` **POST** doesn't document "invalid position format", returned when position parses to an array of size ≠ 3.

gzotti:
Fixed also these.

kutaibaa-akraa:
@gzotti 

I noticed that three endpoints from your `scenery3d plugin` have not been documented. Their documentation depends on the plugin's availability—specifically, its activation at startup. The missing endpoints are:

- `GET  /api/scenery3d/listscene`
- `POST /api/scenery3d/loadscene` (requires an `ID` parameter) save me if I'm wrong!
- `GET  /api/scenery3d/scenedescription/` (requires an `ID` parameter) (I didn't test) 

Should we document these for the sake of consistency? They are currently absent from the "[RemoteControl plugin HTTP API description](https://stellarium.org/doc/26.0/remoteControlApi.html)", and there is no reference to them in the [3D Sceneries plug-in](https://stellarium.org/doc/26.0/group__scenery3D.html) documentation either.

Additionally, I realize that I also omitted these endpoints from `apiExplorer.js`.
Have a good day

plugins/RemoteControl/webroot/docs/stelOpenAPI.yaml
kutaibaa-akraa, line 645: Missing  responses  for /view/skyculturedescrition/

kutaibaa-akraa, line 611: Missing  responses  for /view/landscapedescription/

kutaibaa-akraa, line 695: Returns the list of regions. If the optional `planet` parameter is provided,
        only regions belonging to that planet are returned. If no planet is specified,
        all regions from all planets are returned

gzotti: fixed

gzotti: fixed

gzotti: fixed (7188b3ba1fec4ac7b1964dc645e755a8317fe16d)

PR #5088, opened 2026-08-26
Make Lightweight atmosphere model the default

10110111:
So Lightweight atmosphere model has been rather stable for me. I think it can be made the default model.


gzotti:
OK to become the new default. Just SUG 11.2.3 needs a 1-line update as well. And maybe we should add that for those on weak field laptops on batteries, Preetham is still a bit less energy consuming. (But lightweight is a great compromise!)

alex-w:
OK for me

gzotti (approved):
OK, thanks.

PR #5113, opened 2026-09-14
Improve legibility of text esp. in screenshots.

gzotti:
### Description

@sushoff asked me to investigate options for clearer screenshot printing. 

Here are attempts to add additional darkness behind the Infotext.

1. make the dropshadow configurable: blur radius and opacity are now configurable. However, The drop shadow just won't get really dark and large.
2. set the HTML with dark background. But there are issues with Qt's HTML rendering. 

found:
- HTML font tag is actually ~20 years deprecated. use style attributes instead.
- background color can only be set by a div/span combination,
- table font and background color 'inherit' does not work

Ideas:
- [x] Use dark background only for night screenshots (logic from daylight text color)
- [ ] GUI switch vs. config.ini?

This needs some care also with our Infotext postprocessing etc.

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Operating system: Win11
* Graphics Card: Geforce, but should not matter

## Checklist:


gzotti:
I face a problem in that text color attributes are swallowed by Qt's rendering of HTML tables. These need dedicated CSS attributes color and (where needed) background-color, so the colors will be set several times. 
StelObject::getInfoColor() and overrides will need to consult StelCore which would be allowed to override the object colors with forced or daylight colors. A problem is the occasional mix/misuse of labelColor and InfoColor, this may switch some colors.  
A technical issue is the darkened background. Here I think the easiest approach is changing getInfoColor to return a Vec4f with the last element influencing opacity of the text background. In its "daylight" analysis, StelCore could decide (if user wants) to set a darkened background for night screenshots.  
Another issue is stripping away color when InfoString::noFont is requested. (This was obviously named after the long-deprecated font tag that enclosed the HTML and should probably be renamed noColor.) So, instead of not adding the font/color group, we just need to strip out the CSS color style tags via regex.

10110111:
Why do you even define the style inline when you could define a bunch of classes in a separate CSS file and give the elements the class name? E.g. instead of 
```html
<table style='margin:0em 0em 0em -0.125em;border-spacing:0px;border:0px;color:%1;'>
```
you'd use
```html
<table class='info-string'>
```
and have this in the common CSS:
```css
.info-string {
    margin: 0em 0em 0em -0.125em;
    border-spacing: 0px;
    border: 0px;
    color: #YOURCOLOR;
}
```

gzotti:
Thanks. The inline CSS was already here, so I just expanded it. I am actually not familiar with interacting with external CSS: The YOURCOLOR will have to change during infostring generation/before display. How would I set this? In addition, the HTML surrounding the tables also needs the color/background. (This is now done by the span element)

10110111:
Looks like you can set the style sheet for `InfoPanel` using `document()->setDefaultStyleSheet(css);`. This should be done *before* calling `setHtml(infoHTML)`, it won't apply to the existing text.

As for `YOURCOLOR`, you can use a `%1`-style placeholder for it in the common CSS, and replace it before applying this CSS. This will be the single place you ever define this color.

10110111:
> Unclassified text still has no color.
> Back to wrapping in font tags?

What kind of text? Something out of any tags? Maybe you could wrap such text in `<p class=...>` or `<span class=...>`?

gzotti:
Yes, I reactivated the div class=.... However, the background-color is not used. Apart from that it seems to work now.

src/gui/StelGui.cpp
```diff
@@ -1553,7 +1553,7 @@ void StelGui::copySelectedObjectInfo(void)
 {
 	const auto cb = QGuiApplication::clipboard();
 	const auto md = new QMimeData;
-	const auto colorRE = QRegularExpression(" color=\"?#[0-9a-f]+\"?");
+	static const QRegularExpression colorRE("color: ?#[0-9a-f]+;?");
```
10110111: I guess this has no effect now. In fact, now that the CSS is separate, there's no need to do anything here anymore, the text should have default color automatically.

src/core/modules/Planet.cpp
```diff
@@ -745,7 +745,7 @@ QString Planet::getInfoString(const StelCore* core, const InfoStringGroup& flags
 	if (flags&Distance)
 	{
 		if (withTables)
-			oss << "<table style='margin:0em 0em 0em -0.125em;border-spacing:0px;border:0px;'>";
+			oss << QString("<table class='info-string'>");
```
10110111: The `QString(...)` wrapping is no longer necessary at this iteration. Same in other instances below.

src/gui/StelGui.cpp
gzotti: Ah right, the only place where color is now written is the InfoPanel. 
That would mean, the noFont InfoStringGroupFlag can be removed, and all color removal regexp code as well. (The HTML InfoString is displayed without color in the RemoteControl UI. Will check in next cleanup round.)

src/gui/SkyGui.cpp
```diff
@@ -35,7 +35,7 @@
 #include <QRegularExpression>
 #include <QGraphicsDropShadowEffect>
 
-InfoPanel::InfoPanel(QGraphicsItem* parent) : QGraphicsTextItem("", parent)
+InfoPanel::InfoPanel(QGraphicsItem* parent) : QGraphicsTextItem("", parent), opacity(0)
```
10110111: It would be better to set the default value inline. It would make it visible immediately at the declaration of `opacity`, and avoid possible future "X will be initialized after Y" warnings.

src/gui/SkyGui.cpp
```diff
@@ -97,11 +98,56 @@ void InfoPanel::setTextFromObjects(const QList<StelObjectP>& selected)
 	}
 	else
 	{
+		static const QString css(
+			    "h2 {"
+				      "color: %1;"
+				      "background-color: rgba(0,0,0,%2%);"
+				"} "
+			    ".info-string {"
+				"color: %1;"
```
10110111: Why not indent contents as in `h2` above?

src/gui/StelGui.cpp
```diff
@@ -1553,8 +1553,12 @@ void StelGui::copySelectedObjectInfo(void)
 {
 	const auto cb = QGuiApplication::clipboard();
 	const auto md = new QMimeData;
-	const auto colorRE = QRegularExpression(" color=\"?#[0-9a-f]+\"?");
-	md->setHtml(skyGui->infoPanel->getSelectedHTML().replace(colorRE, ""));
+	// We must trim away the color and background tricks from the infoPanel
+	QString htmlInfo=skyGui->infoPanel->getSelectedHTML();
```
10110111: Instead of doing all the hacks below it would make more sense to store `InfoPanel::infoHTML` without any styling, and only apply the styling to a temporary variable that would be passed to `InfoPanel::setHtml()`. Then `InfoPanel::getSelectedHTML()` will return the original unstyled HTML, while on the screen it will have the CSS applied.

gzotti: Actually, indeed. Again, tried to keep the previous style.  I see two options:

- keep infoHTML, set the "pure" infoString and decorate with div and span only in setHTML.
- Keep the current trim style and remove the infoHTML variable. The HTML can be retrieved via QGraphicsItem::toHtml(), or not?

Why did we then need the infoHTML in the first place?

src/gui/SkyGui.cpp
gzotti: You mean "inline" as in "in the header file"? We never did that. -- Ah, that's C++11. But I prefer the old way.
Actually I can remove it entirely as it is then set from config.ini.

src/gui/StelGui.cpp
10110111: Trimming looks like a kludge, so it's not the first option.

As for why, I wanted to keep the sub-/superscripts first of all, which get flattened in plain text. And the prominent header as a bonus. But not the colors.

Not sure about the bold font, which the original commit message (64f09a0add45f08e416a6fd183fe8a2c43ed3753) mentions. Is it preserved? I don't see any mentioning of boldness/weight in the common CSS of this PR, so I guess it's hidden in some legacy HTML tags.

src/gui/SkyGui.cpp
10110111: > But I prefer the old way.

And this has already lead to annoying compiler warnings several times. Plus, in some cases these warnings may be correct (if initialization of a member depends on previous member being already initialized), because the order of initialization is defined in the declaration, not in the initializer list.

src/gui/StelGui.cpp
gzotti: OK I tried: 
- getSelectedHTML(): whatever we deliver
- toHTML(): huge HTML including DOCTYPE, default CSS etc.
- getSelectedText(): Plaintext info
- toPlaintext(): identical to getSelectedText.

--> We could remove getSelectedText, but keep infoHTML, and decorate with div/span only where needed.

10110111: > We could remove getSelectedText

Better keep it, it provides some encapsulation. `toPlainText` is inherited from `QGraphicsTextItem`, it's not our API. We may want to tweak `getSelectedText` later, e.g. by replacing subscripts with Unicode subscripts or whatnot (not that I'm advocating for this, just an example).

> but keep infoHTML, and decorate with div/span only where needed

Yes, my idea is to have `infoHTML` as the base that we could return with minimal (or no) tweaks.

src/gui/SkyGui.cpp
gzotti: And I am aware of this and keep cleaning up these warnings (fix order of initializer list) when I see them. In this case there was no warning, only correct initialisation (that was then changed again, so can be said redundant).

alex-w:
I hope copying data via clipboard is still alive?

gzotti:
> I hope copying data via clipboard is still alive?

Of course. I tested both types, HTML in LibreOffice and plaintext in a text editor.

gzotti:
Any more comments/additions?

10110111:
The title background appears to be too wide:

stellarium-021

10110111:
Negative opacity fails to change the text color, leaving it unreadable:

stellarium-022

10110111:
> The title background appears to be too wide

There's an interesting way to combat this, but there is a drawback. The patch:
```diff
diff --git a/src/gui/SkyGui.cpp b/src/gui/SkyGui.cpp
index 534a28096c..9b9c44f4f0 100644
--- a/src/gui/SkyGui.cpp
+++ b/src/gui/SkyGui.cpp
@@ -31,6 +31,7 @@
 #include <QTimeLine>
 #include <QGraphicsSceneMouseEvent>
 #include <QSettings>
+#include <QTextCursor>
 #include <QTextDocument>
 #include <QRegularExpression>
 #include <QGraphicsDropShadowEffect>
@@ -101,18 +102,15 @@ void InfoPanel::setTextFromObjects(const QList<StelObjectP>& selected)
         static const QString css(
             "h2 {"
                   "color: %1;"
-                  "background-color: rgba(0,0,0,%2%);"
                 "} "
             ".info-string {"
                   "color: %1;"
-                  "background-color: rgba(0,0,0,%2%);"
                 "} "
             "table.info-string {"
                   "margin: 0em 0em 0em -0.125em;"
                   "border-spacing: 0px;"
                   "border: 0px;"
                   "color: %1;"
-                  "background-color: rgba(0,0,0,%2%);"
                 "} ");
 
         // just print details of the first item for now
@@ -120,7 +118,7 @@ void InfoPanel::setTextFromObjects(const QList<StelObjectP>& selected)
         StelCore *core=StelApp::getInstance().getCore();
 
         Vec3f color = selected[0]->getInfoColor();
-        QString opacity=QString::number(this->opacity);
+        int bgAlpha = this->opacity * 255 / 100;
         if (StelApp::getInstance().getFlagOverwriteInfoColor())
         {
             // make info text more readable...
@@ -130,21 +128,27 @@ void InfoPanel::setTextFromObjects(const QList<StelObjectP>& selected)
         {
             // make info text more readable when atmosphere enabled at daylight.
             color = StelApp::getInstance().getDaylightInfoColor();
-            opacity="0";
+            bgAlpha = 0;
         }
         // A negative opacity forces dark background even for daylight scenes.
         if (this->opacity<0)
-            opacity=QString::number(qAbs(this->opacity));
+            bgAlpha = qAbs(this->opacity) * 255 / 100;
 
         infoHTML = selected[0]->getInfoString(core, infoTextFilters);
         selected[0]->removeExtraInfoStrings(StelObject::AllInfo);
 
-        document()->setDefaultStyleSheet(css.arg(color.toHtmlColor(), opacity));
+        document()->setDefaultStyleSheet(css.arg(color.toHtmlColor()));
 
         // We need the span to set the background opacity.
-        setHtml(QString("<div class='info-string'><span style='background:rgba(0, 0, 0, %1%);'>").arg(opacity)
-            + infoHTML
-            + "</span></div>");
+        setHtml("<div class='info-string'>"
+             + infoHTML
+             + "</div>");
+        QTextCursor cursor(document());
+        cursor.setPosition(0);
+        cursor.movePosition(QTextCursor::End, QTextCursor::KeepAnchor);
+        QTextCharFormat fmt;
+        fmt.setBackground(QColor(0,0,0, bgAlpha));
+        cursor.mergeCharFormat(fmt);
     }
 }
 
```
The drawback: for some reason `<b>` and `</b>` result in "vertical line" artifacts when |opacity| is not 100%:

stellarium-023

gzotti:
> The title background appears to be too wide:

I have seen it. But I have no idea what to do here.

The requested-for application of this manual-only setting is to create more legible screenshots at night, i.e., to cover stars and other sky elements in the text area. For daylight screenshots, you must obviously use "override text color" and set a daylight color other than black. I don't fight any 'black on black' situation the user can mitigate. The daylight application (negative opacity) is ugly to begin with and I don't know if anybody ever would use it (It may help visually impaired, though), but was easy to add.

gzotti:
> QTextCursor cursor(document());

Interesting hack. But it's a new hack, not what I call elegant.  
Maybe the h2 issue stems from the rtl/ltr settings that might extend the width to the maximum width of the infoPanel. Would setting the background inside a span within the rtl/ltr clauses help? Does it really matter?

10110111:
> Maybe the h2 issue stems from the rtl/ltr settings that might extend the width to the maximum width of the infoPanel. Would setting the background inside a span within the rtl/ltr clauses help?

I think the header takes the whole width regardless of rtl/ltr settings. The styled span inside the header (with unstyled header) should fix it, though I didn't check.

> Does it really matter?

It's a matter of completeness of the feature.

gzotti:
I played around with defining spans inside the Planet::getInfoLabel() and other attempts that could have helped assuming the rtl/ltr cause. Seems not to help. But I also saw how broken the whole InfoString looks in Arab!  

At this point I declare the long dark title bar a feature. Currently this PR feature has one user.

10110111:
Hmm, something strange is going on. I have black text on black background even during the night with the following settings:
```
info_panel_opacity                              = -100
info_shadow_opacity                             = 0
info_shadow_radius                              = 20
```
stellarium-022

gzotti:
With the latest commit?

10110111:
Tested with 015e1c6353aef7829dad1d63355b98b2df8faab3. Before it there was no such issue.

gzotti:
Alright, my daylight color is yellow, yours is black. Now I hope your OverwriteColor is not black.

10110111:
> Alright, my daylight color is yellow, yours is black. Now I hope your OverwriteColor is not black.

The question is not what "my" colors are, but rather what the defaults are. I don't remember to have changed colors.

At least, with the latest commit things seem to look OK.

10110111 (approved):

PR #5107, opened 2026-09-05
Reduce Moon visibility in daylight

Atque:
This branch gets rid of the burnt-out Moon in daylight skies. Works equally well in Preetham and ShowMySky. See the screenshot below for a comparison.

Fixes #3067 

### Screenshot:
new_moon_branch

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update
- [ ] Housekeeping

**Test Configuration**:
Fedora 44
Qt 6.11.2

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [ ] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #5114, opened 2026-09-14
Show local sky luminance in Pointer Coordinates plug-in

### LINKED ISSUE #1404: Displaying the value of the luminance of the sky
Feature request:

Dear all,

It would be extremely useful for users interested in visual issues to know the value of the luminance (brightness) of the sky at the alt-az points the celestial objects are located. This could perhaps appear as an additional data when clicking on an object (at the end of the list with its magnitude, coordinates, et al)

As far as I could see in the user's manual, this luminance is already calculated by Stellarium, in order to determine the visibility of an object against the sky background (natural or light polluted). According to the manual "The skybright class The class governs physical sky brightness values based on Bradley Schaefer’s VISLIMIT.BAS basic source code (Schaefer, 1998). The basic sources are available on the Sky and Telescope web site" 

Displaying the sky luminance in the SI units of cd/m^2 (easy conversion from Lamberts) would really be a very welcome addition. Please feel free to contact me for further info: salva.bara at usc.gal

Atque:
### Description
This adds the ability for users to use the Pointer Coordinates plug-in to estimate sky luminance in cd/m².

- We expose a common API through `Atmosphere` and `LandscapeMgr`, as to keep the plug-in independent from the individual atmosphere models
- We use the luminance reported by the respective atmosphere model, with interpolation matching their rendering paths
- To ease workload on the GPU, we only update 5 times per second

This is the first step towards more accurate twilight/moonlight simulation.

Code generated by Codex (GPT-6 Astra). I've tested the code and it runs smoothly on my system and seems accurate (of course, I haven't verified it).

Fixes #1404 (which seems to have stalled)

### Screenshot:
bild

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Fedora 44
Qt 6.11.2

## Checklist:


10110111:
Hmm, the approach seems a bit ugly in all atmospheres, and especially in the Lightweight one. And in ShowMySky the code tries to catch `ShowMySky::Error` from a pure OpenGL code, which can never throw this exception.

Also, it seems to sample only the atmosphere, so e.g. a landscape will be transparent to this code, right?

I'm planning to restart my attempt to implement linear-light rendering for 26.4, then #5107 will be solved automatically, and this PR will become doable in a much cleaner way.

Atque:
> Hmm, the approach seems a bit ugly in all atmospheres, and especially in the Lightweight one. And in ShowMySky the code tries to catch `ShowMySky::Error` from a pure OpenGL code, which can never throw this exception.

Hm, okay. Are you suggesting we should enable it only for SMS?

> Also, it seems to sample only the atmosphere, so e.g. a landscape will be transparent to this code, right?

Yes, so hovering below the horizon, on the ground for instance, reports the mirrored atmosphere's luminance.

> I'm planning to restart my attempt to implement linear-light rendering for 26.4, then #5107 will be solved automatically, and this PR will become doable in a much cleaner way.

Great to hear! But, should I keep this open? Put it as draft? I have no clue if the code could be useful in the future.

10110111:
> But, should I keep this open? Put it as draft? I have no clue if the code could be useful in the future.

I guess this could stay a draft until I actually do implement what I'm planning (or fail).

10110111:
> Are you suggesting we should enable it only for SMS?

No, it would be useful to be able to compare the outputs of different models numerically.

Atque:
Just a thought while playing around; after converting cd/m² to mag/arcsec², I noticed that ShowMySky shows a zenith luminance of well above 30 at night, which seems unlikely as the darkest measurements of an SQM usually yield about 22 from the pristine "Bortle 1" locations. Websites such as [Light pollution map](https://www.lightpollutionmap.info) also present their darkest locations as such.

Is it the lack of airglow and the combined light of the Milky Way, zodiacal light and starlight that contributes, none of which SMS takes into account?

10110111:
> Is it the lack of airglow and the combined light of the Milky Way, zodiacal light and starlight that contributes, none of which SMS takes into account?

Right, see https://github.com/Stellarium/stellarium/blob/69888f4f47af2aa4d1c9d14ec3853e874e253c73/src/core/modules/AtmosphereShowMySky.cpp#L752 and https://github.com/Stellarium/stellarium/blob/69888f4f47af2aa4d1c9d14ec3853e874e253c73/src/core/modules/AtmosphereShowMySky.cpp#L726, where this is manually inserted, simply copied from Preetham code.

I did (and still do) want to add support for simulating airglow, but I'm stuck at [this question](https://earthscience.stackexchange.com/q/26467/12360).

PR #5116, opened 2026-09-17
Add extinction colors from ShowMySky transmittance data

### LINKED ISSUE #1223: More detailed extinction coloring
Currently, the extinction color seems to be generalized over the whole disk based on the center point of the disk. This is especially noticeable when the moon is being enlarged. In some photos, we can see different shades of red in a setting moon:
https://media.gettyimages.com/videos/moonset-over-downtown-area-near-n-seoul-tower-and-cheongdamdaegyo-video-id1219291937?s=640x640
https://apod.nasa.gov/apod/ap051125.html

Also, perhaps add at least some reddening of stars (possibly only the bright ones). I've seen Altair close to the horizon, very red, once.

Atque:
This PR adds extinction colors from transmittance data of the ShowMySky model, which appear more accurate than the current ad hoc method of reddening objects low in the sky. It's applied on SMS and Lightweight atmosphere models, and is not intensified with increased coefficient set by the user. Preetham model is left untouched.

It also takes into account per-pixel extinction coloration. This is not very visible under normal circumstances at sea level, but when enlarging the Moon beyond sanity or when observing from high altitude, it becomes obvious.

Fixes #1223

Code generated by Codex (GPT-6 Astra set to `high`). To successfully run it, make sure it's installed by e.g. `sudo make install`.

### Screenshot:
bild

### Type of change

- [ ] Housekeeping

**Test Configuration**:
Fedora 44
Qt 6.11.2

## Checklist:


src/CMakeLists.txt
```diff
@@ -179,6 +179,8 @@ SET(stellarium_lib_SRCS
      core/planetsephems/htc20b.c
      core/planetsephems/htc20b.h
      core/modules/AtmosphereLightweight.cpp
+     core/modules/AtmosphereTransmission.hpp
```
alex-w, line 182: Hmm… looks like wrong ordering of including files…

10110111:
* A whole megabyte of extinction data in RGB for one altitude? ShowMySky's transmittance in 16 spectral bands and 32 altitude levels takes the same amount of space!
* For ShowMySky atmosphere this data really should be derived from transmittance that should be provided by ShowMySky, not by an ad-hoc loader of opaque data (ShowMySky never guaranteed that the format will be the same in newer versions!). What's done here is an ugly hack, treating ShowMySky as a closed-source black box.
* From what I remember, RGB transmittance computed naively is generally insufficient for correct colors. At least, some care must be taken to convert spectral transmittance to its RGB values, it's not simply a dot product with the color matching functions. This is because of metamerism of colors.
* In your screenshot the Moon looks weirdly oversaturated at the bottom and very bright for this amount of coloration.
* Not sure this improvement makes much sense when the Moon texture is gray and that of Venus is yellow. Besides, IIRC there was some artificial increase of color saturation for the Sun or something, which would be thrown away by this change. (I don't think it's bad in itself, just worth mentioning.)

I think the ShowMySky-derived extinction is just another thing that's waiting its turn when HDR rendering arrives.

Regarding AI agents, I think you really should actually read what they emit and evaluate it critically. They often do quite crazy stuff—not in a good sense. Just recently I asked Codex (in a different project) to align the solar corona texture to Sun's poles, and instead of requesting Sun's model matrix from app's API to get the orientation of the poles on the screen it connected to the NASA website, got RA/DE of the solar rotation axis from there, added them as an additional dataset to the project, and used it. Human should really supervise the agent, not just check that the code works. Otherwise you're very likely to end up with AI slop.

Atque:
> * A whole megabyte of extinction data in RGB for one altitude? ShowMySky's transmittance in 16 spectral bands and 32 altitude levels takes the same amount of space!
> 
> * For ShowMySky atmosphere this data really should be derived from transmittance that should be provided by ShowMySky, not by an ad-hoc loader of opaque data (ShowMySky never guaranteed that the format will be the same in newer versions!). What's done here is an ugly hack, treating ShowMySky as a closed-source black box.
> 
> * From what I remember, RGB transmittance computed naively is generally insufficient for correct colors. At least, some care must be taken to convert spectral transmittance to its RGB values, it's not simply a dot product with the color matching functions. This is because of metamerism of colors.
> 
> * In your screenshot the Moon looks weirdly oversaturated at the bottom and very bright for this amount of coloration.
> 
> * Not sure this improvement makes much sense when the Moon texture is gray and that of Venus is yellow. Besides, IIRC there was some artificial increase of color saturation for the Sun or something, which would be thrown away by this change. (I don't think it's bad in itself, just worth mentioning.)
> 
> 
> I think the ShowMySky-derived extinction is just another thing that's waiting its turn when HDR rendering arrives.

Hm, all right, I appreciate your honesty, this is clearly much more complicated than I had imagined. I thought this was a good idea, and it looked good on screen. Alas, if we're aiming for physical correctness, then this might very well not be enough. And certainly, if it hinders future work or makes it unnecessarily complicated to alter or circumnavigate, then I don't want to add vibe coded quick fixes.

> Regarding AI agents, I think you really should actually read what they emit and evaluate it critically. They often do quite crazy stuff—not in a good sense. Just recently I asked Codex (in a different project) to align the solar corona texture to Sun's poles, and instead of requesting Sun's model matrix from app's API to get the orientation of the poles on the screen it connected to the NASA website, got RA/DE of the solar rotation axis from there, added them as an additional dataset to the project, and used it. Human should really supervise the agent, not just check that the code works. Otherwise you're very likely to end up with AI slop.

Indeed AI is a powerful tool, and it lets laymen such as myself write actual code, and it obviously needs human supervision. I have also had a few bad experiences, and it's quite annoying when precious tokens are being consumed by misunderstandings on the AI's behalf. But please don't judge this work as AI slop. I can't validate the code itself and judge which approach is the better, because I lack proficiency, but I do test the code before pushing it, making sure it covers my needs.

But I'll close this, as full HDR rendering support is a bit too high for what I can aim for. At least today.

10110111:
> I can't validate the code itself and judge which approach is the better, because I lack proficiency, but I do test the code before pushing it, making sure it covers my needs.

Well it's not great, but you can at least ask the AI agent to do a technical review of your changes (in a new session, so that it doesn't have the context of your development discussion). Particularly it should look for things generally considered bad design, compatibility issues, not being future proof regarding the use of libraries, add your own ideas. Even better if you can ask multiple different AI tools to do the review (e.g. add Claude Code to Codex).

This will not raise my concern about the size of your additional data, but I think it should find the loader of the external library's data ugly (though I didn't check, maybe they aren't smart enough yet).

PR #5124, opened 2026-09-20
Disable selection for ShowMySky when build does not enable it.

gzotti:
### Description

A build where ENABLE_SHOWMYSKY is off should not list it as available in the AtmosphereDialog.
This build just removes the entry from the ComboBox.

Fixes # (issue)

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?

**Test Configuration**:
* Graphics Card: irrelevant
## Checklist:


src/gui/AtmosphereDialog.cpp
```diff
@@ -122,6 +122,19 @@ void AtmosphereDialog::createDialogContent()
 		}
 	}
 
+#ifndef	ENABLE_SHOWMYSKY
+	{
+		for (int row = 0; row < ui->atmosphereModel->count(); ++row)
```
10110111: This just duplicates the loop above. Why not simply do
```cpp
#ifdef ENABLE_SHOWMYSKY
    const bool showMySkyEnabled = true;
#else
    const bool showMySkyEnabled = false;
#endif
```
and then use it as an additional condition to `isGLES` above?

gzotti: I finally had a similar solution seconds after switching off for the night...

src/gui/AtmosphereDialog.cpp
```diff
@@ -110,7 +110,11 @@ void AtmosphereDialog::createDialogContent()
 	connectBoolProperty(ui->showMySky_pseudoMirrorEnabled, "LandscapeMgr.flagAtmospherePseudoMirror");
 	connectIntProperty(ui->showMySky_eclipseSimulationQualitySpinBox, ECLIPSE_SIM_QUALITY_PROPERTY);
 
-	if (StelMainView::getInstance().getGLInformation().isGLES)
+	if ((StelMainView::getInstance().getGLInformation().isGLES)
+		#ifndef	ENABLE_SHOWMYSKY
```
10110111: The logic becomes quite convoluted this way, not least because of the hardly visible `n` in `ifndef`. I'd prefer to see an easily understandable variable that says something like `showMySkyEnabled` or `showMySkyAvailable` and have it used in this condition.

10110111 (approved):

PR #2451, opened 2022-05-19
Added visualisation Earth umbra and penumbra for satellites

alex-w:
### Description
Added visualisation Earth umbra and penumbra for artificial satellites at fixed distance or at distance of selected satellite (distances are started counting from Earth's surface).

Fixes #2360 and #1246 (issue)

### Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] This change requires a documentation update

## Checklist:

- [x] My code follows the [code style](http://stellarium.org/doc/head/codingStyle.html) of this project.
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation (header file)
- [ ] I have updated the respective chapter in the Stellarium User Guide
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #5126, opened 2026-09-21
Fix black artifact for satellites in Earth's umbra

MARS02isfromMARS:
Fixes black opaque circle artifact when satellite is in Earth's umbra.

### Description
When a satellite enters the umbra, `sinThetaU` can become negative at certain Sun-Earth-satellite geometries, resulting in negative `umbraRadius` / `penumbraRadius`. This was rendered as an inverted, solid black circle.

This PR clamps both radii to >=0 with `std::max(..., 0.0)` in `getEclipseShadow()` (Satellite.cpp ~L899 and L911), preventing the inverted shadow.

Related to #2451, #4615 and the 26.1 penumbra visualization.

### Type of change

### How to test
1. Enable Satellites, show ISS
2. Fast-forward to eclipse entry (umbra)
3. No opaque black disk should appear — satellite should fade using invisible color mix.


plugins/Satellites/src/Satellite.cpp
```diff
@@ -896,7 +896,7 @@ Vec4d Satellite::getUmbraData(const double rhoE)
 	// geocentric distance of shadow circle towards antisun
 	const double umbraDistance=rhoE*cos(mu);
 	// radius of shadow circle
-	const double umbraRadius=rhoE*sin(mu);
+	const double umbraRadius=std::max(rhoE*sinThetaU, 0.0);
```
10110111: Why does `mu` change into `sinThetaU` in the case where `std::max` is a no-op? Similar question to the other hunk of this patch.

MARS02isfromMARS: Before fix - ISS in umbra shows opaque disk artifact (Lyra/Vega field)
This PR makes it fade correctly.

gzotti: This does not answer why you replaced sin(mu) by sinThetaU.

MARS02isfromMARS: > Why does `mu` change into `sinThetaU` in the case where `std::max` is a no-op? Similar question to the other hunk of this patch.

Thank you for review! You are right - in this hunk I meant to clamp the final radius, not change mu.

The artifact I saw (screenshot above in Lyra field) is an opaque disk when ISS is in umbra. It appears when `rhoE * sinTheta` becomes very small / negative near the umbra boundary due to floating point.

I will revert this hunk to keep `mu` and only add the clamp if `mu` can be <0, or investigate the actual draw call in `drawPoint` where penumbra mix is done. Could you point me to where opaque disk is drawn? I can update patch.

Sorry for confusion!

alex-w: sinThetaP sinThetaU are not declared

gzotti:
I have not fully re-studied the code and model in these days. a negative umbra radius sounds like annular eclipse geometry, but for this the satellite has to be very far away. Can you provide a screenshot where this happens? And a negative penumbra? Should that be on the sun side of earth, or actually behind the sun?

MARS02isfromMARS:
Hi @alex-w and @gzotti - all 17 checks are now green (AppVeyor finally passed this morning).

Fixed per your feedback: correct sinThetaU/sinThetaP naming + qMax clamping for umbra.

From your email last night , understood you are focused on TelescopeControl & ASCOM Alpaca. Happy to help test from Bangladesh if needed.

Ready for re-review when you have a moment. Thank you for your time!

- Marshad (MARS02isfromMARS)

MARS02isfromMARS:
Hi @gzotti thanks for asking, that helps me explain better.

The screenshot I attached near Vega at 16:12 - at that time Sun was on opposite side of Earth. So it's Sun -> Earth -> ISS. ISS is behind Earth in umbra. It should make a small round shadow, but current code makes that big dark circle that never moves. That big circle is the bug.

You are right that negative radius can mean annular eclipse, but that only happens very far, beyond 1.4 million km. ISS is only 400km and Moon is 384k km, still inside umbra. So negative here is just math error when sinTheta goes wrong. It flips the cone.

And for negative penumbra on sun side, yes that should be 0, no shadow at all when satellite is in front of Earth. My qMax(0.0) does that.

So my fix is simple - if radius goes negative, make it 0. Then we get no inverted big square, just small correct shadow or no shadow when it should be annular.

I have before and after video if you want, before shows big square covering Earth, after shows small circle.

Let me know if I should add a comment in code to explain this clamp.

Thanks!
Marshad
Screenshot_2026-09-21-00-23-40-18_a23b203fd3aafc6dcb84e438dda678b6

alex-w:
@MARS02isfromMARS but you see artifacts from telescope on the screenshot from Stellarium Mobile

10110111:
> So negative here is just math error when sinTheta goes wrong. It flips the cone.

Doesn't this mean that sinTheta is the one to be corrected? What you're doing then looks like a kludge, no?

BTW, you should wait until your screenshot gets uploaded, otherwise we only get the never-finalizing "uploading" statuses instead of the actual screenshots.

gzotti:
Yes, this screenshot is not from the desktop version, and these artifacts won't go away even if you switch off the satellite display completely, as they have nothing to do with satellites.  These telescope secondary mirror images are close to bright stars like Vega, Spica, Arcturus, Sirius etc. in the DSS surveys, as part of a photographic image. They don't move and won't go away. We can do nothing about them. (Apart from manually painting them away...)

MARS02isfromMARS:
@10110111 Fixed as suggested, thanks for catch!

- Root cause fix: clamp `sin(theta)*rhoE` with `qBound(0.0, ..., 1.0)` before `asin()` so cone never inverts when satellite deep in umbra
- Safety: `qMax(0.0, rhoE*sin(mu))` for radius

This should give small proper shadow instead of giant black square. Let me know if looks good.

10110111:
No, this makes no sense. Have you even tried to follow the logic of computation? Why did you remove division by `rS`? Why does the clamping even have to be done—did you check that the negative values you get are simply rounding errors, and not results of an algebraic mistake?

gzotti:
Repeat: Please send a screenshot created with the desktop edition of Stellarium that shows the problem: a satellite, selected and showing lots of data, especially its distance, and showing a 'black opaque circle artifact'. Note date, time and location.

gzotti:
You can add ASSERTs and run a debug build to see if the radii ever become negative. And show a screenshot from desktop Stellarium that illustrates the actual problem.

PR #5128, opened 2026-09-22
Backup existing SC when overwriting using SCM

xLPMG:
When editing a sky culture using the SCM and then saving it back to disk, thus overwriting the existing files, the original sky culture directory is first deleted and then the new one is saved. Should saving the new sky culture data fail for some reason, the old original sky culture is gone too. This PR fixes this with a backup mechanism:
- save new SC to temp dir, only continue if this was successful.
- next, backup original SC to backup file
- then, try to move temp to final file: 
    fail? -> restore original from backup
    success? -> delete backup

### Type of change

- [ ] Housekeeping

### How Has This Been Tested?
Edited an existing sky culture and save it again, then watched the file system if the temp and backup files are created.
If the file browser doesnt update fast enough, one can also edit the code to keep the temp/backup files from being deleted.

**Test Configuration**:
MacOS 27.0

## Checklist:


gzotti:
Thanks!
Maybe even optionally keep the backup? Add a checkbox "Keep backup", and inform the author in a MessageBox: "The old version is available at <FULL_PATH>". Of course, backups should not overwrite each other, so take care for attaching "_old" , but then check whether "<currentName>_old" exists already, and attach a number ("<currentName>_old(1)" etc.).  OTOH, I don't know now, would editing "<name>_old" later even be allowed, so that its backup could become "<name>_old_old"? Or is dirname==ID with all consequences of names inside index.json? Then probably only one backup level makes sense, and at best, a warning box "Overwrite old backup?"  could appear, and names would be "<currentName>_old" "<currentName>_old(1)", "<currentName>_old(2)". and the folders themselves would be invalid as skyculture because dirname!=ID.

plugins/SkyCultureMaker/src/gui/ScmSkyCultureExportDialog.cpp
```diff
@@ -119,131 +139,200 @@ bool ScmSkyCultureExportDialog::exportSkyCulture()
...
 		}
+		skyCultureDirectory = QDir(tempDir->path());
+	}
+	else
+	{
+		skyCultureDirectory = finalDirectory;
 	}
 
-	// Create the sky culture directory
+	// Create the sky culture directory (already exists for the temporary directory case)
```
10110111: It would be more sensible to set `createdDirectorySuccessfully` to `tempDir->isValid()` in the branch where it's created, and do the `skyCultureDirectory.mkpath(".")` call in the `else` branch of `isOverwrite`, also setting the flag. Then no such comment will be needed.

plugins/SkyCultureMaker/src/gui/ScmSkyCultureExportDialog.cpp
```diff
@@ -119,131 +139,200 @@ bool ScmSkyCultureExportDialog::exportSkyCulture()
...
+			return false;
+		}
+
+		// move the freshly exported sky culture into place
+		if (!QDir().rename(tempPath, finalPath))
+		{
+			qWarning() << "SkyCultureMaker: Failed to move exported sky culture into place at" << finalPath
+			           << ". The backup will be restored.";
+			// Roll back to the original sky culture.
+			QDir().rename(backupPath, finalPath);
```
10110111: This can also fail and need a message to the user.

plugins/SkyCultureMaker/src/gui/ScmSkyCultureExportDialog.cpp
```diff
@@ -119,131 +139,200 @@ bool ScmSkyCultureExportDialog::exportSkyCulture()
...
+	// exported directory in for the existing one, keeping a temporary backup so a failed swap can be rolled back.
+	QString keptBackupPath;
+	if (isOverwrite)
+	{
+		tempDir->setAutoRemove(false);
+		const QString finalPath  = finalDirectory.absolutePath();
+		const QString tempPath   = skyCultureDirectory.absolutePath();
+		const QString backupPath = finalPath + ".scmbak";
+
+		// remove potential leftover backup from a previous failed run
```
10110111: From a previous failed run? Wouldn't a normal backup also have such a name? Then exporting twice would ruin the original version of the SC.

I think a better way would be to use `.scmbak`, `.scmbak2`, `.scmbak3` etc. for each new backup if the chosen name is already used by an existing directory.

10110111: Ah, I hadn't read to the point where the final name was chosen. Still, isn't there a scenario where the backup from the previous attempt was the only remaining copy of the original version of the SC?

plugins/SkyCultureMaker/src/gui/ScmSkyCultureExportDialog.cpp
```diff
@@ -119,131 +139,200 @@ bool ScmSkyCultureExportDialog::exportSkyCulture()
...
+			const QString oldPath = makeUniqueBackupPath(finalPath);
+			if (QDir().rename(backupPath, oldPath))
+			{
+				keptBackupPath = oldPath;
+			}
+			else
+			{
+				qWarning()
+				        << "SkyCultureMaker: Failed to keep backup at" << oldPath << "- discarding it.";
+				QDir(backupPath).removeRecursively();
```
10110111: I don't quite follow the logic here. Where does the backup end up in this case? It surely exists by this point, since we just wanted to rename it—but instead of keeping it in a temporary-named directory we destroy it completely?

